PR #64, opened 2018-11-01
Add prettier config file

glauberramos:
Add Prettier config file so all files keep the same format.


BenGamma:
Hi Glauber,

First, thanks for your contribution :)

We were using the default settings until now with prettier, but you are right, better to have our own ✌️

PR #146, opened 2018-11-26
feat(functions): adding image converter support

nzchicken:
@chrismclarke I don't have the contents of the functions/config directory, so I can't deploy the function. Are you able to deploy this? I've fixed all linting errors so it should just deploy. Also, feel free to just send me the config on slack and I'll push and test it from here.

### DIFF
--- a/functions/package.json
+++ b/functions/package.json
@@ -13,9 +13,10 @@
   "dependencies": {
     "axios": "^0.18.0",
     "body-parser": "^1.18.3",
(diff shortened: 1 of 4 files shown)


chrismclarke:
Hey man, sorry it’s taken so long to get round to this, had an intense burst of real-life stuff to do and then a bit of housekeeping to do on the repo more generally before being ready to merge. Will send config via slack

Issue #177, opened 2018-12-26
Remove uploaded files when documentation not saved


chrismclarke:
child of #3
Did partial implementation using refs and uploadedfile component, however doesn't delete full folder reference on server (leaves empty folder), and messy when requiring second set of refs for steps, so propose trying to do something server-side instead

BenGamma:
Where are we on this. Still sounds relevant to me to implement.

chrismclarke:
no longer relevant as all files uploaded on submit

PR #178, opened 2018-12-26
#118   image delete

### LINKED ISSUE #118: allow deleting image after upload
related to #74

### LINKED ISSUE #153: Documentation: pics bug
I was trying to upload a new documentation but I encountered multiple bugs when uploading images. Couldn't seem to work out a pattern there. Images previously uploaded would suddenly disappear (multiple at a time). Had to go back and forth to check all ics were still there. Super weird.

child of #3

chrismclarke:
Implements file deletion option for uploads (this time including fully-merged docs refactor code)
closes #118
closes #153

## Breaking changes
-	‘CoverImage’ element no longer used, instead we use the UploadedFile element to render the image preview (could use the CoverImage element inside of it)

## Additional changes
-	Tutorial file upload now takes multiple files – it made sense to implement this way, but if we really only want a single can revert back
-	File info preview given for tutorial files when uploaded
-	Converted FirebaseFileUploaderField to use typescript – quite simply the js was harder to implement (e.g. no autocomplete) and had incorrect code (e.g. saying props existed on it which didn’t – fine for current implementation but threw errors trying to implement in slightly different context)
-	Fix tags select - built react-final-form bindings for component (so can be intergrated as custom field component) and fixed css overrides from elements file (wasn't implementing due to injection order)

## Additional notes
Deleting automatically on page exit:
started to implement a method which tracks refs of all uploaded files, and then if the user doesn’t save the tutorial calls the ‘delete’ method on all of them. Was working fine except then I also needed to implement on step images as part of refactor – it started to feel clunky/messy so I abandoned the approach and instead will try to do something in the future server-side that deletes all uploaded files associated with tutorial idea (created issue #177 )

fieldarray vs array mutator methods:
like what you’ve done for steps field array and uploaded field variant. Decided was overkill to do same implementation for multiple file display and upload so stuck to using simpler array mutators


chrismclarke:
Updated request, have fixed the cover image not displaying bug (integrating changes from uploaded file delete button to form) and removed/commented out additional code no longer used

PR #190, opened 2018-12-27
tools(travis): changing branch names + setting run specificity + security

nzchicken:


nzchicken:
@chrismclarke I changed the CI script to stop echoing out the firebase token, since anyone with travis access could see our production key there.

PR #199, opened 2018-12-29
Image resizing bug#197

nzchicken:


chrismclarke:
I'm still seeing errors on all the calls
https://console.firebase.google.com/u/0/project/precious-plastics-v4-dev/functions/logs?search=&functionFilter=imageResize

Is this because it's still the old code deployed?

chrismclarke:
image

Also you can get full logs and summaries from the firebase console (in case you hadn't already seen)

chrismclarke:
Also I see you've changed the scripts, good to know % doesn't work in mac-land (I know you have $ but assumed both worked).... There's also a useful npm library called [cross-env](https://www.npmjs.com/package/cross-env) which I usually use in these types of situations, although for now agree it's a lot simpler to just use the known relative path

nzchicken:
@chrismclarke whoops, needed to check if key existed on object. This has been a real pain to debug this thing, since the original code was overwriting the file, where as we are storing a backup. When storing the backup and overwriting the original, it fires addition onFinalize events, which is what i'm using the object.metadata to store and ignore.

% is only supported by windows, $ is supported by unix/bsd based (i.e. linux + mac, etc)

chrismclarke:
image

PR #225, opened 2019-01-16
Stop confusing npm with yarn

PepijnSenders:
I removed the `package-lock.json` file, since we apparently use `yarn` in this project and also changed `npm run start` to `yarn start` in the `CONTRIBUTING` file to avoid confusion.


chrismclarke:
Thanks @PepijnSenders , I'd be amazed if this is the last we see of the package lock (some people may still choose to use npm and I don't have a problem with that), but makes sense for the docs to be consistent.

Whilst we're at it do you have any experience/advice for how we can avoid lock file conflicts? A while back I tried making all dependencies explicit and getting all users in sync but they still kept arising.

PepijnSenders:
@chrismclarke I think it's more stable to use either one, the `yarn.lock` file contains exact references to the packages, so when anyone installs again they will get exactly the same modules as anyone else. Also yarn automatically fixes merge conflicts in the lock file when you just run `yarn` works really well 😉

PR #231, opened 2019-01-21
fix back button display - broken image on step

BenGamma:


package.json
```diff
@@ -27,6 +27,7 @@
     "md5": "^2.2.1",
     "mobx": "^5.5.2",
     "mobx-react": "^5.3.6",
+    "moment": "^2.23.0",
```
PepijnSenders: Do we really need `moment` the package is humongous :/ I can recommend using [date-fns](https://date-fns.org/) instead!

src/components/Button/elements.js
```diff
@@ -7,7 +7,7 @@ const getBorder = ({
   basic
 }) => {
   if (basic) {
-    return `borner: 1px solid black;`;
+    return `border: 1px solid black;`;
```
PepijnSenders: 😂

PepijnSenders: Maybe we should start implementing `stylelint` to prevent this, not sure if it would have caught this though.

src/pages/Discussions/ListRow/ListRow.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <Title href={'/discussions/post/' + post._id}>{post.postTitle}</Title>
+          <TagsContainer>
+            {post.tags.map((tag, j) => (
+              <Tag key={j}>{tag}</Tag>
+            ))}
+          </TagsContainer>
+        </TitleAndTagsContaier>
+        <InteractionNb>
+          {post.postType === 'discussion'
+            ? post.commentCount + ' comments'
```
PepijnSenders: Maybe we should check for the length of `1` and `0` as well.

package.json
chrismclarke: Agreed! We don't really need any package and should just be able to use vanilla javascript Date objects. 

I see in discussions it is only used to calculate the time between original post and now, so that is likely just 
const now = Date.now() 
const posted = new Date(postDate)
const diff = now-posted
And convert to whatever format preferable

PepijnSenders: Even better 😉

PepijnSenders: I'll create a PR to update it.

src/components/Button/elements.js
chrismclarke: I think the styled-components implementation needs a bit of work, we shouldn't have this many standalones and should in the very least be extending on a base system. I'll put together a proposed spec and see what people think. @PepijnSenders , feel free to provide an example of the system you are proposing also so that we can better understand and judge.

chrismclarke: I think there's no point adding another layer until we sort out the underlying base, but looks like it could be interesting once that is done so thanks. Will add an issue so we can come back to the discussion.

PepijnSenders: You can replace this entire base elements layer with a styled system and then you also wouldn't need a file that contains GreenButton, LargeGreenButton etc. You would just have one export and then every component is configurable by itself.

package.json
chrismclarke: Great! if possible double-check how it's currently implemented in the discussions component to keep fidelity with what @BenGamma is working on

src/components/Button/elements.js
chrismclarke: I definitely like the idea, I've mentioned you in the thread on github so hopefully @BenGamma will review (he's been taking the frontend lead) and move the discussion along

PR #239, opened 2019-01-28
Remove moment.js

PepijnSenders:
Moment.js was added in this PR: https://github.com/OneArmyWorld/onearmy/pull/231. I exchanged it for a `date-fns` implementation and added a test. @chrismclarke let me know when you add the design document and then I'll add an entry for using `date-fns` when you're doing date calculations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "file-loader": "0.11.2",
     "final-form": "4.10.0",
     "final-form-arrays": "^1.1.0",
-    "firebase": "^5.8.1",
+    "firebase": "5.8.1",
     "fork-ts-checker-webpack-plugin": "0.2.8",
     "fs-extra": "3.0.1",
     "html-webpack-plugin": "2.29.0",
--- a/package.json
+++ b/package.json
@@ -12,13 +12,14 @@
     "case-sensitive-paths-webpack-plugin": "2.1.1",
     "chalk": "1.1.3",
     "css-loader": "0.28.7",
+    "date-fns": "^1.30.1",
     "dotenv": "4.0.0",
     "dotenv-expand": "4.2.0",
     "extract-text-webpack-plugin": "3.0.2",
     "file-loader": "0.11.2",
     "final-form": "4.10.0",
     "final-form-arrays": "^1.1.0",
-    "firebase": "5.5.8",
+    "firebase": "^5.8.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -12,13 +12,14 @@
...
     "css-loader": "0.28.7",
+    "date-fns": "^1.30.1",
     "dotenv": "4.0.0",
     "dotenv-expand": "4.2.0",
     "extract-text-webpack-plugin": "3.0.2",
     "file-loader": "0.11.2",
     "final-form": "4.10.0",
     "final-form-arrays": "^1.1.0",
-    "firebase": "5.5.8",
+    "firebase": "^5.8.1",
```
PepijnSenders: There was a `grpc` bug with my node version.

chrismclarke: until we have better testing it might be safer to use pinned. happy to go with 5.8.1

PepijnSenders: Ok, can pin it here, yarn already pins it already though.

chrismclarke:
I'm not exactly sure which design document you are referring too...
Also I saw the build fail (typescript types) but see another commit just come in so guessing that's probably the fix

PepijnSenders:
@chrismclarke I meant this one: https://github.com/OneArmyWorld/onearmy/pull/235/files#r251348530

package.json
chrismclarke: haha, yeah I struggle to put my full faith in yarn + humanity (I'm pretty sure it was you who removed the npm lock file, let's see if it creeps back in.... also somebody always finds a way for yarn conflicts). but thanks!

chrismclarke:
Oh, I was planning to leave it as a google doc and just link for comments, but I take your point on tracking changes. Will do it now

chrismclarke:
As I don't want to have to keep two sources of information in sync (and think it's important to sitll retain the google doc for overall platform) I've moved the 'deep-dive' (/things that would be of more interest to devs) to a wiki page: https://github.com/OneArmyWorld/onearmy/wiki

Will try to populate a bit more this week, I also have no idea how easy it is for other people to add what they might want so perhaps you can let me know your thoughts

package.json
chrismclarke: that's good to know about yarn, will sleep easier tonight so thanks!

Issue #247, opened 2019-01-30
[how-to] cost saved as NaN


BenGamma:
When displaying a how-to, the cost in the description can be saved as NaN. This appens when the user enters a value with the currency or other caracters than numbers.
See :
<img width="1173" alt="capture d ecran 2019-01-30 a 14 48 04" src="https://user-images.githubusercontent.com/4993215/51985453-283c8b80-249e-11e9-9f2e-95ca315dddb4.png">

For now it should only be saved as number.
We need to perform a check before saving the datas that the `tutorial_cost` field is a number and not a string.

PR #249, opened 2019-01-30
#247. Fixed `tutorial_cost` formatting.

chupakabra:


chupakabra:
@tx4x, probably, I just figured `parseFloat` will derive the first look-like numeric substring. I rethought it and decided that probably showing validation error will be better to be certain.

p.s.: I tried to make the minimum changes as have minor experience with typescript (just to not break existing stuff)

tx4x:
@chupakabra , i see. keep in mind TS isn't doing anything for you at run-time. think of it more of a annotation system to catch problems at build-time.
i think you may have a look also at local specifics when entering prices (ie: using dots for thousands notation), possibly overkill ... https://github.com/adamwdraper/Numeral-js/blob/master/src/numeral.js

Issue #232, opened 2019-01-22
Best way to track views on post


BenGamma:
We'll need a view counter of each post on the discussion section.
What is best way to achieve this ?
- each time a post page is requested we check the IP adress, if it is unique IP then we increment the `viewCount` value ? That would require to save all IP adress from post page display, may be a lot of data.
- Maybe use a session variable in the cookies ?

chupakabra:
Think using cookies is a better option (if we don't care about views of anonymous users) as IP address doesn't seem unique and representative enough (the same user may view it from different locations or few users may use the same IP address or user may use Tor, etc.) and s you said we will have to store a lot of data.

Found some package which could be useful: https://www.npmjs.com/package/react-count, but didn't figure if it distinguishes users.

Anyway, I'm going to spend some time to try it. 
cc @mattia-io

mattia-io:
@chupakabra brilliant. Keep us updated on this :)

BenGamma:
@chupakabra If at some point you need to access the database, just DM on slack and I'll open access for you :v:

chrismclarke:
If you're thinking of cookies then I'm guessing google analytics would save a lot of time (will need GDPR compliance). 

Personally I would want us to collect only the minimal amount of data that we actually need (I think lots of people in our community are quite opposed to big-data culture), so before going too far it might be good to try define what that is. I'm guessing @mattia-io might have some thoughts on this too.

The other option is to write directly to the database, either via a cloud function or client-side. This doesn't scale quite so well as the same document typically only likes 1-write per second or so, but there are ways around it if we thought there would be need.

A couple links that might be of use:
https://medium.com/google-cloud/tracking-site-visits-on-react-app-hosted-in-google-cloud-using-google-analytics-f49c2411d398
https://firebase.google.com/docs/firestore/solutions/counters
https://tonyxu.io/posts/2018/use-firebase-cloud-function-to-count-website-visitors/

chupakabra:
I struggled a bit with `react-count` applying and tried to set up google analytics tools, but unsuccessfully for both.

-  for `react-count` there was some troubles with `OnlineCount` tag
-  for google analytics I tried different manipulations and debug tool showed that everything seems is fine, but no data neither users nor some events in Analytics overview.

And then was 'wait a minute...', just figured that we don't have any authentication system and don't store any session keys, do we?

mattia-io:
[...]

PR #261, opened 2019-02-09
#232. Added views count tracking by means of Google Analytics.

chupakabra:
I decided to focus on implementing this feature by means of Google Analytics, but faced few issues:

1. I didn't manage to generate JWT access token as it was done in `functions/src/databaseBackup.ts`, there were some troubles in `google-auth-library` and particularly in `pify`, exactly like described here:
https://stackoverflow.com/questions/54468809/what-is-causing-typeerror-expected-input-to-be-a-function-or-object-issue. So, just used token generated via sample: https://developers.google.com/analytics/devguides/reporting/core/v4/rest/v4/reports/batchGet (sorry, I know that's not good, but spent a lot of time to figure out that libs issue)

2. Seems there is some delay in info updates on google server: updates appear in Analytics dashboard panel in a minute and return in response in 5 minutes! 

Conclusion: I didn't try firebase solution, but seems to me that might be a better option


chupakabra:
Fixed JWT token issue. Used `google-oauth-jwt` for that.

chrismclarke:
Just taken a look, I really like the implementation, I'm wondering what would be the best way to add a bit more separation between components and analytics (move the logic out). I'm thinking we could either create a global analytics store which can handle all request logic and pass back through the component, or creating an api endpoint on our functions to handle logic, or possibly both. What do you think?

chupakabra:
@chrismclarke, hey ) Thanks for the review! 

Regarding moving analytics functionality to separate store I think it depends if we are going to use analytics more extensively. For now I just wanted to move google client authorization functionality to `functions`, but faced the issue with importing (tried to link, but it complained about import anyway), going to figure out it later today.

I'm planning on trying approach with storing the views count in db this week as this one doesn't seem optimal, it has some delay with updating info on google server. But I'm still not sure. Just want to compare.

chrismclarke:
Thanks for all of your work on this!

What was the issue with importing? Did you have a .js file with an import statement throwing protected keyword? if so you can add an .eslintrc config (did this today for the main repo facing similar problems)

Awesome to have both options for comparison, keen to hear what you think when you've had the chance to take a look.

chupakabra:
> @chrismclarke, hey ) Thanks for the review!
> 
> Regarding moving analytics functionality to separate store I think it depends if we are going to use analytics more extensively. For now I just wanted to move google client authorization functionality to `functions`, but faced the issue with importing (tried to link, but it complained about import anyway), going to figure out it later today.
> 
> I'm planning on trying approach with storing the views count in db this week as this one doesn't seem optimal, it has some delay with updating info on google server. But I'm still not sure. Just want to compare.

I've got this:
```
Module not found: You attempted to import /home/chupakabra/onearmy/functions/src/gAPIJwtAccessToken which falls outside of the project src/ directory. Relative imports outside of src/ are not supported. You can either move it inside src/, or add a symlink to it from project's node_modules/.
```
tried to add yarn link to `functions` from `node_modules`, at least saw such suggestions on forums, but had the same error

chrismclarke:
Ah I'm following, firebase gives a way of calling cloud functions from the frontend source via 'callable functions' https://firebase.google.com/docs/functions/callable. You can see an example of 'removeStorageFolder' from `functions/src/index.ts` which I created to test this. You can then import functions from the main `src/utils` package and use to trigger the api call.

chupakabra:
> Ah I'm following, firebase gives a way of calling cloud functions from the frontend source via 'callable functions' https://firebase.google.com/docs/functions/callable. You can see an example of 'removeStorageFolder' from `functions/src/index.ts` which I created to test this. You can then import functions from the main `src/utils/firebase` package and use to trigger the api call.

Sorry, I didn't manage with that. Suddenly have a lot of syntax errors in typescript for `functions/src/index.ts` and can't deploy command to firebase (firebase: no such command), so it doesn't see any exported command from `src/utils/firebase` as a consequence. 

Unfortunately, have no time to figure out in detail with that now, so if it's not critical can we just copy `config.json` to `src/config/config.json`? Sorry

chrismclarke:
Not your fault, I'm not near a computer now but I'll send over config file later and publish the dev one to the repo (no reason why can't be public). Sounds like I probably could give a quick tidy up to the file structure too

chrismclarke:
ok file now populated if you want to merge from master

chrismclarke:
Hey @chupakabra , quickly checking in, is this now good to merge or are you still working on it?

chupakabra:
> Hey @chupakabra , quickly checking in, is this now good to merge or are you still working on it?

Hey

I didn't succeed with trying to store counter directly to firebase via function and probably this solution will be optimal than storing it in db anyway. So, the answer is yes, it can be merged. 

P.S.: Let me know if I should add permissions for other service accounts in analytics or we can create different analytics project.

chrismclarke:
Great, I'll merge it tomorrow alongside any other outstanding PRs.

In terms of analytics accounts, I've just created an issue (#278) to create a master email that will be used by the project and we should share those types of things with. Will let you know when that is done and you can either add as an admin or transfer ownership.

chrismclarke:
oh, also just for future ref I'll be removing the backend config files in favour of passing environment variables (see #277 )

chupakabra:
> oh, also just for future ref I'll be removing the backend config files in favour of passing environment variables (see #277 )

Good one! It seems not really secure indeed.

chrismclarke:
Haha, no it's not! goes against my trusting nature :p . But makes sense to have dev and prod using mostly the same pipelines and hopefully shouldn't impact on other devs workflow. Will refactor existing functions to use the variables and hopefully should be clear how others can do the same.

chrismclarke:
Just as a quick update, I merged the changes and am now doing a bit of refactoring to call from the cloud functions instead of direct (and pass through variables more securely). I think I've figured out the problem with calling functions (CORS request as the functions sit on a different server, so trying various ways to get them to play more nicely, if successful will let you know)

chrismclarke:
@chupakabra , we now have a google setup for the project, so if it is possible to transfer ownership please send it to hello@onearmy.world. If not possible I'll just create another analytics account and add you as an admin.

chupakabra:
> @chupakabra , we now have a google setup for the project, so if it is possible to transfer ownership please send it to [hello@onearmy.world](mailto:hello@onearmy.world). If not possible I'll just create another analytics account and add you as an admin.

@chrismclarke, I can add admin permissions to [hello@onearmy.world](mailto:hello@onearmy.world) (what I did), but looks like I can't change ownership, guess cause it is not tied to my google account. So, probably it's better to create another analytics account to be sure.

chrismclarke:
Sure thing, shall do thanks

Issue #234, opened 2019-01-24
How-to frontend


mattia-io:
@charleenfei is working on this issue 🚀

The How-To page was designed in a rush to make sure we could at least gather content from users. We now have a more consistent UX design ready to be implemented. This will undergo a final design retouch later in the year. 

![Screenshot 2019-01-27 at 21.49.51.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5beadf2e66e72100136491d9/125516c66e82c728ace21e0d46db978826878dba87e6ab03f60da1cd6416763e7b53e32a64c5b179f116224f5a5e5cb053181d1ba6acda6fb4bc2e75d80d58be8064597c946f14adbb821fb950bf218e4c87b7ae965e82b96371b8d6d6056b5c17129bfe553fa19c4fc05ef449e34d53a6113171ab.png)

Assets:
https://drive.google.com/drive/folders/1NnnIEpKY_UrWWCgB0CsavkUJUtZdsQ3g

Prototype:
https://sketch.cloud/s/KnZgA/WK2amx2/play

Recordit:
http://recordit.co/QrIFynXFIn

child of #3

tx4x:
me as newbie would love to know a few things just : 
1. use @material-ui/core/createStyle over 'styled-component' ? 
2. where to add end-points in the back-end
3. patterns and code quirks ? ie: do we should use react's newer features such as 'useState' and 'useContext' in regard of more clean code, ie: functional programming (ramda?) and point free programming, see more https://www.youtube.com/watch?v=wXLf18DsV-I 
4. localization : where , what ,...how
5. policy when adding dependencies ? 

that's it for now ;-)

chrismclarke:
1. Please describe the benefit/drawbacks of your suggestion so we can open it to discussion
2. There is a functions folder, see the index.ts for different ways backend can be interacted (http api, cron, triggers and called directly)
3. We do not have strong code requirements (a limited set is in the contributing guidelines). Essentially we expect all programmers to be responsible for writing their own clean code. If for any reason somebody feels a particular convention should be site-wide they should put the case forward as to why, and ideally be willing to support/review code of others trying to implement the same.
4. We started to create a module, however it's now been put in the backlog in favour of more pressing issues (quite frankly we're not sure whether we would have the capacity to keep a multi-language site well-maintained, especially in the early stages where lots of things are constantly changing
[...]

PR #262, opened 2019-02-10
HowToList component styling

charleenfei:
this is just a simple PR containing style changes for the HowToList component.

I also refactored a bit the way the component is rendered to make it easier to work with in future changes.

The followup PR will contain the changes for the HowTo.content.tsx component as indicated by the new UI styles brief.


BenGamma:
Hey Charly,
I was probably not clear on the mockups but don't remove the "create how-to" button for now, it's still needed.
Here the main focus is the card display and the card grid.

charleenfei:
> Hey charly, thanks for the PR, a couple quick things:
> 
> 1. I see you've removed the 'ClampLines' component. This was originally designed to ensure that all descriptions fill the same number of lines no matter how long (displaying a '...' if the description is too long). Is there a reason to remove (do you have a better implementation in mind?)

Gotcha, I overlooked that it was being used for this, I can move it back.

> 2. I'm happy to see the styles and render methods moved out to make it easier to work with (👍 ). In the future we would likely want to move it out of the file entirely to implement styled components instead (however no need to worry too much for now, just to make you aware).

noted!

> 3. I see the new font. I'm not sure if we've had full spec from Mattia for exactly what fonts he wants in the project (you probably know more than me at this point), however for now I suggest fine to leave in and will open discussion in a separate git issue. Although would you mind trying to see how easy it would be to import into the build instead of linking? By having it part of the bundle we are not reliant on external servers, improving any future offline build we might want and also avoid flash-of-invisible-text if server slow (I assume at least). Example here:
>    https://stackoverflow.com/questions/41676054/how-to-add-fonts-to-create-react-app-based-projects

Yep, will try to have it imported into the build.

charleenfei:
> Hey Charly,
> I was probably not clear on the mockups but don't remove the "create how-to" button for now, it's still needed.
> Here the main focus is the card display and the card grid.

will put it back in! like you said, in the mockups the button was no longer there (I had just assumed it would be part of the top bar component).

charleenfei:
> 1. I see you've removed the 'ClampLines' component. This was originally designed to ensure that all descriptions fill the same number of lines no matter how long (displaying a '...' if the description is too long). Is there a reason to remove (do you have a better implementation in mind?)

I've just checked again the reasoning for this -- it's because the description in the how-to list has been removed in the new design files.

BenGamma:
Hello here !
I just review the PR and I'm gonna close for now, as I saw it was still a work in progress.
You can still add commits to it but please reopen when you think that it will close the issue #234  ! Thanks 🙌

Issue #212, opened 2019-01-11
change top section text in onboarding bar (dark blue bar below navigation) depending on nav


BenGamma:
Depending on your navigation, we'll display a different onboarding bar under the top navigation, to inform the user on which section he is.

<img width="1407" alt="capture d ecran 2019-01-11 a 12 26 13" src="https://user-images.githubusercontent.com/4993215/51031218-287ef080-159c-11e9-9c5d-83410a9cc9de.png">

The text contained in the onboarding bar is not figured out yet, so it can just be the name of the section e.g. (newsfeed, how to, discussion)

child of #106

mattia-io:
I think we can wait on this for the final UI from nico.. navbar is ok as it is for now with the raw functionalities in place

suvya:
I can take this up @mattia-io

mattia-io:
Nice one @suvya 🙌

mattia-io:
@suvya any rough idea how long this would take?

PR #263, opened 2019-02-11
 #212 - change top section text in on-boarding bar 

suvya:
**Enhancement** - Add a blue bar on the header to display text based on the menu being clicked by the user.
* Add formLink to display the selected link description on page
* Add labelText to LocalStorage to persist the state
* Add onClick event to links to pass the clicked item
* Add handleOnClick function to set labelText state

Note: This PR just displays the name of the menu being clicked on the bar. This needs to be changed to the respective description once the text is finalized. There will be no bar displayed on fresh load.  

@mattia-io @BenGamma


chrismclarke:
Also I'm not sure what was the best way to append onto your existing commit fork, so the addition has landed here:
https://github.com/OneArmyWorld/onearmy/tree/suvya-highlight-onboarding-bar

Please merge into your fork before adding more code

BenGamma:
We didn't ear back from you @suvya so we keep on going. Don't hesitate to ask question if you feel stuck next time, we'll try our best to answer and produce good work !
Merge with #291, closing.

PR #264, opened 2019-02-11
Button disabled state

tx4x:
PR to add a 'disabled' state to the core button, needed to mark buttons active or inactive in dialogs or forms.


tx4x:
@chrismclarke : ups, sorry! if you need it different let me know, i am also curious about ui-basics such as keyboard navigation and/or focus management and policy

chrismclarke:
makes sense to me, merged

Issue #186, opened 2018-12-27
allow toggle between community and public sites


chrismclarke:
child of #106
requires #184

mattia-io:
Hey boys,

Yhe I thought the behaviour here was clear, that's why I didn't comment explanations. The behaviour is just like you've described @BenGamma with the addition that on https://onearmy.world there is a big button "go to community" (temporary text :) that will bring both logged in and not logged in users to https://community.onearmy.world

Any more info needed here?

BenGamma:
Gotcha, I was looking for the link to the mockups, but just found it 😅 https://drive.google.com/drive/u/0/folders/1OyJmh39ojHJxs6otETNkYvDUVpAh-3aN

In a close futur we will probably want to onboard user with the user login procedure after clicking on the "go to community" button.

But for now what we'll do is :
* display the Public site on *onearmy.world*
* on click on "go to community" button, redirect to *community.onearmy.world* subdomain
* To display the associated page & navigation depending on the subdomain you can found an example in `src/pages/index.tsx`

chrismclarke:
we're quite a way off implementing this so closing for now

Issue #184, opened 2018-12-27
create global site navbar


chrismclarke:
*Updated 06-02-19 (Mattia)*

Overview:
We need to create the navbar for the global site https://onearmy.world/ 
This navbar will enable users to navigate to:
-Projects
-About
-Community

![Screenshot 2019-02-06 at 09.43.23.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5beadf2e66e72100136491d9/125516c66e82c728ace21e0d46db978826878dba87e6ab03f60da1cd6416753e7952e32a64c5b371f116284f5d5c5cb053181d1ba6acda6fb4bc2e75d80d58be8064597c946f14adbb821fb950bf218e4c87b7ae965e82b96371b8d6d6056b5c17129bfe5a3fa19c4fc15cf249e74c52a2153f71ab.png)

Latest prototype:
Prototype
https://sketch.cloud/s/KGerV/VrVvaPw/play

Originals + Screens
https://drive.google.com/drive/u/0/folders/1OyJmh39ojHJxs6otETNkYvDUVpAh-3aN

Epics:
https://docs.google.com/spreadsheets/d/1pkLRKCbQiJOtQwWEhVNgSTvDWf5SnVAz10vMo4k-LNg/edit#gid=0

Any questions post in the comments

child of #106

mattia-io:
@harrier91 is working on this 💪
Make sure to check this other issue related to this #186

chrismclarke:
cool, pr sent to put login system back in and will leave to @harrier91 for styling (I'm hoping that's the same Andrea that I've been talking to on slack right?)

chrismclarke:
wait my bad, Andrea is community navbar... all cool

chrismclarke:
any update on this @harrier91?

mattia-io:
@harrier91 I am gonna have to assign this issue to someone else for now unless you will confirm you want to deliver this

harrier91:
Some changes I did were merged already into master, I'm not sure what more should be done about it. Which is the definition of done of this task? Because I think some pages are not ready yet

mattia-io:
Hey @harrier91 thanks!

@chrismclarke is there anywhere we can check this? 
Have the global site pages (even placeholders) already been created?

BenGamma:
@harrier91 I can't find any PR from you, can you point us to your changes ?

[...]

PR #268, opened 2019-02-13
integrate login system back into header

### LINKED ISSUE #185: link login system to navbars
child of #106

chrismclarke:
@BenGamma , this is just a small commit to bring the login system back into the header and some code refactoring. 

I used the styled-component button however wanted to add an onClick event to it which was not supported by default. Instead of creating an extra 'onClick' optional prop I've added a slightly complicated looking typescript extension. This is copied from react's default implementation of a <button /> element, so I think it is correct for our element which looks to extend this. It will automatically allow for any button properties, component children etc.

Please take a quick look when you can (I know somebody else was planning to work on further navbar styling so this will need to be committed first). Full list of changes as:

- integrate login system back into header
- extend types to styled-component button
- refactor login from to run from user store
- remove login container
- add optional app color type list

closes #185


BenGamma:
Good for me.
I guess you are talking about #184 & #186 ? For the toggle between public and community site we will probably display the login option only on public site, and access to the community site only once logged in, but that is an further behavior that we will manage later on.

PR #273, opened 2019-02-14
Tiny MCE editor component

tx4x:
This adds the Editor component (only). There are still pending questions and todos but for now you can use it to get an idea. If you need things changed, no problem !


tx4x:
@chrismclarke or @BenGamma, here the first slice,. I placed a debug page at http://localhost:3000/debugEditor (dev mode only)

chrismclarke:
amazing, will test and merge tomorrow
Thanks again

Issue #272, opened 2019-02-14
Add dev site notice


chrismclarke:
would probably be good to have a small message somewhere stating that this is the dev site, and any data shared is openly available. This wouldn't be shown on live site, but the code might be adapted for cookie policy or any other requirements

BenGamma:
#272 added a little icon on top of the page that lived in `src/components/Dev/DevNotice.tsx` based on the isDebug node variable.
What would be nice to have is on hover a little tooltip with this indication :
"This is the dev site and all datas posted here can be public"

mattia-io:
@sanjuskm05 is now working on this ⚡️

sanjuskm05:
Submitted the pull request. Whenever possible, please review and merge it

chrismclarke:
reviewed, minor changes requested, please update the PR when complete. Thanks

PR #275, opened 2019-02-18
#272 dev notice

tx4x:
adds simple exclamation icon to the top bar if src/config.ts::isDebug is true (process.env.NODE_ENV === 'development').

@todo: DevNotice styling


chrismclarke:
Looks good, just in keeping with our current workflow we'll plan to test and merge tomorrow

BenGamma:
Thanks, I merge the DevNotice component into master with #311 
I left the commits related to the readme / machines and the button disabled state because they are not related to #272

PR #282, opened 2019-02-18
#232  track post views  firebase

chupakabra:
Though we decided that the solution with using Google Analytics tools seems better, I decided to experiment a bit with saving counter in db.

Comparing with GA it seems updates counts much faster, and it took about ~100-700ms from what I saw in logs: https://console.cloud.google.com/logs/viewer?project=precious-plastics-v4-dev&authuser=0&resource=cloud_function%2Ffunction_name%2FhitPostCounter%2Fregion%2Fus-central1&minLogLevel=0&expandAll=false&timestamp=2019-02-18T18:38:27.113000000Z&customFacets=&limitCustomFacetWidth=true&interval=PT1H&scrollTimestamp=2019-02-18T18:28:13.934150314Z&dateRangeUnbound=both&dateRangeEnd=2019-02-18T18:38:15.037Z

Though still not sure how to apply distributed counters solution in our case and how to optimize writing operations, that's all seems not optimal at all in terms of scalability.

Regarding making countings per unique user, probably we could use jwt tokens, not sure.


src/pages/Discussions/Discussions.tsx
```diff
@@ -9,27 +9,33 @@ import ListRow from 'src/pages/Discussions/ListRow/ListRow'
...
+  discussionStore: allStores.discussionStore,
+}))
 @observer
 class DiscussionsPageClass extends React.Component<IProps, any> {
   constructor(props: any) {
     super(props)
+
+    // initial state
     this.state = {
-      posts: DISCUSSIONS_MOCK,
```
kopecms: I would delete DISCUSSIONS_MOCK from imports.

src/models/discussions.models.tsx
```diff
@@ -0,0 +1,6 @@
+export interface IDiscussion {
+  avatar: string
```
kopecms: mock model had diffrent attributes so this probably breaks orderListBy in src/pages/Discussions/Discussions.tsx

src/pages/Discussions/Discussions.tsx
```diff
@@ -9,27 +9,33 @@ import ListRow from 'src/pages/Discussions/ListRow/ListRow'
...
   }
 
   public async componentDidMount() {
     // load mocks
     console.log('mocks:', DISCUSSIONS_MOCK)
+    await this.props.discussionStore.getDiscussionsList()
+    this.setState({posts: this.props.discussionStore.allDiscussions})
   }
 
   public orderListBy(orderType: string) {
```
kopecms: the funciton could be simplify to something like (or could be deleted becouse I think it will stop working with this commit):
 public orderListBy(orderType: string) {
    let sortedList = this.state.posts.slice().sort((a, b) => {
      return b[orderType] - a[orderType]
    })
    this.setState({ posts: sortedList })
  }

chrismclarke:
Thanks for the code and feedback both, I've updated the mock and models on the /module/discussions branch so I think this can stay closed. Am planning to do some work on the discussions store tomorrow so will migrate the code above then

PR #330, opened 2019-02-21
323 create discussions page

BenGamma:
A basic component for the createPost page with two fields :
* title of the post
* editor with post content

and a button to save the tutorial in database.


chrismclarke:
looks like it should be mostly good, I'll review this evening...

chrismclarke:
Made one minor change (rename db to afs as I'm currently refactoring how the database works), in another update I'll move the code that interacts with the database out of the component and to the discussion store but for now it works fine.

PR #332, opened 2019-02-22
Module/discussion backend

chrismclarke:
- add discussion store
- backend database methods
- bind db data to discussions
- fix broken fields

REQUIRES DEPENDENCY INSTALL (mobx downgrade from 5.x to 4.x)


src/pages/Discussions/PostList/ListRow/index.tsx
```diff
@@ -31,15 +31,14 @@ export default class ListRow extends React.Component<IProps, IState> {
...
-
+  public durationSincePosted(postDate: Date) {
+    const daysSince: number = differenceInDays(new Date(), new Date(postDate))
     return `${daysSince} days`
   }
-  public postViewReactGA(postId: string) {
-    ReactGA.initialize(GOOGLE_ANALYTICS_CONFIG.trackingCode, { debug: true })
-    ReactGA.ga('send', 'pageview', '/discussions/post/' + postId)
-  }
+  // public postViewReactGA(postId: string) {
```
kopecms: commenting out code is not a good practise in my opinion. it's better to leave link to commit in related issue description and delete code because for new comer it does not make any sens

src/pages/Discussions/PostList/index.tsx
```diff
@@ -9,82 +9,67 @@ import ListRow from 'src/pages/Discussions/PostList/ListRow'
...
 
-  public orderListBy(orderType: string) {
-    let sortedList = []
-    switch (orderType) {
+  public orderList(list: IDiscussionPost[]) {
+    // mobx observable must first be sliced in order to sort properly
+    list = list.slice()
+    switch (this.state.orderBy) {
       case 'repliesCount':
-        sortedList = this.state.posts.sort((a, b) => {
```
kopecms: it is a lot of code repetition. you could pass to orderBy _commentCount etc and just access posts attributes by key. if you can substract timestamps it would cool ; ) also I am new to React and I wonder why this attributes seems to be private with underscore?

chrismclarke: yeah we have a spec for a better system, just tidying what was already there but point taken.
As for underscore, that is things that will be populated automatically by the server (not private, but could be made if building a separate class for discussions)

src/pages/Discussions/PostList/ListRow/index.tsx
chrismclarke: agreed, hadn't wanted to remove entirely so that the original author would know where it is (assuming they will be the one doing the refactoring to move it to cloud functions) but will do so and just send a message

src/pages/Discussions/PostList/index.tsx
tx4x: @kopecms most of the render code, inc. the sort functions will all go away soon with my next PR :-)

src/pages/Discussions/PostList/ListRow/index.tsx
tx4x: also this file will be removed by another PR soon

Issue #348, opened 2019-02-24
Avatar showing when not logged in


chrismclarke:
Should display a login button to open modal instead

blocks #347

PR #350, opened 2019-02-24
Module/discussion backend

chrismclarke:
Large update
- create shared store to replicate core functionality between modules more easily
- refactor and build out discussion store
- bind discussion store to list and create pages
- modify discussion create page to better bind to editor as field value
- create placeholder discussion view page and bind to store
- add basic functionality to handle post comments

To do:
- create avatar component that will load a user's avatar given id  (#351)
- fix login component (#348)


chrismclarke:
Apologies in advance, a lot of work required to build discussions back-end in a sensible way and integrate for testing into existing templates. But should have most desired functionality

chrismclarke:
Ok, stopping adding commits now (!)

src/models/discussions.models.tsx
```diff
@@ -1,27 +1,26 @@
...
+  tags: string[]
+}
+
+export interface IDiscussionPost extends IPostFormInput, IDbDoc {
   _lastResponse: firestore.Timestamp | null
   _commentCount: number
   _viewCount: number
   _usefullCount: number
   _last3Comments: any
   // user id only saved for created by
```
kopecms: maybe move or delete this comment?

src/pages/Discussions/index.tsx
```diff
@@ -18,7 +19,7 @@ export class DiscussionsPage extends React.Component<any> {
       <Switch>
         <Route exact path="/discussions" render={props => <PostList />} />
         <Route exact path="/discussions/create" component={PostCreate} />
-        {/* <Route path="/discussions/:slug" render={props => <Post {...props} />} /> */}
+        <Route path="/discussions/:slug" component={PostView} />
 
         {/* <Redirect to="/discussions/list" /> */}
```
kopecms: seems like something this one also can be deleted (I know this does not bring much but I am trying ;p )

src/stores/Discussions/discussions.store.tsx
```diff
@@ -2,64 +2,101 @@ import { observable, action } from 'mobx'
...
+      values,
+    )
+  }
+
+  @action
+  public async deleteDiscussion(discussion: IDiscussionPost) {
+    return Database.deleteDoc(`discussions/${discussion._id}`)
   }
 
   public async saveDiscussion(discussion: IDiscussionPost) {
```
kopecms: this one should also be an @action no?

src/pages/Discussions/index.tsx
chrismclarke: haha, no it's really useful to know somebody has passed an eye over (and that it's not a total mess!). Will remove

src/pages/Discussions/PostView/PostView.tsx
```diff
@@ -0,0 +1,106 @@
+import * as React from 'react'
```
kopecms: also question about naming convention which should be used: why not index.tsx like rest of files in this module?

src/pages/Discussions/PostCreate/index.tsx
```diff
@@ -1,97 +1,81 @@
 import * as React from 'react'
 import { RouteComponentProps } from 'react-router'
-import { Form, Field } from 'react-final-form'
+import { Form, Field, FieldProps, FieldRenderProps } from 'react-final-form'
```
kopecms: not used anywhere ;p

src/stores/Discussions/discussions.store.tsx
```diff
@@ -2,64 +2,101 @@ import { observable, action } from 'mobx'
...
-    } else {
-      this.activeDiscussion = discussion
+  public async createDiscussion(values: IPostFormInput) {
+    console.log('adding discussion', values)
+    const discussion: IDiscussionPost = {
+      ...Database.generateDocMeta('discussions'),
+      _commentCount: 0,
+      _last3Comments: [],
+      _lastResponse: null,
+      _usefullCount: 0,
```
kopecms: spelling is wrong ;( this will remain in db if not fix now

src/pages/Discussions/index.tsx
kopecms: ; )

src/stores/Discussions/discussions.store.tsx
chrismclarke: ah yes, thanks hadn't noticed that one (will only be accessed directly once we have some sort of edit existing functionality...). Actually it probably makes more sense for only the `saveDiscussion` action to be available and that can handle the logic to swap between populating all the meta for a new discussion vs creating a new one (adding now...)

src/pages/Discussions/PostView/PostView.tsx
chrismclarke: I had 4 `index.ts` files open in my editor which I actively needed and it was driving me insane! I think this discussion has gone back and forth a little bit but not fully resolved
(https://github.com/OneArmyWorld/onearmy/issues/32)

Personally I much prefer saving `index.ts` for a hoc only (e.g page entry point) but aware other people will differ in opinion.

src/pages/Discussions/PostCreate/index.tsx
chrismclarke: Thanks, removed the fieldProps, the renderProps are used in the EditorField interface (to know that when binding to the form a number of props include `input.value` are made available)

src/stores/Discussions/discussions.store.tsx
chrismclarke: Thanks, I noticed last week and forgot to change, will do now

src/models/discussions.models.tsx
chrismclarke: Thanks, yeah I wanted to make it clear that we wouldn't want to pass a full user object (a lot harder to keep in sync) and just the userId but will update to more explictly say type :userId and define that as string.

PR #357, opened 2019-02-25
sign up user on form submit

kopecms:
done:
- legacy_id and legacy_registered marked as not required (I do not know if its needed for new users(?))
- creating entry in Authentication and users/<email> on submit

still todo:
- email verification (now user is automaticly logged in after sign up)
- handling error while sign up proccess (delete create entry in Authentication?)
- handle additional fieds in form (avatar, nickname itd..)
- add some tests

### DIFF
--- a/src/mocks/discussions.mock.tsx
+++ b/src/mocks/discussions.mock.tsx
@@ -1,5 +1,4 @@
-import { firestore } from 'firebase/app'
-
+import { toTimestamp } from 'src/utils/helpers'
 import {
   IDiscussionPost,
   IDiscussionComment,
@@ -344,7 +343,3 @@ const commentsTemplate = [
     },
   },
 ]
-
-function toTimestamp(dateString: string) {
-  return firestore.Timestamp.fromDate(new Date(dateString))
-}
--- a/src/mocks/user.mock.tsx
+++ b/src/mocks/user.mock.tsx
@@ -1,11 +1,14 @@
 import { ILegacyUser, IUser } from 'src/models/user.models'
+import { toTimestamp } from 'src/utils/helpers'
 
 export const MOCK_USER: IUser = {
   verified: true,
-  _created: new Date(),
-  _modified: new Date(),
+  _id: '123',
+  _deleted: false,
+  _createdBy: '123',
+  _created: toTimestamp('Friday, January 2, 2015 12:59 AM'),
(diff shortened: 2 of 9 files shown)


chrismclarke:
As for the still to-dos:
1. Firebase has an in-built system for email verification which is super easy to use so can hopefully make this part quite quick. Once up and running it might be worth shooting a quick message over to mattia to make design edits to the email template that is sent out.
https://firebase.googleblog.com/2017/02/email-verification-in-firebase-auth.html

2. I think this is a fairly unlikely case scenario (that connection drops within the milliseconds between). But still, I think what might be easiest to implement and more useful in the future is just a form that can popup if there's more information that we want captured (in fact probably just a profile edit page). The main thing is to have the first auth operation succeed, if we don't capture all the meta we want after then that's no biggie, we can just ask them to add it another time. The caveat here will be to check methods existing on the authstatechange handler (which automatically checks for the user doc on login)

3. This can all be handled on the profile page as mentioned above. Perhaps it would be good to redirect to it straight after Auth success? Technically we don't have to capture any additional info during signup and could leave it to just username and pw, and then once verified prompt user to update their profile? What do you think?

4. This would be useful, although probably best to test the overall flow using e2e (cypress once implemented). As part of this you'd most likely be creating and deleting dummy users, info to do that is here:
https://firebase.google.com/docs/auth/web/manage-users#delete_a_user

kopecms:
Sure I will modify it ; ) Yeah it will be better to just extend edit profile page. I am not sure about authentication without email activation though. Entry in Authorization is created after calling sign up functions and I saw that FirebaseUser has isVerified flag or something but login shouldn't be possible without activation, should it?

kopecms:
I have also deleted login attribute which seems to be redundant

chrismclarke:
Login without confirmation is the default behaviour, so if we don't want to allow users to use without we would need to handle it at the component level (e.g. some sort of `please confirm email` component that stalks them round the site)

chrismclarke:
nice one thanks, I've copied you into the main module thread so hopefully you'll get update on latest mocks etc (mattia thinks should be done at some point this week). I'm assuming you're happy to mostly take the lead on the overall user module (sign up page, backend store etc.), but if you'd prefer to work on anything else just let me know and we can split up the tasks between other devs.

Issue #351, opened 2019-02-24
Create user avatar component


chrismclarke:
Want a simple component that will:
1) Take a userid (email), and look up their avatar from the database
2) Render as a small image

Note (to self), the user avatar might need to be stored in another database location to rest of profile depending on security rules (assume only users will have read/write access to their own doc, although could perhaps try figure out a way to give read permission to certain fields)

kopecms:
About this read/write access to only own doc, don't we want to allow ready other users profile data like display name country (email perhaps)?

chrismclarke:
I think what most of this comes down to is whether users will have any private information saved. At this point I'm assuming mostly no, all the information saved in the user doc is what they choose to put into their public profile, however both the id of the doc and a field within contain the email address the user signed up with which they probably don't want to share.

So either we remove reliance on emails, keep username keys and remove from field, or we have all of the user docs readable only by the user and add a separate api endpoint on the server that can get things like avatars for a given user.

chrismclarke:
The reason any of this matters is because documents retrieved from the database are cached offline to local storage (indexedDB), so a user could easily use any of that data even if not explicitly exposed within the interface, e.g.
![image](https://user-images.githubusercontent.com/10515065/53571833-bc634680-3b61-11e9-99d7-4022ec3447d4.png)

kopecms:
Maybe deleting email address from users doc is not that bad idea (we can add private data doc later when needed). Displaying discussion or any other entity anyway require downloading users public profile. So maybe I will replace email with id generated when creating Authentication profile? And add avatar and other public data to /users doc?

chrismclarke:
I think that probably makes the most sense at this point. I'm guessing we could also save the user doc using that same id.

BenGamma:
created new related issue #399 for using user id instead of email since Avatar component is now created and merge

Issue #216, opened 2019-01-14
Module - Tags and Filters


mattia-io:

mattia-io:
**Updated 14-01-19 (Mattia)**

**Overview:**
The filter bar is used across the community site to allow users to filter content on the following pages: How-To, Discussions, Events and Maps. 

The filter bar includes the Project selection dropdown, the category and tag dropdown. It should also include search. 

**Latest design resources:**
Prototype
https://sketch.cloud/s/KnZgA/m1O53gL/play

**Originals + Screens**
(coming soon)

**Epics:**
https://docs.google.com/spreadsheets/d/1pkLRKCbQiJOtQwWEhVNgSTvDWf5SnVAz10vMo4k-LNg/edit#gid=0

Any questions post in the comments

BenGamma:
This is on hold for now. Will reopen if appropriate in a close futur.

PR #358, opened 2019-02-25
Minimal implementation with devextreme-reactive/react

tx4x:
@TODOS: 
- apply filter system
- create missing pages when clicking on certain list columns as avatar, comments, tags
- search
- integrate server paging
- default avatar

### DIFF
--- a/src/pages/Discussions/PostList/ListRow/elements.js
+++ b/src/pages/Discussions/PostList/ListRow/elements.js
@@ -56,7 +56,6 @@ export const Tag = styled.span`
 
 export const InteractionNb = styled.span`
   font-size: 0.8em;
-  margin-left: 20px;
   width: 10%;
   display: inline-block;
   vertical-align: middle;
@@ -70,10 +69,9 @@ export const ViewCount = styled.span`
   vertical-align: middle;
 `
 
-export const UsefulCount = styled.span`
+export const UsefullCount = styled.span`
   font-size: 0.8em;
   padding: 20px;
-  margin-left: 30px;
   display: inline-block;
   vertical-align: middle;
   border: 1px solid black;
--- a/package.json
+++ b/package.json
@@ -3,6 +3,10 @@
   "version": "0.1.6",
   "private": true,
   "dependencies": {
+    "@devexpress/dx-grid-core": "^1.10.1",
+    "@devexpress/dx-react-core": "^1.10.1",
+    "@devexpress/dx-react-grid": "^1.10.1",
+    "@devexpress/dx-react-grid-material-ui": "^1.10.1",
     "@babel/core": "7.2.2",
     "@ephox/sand": "^2.0.11",
     "@material-ui/core": "3.7.1",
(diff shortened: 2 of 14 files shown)


tx4x:
image

chrismclarke:
Great thanks, there's quite a bit to look at here so I'll sit down with it properly this evening

src/pages/Discussions/PostList/ListRow/elements.js
```diff
@@ -70,10 +69,9 @@ export const ViewCount = styled.span`
   vertical-align: middle;
 `
 
-export const UsefulCount = styled.span`
+export const UsefullCount = styled.span`
```
kopecms: isn't 'useful' the correct version? ;p

tx4x: yeah, will correct and notify here

chrismclarke: I think there's gonna be a few changes since latest merge so no need to worry to much, I'll pick them up on review and merge

tx4x: right, please also ignore Discussions/common/filter : it's not used, yet. my bad :-)
g

chrismclarke: no probs... taking a look now

chrismclarke:
Additional comments
* apply filter system
- would be great to have in, although might have to hold off until we get the tags system in, which would be the basis for most simple filtering. I'm not sure the current status of the spec, I think a little bit of organisation is probably needed on my side to bring together existing mocks with thoughts people have, if you want to be part of the discussion and help push it forward then #216 is a sensible place for it

* create missing pages when clicking on certain list columns as avatar, comments, tags
I'm assuming we can keep implementation pretty simple to begin with and all columns just go to the discussion view page, and then can think later about extra interactions that might be useful based on feedback

* search
This should be linked to algolia which I think you planned to investigate. It might make sense to first implement on the how-to page as that is closer to being shipped out and would avoid lots of people working on the same module. Of course we should view the implementation as portable across modules (really filter/search are a module in their own right, cutting across the site)

* integrate server paging
This might not actually be necessary. I've been exploring a couple different approaches and right now have opted for a system where all data is pulled from the server, cached, and then just updated when changes are made (kept in sync with live db rather than constantly querying). What this costs in terms of population of potentially unwanted info (e.g. loading a minimum amount of information about all discussions, not just the first 10 or so) I think will actually be saved from repeatedly querying what is essentially the same data. So for now I think we can assume the pagination you've put in is sufficient, and can revisit once we have a bit more data to consider

* default avatar
I've created an issue #351 which will be a separate avatar component that can be used (so doesn't necessarily need to be coupled with work on this module). The plan is for user ids to be persisted to the discussion/comment/etc. and then the avatar url will be fetched from the database (makes handling user profile changes easier as otherwise need to find a way to trace all documents with references to the avatar that needs updating). 

In addition I might take a quick look at the interfaces as currently we have a few different ones floating around which could likely be brought together (e.g.  IPostInfos and IDiscussionPost). At the same time want to get rid of some of the specific references to 'discussions' and just refer to more general post and comment interfaces throughout most. 

Thanks for all your work on this though, I'll add a reference to this thread in the main module and we can handle follow-up in separate issues

PR #364, opened 2019-02-26
Confirmation and reset password emails

kopecms:
still todo: 
- cutom pages/handling for bouth actions
- checking if account verified before login


kopecms:
so with this commit accounts should have verfied email address. I have problem with utilising verified field from IUser which means that every user need entry in Authentication on firebase (which is not a case right now).

chrismclarke:
cool, I'm pretty swamped today but will try take a look either tomorrow or thursday

PR #371, opened 2019-03-01
show link to email confirmation view when one is not verified

kopecms:


kopecms:
I will do it different way. Closing for now

Issue #20, opened 2018-09-28
Create user registration component


chrismclarke:
Should have place to specify required info: email, pw (+confirmation). Additional info likely also such as first name, last name (see what is currently in Dave Hakkens signup). This component should also double as profile update, providing option to either register or update info.
child of #8

chrismclarke:
note, will also want to populate db with user meta

kopecms:
I would love to start proggress on it ; ) @mattia-io @chrismclarke ok?

chrismclarke:
That would be awesome! I'm not sure if the full mockup has been made for this yet but will check and get back to you. If not you could still start building some of the core functionality. I'd suggest using react-final-form for form data handling as that is what is already in the project

chrismclarke:
Another good place to start work might be a profile picture uploader/editor. We have a couple different implementations of file uploads which could work totally fine (see the how-tos), or if you wanted a richer experience you could explore options for a simple image editor (e.g. cropping). Totally optional though, will need to check with @mattia-io whether this would be a core feature wanted on the platform (I know he's away for the next couple days but should be back by early next week)

kopecms:
Should sing up form be on separete page? Is there way to access firebase database localy (I haven't look much into this technology yet)? And what about workflow: should PR be full feature closing issue or issue can have some PRs?

chrismclarke:
I  think to streamline development and testing it probably makes sense to have a separate page, we could always move later if needed.

For the dev site firebase is setup and ready to work with, you can see how some pieces of work interact directly with it via import of the `utils/firebase` helper, although I'm slowly trying to migrate to use a middle layer `Database` object which data stores will interact directly with. There is already a user.store which you could build on.

If confused the difference between LegacyUser and User models, that is because we will be migrating user logins and profiles from the old site. This requires a user to attempt login first using their wordpress username and password. This is then cross-checked against the legacy user database collection (which stores an encrypted version of their password, so passes the pw input through the same encryption protocol to see if output matches as a confirmation that this person is the owner of the old account)

As it's a fairly isolated piece of work feel free to handle PR however best works for you. We make an active effort to review all PRs each Tuesday (and whenever else we can find time). This is probably easier if the PRs aren't huge, but again we wouldn't want to be a bottleneck for your work so just use your judgement. If there are multiple PRs a naming system like feat/user-registration, feat/user-registration-some-part would probably make our lives a bit easier too

chrismclarke:
[...]

PR #373, opened 2019-03-03
Login flow

### LINKED ISSUE #22: Email verification
Firebase should allow for email verification, but will want login component to reflect what stage of the process they are in (possibly giving option to resend confirmation or change info provided)
child of #8
depends on #19

### LINKED ISSUE #370: Authenticated user not shown as logged in
child of #8
I log in successfully however the sign in modal doesn't dismiss and I can still click the button to log in. I'm assuming this is because my email isn't verified? Would be good to have a clearer indication that is the case and mechanism to resend verification email.

image

(screenshot, successful login/auth user state change but modal and log in button still present)

kopecms:
- using layouts defined in components/Layout/
- displaying error message and resend confirmation email button when account not verified
- display success message after submitting sign up form
- creating custom page for handling activation and password reset (when merged change for url in firebase templates is required)
- using same form styles inside login modal (so login form and reset password form look the same)

close #370, close #22
relates to #8, #20


src/pages/common/User/Detail.tsx
```diff
@@ -2,34 +2,43 @@ import * as React from 'react'
 import { IUser } from 'src/models/user.models'
 import { Label } from 'src/components/Form/Fields'
 import AvatarPic from '@material-ui/core/Avatar'
+import { FlexContainer } from 'src/components/Layout/FlexContainer'
+import { space } from 'src/themes/styled.theme'
+import styled from 'styled-components'
+
+const FlexDetail = styled(FlexContainer)`
```
BenGamma: Here instead of creating a const and importing `space` and `styled` you can directly add the styles in props of FlexContainer like so :
```js
<FlexContainer p={0} justifyContent={'space-between'} alignItems={'center'}>
```
`p={0}` will automatically refer to the space[0] value in the theme.

src/pages/common/Login/LoginForm.tsx
```diff
@@ -158,41 +125,41 @@ export class LoginFormComponent extends React.Component<IProps> {
               {this.state.showResendConfirmationButton ? (
                 <Button
                   onClick={this.resendConfirmation}
-                  fullWidth
-                  variant="contained"
-                  color="primary"
-                  style={styles.submit}
+                  width="100%"
```
BenGamma: you can use a ratio value here like `width={1}`
https://styled-system.com/api/#width
will be useful for responsive style when implementing later on

kopecms: done

src/pages/common/User/Detail.tsx
kopecms: done

BenGamma:
This looks good to me, but since @chrismclarke reviewed & merged the first part of this login flow, I'll wait for his review before merging with master.
Thanks for the work @kopecms !

PR #383, opened 2019-03-06
Adding storybook as component explorer for design system

rudolphfunk:
…d button component to rebass, added some documentation for theme


BenGamma:
implemented with #388 thanks for the good work @rudolphfunk !

PR #384, opened 2019-03-07
Module profile

kopecms:
Part is the same as in Login Flow (#373) so only last 4 commits are new (I didn't want it to be the same PR as #373 since this one propose not discussed changes into storing users so probably can be dismissed)
done:
- use Authenicated user profile uid as key in /users doc -> as @chrismclarke mentioned in #351 we would probably like to keep separately email address and other public profile data which is not posible with email as a key for this doc
- create current user and user profile components


chrismclarke:
Thanks, I will aim to review this by Tuesday

chrismclarke:
Apologies, reached the end of the day without finding a spare moment but will put it as a top priority for tomorrow

PR #385, opened 2019-03-07
351 create avatar component

BenGamma:
I created an Avatar component with the prop `userEmail`. The prop is not mandatory so if you just use the avatar component without prop, it will display a placeholder user Icon. In a close futur this placeholder should be a proper image.
As a quick solution I added an `avatar` field in the users collection in firebase with a link to our github avatar.
I added the field by hand for chris and I, there is probably a better way to do it.

TODO :
* automatically add the field for new created user.
* add the avatar to the `/profile` section with the ability to modify it
* decide where to save the avatar. To begin I think that firebase storage would be ok, but it's open for discussion

### DIFF
--- a/src/components/Icons/index.tsx
+++ b/src/components/Icons/index.tsx
@@ -9,6 +9,7 @@ import {
   MdCheck,
   MdArrowBack,
   MdKeyboardArrowDown,
+  MdAccountCircle,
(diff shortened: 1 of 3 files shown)


BenGamma:
This PR closes #351

Issue #359, opened 2019-02-25
Add backend function to populate total discussion responses when comments added and deleted


chrismclarke:
child of #213
Currently when a user creates a post it contains a `resonses:` field which will track the number of responses. 

When responses are created they are added to a subcollection (e.g. `discussions/${discussion._id}/comments/${comment._id}`), and so the original document has no way of knowing how many responses there are without running a query to get them all (which is inefficient/expensive, as the full subcollection documents aren't required on a list view page).

So the usual solution is to add a cloud function, which will respond to a trigger when the subcollection doc is created/deleted to increment or decrement the parent doc field

some examples which might help get the idea:
https://angularfirebase.com/lessons/firestore-cloud-functions-data-aggregation/
https://firebase.google.com/docs/firestore/solutions/aggregation
https://firebase.google.com/docs/firestore/solutions/counters
(although definitely won't need full distribution counter, single counter will be fine)

chupakabra:
I think I will focus on that this week
@chrismclarke, @mattia-io

mattia-io:
NIce one @chupakabra 🙌

PR #390, opened 2019-03-09
#359 func total discussion responses

chupakabra:


chrismclarke:
Thanks, I will aim to review this by Tuesday

chrismclarke:
Apologies, reached the end of the day without finding a spare moment but will put it as a top priority for tomorrow

chrismclarke:
Note, I've added a placeholder `syncCommentsCount` function for now so that the master build still works (otherwise, because it does not appear in the master branch yet it asks firebase whether it should be removed from the server on each build, and the build fails because you cannot type Y/N during the CI process)

chrismclarke:
actually I've just merged it in for now as adding a placeholder was just awkward and we will still want to keep this code to verify (and can add another bit of code to increment the counter)

chupakabra:
@chrismclarke, sorry, read your comments with some delay

BenGamma:
is this closing #232 ?

chrismclarke:
Sorry for the slow response. I don't think the change was intentional (possibly just whilst testing), no I think you're right that it makes more sense to call from trigger. yes I think simple increment/decrement function would be ideal for the near-future

PR #408, opened 2019-03-15
fix button component

kopecms:
I can not see any difference without defaultProps but its presence cause Button component to be a div instead of button and breaking submitting form

### DIFF
--- a/src/components/Button/elements.js
+++ b/src/components/Button/elements.js
@@ -26,6 +26,12 @@ export const StyledButton = styled(BaseButton)`
   ${baseStyles}
 `
 
+StyledButton.defaultProps = {
+  as: 'button',
+  className: 'button',
+  variant: 'primary',
+}
+
 export const Label = styled.span`
   display: block;
   flex: 0 0 auto;
--- a/src/components/Button/elements.js
+++ b/src/components/Button/elements.js
@@ -15,7 +15,6 @@ const baseStyles = css`
   justify-content: center;
   font-weight: 600;
   white-space: nowrap;
-  width: fit-content;
   word-break: keep-all;
   cursor: ${props => (props.disabled ? 'not-allowed' : 'pointer')};
 `
--- a/src/components/Button/elements.js
+++ b/src/components/Button/elements.js
@@ -27,11 +27,6 @@ export const StyledButton = styled(BaseButton)`
   ${baseStyles}
 `
 
-StyledButton.defaultProps = {
-  className: 'button',
-  variant: 'primary',
-}
-
 export const Label = styled.span`
   display: block;
   flex: 0 0 auto;
--- a/src/components/Button/elements.js
+++ b/src/components/Button/elements.js
@@ -1,6 +1,4 @@
-import React from 'react'
 import styled, { css } from 'styled-components'
-import { variant, color, space, width } from 'styled-system'
 import { Button as RebassButton } from 'rebass'
 
 const baseStyles = css`
@@ -19,9 +17,6 @@ const baseStyles = css`
   word-break: keep-all;
   cursor: ${props => (props.disabled ? 'not-allowed' : 'pointer')};
 `
-const colorsVariant = variant({
-  key: 'buttons',
-})
 
 export const BaseButton = styled(RebassButton)`
   ${baseStyles}
--- a/src/components/Button/elements.js
+++ b/src/components/Button/elements.js
@@ -1,6 +1,7 @@
+import React from 'react'
 import styled, { css } from 'styled-components'
 import { variant, color, space, width } from 'styled-system'
-import { Button as BaseButton } from 'rebass'
+import { Button as RebassButton } from 'rebass'
 
 const baseStyles = css`
   text-transform: uppercase;
@@ -22,15 +23,9 @@ const colorsVariant = variant({
   key: 'buttons',
 })
 
-export const StyledButton = styled(BaseButton)`
+export const BaseButton = styled(RebassButton)`
   ${baseStyles}
 `
-
-StyledButton.defaultProps = {
-  as: 'button',
-  className: 'button',
-  variant: 'primary',
-}
 
 export const Label = styled.span`
   display: block;
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -18,11 +18,6 @@ const StyledButton = ({ children, ...props }) => (
   <BaseButton {...props}>{children}</BaseButton>
 )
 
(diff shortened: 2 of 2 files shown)


src/components/Button/elements.js
```diff
@@ -27,11 +27,6 @@ export const StyledButton = styled(BaseButton)`
   ${baseStyles}
 `
 
-StyledButton.defaultProps = {
-  className: 'button',
-  variant: 'primary',
-}
```
BenGamma: That's a strange one, probably because we are not applying the default props at the right place.

Can you then add this defaultProps to the exported Button in `src/components/Button/index.tsx` ?

kopecms: done ;)

rudolphfunk:
I think the button to div was introduced when I moved the button to extend rebass button. See https://github.com/OneArmyWorld/onearmy/pull/383/commits/c540b4fa9b48a017a2c90772027f2adff47b4e23. My bad, sorry.

src/components/Button/elements.js
rudolphfunk: `<Buttons />` in [Rebass](https://github.com/rebassjs/rebass) are all extended from the Box which is a `<div/>`

It looks like if you add the `as` key to default props you can change the element type:

```
StyledButton.defaultProps = {
  ..
  as: 'button'
  ..
}
```

See: https://github.com/rebassjs/rebass/blob/32b6a7dbc469b6d60862e5ab38b9a0ac85fa6279/src/index.js#L137

Tested locally and seems to work.

src/components/Button/elements.js
```diff
@@ -15,7 +15,6 @@ const baseStyles = css`
   justify-content: center;
   font-weight: 600;
   white-space: nowrap;
-  width: fit-content;
```
kopecms: still unable to pass width when this is here

src/components/Button/index.tsx
```diff
@@ -24,7 +24,7 @@ type BtnProps = IBtnProps &
   disabledProp
 
 export const Button = (props: BtnProps) => (
-  <StyledButton {...props} px={3}>
+  <StyledButton px={3} {...props}>
```
kopecms: I think this way it is possible to override px

src/components/Button/elements.js
kopecms: oo cool thanks ; )

src/components/Button/index.tsx
martisj: This makes sense 👍

src/components/Button/elements.js
```diff
@@ -28,6 +27,7 @@ export const StyledButton = styled(BaseButton)`
 `
 
 StyledButton.defaultProps = {
+  as: 'button',
```
martisj: This looks to me to be a bit of a "hacky" solution 🤔 I happened upon this issue in the rebass https://github.com/rebassjs/rebass/issues/574

Following that example, this seems to work on my preliminary test:
```jsx
import { Button as RebassButton } from 'rebass'
const BaseButton = styled(RebassButton)`
  ${baseStyles}
`
export const StyledButton = ({ children, ...props }) => (
  <BaseButton {...props}>{children}</BaseButton>
)
```

src/components/Button/elements.js
martisj: @rudolphfunk doesn't it seem a bit weird that we have to set this when we are extending that component in the first place?

src/components/Button/elements.js
rudolphfunk: @martisj I agree, this seems to be the way to go. Changing the `as` key in the `defaultProps` does seem a bit strange. The Rebass [documentation](https://rebassjs.org/getting-started#extending) on extending doesn't specify that it is necessary to wrap the component in a Stateless Functional Component. Doesn't ESLINT complain about spreading the props in the function argument?

src/components/Button/elements.js
rudolphfunk: After further investigation it seems that setting `defaultProps` on a [Rebass](https://rebassjs.org) component is what causes the element to render as a `<div/>` instead of a `<button />`

kopecms: so what should I do?

src/components/Button/elements.js
martisj: I converted it to typescript so no complaints from Eslint 😂

src/components/Button/elements.js
martisj: What about putting defaultprops on the `Button` component that extends `StyledButton`?

kopecms:
```
<StyledButton variant="primary" className="button" px={3} {...props}>
    {props.icon && <Icon glyph={props.icon} />}
    <Label>{props.children}</Label>
  </StyledButton>
``` like this?

File: src/components/Button/index.tsx [hunk 5]
```diff
@@ -24,7 +28,7 @@ type BtnProps = IBtnProps &
   disabledProp
 
 export const Button = (props: BtnProps) => (
-  <StyledButton {...props} px={3}>
+  <StyledButton variant="primary" className="button" px={3} {...props}>
     {props.icon && <Icon glyph={props.icon} />}
     <Label>{props.children}</Label>
   </StyledButton>
```
rudolphfunk: @kopecms - please as above leave `defaultProps` in `src/components/Button/elements.js` deleted. And add `defaultProps` to `src/components/Button/index.tsx`. Then  `<Button />` that wraps `<StyledButton/>` should render a `<button/>` element. 

For example:

```Javascript
/*src/components/Button/index.tsx line 29 */
export const Button = (props: BtnProps) => (
  <StyledButton {...props} px={3}>
    {props.icon && <Icon glyph={props.icon} />}
    <Label>{props.children}</Label>
  </StyledButton>
)

/* Add defaultProps on <Button/> instead of <StyledButton> */
Button.defaultProps = {
  className: 'button',
  variant: 'primary',
}
```

kopecms: done

BenGamma: Here I remove the px={3} props, as it is a default props of rebass button [see here](https://github.com/rebassjs/rebass/blob/32b6a7dbc469b6d60862e5ab38b9a0ac85fa6279/src/index.js#L136)

kopecms: so delete it or move it back to place where it was?

BenGamma:
I just created #417 to implement all these changes and add theme values, thank you all for your contributions on this !

Issue #266, opened 2019-02-12
Module - Automated Tests


chrismclarke:
Currently our system of testing relies on confirmation of successful build during CI pipeline process, use of typescript models to enforce sensible data structures, code reviews on PRs, manual quick testing and general common sense. 

Whilst this is reasonable for small projects and few contributors we're quickly moving beyond what is time-efficient or reasonably robust. I would suggest in the first instance we create a few simple e2e tests to make sure core platform features work as expected, and provide a basis to build on as more things crop up.

Open to suggestions on favourite tooling to do so, or anything else we might want to implement

tx4x:
hey, the most leg work is done [here](https://github.com/tx4x/test-stack/); basically a combination of puppeteer and cucumber.

in short, 
1. you write all the cucumber step definitions like [this](https://github.com/tx4x/test-stack/blob/master/test-suite/step-definitions/recorder/recorder.step.ts) 
2. you reference the steps as [this](https://github.com/tx4x/test-stack/blob/master/test-suite/features/record.feature), in any order you like, and also pass parameters to it.

this runs in the CI as well, the stuff outside of './test-suite' is just fancy CLI tooling to fire up sample page

currently there is code in place to measure/intercept network requests as well, so you can add here your KPI for app size and what not. I still have somewhere a full chrome event log parser to analyze other things like LongTask-Reports, rage clicks but I left this to google lighthouse later on.

let me know, i can adjust this quickly for the current dev ops, it's BSD-3 so do whatever you like :-)

A typical cucumber test would look as follows:

Background:
**Given** I am connected to the one-army setup at 'your url here'

@debug
**Scenario**: User can not create poll
**Given** a user with less than '100' points
**Then** the poll button is not shown

I love it however, you can pipe this reports than also into automatic issues

chrismclarke:
Cool! Yeah I'm pretty fixed personally on puppeteer, I've used it with a few other projects and really like it (and am not aware of any widely used alternatives for e2e)

I know react comes with jest, I've also used Mocha (+chai) and a bit of Jasmine before which I had no issue with. I can see cucumber is pretty popular too but haven't ever used. I really like the look of what you've described.

My gut feeling at the moment is:
**Unit** - Jest + Enzyme (although not implemented for all components)
**E2E** - Puppeteer + cucumber

We could also push the boat out a little with the like of:
**Lighthouse** - we could link into their cli tools to run automated, this would be pretty straightforward and could be quite cool as an additional performance metric
[...]

PR #411, opened 2019-03-15
Module/266 automated tests static analysis

rudolphfunk:
## Background
In response to #266, starting implementation of Automated Testing strategy inline with the 'Testing trophy':

testing trophy

Static Analysis is the foundation. Unit tests are fast and cheap. Integration tests are the best balance of confidence and speed.

## Reasoning
- Linting makes more sense when run before committing your code. By doing so you can ensure no errors go into the repository and enforce code style. [Prettier](https://prettier.io/) config file is already in project, but it is not a dependency. [eslint](https://eslint.org/) and [tslint] (https://palantir.github.io/tslint/) are dependencies and configured. These relying on contributors having their code editors configured correctly to use these formatter and linters, which can be unpredictable and can result in bad habits creeping into the project.
- running a lint process on a whole project is slow and linting results can be irrelevant. Ultimately you only want to lint files that are to be committed. Therefore [lint-staged](https://github.com/okonet/lint-staged) has been introduced to the project. [Husky](https://github.com/typicode/husky) handles the pre-commit hook.
- defaultSeverity has been raised to error in tslint.json to prevent code that does not pass linter from being introduced to the project

## Next steps: 
- configure [Jest](http://jestjs.io) and [react-testing-library](https://github.com/kentcdodds/react-testing-library)
- investigate the feasibility of adding [Cypress](https://www.cypress.io/) to the project

### DIFF
--- a/.huskyrc
+++ b/.huskyrc
@@ -1,0 +1,5 @@
+{
+  "hooks": {
+    "pre-commit": "lint-staged"
+  }
+}
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,0 +1,14 @@
+{
+  "linters": {
+    "*.{ts,tsx}": [
+      "tslint --project tsconfig.json --config tslint.json"
+    ],
+    "*.+(js|jsx)": [
+      "eslint"
+    ],
+    "*.+(js|jsx|json|yml|yaml|css|less|scss|ts|tsx|md|graphql|mdx)": [
+      "prettier --write",
+      "git add"
+    ]
+  }
+}
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -3,7 +3,7 @@
     "*.{ts,tsx}": [
       "tslint --project tsconfig.json --config tslint.json"
     ],
-    "*.+(js|jsx)": [
+    "*.{js,jsx}": [
       "eslint"
     ],
     "*.+(js|jsx|json|yml|yaml|css|less|scss|ts|tsx|md|graphql|mdx)": [
--- a/.storybook/wInfo.ts
+++ b/.storybook/wInfo.ts
@@ -1,11 +1,12 @@
-import { withInfo } from '@storybook/addon-info'
+import { withInfo } from '@storybook/addon-info';
 
-const wInfoStyle = {}
+const wInfoStyle = {};
 
-export const wInfo = (text?: string) =>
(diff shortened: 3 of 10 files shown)


BenGamma:
Thanks for this PR @rudolphfunk, tests are something we want to implement since quite some times now !
Could you add a bit more informations on how the test would consist ?
Do we need to run a specific command ?
For now these commits seems related only to pre commit linting right ?

rudolphfunk:
@BenGamma this PR only covers static code analysis running [TSLINT](https://palantir.github.io/tslint/) and [ESLINT](https://eslint.org/) and [Prettier](https://prettier.io/). Contributors should follow the usual workflow:

1.  Make their changes in their working copy (local repo)
2. Stage these changes once ready to commit `git add <path-to-file>/<filename>`
3. Commit their changes `git commit -m '<commitMessage>'`, upon which the precommit hook will trigger `lint-staged` to run the linters and formatter against the staged files.

The advantage of running these checks locally is that CI does not need to catch them, therefore shortening the feedback loop for developers.

> Could you add a bit more informations on how the test would consist ?
I think here you are referring to the unit or integration tests. I will get to these later and of course provide more guidance in the docs

> Do we need to run a specific command
No ( see steps above)

> For now these commits seems related only to pre commit linting right ?
Right you are

rudolphfunk:
Reopening PR

chrismclarke:
Hey, massively appreciate the effort and clear communication.

Could you rollback the default severity? This was quite a pain point in discussions a couple months back, summarised here: https://github.com/OneArmyWorld/onearmy/wiki/Typescript

chrismclarke:
Also happy to see you're also looking at cypress, we had some discussions more widely here: https://github.com/OneArmyWorld/onearmy/issues/266

rudolphfunk:
@chrismclarke, I have made the requested changes:

> In terms of linting I like the approach and agree, with request to rollback default severity.

Done

> I see there's also lots of changes to icons etc., please could you clarify, are these just to enforce the conventions specified? (i.e. no breaking changes)

This was prettier re-formatting my code. I had prettier disabled in my editor (settings from another project) and therefore the files were not formatted properly. I have reverted these changes to keep the PR clean. These files will get cleaned up in the future when changes are made and formatting rules are enforced.

chrismclarke:
All sounds and looks great, thanks again (merged)

PR #422, opened 2019-03-24
Fix typo

martisj:


BenGamma:
This looks like a duplicate from #423 or should I consider it ?

martisj:
Ah I know, sorry it looks like a duplicate, it is not a duplicate. I can combine them to make it easier, would that help?

BenGamma:
Yes would be great, thanks

martisj:
Putting this fix in #423

Issue #448, opened 2019-04-21
[enhancement] - Update user avatar methods


chrismclarke:
**Is your feature request related to a problem? Please describe.**
During user module rewrite most avatar-related methods became deprecated and need to be put back in (to match improved workflows and conventions).  Will try get it done this week.

BenGamma:
@chrismclarke is this issue still to fix ? There was some work done on avatar component in #782, maybe that fixed the issue here ?

chrismclarke:
Given that we're no longer using profile avatars (instead member pins) I'm guessing it's safe to close (can always revisit if we decide to put back in)

Issue #447, opened 2019-04-21
[nit] - replace instances of material form with react


chrismclarke:
**Is your feature request related to a problem? Please describe.**
When we remove material ui we will also lose form elements that had been used in some form binding (e.g. `FormControl` elements). Their child input elements behave differently to react-final-form `Field` elements (onChange returns a value instead of object). To save confusion it would be good to go and check these instances and put in consistent way.

chrismclarke:
A bunch of work done on this previously, but still a few outstanding things to go (some legacy code for removal, so existing code pending update) so will keep open. Removing myself as assignee and replacing with @BenGamma who has been working on this

chrismclarke:
Update - still a handful of stuff there so will keep open

Issue #445, opened 2019-04-21
[bug] - legacy users requires upgrade


chrismclarke:
**Describe the bug**
With new changes to auth system all legacy users will be required to have a username field on their profile which is linked to auth display name. Should look into temp system as a way to ease migration for existing users (alt would be just to have everyone create new accounts). Could be linked to future system for existing user to change their username.

chrismclarke:
Assume users using new sign-up, or old issues handled case by case. Closing

Issue #443, opened 2019-04-19
[optimisation] - reduce number of server queries for user profiles


chrismclarke:
**Is your feature request related to a problem? Please describe.**
When displaying user content individual db requests are made to the server to get info about the author (e.g. user name and avatar url). This can get quite heavy if viewing lots of pages

**Describe the solution you'd like**
First make any requests from cache and see if user profile already known. Run extra request after to see if new data available (could try search for users where id=xyz and modified>last known update date)

**Describe alternatives you've considered**
Leave it as is, shouldn't be an obvious issue until user base grows substantially. Possible use of alternate dbs to store this info for quick lookup (excessive and still request-heavy), embedding directly into content (hard to keep up to date)

chrismclarke:
Ok, think I've figured out best option for now is:

1. Store all users in the database with their username (/@mention-name) as the key. This means no db lookup required to give author of content (stick with mention-name instead of fullname/nickname etc)

2. Store all user avatars in a single storage folder using userName as an identifier (e.g. avatars/chris-m-clarke.jpg or avatars/chris-m-clarke.thumb.jpg)

This effectively bypasses all requirements to make db requests when providing initial data (could run query on full document open for example, in which case may want to consider caching suggested above). But for now this should be sufficient so will close when PR made from `user-profiles` branch (where currently implemented)

PR #449, opened 2019-04-21
Module/user profiles

chrismclarke:
Substantial overhaul of user profiles (sorry! lots of tightly and loosely coupled components and schemas). Suggest any review take just small part of what is included and see if works as hoped

Huge thanks to @kopecms for all the early code that formed a basis for user registration and profile forms. 

**Included**
- User sign up/login forms and profile page with edit and change pw
- Change to how user ids are stored and used by db (see #443), essentially now all users are stored by their userID within the database to make it possible to access @mention and avatars without db requests
- Overhaul of user migration methods (pulling data from DaveHakkens.nl) - note testing will show 'internal error' when not on live due to localhost cors limitations, now automatically picks user ids from the site on a daily basis and can pull user profile data from the site on request.
- Tidying/improving workflow for backend functions
- Minor bug fixes (e.g. console import, functions linting, extra style variables etc.)

**Todo**
- Add avatar import, change and upload methods (including server-side functions to compress and make thumb copies) #448
- Add automated testing for user workflows
- Add ability to delete a user account (will definitely be required when automated testing)

**Known issues**
- Registration and login forms use m-ui components which will need to be changed (#447)
- Legacy users won't be in correct format (#445)


BenGamma:
Looks great thanks !
I'm just a little confused about the userflow on the profile page, I didn't understand all the fields/buttons. We'll clarify this week, need to separate the fields / features for release v0.3.0 (which is a simple implementation of the login) and the extra ones (mentions, importing data from davehakkens.nl, avatar).

Issue #459, opened 2019-04-25
[feature request] DIsplay how-to title in breadcrumb


BenGamma:
**Describe the solution you'd like**
Display the how-to title in the breadcrumb like so `How-to > How-to-title`

PR #515, opened 2019-06-26
#459 Add breadcrumb for how-to page

kopecms:
Just want some feedback. If it is ok I would add breadcrumb for all pages and also links for higher levels (?)


davehakkens:
Sounds good! However at the moment only really makes sense on the how-to pages, since they go a layer deeper.

kopecms:
Ok so: breadcrumb for how-to and discussions (I removed displaying a description defined in PageList.tsx), Link to higher level.
Closes #459

davehakkens:
cool. we review pull requests on Tuesday, also don't forget to check the monthly news 👊

BenGamma:
And dave is talking about the monthly news because we mentionned you in it 🤗

PR #516, opened 2019-07-01
Feedback in header

davehakkens:
My first attempt in react. Probably shitty. Sorry for that. Wanted to test pull requests 🤙
I've added a url for the feedback form in the top header

### DIFF
--- a/src/pages/common/Header/CommunityHeader/CommunityHeader.tsx
+++ b/src/pages/common/Header/CommunityHeader/CommunityHeader.tsx
@@ -82,12 +82,36 @@ export class CommunityHeader extends React.Component<IProps, IState> {
     const user = this.injected.userStore.user
     return (
       <div>
-        <TopLink target="_blank" width={1} href="https://build.onearmy.world">
-          <Text bg="blue" color="white" py={2} medium>
-            This is the alpha version of onearmy platform, click here to help
-            building it.
-          </Text>
-        </TopLink>
+        <Text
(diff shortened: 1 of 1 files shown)


src/pages/common/Header/CommunityHeader/CommunityHeader.tsx
```diff
@@ -82,12 +82,36 @@ export class CommunityHeader extends React.Component<IProps, IState> {
...
-          </Text>
-        </TopLink>
+        <Text
+          style={{ textAlign: 'center' }}
+          bg="blue"
+          color="white"
+          py={2}
+          medium
+        >
+          This is the alpha version of our new community platform, click here to
```
nzchicken: I would maybe add a `&nbsp;` at the end of the line here, and remove the `{' '}` below. Reason is that you don't want the space character to be part of the link, and also `{' '}` is a little hacky 👍

There is another way to also achieve this using styled components by wrapping the TopLink and then you can apply some styling there instead. e.g.

```typescript
const LinkWithEmphasis = styled(props => <TopLink {...props} />)`
  text-decoration: underline;
  color: red;
  padding: 0px 5px;
  
  &:hover {
    color: blue;
  }
`
```

And then you just use it in place like so:
```
<LinkWithEmphasis
  target="_blank"
  href="https://build.onearmy.world"
/>
  help code it.
</LinkWithEmphasis>
```

Not sure how this fits in with the coding guidelines though :-D

Issue #144, opened 2018-11-23
Module - Map


chrismclarke:
## Update 28-08-19 (Chris)

V1 merged, known todo:

**Frontend Dev**
- [x] Take My Pin button somewhere
- [ ] Debug weird problem where map resets to the center for some odd reason
- [x] Add additional fields for cover/hero image in user profile
- [ ] Map pin preview within profile settings edit page
- [ ] Limit text shown in user popup preview
- [ ] Fix popup open inconsistency (e.g. change page, return, try open)

**Backend Dev**
- [x] Decide on sensible DB strategy (#518 )
- [x] Hook up data from firebase (map and pins + filters)
- [x] Add sync methods for pins to realtimedb
- [ ] Update user lastActive times on login
- [ ] Improve bindings to user profile (see notes in #550)

## Update 11-01-19 (Mattia)

**Overview:**
We've had a map on Precious Plastic for 1 year and a bit and it has been one of the most useful and used tools we've released so far. It really makes collaborating and finding people nearby easy and straightforward. For One Army we want to create an even more powerful map to help our community to meet and collaborate. 
The map should host current projects and be developed to allow future projects to be hosted on it. For this version it should be developed to host Precious Plastic and Project Kamp. 

**Latest design resources:**
Prototype
https://sketch.cloud/s/KnZgA/m1ME5W9/play

**Recording**
http://recordit.co/1CvywlW7q3

**Originals + Screens**
https://drive.google.com/drive/u/0/folders/11vyG5sd5DeaTSyB-q6WgUxuaDIEhY5Wr

**Epics:**
https://docs.google.com/spreadsheets/d/1pkLRKCbQiJOtQwWEhVNgSTvDWf5SnVAz10vMo4k-LNg/edit#gid=0

Any questions post in the comments

chrismclarke:
now things being handled in smaller issues so feels less relevant to keep open. Closing.

PR #517, opened 2019-07-08
feat/module/map/frontend

nzchicken:
First cut of front end for the map.

TODO:
- Hook up data from firebase (map and pins + filters)
- Take My Pin button somewhere
- Debug weird problem where map resets to the center for some odd reason


src/pages/index.tsx
```diff
@@ -65,9 +65,11 @@ export class Routes extends React.Component<any, IState> {
                           title={page.title}
                           description={page.description}
                         />
-                        <PageContainer>
+                        {/*<PageContainer>
```
nzchicken: I've just realised that I committed this change. For the map, I think it's better that there is no page container, but unfortunately there's no "easy" way to do this one. This should probably be looked at since it currently is going to mess up the style on the rest of the pages etc.

src/pages/Maps/Content/MapView.tsx
```diff
@@ -0,0 +1,86 @@
...
+        zoom={zoom}
+        maxZoom={18}
+        style={{ height: '100%' }}
+        onMove={this.updateBoundingBox}
+      >
+        <TileLayer
+          url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
+          attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
+        />
+        <div>Testing</div>
```
chrismclarke: Can remove this

src/pages/Maps/Content/Sprites.tsx
```diff
@@ -0,0 +1,29 @@
+import L from 'leaflet'
+import './sprites.css'
+import { IMapPin } from 'src/models/maps.models'
+
+// mmm curry
```
chrismclarke: :yum

src/pages/index.tsx
chrismclarke: Agreed,  I think it makes sense that the map component defaults to 100% height and general page component ensures this fits with the header

src/stores/Maps/maps.store.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    return this.pinData.map(({ id, location, pinType }) => ({
+      id,
+      location,
+      pinType: filterMap[pinType],
+    }))
+  }
+
+  // Caching pinDetails in a map to reduce database calls. We don't want to cache
+  //  this using firebase since this data could change over time
```
chrismclarke: Yeah I'm still really struggling to think about what the best strategy is. Firestore is really poorly optimised for these types of operations (costs scale linearly with results retrieved from a query), so getting 1000 small files is really expensive compared to 1 large, but having many rows of data makes a lot more sense for the map.  Will write-up some possible alternate approaches in another issue when thinking about the backend for discussion.

chrismclarke: issue opened: #518

nzchicken:
I've just pushed a new couple of commits to fix the comments and the bind stuff. I don't know why I didn't think of that earlier. I was running into this context problems, and just defaulted to the old way of doing it, whoops.

@chrismclarke It's not currently possible to put the popups over the controls as the mapview and controls are parent to the map component (i.e. they sit in different trees). It could be achieved by having them in the same tree, but it could also be solved by centering the map on the pin when it is clicked also. Would need to double check with Saskia about the UX on this one of course, since i don't think floating over the controls makes sense either (you're blocking user input then).

As for the multi-select, you'll see that the default `react-select` (with isMulti option) doesn't work to how the current designs dictate, thus the extra include. It is MIT license though, so you could just take it, remix out the things you don't want and put it somewhere in the code base. I'm currently having to use !important css tags since the imported select doesn't allow some style overriding which is unfortunate (I would argue though, this is really a future problem, not a problem for now though as getting it done is probably higher up there).

Similar story goes for ItemCounter and ItemLabel components. They are super localised for the use in the multi select, so I don't see any value in trying to make them generic (unless you've got plans to use them somewhere else of course, but I'd just make the GroupingFilter component more generic in that case).

I can imagine that the page container might want to have different sizes based on the content that lives within. It could be achieved by passing the route name as a prop to the component, and that way you can just use a styled component prop to toggle the styles you want. I'll make another PR to show you what I mean

src/pages/index.tsx
nzchicken: Proposed change in #519 for changing the maxWidth requirement or not. Resolving this for now since I just committed this part out.

chrismclarke:
> It's not currently possible to put the popups over the controls
Thought would most likely be that way, longer term solution might be to have the search box expand/contract as a popup is toggled but for now not to worry.

> As for the multi-select...
Cool, it's still super lightweight (5kb) so not worried, more a case that I really don't like the existing `react-select` component so might be worth moving everything over. But again, really not a priority. Agree similarly for `!important` styles and styled components, the whole module is nicely organised/contained so I don't think there will be any real confusion as to where things come from.

I've collated other bits from here into #144 as a summary for whenever there's a map v2 push, and will add there with any feedback that comes or updates on specific issues. Thanks again @nzchicken for all your work on this.

PR #519, opened 2019-07-10
feat(container): pageWidth characteristics on page containers

nzchicken:
Not sure if this maps a lot of sense to do it this way, but I can imagine some pages want to have a "cover" format. I'm not sure on the right words to use, or whether you want to achieve it this way, but just an idea. The main use case is for something like Maps where you want it to cover the whole width of the page, and not be boxed into the container size.

### DIFF
--- a/src/pages/PageList.tsx
+++ b/src/pages/PageList.tsx
@@ -14,6 +14,7 @@ export interface IPageMeta {
   title: string
   description: string
   exact?: boolean
+  fullPageWidth: boolean
 }
 
 const howTo = {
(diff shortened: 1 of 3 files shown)


src/pages/PageList.tsx
```diff
@@ -14,49 +14,57 @@ export interface IPageMeta {
   title: string
   description: string
   exact?: boolean
+  fullPageWidth: boolean
```
chrismclarke: Should this be optional? (`fullPageWidth?:boolean`)

I think it's right how the default behaviour is the standard max-width value, and this is essentially an optional override. We could also remove the `fullPageWidth: false` from any of the how-tos not implementing this feature

nzchicken:
@chrismclarke I've pushed the optional change, is this what you meant?

PR #539, opened 2019-08-16
Refactor/timestamps

### LINKED ISSUE #343: Handle date and timestamp objects more gracefully
By default whenever firestore saves a date object it converts to a timestamp to support nanoseconds for use in queries (recent behaviour change). Either want to update interfaces to acknowledge this is the data format, find a way to disable this behaviour, or add a conversion for required fields on data return.

chrismclarke:
This PR mostly removes references to timestamps throughout the platform, as they have been becoming increasingly difficult to work with (see #343, closes #343)

**What has been done:**

1. `functions/...`
Fix backend db backup function. At same time a few minor commits to improve the general workflow for people working with the backend (documented in backend readme)

2. `src/mocks/...`
Replace any reference to timestamps in mocks. Add `db.mock` to generate mock db metadata (`_id, _created, _modified`)etc

3. `src/models/...`
replace references to firestore timestamps with `IsoDateString` - this is simply the same as a string format, but stated as a reminder that all dates should now be stored in the ISOString format (e.g. `2019-08-03T10:24:05.585Z`). This includes non-db meta fields, specifically `user.year`,  `event.date` and `discussion.lastResponse` fields.

4. `src/pages/...`
Replace methods that operate on timestamps with ones to operate on date strings

5. `src/stores/...`
Replace methods as above, and change all database endpoints to be prefixed with `v2_`  so that data can be safely migrated to the new version

5. `src/utils/...`
Remove deprecated methods for generating/converting timestamps

6. `*not in code*`
- database backed up
- all existing collections migrated to `v2_...` collection

**Todo**
- A reasonable amount of testing has been done - the build works, how-tos and events seem to display correctly with date fields filled out, however it would be good if another person could give a once over in case anything has been missed.

- Until this goes live on staging any data added to the database will not be reflected in the new version. Therefore the db upgrade backend function should be run again once live to ensure no data is lost

- Before merge into production similar backup and testing of the database should be done

- After merge into production again the db upgrade should be run to ensure no data is lost


BenGamma:
Merge, would be deploy soon on https://dev.onearmy.world
Once deploy, you can take care of running the different script to do the upgrade on the db ? Then I will do a session of tests as you suggest, to double check 🤓

PR #540, opened 2019-08-18
Backend/db restructure

### LINKED ISSUE #522: Migrate DB methods to use Dexie instead of firestore for caching
Dexie provides a really useful wrapper around local indexeddb, and offers a more versatile solution for offline-first data approaches. The firestore cache can only be used with firestore methods, making it harder to migrate to other dbs in the future and doesn't support use with firebase realtime which is currently required for the map. 

In addition Dexie provides a nice query syntax which could be used throughout the site.

chrismclarke:
This PR lays the groundwork for changes to the overall database structures and how the database works with module stores. None of the existing modules have been changed to use the system yet, but instead will be trialled with the map backend before slowly migrating others.

The main motivation for the changes are:
1. Make it easier to migrate away from firestore via a better 'abstract' middle database layer and consistent naming convention
2. Make database operations efficient in terms of what is retrieved online vs from a cache, adding Dexie as a wrapper for indexeddb to remove reliance on firestore for caching.
3. Provide options for caching larger queries online for efficient retrieval (e.g. all map pins) - adding firebase realtime database for now as a solution for this (or something like redis in the future).
3. Use a single database throughout the site instead of calling new database objects from each store
4. Allow stores to share information with each other in a tidier way (e.g. the current logged in user)

Changes:
**functions/...**
Added backend functions that handle a daily sync of the main database queries to the serverCache. Currently this consists simple of all public top-level collections, such as `v2_mappins`, `v2_howtos` etc. The full collection can be downloaded in a more efficient way from this cache. In the future it could also hold more complex queries.

**package.json**
Add Dexie database, which is a wrapper for the browser indexeddb database and will be used to cache data offline. Previously this had been handled by firestore's own indexeddb methods however dexie provides a lightweight, more flexible solution

**src/index.ts**
The way stores are injected has been modified slightly, so that instead of separate stores for every module there is one 'root' store which has all other stores within it. This makes it easier to share information between stores, or have a single database open and made available to all stores. This pattern is recommended by Mobx here: https://mobx.js.org/best/store.html

image

i.e., if the HowtoStore wanted to access information about the user it could request
```
this.rootStore.stores.userStore.user
```
Or access to the database would be via
```
this.rootStore.db
```
(instead of instantiating a new database object inside of every store, which get's messy when dealing with indexeddb in particular)

**src/stores/...**
All stores have had their constructor methods changed to allow the injection of the root store. Currently this does not change how any of the stores behave, but in the future will allow for stores to access the newer database structures

**src/stores/database/index.ts**
This is home to the new database system. It is not currently in use by any of the stores but will be in the future. The main `index.ts` file provides the database class for the platform to interact with. It sits between specific database clients (such as firestore), and offers what should be a simple and type-safe way to interact with the database. The main methods will be described in separate documentation, but consists mostly of collection and doc objects, and methods to read (once via `get()`, or in realtime via `stream()`), query or write. 

An example to retrieve a list of all howTos (checking first cache and then online), and continuing to emit and cache updates as they happen would be:
```
db.collection('v2_howtos').stream()`
```
Or accessing a particular user profile would be
```
db.collection('v2_users').doc('chris-m-clarke').get()
```
Or getting a document(s) with a specific field-value, such as a howTo by slug
```
db.collection('v2_howtos').getWhere('slug','==','my-first-howto')
```
Some extra logic is in place for handling calls to cache vs live database, which again will be outlined in separate documentation

**src/stores/database/types**
This is the start of a better system of documentation for the overall database, so that instead of lots of inline comments a specific set of guidance and examples can be made available with code completion in vscode, or auto-generated documentation. E.g., hovering over `DBClients` gives more details:

image

**src/stores/database/clients**
This provides bindings for specific database clients, in a consistent way to make it easier to add different databases in the future. Currently there are bindings for firestore, firebase realtime db, and dexie.  An example for Dexie is:
image
(The references to `T` or `T & IDBDoc` are simply placeholders in case a user wants to specify the type of data expected at a given database endpoint (replacing T with IHowTo for example), and informing the user that they will get that type of data back with additional metadata populated by the database)

Any other databases can be added by providing similar bindings for the core CRUD methods, which are then utilised by the main Database object.

There are also a few small utils functions to help with some common logic, database initialisation etc.

closes #522


chrismclarke:
The v2 stuff was all part of pull #539, where all the data had to be changed to remove firebase firestamps and replace with regular string dates. But it would definitely be worth adding some docs as to when/why you might need to do this (basically a cheap way to handle breaking changes, instead of lots of `if(typeof date === 'string')...` etc., easier to just have a fresh data source to work from. Its also worth mentioning how Dexie indexes work (by default firestore creates indexes on every field to allow quick searching, but dexie requires manual indexes so that data can quickly be ordered by, for example, date modified of value of the `slug` field)

chrismclarke:
Thanks @catx23 , not sure exactly what you mean by shortening - are we talking easy case removing the word 'store' or something else (e.g. default exports)? Or using a better system of generators to create on the fly perhaps? I've never gotten around to learning/understanding functional programming, I feel like it's a powerful paradigm but one that definitely takes a shift in mindset. I know mobx as the tool for stores/state management is built on functional programming design principles but I don't know much about architecture for pairing with functional components.... one day...

catx23:
@chrismclarke , yup; as i remember; the suffix 'store' should be removed; eg: by the stores/index import; should be clear enough though.
I think I left a few links on Discord but it's all gone somehow but essentially I think it's good idea to start blending in slowly on the side (FB promotes that too), at any cost. There is lots of code you don't need to write; despite it's just script though anyways. For instance the default constructors, if you look at Angular or more recent TS code, there is no need for 'this.var = arg ? arg : default' in the constructor. I messed my code for a while like this too til they made this work without. Possibly this sort of mistakes can be avoided with a general font-end vision document ... 
Anyways, peanuts & cosmetics though , keep going, just saying that around 10-30 % of the code could be removed :-)

PR #544, opened 2019-08-21
UI changes

paulpichot:
- Added 2 fonts (Varela Round & Inter) and font faces + Font family -> Varela Round in Global body
- Created 2 components : EventCard and HowToCard
- Created Flex that extends Flex rebass with Matthia's Design System titles
- Fixed responsive of how to and image gallery style
- Removed the material UI loading progress and replaced simply by "Loading..."
- Styled the map popup


src/pages/common/Login/Login.tsx
```diff
@@ -8,6 +8,7 @@ import { LoginForm } from './Login.form'
 import { SignUpForm } from './SignUp.form'
 import { ResetPWForm } from './ResetPW.form'
 import { Box } from 'rebass'
+import { litleBtn } from '../../../components/Button/index'
```
BenGamma: Here you don't need to import the prop, it must be automatically available on the component

PR #548, opened 2019-08-28
Howto filters

catx23:
added missing example code for the filter system, enabling filtering how-tos by multiple properties

@BenGamma : for some reason, the filter system works but the how-to list page isn't correctly updating.
have fun

### DIFF
--- a/src/pages/Discussions/common/filters.ts
+++ b/src/pages/Discussions/common/filters.ts
@@ -173,8 +173,7 @@ export const operationPredicates = {
     */
   },
   has: (value, filter) => {
-    const ret = value.find(v => v.indexOf(filter.value) !== -1) != null
-    return ret
+    return value.find(v => v.indexOf(filter.value) !== -1) != null
   },
   hasNot: (value, filter) => value.indexOf(filter.value) === -1,
 }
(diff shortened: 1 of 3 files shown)


BenGamma:
Hey thanks for the examples ! We were at the same time working on the same things, I just make a PR with the sorting system by tags working in #549

catx23:
@BenGamma Could you please know when you discard/remove the rest of my code as well ? It may not be important to you and I understand I am not part of the club or the management, etc... but I spent weeks on onearmy and I am a little confused about how to help at all. Because if it's like that, it would be cool to know at least so I can go ahead with a fork instead, that would eventually also preserve the invests done. I am fully aware that things change over time but yet, that were precious 6 weeks for me. 

thank you.

BenGamma:
Sorry if you feel like these weeks were a lost, I've put a lot of effort into things that have been changed or discard also. But that's how it is, you have to understand that this project is still in it's very early stage and directions can change over time.

If I understand well you want me to remove your code (Editor, discussionList and filters ) ?

catx23:
Hi,
thanks for your response. I don't think this is 'like this' but ok, lesson learned ...

And no, I didn't ask to remove code; just for a simple notification when it's going to be removed so I can start working on a fork instead. 

Please understand; we also gave up lots of other opportunities to get things done for one-army and from what I see, now you have a very specific filter solution instead of a generic, tested and more powerful one. Same for the editor. It was built on PP experience in and now we have not even a simple rich text editor. That actually breaks my heart to see this happening but ok; I guess I have to get used to it :-)

thanks again.

PR #549, opened 2019-08-28
How to sorting by tags

BenGamma:
In howtoList, added the TagsSelect component and allow the sorting of the howtos by tags.

### DIFF
--- a/src/stores/Howto/howto.store.tsx
+++ b/src/stores/Howto/howto.store.tsx
@@ -52,7 +52,7 @@ export class HowtoStore extends ModuleStore {
     this.uploadStatus[update] = true
   }
 
-  @computed get currentHowtos() {
+  @computed get filteredHowtos() {
     // Check if this.selectedTags is empty
     if (
       Object.keys(this.selectedTags).length === 0 &&
--- a/src/stores/Howto/howto.store.tsx
+++ b/src/stores/Howto/howto.store.tsx
@@ -1,4 +1,4 @@
-import { observable, action } from 'mobx'
+import { observable, action, computed } from 'mobx'
 import {
   IHowto,
   IHowtoFormInput,
@@ -19,7 +19,10 @@ export class HowtoStore extends ModuleStore {
   @observable
   public allHowtos: IHowto[]
   @observable
+  public selectedTags: ISelectedTags
+  @observable
   public uploadStatus: IHowToUploadStatus = getInitialUploadStatus()
+  rootStore: RootStore
 
   constructor(rootStore: RootStore) {
     // call constructor on common ModuleStore (with db endpoint), which automatically fetches all docs at
@@ -28,6 +31,8 @@ export class HowtoStore extends ModuleStore {
     this.allDocs$.subscribe(docs => {
       this.allHowtos = docs as IHowto[]
     })
+    this.rootStore = rootStore
+    this.selectedTags = {}
   }
 
   @action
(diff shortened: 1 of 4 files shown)


src/stores/Howto/howto.store.tsx
```diff
@@ -46,6 +52,27 @@ export class HowtoStore extends ModuleStore {
     this.uploadStatus[update] = true
   }
 
+  @computed get currentHowtos() {
```
chrismclarke: I think it will probably clearer to call them `filtered` howTos instead of `current`, unless there was another particular reason you wanted to call them that

src/stores/Howto/howto.store.tsx
```diff
@@ -46,6 +52,27 @@ export class HowtoStore extends ModuleStore {
...
+    if (
+      Object.keys(this.selectedTags).length === 0 &&
+      this.selectedTags.constructor === Object
+    ) {
+      return this.allHowtos
+    } else {
+      // If there is tags return a filtered array that compare howto.tags & the selectedTags
+      const filtered = this.allHowtos.filter(howto => {
+        // needs JSON.stringify to avoid return false when comparing two arrays
+        return JSON.stringify(howto.tags) === JSON.stringify(this.selectedTags)
```
chrismclarke: Nice idea, although it won't work in cases where the how-to has multiple tags (e.g. 3) and the user searches for multiple tags (e.g. 2). i.e.
```
howToTags1 = {
 tagA: true,
 tagB: true,
 tagC:true
}
```
```
searchTags = {
  tagA: true,
  tagC: true
}
```
Instead I would suggest something along the lines of
```
const howToTags = Object.keys(howTo.tags)
const searchTags - Object.keys(searchTags)
```
and then check if all the elements in the searchTags exists in the howToTags
https://stackoverflow.com/questions/16312528/check-if-an-array-contains-any-element-of-another-array-in-javascript

catx23: yeah, just wanted to drop it as example; I don't have time for the details, sorry. Multiple tags are no problem, just add them to the list of filters : 

const filters: IPropertyFilter[] = [
      {
        prop: 'tags',
        filter: {
          op: 'has',
          value: 'this tag',
        },
      },
      {
        prop: 'tags',
        filter: {
          op: 'contains',
          value: 'other tag'
        }        
      }
    ]
 
So, this wasn't meant as PR per-se; just some illustration. The filter system would also allow filtering by user; the same way as shown above.

catx23: I can add more examples to make explicit use of the built-in AND and OR operators but I think the defaults are fine (we used this in advanced apps). 
However, this system allows you also other easy filters, eg:
- recent popular how-tos, 
- filter by one tag only
- by user & search text and/or title
- by time, etc..

you gonna save lots of code & time with this :-)

catx23: ah sorry, i just realized I commented in the wrong place.

BenGamma: Thanks, just updated the PR. I used every() and includes() instead of some(), as some is only to compare one value and we can have multiple tags selected. As suggested [here](https://stackoverflow.com/questions/9204283/how-to-check-whether-multiple-values-exist-within-an-javascript-array) I compared the arrays and it seems to work well ✌️

chrismclarke:
Eurgh, so afterthoughts
```
if (
      Object.keys(this.selectedTags).length === 0 &&
      this.selectedTags.constructor === Object
    ) {
      return this.allHowtos
    }
```
1. What is the reason for the selectedTags constructor check?
```
      this.allHowtos.map(howto => {
        if (howto.tags !== undefined) {
          // encapsulate howtoTags in const to avoid type error
          const howtoTags = Object.keys(howto.tags)
          // filter the howto containing the selected tags
          const isMatching = Object.keys(this.selectedTags).every(val => {
            return howtoTags.includes(val)
          })
          // push the matching howto to filtered array
          if (isMatching) {
            filtered.push(howto)
          }
        }
      })
```

2. `howtoTags = Object.keys(howto.tags)` should be taken outside of the allHowtos as it only ever has to be calculated once, not every time a new how to is tested

3. The main code could probably go in an external lib

4. The code could be a bit more concise with use of ternarys.,e e.g.
```
@computed get filteredHowtos() {
    const selectedTagsArr = Object.keys(this.selectedTags)
    return selectedTagsArr.length > 0
      ? this.allHowtos.filter(howto => {
          const tags = howto.tags ? Object.keys(howto.tags) : null
          return tags ? selectedTagsArr.every(key => tags.includes(key)) : false
        })
      : this.allHowtos
  }
```
But for now it's no biggie, will probably refactor next time working with it (e.g. map pin filtering)

PR #555, opened 2019-08-31
Add Build matrix to CI

wallies:
test against Nodejs 9,10,11,12 so we know issues in the next version and see if we can move from 8 to 11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_script:
   - export REACT_APP_BRANCH=$TRAVIS_BRANCH
   - echo "REACT_APP_BRANCH=$REACT_APP_BRANCH"
 script:
-  - node --max_old_space_size=8192 npm run build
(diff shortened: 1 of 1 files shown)


chrismclarke:
image
Most beautiful thing I've seen all day!
https://travis-ci.org/ONEARMY/community-platform/builds/579168304

Also Zap came back with 0 issues found for the quick scan so assume that's all good.
```
[INFO]            Starting ZAP daemon
[INFO]            Running a quick scan for https://dev.onearmy.world
[INFO]            Issues found: 0
[INFO]            Shutting down ZAP daemon
```

chrismclarke:
I'm guessing it will make the most sense to wait until lighthousebot accepts invite and then make a pr into master to test fully, but in the meantime if there's anything else you were thinking of integrating then just let me know if there's any extra keys/configs that need to be set.

Thanks again

PR #560, opened 2019-09-03
558 filter events by location

### LINKED ISSUE #558: [feature request] filter events by location
Ability to filter events by location

Build on top of the tag filtering system #557 , the possibility to filter by location.

**Build suggestion**
Use the `location.name` field as an other filter on the current event array.

BenGamma:
### created
* LocationSearch component to event list page to allow filtering by location (still need some UI modifications, will be done in 0.6 release)
### updated
* add method `onClear` as props to LocationSearch component to listen to onclear event on field.

close #558

### DIFF
--- a/src/components/LocationSearch/LocationSearch.tsx
+++ b/src/components/LocationSearch/LocationSearch.tsx
@@ -45,7 +45,7 @@ export class LocationSearch extends React.Component<IProps, IState> {
     this.places.on('change', (selected: IAlgoliaResponse) =>
       this.handlePlaceSelectChange(selected),
     )
-    this.places.on('clear', () => this.onClearInput())
+    this.places.on('clear', () => this.props.onClear())
     this.subscribeToInputChanges()
   }
 
@@ -83,10 +83,6 @@ export class LocationSearch extends React.Component<IProps, IState> {
   handlePlaceSelectChange(selected: IAlgoliaResponse) {
     this.userInputRef.current.value = selected.suggestion.value
     this.props.onChange(_resultToLocation(selected))
-  }
-
-  onClearInput() {
-    this.props.onClear()
   }
 
   render() {
--- a/src/components/LocationSearch/LocationSearch.tsx
+++ b/src/components/LocationSearch/LocationSearch.tsx
@@ -15,6 +15,7 @@ interface IProps {
   placeholder: string
   debounceTime: number
   onChange: (selected: ILocation) => void
+  onClear: () => void
 }
 interface IState {
   debouncedInputValue: string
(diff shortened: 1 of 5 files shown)


src/stores/Events/events.store.tsx
```diff
@@ -38,10 +49,31 @@ export class EventStore extends ModuleStore {
...
+    this.selectedLocation = {
+      name: '',
+      country: '',
+      countryCode: '',
+      administrative: '',
+      latlng: {
+        lat: 0,
+        lng: 0,
+      },
+      postcode: '',
```
chrismclarke: Do we need postcode? I would suggest as not all countries use postal codes might be easier to leave out and depend on name/latlng for geolocation

src/components/LocationSearch/LocationSearch.tsx
```diff
@@ -83,6 +85,10 @@ export class LocationSearch extends React.Component<IProps, IState> {
     this.props.onChange(_resultToLocation(selected))
   }
 
+  onClearInput() {
```
chrismclarke: This feels redundant, we can just let `places.on('clear')` call the props onclear method directly

src/stores/Events/events.store.tsx
```diff
@@ -29,7 +32,15 @@ export class EventStore extends ModuleStore {
   }
 
   @computed get filteredEvents() {
-    return this.filteredCollectionByTags(this.upcomingEvents, this.selectedTags)
+    return this.selectedLocation.value !== ''
+      ? this.filteredCollectionByTags(
```
chrismclarke: This doesn't read very naturally. `filteredCollection` implies the list has already been filtered, so instead this should be `filterCollection` to indicate it is an action that will happen (past tense noun vs verb). 

Also nesting one filter within another is very confusing. Move the inside filter out, give it a variable name and use in the second filter operation.

src/stores/Events/events.store.tsx
BenGamma: Would make sense in this case. But I know in the user profil page we will display the complete adress, which may contain a postcode

src/stores/Events/events.store.tsx
BenGamma: I put the first filter in a const, remove the ternaries and used and good old if / else. Way more readable I agree.

src/stores/Events/events.store.tsx
chrismclarke: ok, more `postcode ? ... : ... ` statements it is then.

src/stores/Events/events.store.tsx
chrismclarke: yeah looks a lot better, nice

src/components/LocationSearch/LocationSearch.tsx
chrismclarke: can see this is now gone (+1)

PR #564, opened 2019-09-05
103 how to edit mode

BenGamma:
### WIP on the how to edit mode.
Main work done is :
* create a EditHowto component
* move howto form from `CreateHowTo` to a `Common` folder so it can be used by Create & Edit howto pages
* add props to howtoForm component : `onSubmit()`, `parentType` & `formValues`
* inject existing how-to values from store in the edit form

Known todo :
* show existing images in the files inputs. @chrismclarke do you have an idea of how to do that ? After a quick search I found the [custom store](https://uppy.io/docs/stores/) section in uppy doc, would be something like that ?
* allow only creator & admin to modify the how-to


src/pages/Howto/Content/HowtoList/HowtoList.tsx
```diff
@@ -69,8 +69,8 @@ export class HowtoList extends React.Component<any, IState> {
             </Flex>
           ) : (
             <Flex flexWrap="wrap" mx={-4}>
-              {filteredHowtos.map((howto: IHowto) => (
-                <Flex px={4} py={4} width={[1, 1 / 2, 1 / 3]}>
+              {filteredHowtos.map((howto: IHowto, i) => (
+                <Flex key={i} px={4} py={4} width={[1, 1 / 2, 1 / 3]}>
```
chrismclarke: Using index as key defeats the point of a key as any new items pushed into the array will force the entire list to re-render. It would be much better to use a unique property of the howto, such as id instead.

PR #569, opened 2019-09-11
add profile delete methods and ui

chrismclarke:
**Overview**
Create a button on the user profile page that allows a user to delete their profile
image

Includes a prompt for user to confirm their password before delete
image
This is mainly because firebase requires reauth for sensitive actions, however probably a good safeguard to ensure a user can't easily be deleted if using a shared computer and forgetting to log out

On click the user is removed from the auth database and profile marked as deleted in the database (with all personal information removed)

**Todo**
- delete user avatar if uploaded
- decide how to handle content attributed to them (e.g. how-tos). Currently all content remains, which is probably preferable
- show notification on successful delete (some sort of toast most likely)
- show notification on error (e.g. incorrect pw provided for revalidation
- handle display of a user profile page that has been deleted (user profile still exists, but only with fields `_deleted:true`

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -4,7 +4,8 @@ import styled from 'styled-components'
 
 interface IProps {
   // provide onDidDismiss function to enable backdrop click dismiss
-  onDidDismiss?: () => void
+  onDidDismiss: (data?: any) => void
+  height?: number
 }
 interface IState {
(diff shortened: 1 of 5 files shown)


BenGamma:
Cool that's a nice GDPR compliant feature ! Thanks for that.
Considering the todo you want to keep going on working on this feature and I'll merge now and we create separated issues for the todos points ?

PR #573, opened 2019-09-12
Refactor/how tos

chrismclarke:
**Main Changes**
- change store methods to use v2 db methods
- minor updates to typings to better-define how-to before and after db entry, and to manually provided `_createdBy` as no longer automatically added with db meta. 

**Additional Changes**
- update db query methods to avoid need for composite indexes (previously ordered all records by date modified, however if also searching for a specific field such as slug this would automatically try to order alphabetically, so redundant ordering by both modified and slug)

Tested working with staging `how-to/refactor-test` (`by` field omitted but should now be working)


chrismclarke:
Yeah, it's always going to be a bit confusing because what goes into the database isn't the same as what comes out, and what the user inputs isn't the same as all the information uploaded either. But hopefully these distinctions should be easier to manage.

PR #590, opened 2019-09-22
testing sign up

porcelani:
In response to #266,

moving cy.visit() of  page to test: So we can have in the page constructor some asserts that the page was loaded correctly 

testing sign up and sing in: the features that can not fail

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,9 @@ The project uses [Jest](https://jestjs.io/) to unit Test and [Cypress](https://w
 - Run the component documentation
   `yarn storybook`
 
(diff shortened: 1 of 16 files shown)


BenGamma:
Thank you guys for the work ! Can you give a bit more informations about how to run the test ?

porcelani:
Hi @BenGamma,
After running the local server
  `yarn start`

You already can run the e2e tests locally. It will happen without open de Cypress Test Runner.
 `yarn e2e`
The config of the base URL is here -> https://github.com/ONEARMY/community-platform/blob/master/cypress.json . You can change the URL and execute the tests in another environment.

You can open the Cypress test Runner too running:
`yarn e2e-open`

PR #606, opened 2019-10-02
Fix invalid package name during "npm i"

StephD:
Just a small fix to allow me to install packages

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "react-router": "^5.0.0",
     "react-router-breadcrumbs-hoc": "^3.2.0",
     "react-router-dom": "^5.0.0",
-    "react-router-hash-link ": "^1.2.1",
+    "react-router-hash-link": "^1.2.1",
     "react-select": "^2.4.3",
     "rebass": "4.0.6",
     "resolve": "1.10.0",


StephD:
Mmm, hey :-)
I know travis said No, but...
I just start to practice the contribution to open source project and I don't know if I need to don't something else :-)

Thanks

StephD:
Well @tmhung88,
I checked and the error make no sense to me :-)
My commit is a modification of package.json, not related to the runtime of the app.
Travis error is from the "academy.spec.ts" script.
This one test only the access to "/academy" to see if you can access without login.
It seem that if this test doesn't pass and it is not because of my commit :-D

So I cannot do anything on my side.

I might be wrong also.

Cheers

tmhung88:
@StephD  I did have the same thinking before triggering the build 2 times. However, it still failed but many other pr/commit passed. Hence, the error could be connected to your change. My guess is installing the lib (router-hash-link) could introduce some conflicts in code and throw some exception at runtime. 

FYI: dependencies, react, router .... are libraries an app requires to run. On the other hand, devDependencies are tools/libraries that we need to build an app, for example: testing tools, linting ....

chrismclarke:
That's super weird!

Thanks for the PR @StephD, that's clearly a typo so I I definitely can't think of any reason why fixing it would have a knock-on effect, so will try investigate further.

@tmhung88  - I tried to recreate and was a bit tricky. The first time I ran the test it failed and the second time succeeded. I think the issue is possibly related to some sort of timeout, that the test is running before the server is actually ready (it is the very first test that fails).

I have no idea why it should only start now with this PR, so if you have any more tips to help diagnose/fix they would be most appreciated

chrismclarke:
Testing again and it definitely seems to be related to the delay between the server starting and content being ready to interact with (see https://github.com/cypress-io/cypress/issues/2639). I've added a PR (#635) that perhaps you can review @tmhung88 , and then hopefully get this merged in after.

tmhung88:
> Testing again and it definitely seems to be related to the delay between the server starting and content being ready to interact with (see [cypress-io/cypress#2639](https://github.com/cypress-io/cypress/issues/2639)). I've added a PR (#635) that perhaps you can review @tmhung88 , and then hopefully get this merged in after.

Yep I'm aware of the long start-up time of the app. It can fail academy.spec.ts sometimes. Will merge master into this PR once #635 is merged

StephD:
Cool, happy to participate to this project with this minimal contribution :-D

chrismclarke:
You're officially on the list!
image

:D

Issue #607, opened 2019-10-02
update filters on map


BenGamma:
Link the existing filters on the map with the profile v2 user models.

There is two filters that doesn't depend on each others :

* Filtering depending on the `profileType` field :
<img width="881" alt="Capture d’écran 2019-10-02 à 14 45 29" src="https://user-images.githubusercontent.com/4993215/66045107-512c8900-e523-11e9-8af2-ab70f9b97fc4.png">

* Filtering depending on the `workspaceType` field :
<img width="896" alt="Capture d’écran 2019-10-02 à 14 43 33" src="https://user-images.githubusercontent.com/4993215/66045016-23474480-e523-11e9-98ad-d7d83107bad4.png">

drydenwilliams:
I think this has been taken over by @chrismclarke. Can you confirm please @chrismclarke?

chrismclarke:
The filters are there and I made a few tweaks to how they work, although I'm not sure if I understand what the issue refers to exactly? Is it old work just to bring the filters in? Or something we need to update for them? Happy to do whatever, or close this issue if already done, so just let me know.

BenGamma:
Yes it was old work. Closing now that map with filtering was release in [v0.8.0](https://github.com/ONEARMY/community-platform/releases/tag/v0.8.0)

PR #614, opened 2019-10-04
Update filters on map #607

byronwilliams:


BenGamma:
Hey @byronwilliams thanks a lot ! Happy to see this feature moving forward.
Sorry for the late reply, but can you describe a bit more what you've did in this PR to help reviewing ?

src/pages/Maps/Content/Controls/GroupingFilter.tsx
```diff
@@ -39,55 +41,104 @@ const ItemCounter = styled.span<ItemCounterProps>`
...
 `
 
+const ElWithBeforeIcon = props => {
+  let after
+
+  if (props.ticked) {
+    after = {
+      content: "''",
+      backgroundImage:
+        'url("https://uploads-ssl.webflow.com/5d41aacc625e7f69441ddaff/5d5eaf83aeeaa530fecae293_icon%20checkmark.svg")',
```
BenGamma: Can you mark this with a TODO comment ? will change very soon with our own icon

src/pages/Settings/content/formSections/MapPin.section.tsx
```diff
@@ -12,10 +12,17 @@ import { Map, TileLayer, Marker, Popup, ZoomControl } from 'react-leaflet'
...
 import 'leaflet/dist/leaflet.css'
 import { generatePinFilters } from 'src/mocks/maps.mock'
+import { ILocation } from 'src/models/common.models'
+import {
+  IWorkspaceType,
+  IProfileType,
+  WorkspaceType,
+  ProfileTypeLabel,
+  IUserPP,
+} from 'src/models/user_pp.models'
```
BenGamma: Here I don't understand why there is changes. The map in the user settings will just be used to display the user pin and will not provide any filtering. Is it a miscommunication on the mockups or there is any other reason ?

BenGamma:
See my comments below. I asked a review from @chrismclarke too.
There may be a bit of confusion around the `ILocation` type.
There was a shorten version that only contains fields `lat`, `lng` and `adress` in the maps model interface `IMapPinBase`. This location field is outdated as now all the users locations are saved under the `ILocation` type that lives in `src/models/common.models`

src/pages/Maps/Content/Controls/GroupingFilter.tsx
byronwilliams: This PR is not ready, so I will change the URL before I submit it for review. This PR is not ready to be merged, not for a while.

src/pages/Settings/content/formSections/MapPin.section.tsx
byronwilliams: This is just a file which I will `checkout` again, there are no changes. This will be tidied up before I submit it for review

byronwilliams:
To me the datastructures don't really make any sense. Are you mapping `users` or `mappins`. If you're only interested in displaying `users` then why is there a separate table to store the `mappins`. The `mappins` data structure is not good enough to store the data for the user. So where is the data supposed to live? In the `user` or the `mappin`? How do these relate to locations?

chrismclarke:
Yeah this is a bit of a messy one for sure, because what we really want is mostly info from the user profile page however pulling full profiles from the user page would both be inefficient (a lot more data than actually needed on the map page) and open up potential for abuse (even though user profiles are public we don't want a single mechanism to pull all at once as it would make it too easy to harvest data; of course in theory you could still do this by pulling in all the map pins and making separate requests for full profiles the same way that the pin does, but it's easier to rate-limit that at least). 

So the map pin is designed to do only the bare minimum of what a map pin requires - i.e. a location, a pin type, and a parent id to retrieve full info

I hope that makes sense, if you have any alternate suggestions happy to hear them of course

chrismclarke:
Hey @byronwilliams , what's the latest on this - I see a lot of recent commits so guessing things are headed in the right direction at least. Anything you want help with to get this merged in?

byronwilliams:
@chrismclarke Please feel free to check this out... there are still a few things to finish off, but I'll raise those as separate tickets.

Issue #437, opened 2019-04-18
add text & heading component to storybook


BenGamma:
Write story for text and heading component.

An example can be found in [button component](https://github.com/OneArmyWorld/onearmy/blob/master/src/components/Button/button.stories.tsx)

tanvirraj:
@BenGamma, I want to do this

BenGamma:
Yes just do it ! Again let me know, to avoid two people working on the same thing. Thanks !

tanvirraj:
start working

tanvirraj:
when i run `yarn storybook` i get this error, can't start storybook 
any idea ?
<img width="1427" alt="Screen Shot 2019-05-07 at 7 33 16 PM" src="https://user-images.githubusercontent.com/3768120/57303392-0ab33980-70ff-11e9-950e-085610f9a031.png">

BenGamma:
Looks like there is an issue with storybook 5.0.0 https://github.com/storybooks/storybook/issues/6306

If you can have a look that would be nice :)

tanvirraj:
i will look into it.

jparadasb:
Hello! is someone working on this issue? I took a look at this and was fixed after upgrading storybook to `5.2.1` and changing the webpack.conf as the migration guideline says [https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecate-webpack-extend-mode](url)

> @BenGamma
> when i run `yarn storybook` i get this error, can't start storybook
> any idea ?
> <img alt="Screen Shot 2019-05-07 at 7 33 16 PM" width="1427" src="https://user-images.githubusercontent.com/3768120/57303392-0ab33980-70ff-11e9-950e-085610f9a031.png">

BenGamma:
Oh cool ! No feel free to fix it !

chrismclarke:
closed by #618

PR #618, opened 2019-10-05
#437 add text & heading component to storybook 

jparadasb:
## objectives
✅ Add Text component to storybook
✅ Add Heading component to storybook

## Other things

* Was fixed an issue to start storybook, do not necesary upgrade storybook, just adapt the `webpack.config`

* Was fixed something  related with the definition of `fontSize` using a prop (medium, small, large) that I think was a typo in heading component (but I not sure if that was something intentional in any case it can be reversed)

* I made a factory to handle the Knops using a default props objects (let me know if is useful or not)

## Demo

storybook

### DIFF
--- a/src/components/Heading/index.tsx
+++ b/src/components/Heading/index.tsx
@@ -5,7 +5,7 @@ import Text, { ITextProps } from 'src/components/Text'
 import { HeadingProps as RebassHeadingProps } from 'rebass'
 
 export const large = (props: ITextProps) =>
-  props.medium ? { fontSize: theme.fontSizes[6] } : null
+  props.large ? { fontSize: theme.fontSizes[6] } : null
 export const medium = (props: ITextProps) =>
   props.medium ? { fontSize: theme.fontSizes[5] } : null
 export const small = (props: ITextProps) =>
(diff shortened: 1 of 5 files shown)


src/components/Heading/index.tsx
```diff
@@ -5,7 +5,7 @@ import Text, { ITextProps } from 'src/components/Text'
 import { HeadingProps as RebassHeadingProps } from 'rebass'
 
 export const large = (props: ITextProps) =>
-  props.medium ? { fontSize: theme.fontSizes[6] } : null
+  props.large ? { fontSize: theme.fontSizes[6] } : null
```
BenGamma: nope that was not intentional, thanks for spotting 👀 and fix

jparadasb: Nice, I would like to collaborate with other issues, I'm joining the discord 🤝

BenGamma:
Note to remove the build error for dev with outdated storybook installation :
* delete storybook node_modules `rm -rf node_modules/@storybook/`
* re-install storybook `yarn add storybook`
* run storybook as normal `yarn storybook`

src/components/Heading/index.tsx
BenGamma: Yes cool ! If you like to work on storybook, adding more of our component library (in `src/components`) to it would be great help 🙌

PR #626, opened 2019-10-08
Set up a new firebase environment for cypress

tmhung88:
- Inserted some seed data
- Had Cypress and App run against the same environment, onearmy-test-ci, when executing e2e tests
- Added a test case that is to filter howtos by tags
- To start the app in the testing env, run this command `yarn start:ci`. Please note it runs at a different port which is **localhost:3456**

image


.travis.yml
```diff
@@ -17,9 +17,7 @@ before_script:
   - echo "REACT_APP_BRANCH=$REACT_APP_BRANCH"
 script:
   - yarn test
-  - if [ $REACT_APP_BRANCH != 'production' ];
-    then yarn cy:ci;
-    fi
+  - yarn cy:run
```
tmhung88: @BenGamma I removed the quick fix because this PR has cypress/app run against the same environment whenever tests are executed

src/config/config.ts
```diff
@@ -42,27 +35,31 @@ const e = process.env
...
+      return 'production'
+    case 'master':
+      return 'staging'
+    default:
+      return 'localhost'
+  }
+}
+
+const siteVariant = getSiteVariant(branch, e)
+
```
tmhung88: For every environment, I think, we need to define the site variant as an env variable so that we don't have to if/else on branch which is very error-prone.

PR #635, opened 2019-10-09
extend timeout and add base route to all tests

chrismclarke:
Should hopefully fix issue of first test fail. On local machine all tests now appear to pass when localhost is slow to load the platform.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,7 +1,7 @@
 {
   "baseUrl": "http://localhost:3000",
   "defaultCommandTimeout_comment": "Increase to 15s to make sure a page is completely loaded.",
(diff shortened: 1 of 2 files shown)


cypress.json
```diff
@@ -1,7 +1,7 @@
 {
   "baseUrl": "http://localhost:3000",
   "defaultCommandTimeout_comment": "Increase to 15s to make sure a page is completely loaded.",
-  "defaultCommandTimeout": 15000,
+  "defaultCommandTimeout": 30000,
   "watchForFileChanges": true,
```
tmhung88: @chrismclarke moved the config to this file which is more appropriate.

PR #656, opened 2019-10-10
Fix #640: CreateHowTo page not opening with its url

### LINKED ISSUE #640: Cannot open the create-how-to page with its url
**Describe the bug**
Login with any user. Visited /how-to/create and got redirected to /how-to

**To Reproduce**
f590178669d2954adab8792d6637e4246b012407

**Expected behavior**
- The create-how-to page should be opened as if clicking on the **Create a How-to** button

lauvrenn:
The redirection was caused by the way AuthRoute works. If the user is logged in, they're supposed to be taken to /how-to/create and when they're not, they should be redirected back to /how-to. The problem with that, however, is that the user store takes a while to log in a user after refreshing the page, so by the time it finishes, the page is already rendered with AuthRoute seeing the user as not-logged-in (this isn't the case when we click the 'Create How-To' button on the main page as, by then, the user store has already finished).

I changed the way /how-to/create works similarly to how the /settings subpage is handled; it comes with two shortcomings though. When a user is logged in, this causes the "not-logged" screen to appear for a second until the user store is updated with confirmation of user authentication. Also, when a user is not logged in, they land at the "not-logged" screen, and it'd be better if they were never able to access the /how-to/create page in the first place. It'll be great if somebody has an idea on how to make AuthRoute work. As long as it doesn't though, this can fix the issue. #640


chrismclarke:
Thanks for the PR @lauvrenn! Will review shortly. 

In the meantime, I think the failed tests might be down to previous issues. Can you merge master and push the changes to see if it fixes.

chrismclarke:
Also I think there should already be a test to catch this issue, so could you move it from `cypress/integration/bugs.spec.ts` to the `how_to.spec.ts` file and check if passes?

lauvrenn:
I updated the branch as well as the test, and it passes. There’s a problem with the academy.ts test though, can’t see it now as I’m on mobile, but I’ll try to fix it in the morning

chrismclarke:
Yeah it looks like it just hit a timeout so I wouldn't be too worried about it. Still getting some of the bugs out of the tests (and fully appreciating the irony :p )

PR #659, opened 2019-10-10
Fix/658   duplicate titles how to

### LINKED ISSUE #658: [bug] - Duplicate how-tos not prevented
**Describe the bug**
See title

**To Reproduce**
Try make a how-to with the same title as another, also test available in cypress bugs spec

**Expected behavior**
Form validators prevent

Fix already prepared, issue more for reference

chrismclarke:
- Fix the issue preventing recognition of duplicate how-to
- Update the how-to spec to add more helper methods (less code duplication)
- Update spec to check for duplicate how-tos
- Rewrite spec to create how-to

closes #658

### DIFF
--- a/cypress/integration/how_to.spec.ts
+++ b/cypress/integration/how_to.spec.ts
@@ -7,22 +7,6 @@ describe('[How To]', () => {
     cy.deleteDocuments('v2_howtos', 'title', '==', 'Create a how-to test')
   })
 
-  describe('[Create a how-to]', () => {
-    it('[By Anonymous]', () => {
-      cy.step('Get redirected to /how-to when trying to create')
-      cy.visit('/how-to')
-      cy.logout()
-      cy.visit('/how-to/create')
-        .url()
-        .should('not.include', '/create')
-    })
-    it('[By Authenticated]', () => {
-      const howTo = new HowToPage()
-      howTo.create(title)
-      cy.url().should('include', `/how-to/create-a-howto-test`)
-    })
-  })
-
   describe('[List how-tos]', () => {
     const howtoUrl = '/how-to/make-glasslike-beams'
     const coverFileRegex = /howto-beams-glass-0-3.jpg/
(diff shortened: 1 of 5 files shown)


cypress/integration/how_to.spec.ts
```diff
@@ -30,10 +70,11 @@ describe('[How To]', () => {
       cy.get('[data-cy=more-how-tos]', SKIP_TIMEOUT).should('be.hidden')
 
       cy.step('All how-tos are shown')
-      const totalHowtos = 7
+      // note - in case deletion hasn't worked correctly total can be higher
+      const minHowTos = 7
       cy.get('[data-cy=card]')
         .its('length')
-        .should('be.eq', totalHowtos)
+        .should('be.gte', minHowTos)
```
tmhung88: @chrismclarke the use of == is to quickly detect any other test that inserts how-tos but forgets cleaning up. The downside is that the test becomes more fragile to race-condition scenarios which rarely happen at the moment.

chrismclarke: Yeah, I only changed it because we didn't have the required clean-up in place to support custom title/dynamic how to create and delete. Really the list test only needs to be verifying there is more than 1 (i.e. a list)

Possibly if we include proper functionality on the 'more' or infinite scroll feature the test could check the correct number is displayed, but probably overkill.

cypress/page-objects/how-to-page.ts
```diff
@@ -1,5 +1,117 @@
...
+    this.fillStep(1)
+    this.fillStep(2)
+    this.deleteStep(3)
+  }
+
+  fillIntro(title: string) {
+    cy.step('Warn if title is identical with the existing ones')
+    cy.get('[data-cy=intro-title]').type('Make glass-like beams').blur({ force: true })
+    cy.contains('Titles must be unique, please try being more specific',).should('exist')
+
```
tmhung88: @chrismclarke  As discussed on Cypress practice, #658 should be part of the create-how-to test. Having a new test (it) adds 7 secs more into our test execution.

It's also the reason that Page Object doesn't play well with Cypress. Code in Cypress philosophy is rarely reusable because it's very specific

cypress/integration/how_to.spec.ts
tmhung88: @chrismclarke It's not only about the correctness of functionality but also the testing data set. When we have excessive amount of junk data, the page could get slower and devs cannot pinpoints any issue immediately because it looks messy compared with this pristine state.

pristine-state

PR #661, opened 2019-10-11
add pw reset binding and notification

### LINKED ISSUE #648: Lost password link doesn't work in the Sign-in page
**Describe the bug**
Nothing happened when clicking onto the **Lost Password** link

**To Reproduce**
c9268e39d298cf5f87c35bc7421e4be0adf11bdb

**Expected behavior**
The lost password form should be shown

chrismclarke:
closes #648


cypress/integration/sign-in.spec.ts
```diff
@@ -3,15 +3,31 @@ import { Page } from '../utils/test-utils'
...
+    cy.step('Lost Password sent a reset link')
+    cy.get('[data-cy=email]').clear().type('howto_reader@test.com')
+    cy.get('[data-cy=lost-password').click()
+    cy.get('div').contains('Reset link sent').should('be.visible')
+
+    cy.step('Lost Password cannot sent a reset link on wrong email')
+    cy.get('[data-cy=email]').clear().type('what_is_mypassword@test.com')
+    cy.get('[data-cy=lost-password').click()
+    cy.get('div').contains(`Email doesn't exist`).should('be.visible')
+
```
tmhung88: @chrismclarke as discussed, we need to handle a exception by showing a message.

chrismclarke: Included, along with a 'friendlyMessage' function to replace the default error messages as suggested. This makes the test a bit more vulnerable (expecting exact strings), so instead I changed the notification to include `data-cy='notification'+type` where type can be `error` or `confirmation` and seems to be working as expected for the sign-in spec tests.

chrismclarke:
Exceptions added and updated test. Had to make minor change to travis to stop recording (quota reached, will revert once OSS license granted). But for now all tests passed so merging.

PR #663, opened 2019-10-11
Fixes #639: wrong msg when there are no how-tos

### LINKED ISSUE #639: Don't show loading... when there is no how-to left
**Describe the bug**
**loading...** was always shown even if there was no how-to matching selected tags

**To Reproduce**
fccf3d8ac0d3451a0d35b92e83aa25ca5b4effc9

**Expected behavior**
Should say to users there is no more how-to available
image

lauvrenn:
closes #639

### DIFF
--- a/src/pages/Howto/Content/HowtoList/HowtoList.tsx
+++ b/src/pages/Howto/Content/HowtoList/HowtoList.tsx
@@ -39,7 +39,7 @@ export class HowtoList extends React.Component<any, IState> {
   }
 
   public render() {
-    const { filteredHowtos } = this.props.howtoStore
+    const { filteredHowtos, selectedTags } = this.props.howtoStore
     return (
       <>
         <Flex py={26}>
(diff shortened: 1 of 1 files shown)


chrismclarke:
oh, and I just updated the main description as it's a handy way to automatically close an issue on merge (just in case you weren't aware, or forgot)

lauvrenn:
@chrismclarke the roll is definitely on, I'll see how many of them I'm able to solve :joy: 
Btw, a quick question: should I move the bug tests (which are added to issue descriptions) from `bugs.spec.ts` to `how_to.spec.ts` when I open a PR that solves some issue?

chrismclarke:
I was talking about this with @tmhung88 recently, the only potential issue with doing this for every bug is that it could make the testing quite long, or appear tedious for some bugs which unlikely to re-emerge. 

So it's each dev's own call really. If you think it makes sense great, no harm either adding to the end of the test sequence or integrating with another test if it makes sense (e.g. testing the no-how-to message probably makes sense whilst also testing selecting multiple tags). If in doubt can just leave in bugs for for the time being.

chrismclarke:
Also love the gap year plans! If you're not already there you should definitely join https://climateaction.tech

lauvrenn:
Okay, got it, thanks! :ok_hand: 
Glad you like it hahah. Yep, I'm on their Slack channel, such a cool community! :smile:

PR #670, opened 2019-10-13
Fixes #615: hide event tags with zero results

### LINKED ISSUE #615: [bug] Hide event's tags from filter when there's no result
**Describe the bug**
In: https://community.preciousplastic.com/events
If you select some filters (eg: "cleanup" at the moment of writing this) there's no result shown.

**To Reproduce**

1. Go to https://community.preciousplastic.com/events
2. Select "cleanup" from the tag filters dropdown
3. See error

**Expected behavior**
Filters that returns zero results should be hidden from the list in the dropdown

**Screenshots**
image

lauvrenn:
closes #615

### DIFF
--- a/src/components/Tags/TagsSelect.tsx
+++ b/src/components/Tags/TagsSelect.tsx
@@ -14,7 +14,7 @@ export interface IProps extends FieldRenderProps<any, any> {
   category: TagCategory | undefined
   styleVariant: 'selector' | 'filter'
   placeholder: string
-  showOnlyRelevantTags?: any
+  relevantTagsItems?: ICollectionWithTags[]
 }
 interface IState {
   selectedTags: string[]
@@ -23,7 +23,9 @@ interface InjectedProps extends IProps {
   tagsStore: TagsStore
 }
 
-const filterArrayDuplicates = (array: string[]) => Array.from(new Set(array))
+interface ICollectionWithTags {
+  tags?: ISelectedTags
+}
 
 @inject('tagsStore')
 @observer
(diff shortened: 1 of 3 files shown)


tmhung88:
@lauvrenn could you check why this test case still failed? Put this piece of code in **bugs.spec.ts** to run. The **howto_testing** tag shouldn't be listed if I understand this fix correctly.

```
it('[615]', () => {
    cy.deleteDocuments('v2_howtos', 'title', '==', 'Create a how-to test')
    cy.visit('/how-to')
    cy.get('[data-cy=tag-select]').click()
    cy.get('.data-cy__menu').contains('product').should('be.exist')
    cy.get('.data-cy__menu').contains('howto_testing').should('not.exist')
  })
```

lauvrenn:
@tmhung88 yeah, you're right. The test was failing for `/how-to` because the code fix only worked for the `events` tab, just as it was specified in the issue. I guess it's better to extend the fix to everywhere the select field is used though, so I just restructured the code. The test should now pass (without the wait(1000) function though, it has some issue with that) :ok_hand:

lauvrenn:
@chrismclarke okay, just implemented the requested changes in the latest commit

src/components/Tags/TagsSelect.tsx
```diff
@@ -86,6 +99,21 @@ class TagsSelect extends React.Component<IProps, IState> {
     arr.forEach(el => (selectedJson[el] = true))
     return selectedJson
   }
+
+  // we want to display only those tags that return results, meaning they are used by how-tos, events, etc
+  private _getTagCounts(items: ICollectionWithTags[]) {
+    const tagCounts: { [key: string]: number } = {}
+
+    items.map(
```
drydenwilliams: Can we just do this in a reduce?

lauvrenn: hmm, yeah, I tried to think of how to do that previously and could not come up with a solution. I think reduce's accumulator keeps track of one value -- in this case, it would be an object -- and would keep track of every tag that is added to it, one by one. It would not keep track, however, of multiple tags' values, at least I don't know how to implement that. If you can come up with a solution though, I'll substitute it :ok_hand:

lauvrenn:
@chrismclarke is this one ready for merging?

chrismclarke:
Sounds good to me! (sorry, hard to keep track of PRs pending/post change requests). I've sent an invite to be a contributor on the platform so perhaps in the future we can add tags so you can easily flag whenever something needs review.

PR #672, opened 2019-10-14
Fixes #616: functionality of the 'More Events' button

### LINKED ISSUE #616: [bug] Nothing happens clicking on "More Events"
**Describe the bug**
At the end of the events list, there's the button "More Events" that doesn't do anything

**To Reproduce**

1. Go to https://community.preciousplastic.com/events
2. Scroll down to 'More Events' button
3. Click on the button
4. See error (nothing happens)

**Expected behavior**
If there are more events, then show more events at the bottom of the list
If there are no other events, then hide the button
Hide the button if there are no more events with the currently selected tags

lauvrenn:
closes #616 :tada: 
the comments rely on #670 and can be uncommented once that issue is merged

### DIFF
--- a/src/pages/Events/Content/EventsList/EventsList.tsx
+++ b/src/pages/Events/Content/EventsList/EventsList.tsx
@@ -11,16 +11,30 @@ import TagsSelect from 'src/components/Tags/TagsSelect'
 import { inject, observer } from 'mobx-react'
 import { EventStore } from 'src/stores/Events/events.store'
 import { LocationSearch } from 'src/components/LocationSearch/LocationSearch'
+// import { ISelectedTags } from 'src/models/tags.model'
 
 interface InjectedProps {
   eventStore: EventStore
 }
 
+interface IState {
+  eventsOffset: number
+  // selectedTags: ISelectedTags
+}
+
+const getLimitedEvents = (start: number, end: number, array: IEventDB[]) =>
+  array.slice(start, end)
+const eventsLimit = 5
+
 @inject('eventStore')
 @observer
-export class EventsList extends React.Component<any> {
+export class EventsList extends React.Component<any, IState> {
   constructor(props: any) {
-    super(props)
+    super(props),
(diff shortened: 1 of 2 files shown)


lauvrenn:
@chrismclarke Okay, I think I will give it a try, I've never worked with the virtual rendering component, but I hope I can figure it out :ok_hand: 
By 'too many requests for cover images on first load', do you mean for it to be extended to how `/how-to` is loaded as well? As events don't have images.

src/pages/Events/Content/EventsList/EventsList.tsx
```diff
@@ -68,18 +90,49 @@ export class EventsList extends React.Component<any> {
           </Flex>
           <React.Fragment>
             <>
-              {filteredEvents.length === 0 ? null : ( // *** TODO - indicate whether no upcoming events or data still just loading
+              {filteredEvents.length === 0 ? null : (
```
drydenwilliams: Please can we just do:

```
{filteredEvents.length !== 0 && (
 <div>Div stuff</div>
)}
```

So it's a bit more readable.

lauvrenn: That piece of code is left as it was before because it will be deleted before merging and substituted, see the attached comments

chrismclarke:
> @chrismclarke Okay, I think I will give it a try, I've never worked with the virtual rendering component, but I hope I can figure it out 👌
> By 'too many requests for cover images on first load', do you mean for it to be extended to how `/how-to` is loaded as well? As events don't have images.

Yes, ideally we will want a reusable component/code that can easily be reused to efficiently render large lists, be they event lists, how-tos or anything else. In the back of my mind I was thinking that the 'more' button was also on the how-to page but I think someone else already removed it as suggested as a simpler workaround for this issue.

BenGamma:
Closing for now, as we remove the More button for "create" button

PR #681, opened 2019-10-15
Added tests for Events and common components

tmhung88:
image


tmhung88:
Added tests for common components on the screen: Home page, not-found page, header and user menu

image

chrismclarke:
Retrying build

PR #682, opened 2019-10-15
Fix/646   how to button disable on save and edit test

### LINKED ISSUE #646: View How-To is not disabled while a how-to is being saved
**Describe the bug**
Users were able to click on the **View How-To** button when a how-to was being saved. As a result, it showed the how-to was not found.

image

**To Reproduce**
1a6905469e4e62ad52dd2e59227baf0100d760e1

**Expected behavior**
The View How-to button is disabled until a how-to is completely saved into firebase

chrismclarke:
closes #646 

Frontend
- bunch of refactoring to better recognise and respond to how-to upload status and remove some duplicate code

Tests
- Removed 6000 wait (add methods for correct button availability)
- Add edit how-to test


cypress/integration/how_to.spec.ts
```diff
@@ -306,4 +303,25 @@ describe('[How To]', () => {
       cy.get('div').contains('Please login to access this page')
     })
   })
+
+  describe('[Edit a how-to]', () => {
+    it('[By Owner]', () => {
+      cy.login('howto_creator@test.com', 'test1234')
```
tmhung88: @chrismclarke The test has 2 issues which depends on another test, create-how-to, and execution order (edit needs to run after create). As a result, it's highly fragile and difficult for new devs to debug when it fails as they have no context.

I put back edit tests for some time because there are 2 questions I haven't addressed
- A test flow for edit 

1. Reset a specific document to the default state 
2. Run the edit test
3. Verify the change against firestore

- an api to verify if data is updated correctly. I imagine it'd be cy.queryDocuments('some conditions').should('eq', {updated_data}

PR #691, opened 2019-10-16
593 max image upload

drydenwilliams:
Got it working 😄 

Screenshot 2019-10-16 at 18 43 58


drydenwilliams:
Wawhooo! 

Screenshot 2019-10-16 at 18 43 58

src/components/FileInfo/FileInfo.tsx
```diff
@@ -5,7 +5,7 @@ import { IUploadedFileMeta } from 'src/stores/storage'
 import { FileDetails } from './FileDetails'
 
 interface IProps {
-  file: File | IUploadedFileMeta
+  file: File | IUploadedFileMeta | null
```
gustavomazzoni: As `file` can accept `null`, should we set it as optional instead?
`file?: File | IUploadedFileMeta`

drydenwilliams: 🤔  my TypeScript isn't great so @BenGamma or @chrismclarke what do you think? My initial thoughts are that the file is still mandatory, it might just be a `null` value.

chrismclarke: Either way should be fine, question mark is more correct as a non-specified file would be `undefined`, but because both `null` and `undefined` are falsy values it doesn't make a huge difference.

PR #696, opened 2019-10-16
Added tests for Sign-up page

tmhung88:
Please review #689  before this PR


cypress/utils/test-utils.ts
```diff
@@ -1,3 +1,13 @@
 export enum Page {
   HOME_PAGE = '/how-to'
+}
+
+export const generatedId = (length: number) => {
```
drydenwilliams: Maybe we could use something like `uuid` for this?

src/pages/Events/Content/EventsList/EventsList.tsx
```diff
@@ -78,8 +78,8 @@ export class EventsList extends React.Component<any> {
                 </Flex>
               )}
               <Flex justifyContent={'center'} mt={20}>
-                <Link to={'#'}>
-                  <Button variant={'secondary'}>More Events</Button>
+                <Link to={'#'} style={{ visibility: 'hidden' }}>
```
drydenwilliams: Link around a button, is this correct? Can't we have a link that is styled like a button?

src/pages/Howto/Content/Common/Howto.form.tsx
```diff
@@ -300,6 +300,7 @@ export class HowtoForm extends React.Component<IProps, IState> {
                                 icon={'delete'}
                                 variant={'tertiary'}
                                 sx={{ position: 'absolute' }}
+                                data-cy={'delete'}
```
drydenwilliams: `data-cy={'delete'}` => `data-cy="delete"`

cypress/utils/test-utils.ts
tmhung88: uuid could be an overkill or not suitable for this case. The id here is part of username/email for sign-up tests.

src/pages/Events/Content/EventsList/EventsList.tsx
tmhung88: The button is not working at the moment. I'd like to hide it. There is a separate issue for it.

PR #702, opened 2019-10-17
Adding yup so we can have consistent validation. Closes #694

### LINKED ISSUE #694: Validation for Password & Confirm-Password does not work
**Describe the bug**
When registering a new user, we're able to proceed even if Password is different to Confirm-Password

**To Reproduce**
93a6a68705db90f3200147c0798240debcd8a7e7

**Expected behavior**
- Should stop the registration and show an error message to users.

drydenwilliams:


tmhung88:
Updated Sign-up tests and confirmed #694 is fixed by this PR

Issue #710, opened 2019-10-20
Change asserts of the date that can change with timezone


porcelani:
**Is your feature request related to a problem? Please describe.**
When I execute the Cypress tests in my timezone the tests fail.

**Describe the solution you'd like**
![Screenshot from 2019-10-20 10-00-06](https://user-images.githubusercontent.com/7378479/67161082-7267e480-f32d-11e9-9aaf-473dc462efdc.png)

**Describe alternatives you've considered*
Change assert of day

chrismclarke:
closed with #714

PR #711, opened 2019-10-20
Change asserts of the date that can change with timezone

porcelani:
In response to #710


chrismclarke:
@tmhung88 - Can you take a look at this? I know you made a timezone fix to the testing so would be good to see if this is still relevant/useful to have

PR #712, opened 2019-10-20
Fixes #678: close dropdown on click away in events

### LINKED ISSUE #678: The location dropdown list cannot be closed in the Events page
**Describe the bug**
The location dropdown list didn't go away when clicking ECS or left-clicking on the open space
image

**To Reproduce**
- Go to /events
- Type **suba** in the Location box

**Expected behavior**
The dropdown list should go away when
- Hit the ECS button
- Left click outside the box

lauvrenn:
The dropdown closes when clicking away from the field now. I wonder about implementing the second requirement from the issue though -- 'ESC' as a way of closing the dropdown. Is it still necessary to add even though clicking away works?

### DIFF
--- a/src/components/LocationSearch/LocationSearch.tsx
+++ b/src/components/LocationSearch/LocationSearch.tsx
@@ -112,6 +112,12 @@ export class LocationSearch extends React.Component<IProps, IState> {
     this.props.onChange(_resultToLocation(selected))
   }
 
+  resetInputAndPanel() {
+    this.userInputRef.current.value = ''
+    this.handleInputChange('')
+    this.places.close()
+  }
+
   render() {
     const { styleVariant } = this.props
     return (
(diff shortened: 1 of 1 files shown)


chrismclarke:
Also @lauvrenn - I've just added you as a contributor on the main project repo. It might be useful in the future if you create branches on the main project instead of your fork, that way if(/when :S) there are issues with tests or similar I can ask other contributors to add additional commits onto the existing PR (annoyingly you can't do that when a PR comes from a fork, but can when it comes from a local branch).

lauvrenn:
Oh, you're totally right! As selecting a location closes the panel, I thought the input could be cleared when the user doesn't want to select one and clicks away; but forgot it would impact the creation of events as well. Just removed the reset in the latest commit. 

About the 'x', it's not on the roadmap for autocomplete.js, but as far as I've seen, places.js is used here. Places.js luckily provides the 'x' and it's already in the code, but it's hidden and with disabled functionality. :thinking:  I'll try to work it out in another PR.

And thanks for adding me as a contributor! :100:

PR #732, opened 2019-10-31
Added seed data for settings tests

tmhung88:
- Switched to Electron as the test browser. firestore.set/firestore.add apis don't work with Chrome somehow.


src/pages/Settings/content/formSections/UserInfos.section.tsx
```diff
@@ -134,6 +136,7 @@ export class UserInfosSection extends React.Component<IProps, IState> {
               </Flex>
             ) : (
               <Field
+                data-cy={'cover-images'}
```
drydenwilliams: `data-cy={'cover-images'}` => `data-cy="cover-images"`

Issue #723, opened 2019-10-30
add zIndex to theme value


BenGamma:
At the moment we have multiple zIndex values around the project and it each time a value is added, the dev never know if the new value is going to override an other z-index value, somewhere else in the project.
As suggested by @drydenwilliams we could use a theme value for z-index like :
```
const zIndex = {
 main: 1,
 header: 2,
 footer: 3
modal: 4
}

BenGamma:
closed with #736

PR #736, opened 2019-11-01
Adding zIndex object to theme styling

drydenwilliams:
Closes: [#723)


src/themes/styled.theme.tsx
```diff
@@ -15,6 +15,20 @@ export const colors = {
...
+export const zIndex = {
+  behind: -1,
+  level: 0,
+  default: 1,
+  modalBackdrop: 9,
+  modalContent: 10,
+  slickArrows: 100,
+  modalContainer: 900,
+  modalContainerInner: 900,
+  modalLink: 900,
```
BenGamma: Why not having one value `modal` instead of 3 ? It would be more readable to me.

src/pages/Howto/Content/Howto/Howto.tsx
```diff
@@ -41,7 +42,7 @@ const MoreBox = styled(Box)`
     background-image: url(${WhiteBubble0});
     width: 100%;
     height: 100%;
-    z-index: -1;
+    z-index: ${zIndex.back};
```
BenGamma: I have an error with the `back` property, it doesn't exist in the theme. Should I use zIndex.default instead ?

BenGamma: or `behind` ?

src/themes/styled.theme.tsx
BenGamma: I created a `modal` value so I can merge like this.

PR #740, opened 2019-11-02
Debugging CI build

tmhung88:
I'm opening this pr in order to debug how Cypress behaves on Travis container. Please do not merge


tmhung88:
Updated:
I tried to isolate the failed test, focus_workplace.spec.ts, from the other suites. There was no **Warning: Possible EventEmitter memory leak detected** in Travis build log but it still failed. Hence, we can safely assume the warning is not  the cause

tmhung88:
I had a doubt about loading the route /events and I added 3 short tests which load external/internal 
routes along with the failed tests
```

  it('[Visit google]', () => { 
    cy.visit('www.google.com')
  })
  it('[Visit Howto]', () => {
    cy.visit(Page.HOME_PAGE)
  })
  it('[Visit Event]', () => {
    cy.visit(Page.EVENTS)
  })

```
The result was the 2nd test [Visit Howto] failed and the rest passed. 
https://travis-ci.org/ONEARMY/community-platform/builds/607104424

----

At this point I'm suspecting the bootstrap time of the server. To verify it, I added 2 more same tests into **academy.spec.ts**. If my guess is true, 1st test should fail and the 2nd & 3rd should pass

chrismclarke:
> The plugin start-server-and-test doesn't actually wait for the entire app to be ready   

Literally has one job to do! Super annoying but good you were able to get to the bottom of it. I think the initial await spec sounds like a reasonable workaround, although instead of a standard wait time (60s) could we just await the root element (or a header or whatever) and pass a longer timeout as an option? e.g.
```
cy.get([data-cy=header],{timeout:60000})
```

tmhung88:
> Literally has one job to do! Super annoying but good you were able to get to the bottom of it. I think the initial await spec sounds like a reasonable workaround, although instead of a standard wait time (60s) could we just await the root element (or a header or whatever) and pass a longer timeout as an option? e.g.

There are many workarounds and the easiest one is to let the very first test which is **academy.spec.ts** wait for 60s (cy.wait(60000)). From this point onward, the app should be up for the remaining tests.

tmhung88:
Sum up all issues that I ran into
- Cypress dashboard was broken because record/key were omitted
**https://github.com/ONEARMY/community-platform/pull/740/commits/384f184595226a7506772a69671f096254a96e9b**
- /how-to/{slug} didn't update pages after login 
 **fixed by #744** 
- Recently the server bootstrap has been significantly slower
**Added 60s into academy.spec.ts 5ae2cca2bf4d809a09ae5dee9999ef3216987a4cf**
- Cypress doesn't clean up localStorage between tests in a suite. Logout is required if tests are needed to run with different users
**Logout between tests within a suite ae2cca2bf4d809a09ae5dee9999ef3216987a4cf**
- Firestore should be updated before a page is load because old data is likely returned if the update is performed when the page is being loading
3506bb622cd045dc8c2f35190d7c3a3bddadba9a

chrismclarke:
Thanks for the summary, I don't think I ever would have realised the `-- --` needed for nesting the cypress record bug! Is it ready for merge or are there still some bits you're working on?

PR #742, opened 2019-11-03
Map/cc edits

chrismclarke:
Last few changes to:
- Change cluster icon (on request from @davehakkens )
- Simplify some of the css/custom marker code and improve display for larger icons (previously text didn't fit)
image


chrismclarke:
Ah, it seems like when a profile is updated it doesn't correctly update the `v2_mappins` entry. Will look into now

chrismclarke:
OK, it needed a fairly big update to better handle the different types of pins and interacting with profiles, but I think all should be working as intended now. 

NOTE - as the `v2_mappins` data fields had to change you won't see any of the old pins (but can generate a new one if you make any edits to your profile). If we want to migrate existing pins for the new format we could write a quick function to do so, however I think it's probably not necessary as there are only 2 users on live with pins (you and dave), and dev site users can just update their profile

PR #744, opened 2019-11-04
Observe userStore.user to render UI on login

### LINKED ISSUE #692: Edit button is not shown after login with firebase api
**Describe the bug**
Most the time, the Edit button was not shown after login done via firebase SDK within Cypress test. However, if we use **cy.wait(1000)**, it'll appear.

**To Reproduce**
Run the Cypress test in this commit f11d33ce9f8a2f084283d8347a08d9ffc6a3a641

**Expected behavior**
 The Edit button should be shown even if it's slow

tmhung88:
the cause #692 is we didn't subscribe userStore.user to render on change
closes #692


tmhung88:
@chrismclarke the build passed for this PR. Check out the log https://travis-ci.org/ONEARMY/community-platform/builds/607174874

I don't know why Travis gets stuck at pending

Issue #750, opened 2019-11-06
Unable to create a new event


tmhung88:
**Describe the bug**
An required error on the event date prevented users from publishing a new event
![failed_create_event](https://user-images.githubusercontent.com/2813016/68270106-d957fe00-008e-11ea-8355-0f4ab47ede51.png)

**To Reproduce**
Run the events.spec.ts suite

**Expected behavior**
Should publish an event when all fields are inputted

PR #751, opened 2019-11-06
Fixed setting map pin/cover image

tmhung88:


tmhung88:
@chrismclarke There are 2 remaining failed tests that are caused by recently merged prs. I've logged issues for them
#750 Test for creating an event
#752 Test for workplace-focused Settings

PR #743, opened 2019-11-04
Storybook components

caiokf:
Adding stories for other components on Storybook. Covered in this PR:
- [x] Text
- [x] Heading
- [x] Icon
- [x] EventCard

PR #769, opened 2019-11-14
Refactoring text component

caiokf:
Dependent on #743, please merge that one first.

### DIFF
--- a/src/components/Text/index.tsx
+++ b/src/components/Text/index.tsx
@@ -19,8 +19,10 @@ export interface ITextProps {
   medium?: boolean
   small?: boolean
   superSmall?: boolean
+
   // keyof colors returns full object prototype, include typeof for just named keys (i.e. color list)
   color?: keyof typeof theme.colors
+
   // clip forces text to fill max 1 line and add '...' for overflow
   clipped?: boolean
   preLine?: boolean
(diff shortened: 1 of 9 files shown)


drydenwilliams:
Awesome Storybook work 👍 just want to make sure there will be no breaking changes with this refactor?

src/components/Text/index.tsx
```diff
@@ -29,94 +31,109 @@ export interface ITextProps {
...
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    overflow: hidden;
+  `}
+
+  ${props =>
+    props.preLine &&
+    `
+     white-space: pre-line;
+  `}
```
BenGamma: Yep that's more clear thanks. I was wondering though if you had an idea to reduce the number of possible props to the text component ?

caiokf:
@drydenwilliams So far this text refactoring would have no breaking change I assume, since the interface of the component (props, etc) haven't changed. Once I reply to @BenGamma about minimizing the amount of props, that might be something we need to go back and fix all uses of the component.

> Awesome Storybook work 👍 just want to make sure there will be no breaking changes with this refactor?

src/components/Text/index.tsx
caiokf: @BenGamma  

What we have at the moment:
```uppercase?: boolean
inline?: boolean
regular?: boolean
txtcenter?: boolean
capitalize?: boolean
bold?: boolean
txtRight?: boolean
large?: boolean
medium?: boolean
small?: boolean
superSmall?: boolean
color?: keyof typeof theme.colors
clipped?: boolean
preLine?: boolean
tags?: boolean
auxiliary?: boolean
paragraph?: boolean
```

What I think we can reduce it to:
```
color?: keyof typeof theme.colors
large?: boolean
medium?: boolean
small?: boolean
superSmall?: boolean
align?: string
bold?: boolean
uppercase?: boolean
capitalize?: boolean
clipped?: boolean
```

Plus, I would have the following separate components for the predefined styles:
```
ParagraphText
AuxiliaryText
TagsText
```
Those can simply be ``export default styled(Text)`color: blah; font-size: bleh;` ``  exported as a different component.

Then we have left `inline?: boolean` which maybe every `Text` could be by default, and if needed to be a block element, just wrap it in a div. 

What do you think?

BenGamma: Yes props make sense 👍 
I like the separate component for `ParaphText`, `AuxiliaryText`, `TagsText`.

How I would see it is in the `src/components/Text/index.ts` having the `Text` component exported by default and then have exported const `ParaphText`, `AuxiliaryText`, `TagsText`.

BenGamma:
> @drydenwilliams So far this text refactoring would have no breaking change I assume, since the interface of the component (props, etc) haven't changed. Once I reply to @BenGamma about minimizing the amount of props, that might be something we need to go back and fix all uses of the component.
> 
> > Awesome Storybook work 👍 just want to make sure there will be no breaking changes with this refactor?

Yes that would require a little bit of refactoring and UI check, but should be ok 🙆‍♂

src/components/Text/index.tsx
caiokf: i will go ahead with that

BenGamma:
@caiokf I pushed some minor changes to be able to compile on `yarn start`.
Should I merge now or you'll do the changes we mentioned in comments on the same branch ?

chrismclarke:
As this is now quite old (not sure whether overlooked, in which case apologies, or there were issues) and has conflicts I'm going to close here, but feel free to re-open if still relevant and resolved

PR #782, opened 2019-11-17
User badge fallback image

### LINKED ISSUE #775: display user badge in header
Instead of the current user logo in the header :
Capture d’écran 2019-11-15 à 14 44 10

Display the profileType badge as in the mockups :
Capture d’écran 2019-11-15 à 14 45 16
Default would be the member badge, then depending on what user choose "community-point", "machine-builder", "collection-point" or "workspace".

There is already an example of a similar behavior in the fallback cover image of the map popup.

lauvrenn:
closes #775 
I added the fallback image and restructured the way workspace icons are retrieved. This way, the non-decorated icons are no longer based in components related to the map, which makes choosing between the two types of workspace icons in other parts of the app easier.

### DIFF
--- a/src/components/Avatar/index.tsx
+++ b/src/components/Avatar/index.tsx
@@ -2,8 +2,7 @@ import React from 'react'
 
 import { Image, ImageProps } from 'rebass'
 import { inject, observer } from 'mobx-react'
-import { UserStore, getUserAvatar } from 'src/stores/User/user.store'
-import { ProfileTypeLabel, IProfileType } from 'src/models/user_pp.models'
+import { ProfileTypeLabel } from 'src/models/user_pp.models'
 import Workspace from 'src/pages/User/workspace/Workspace'
 
 interface IProps extends ImageProps {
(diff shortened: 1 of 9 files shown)


src/components/Avatar/index.tsx
```diff
@@ -34,6 +37,12 @@ export class Avatar extends React.Component<IProps, IState> {
   // user updates their avatar (same url so by default will now be aware of change)
   componentDidMount() {
     this.getAvatar(this.props.userName)
+    this.getFallbackImage(this.props.profileType)
+  }
+
+  async getFallbackImage(type?: ProfileTypeLabel) {
```
drydenwilliams: Does this need to be `async`?

lauvrenn: thanks, fixed :ok_hand:

BenGamma:
I like that you separated the code to get workspace type in a dedicated file. It improve clarity.

I added a small commit to remove a deprecated way of getting an avatar picture. You couldn't know, it's part of a little bit of legacy code.
Now the badge is the main avatar image, not the fallback.

Thanks for the work @lauvrenn 💪

PR #788, opened 2019-11-20
Fix #762: profile links

### LINKED ISSUE #762: Incorrect contacts/links in Profile
**Describe the bug**
All links/contacts were prefixed with http://localhost:3456/u/<link>
image

**To Reproduce**
Open a profile that has contacts & links. Click on any of them and users are taken

**Expected behavior**
- Users should open contacts/links correctly

lauvrenn:
closes #762


src/utils/validators.ts
```diff
@@ -0,0 +1,19 @@
+import isUrl from 'is-url'
+
+/****************************************************************************
+ *            General Validation Methods
+ * **************************************************************************/
+
+const validateUrl = async (value: any) => {
+  console.log('val here')
+  console.log('val ', value)
+  return value ? (isUrl(value) ? undefined : 'Invalid url') : 'Required'
```
drydenwilliams: Please could we not use nested ternaries. Would prefer something more readable.

src/utils/validators.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+const addProtocol = (ev: any) => {
+  const val = ev.target.value
+  console.log('addprot ', val, ' ', ev.target.initialValue)
+  ev.target.initialValue = val.indexOf('://') === -1 ? 'http://' + val : val
```
drydenwilliams: should it be a template literal `http://${val}` also maybe: `://${val}` as it should be on `https://`?

lauvrenn: yep, it was in progress so I hadn't changed it yet then

src/utils/validators.ts
lauvrenn: I just moved this func from somewhere else and didn't mess with it, but I get you. This particular ternary is moderately readable though I think.

lauvrenn:
thanks! Oh, I didn't realize there were other types than links there :no_mouth: thanks for fixing that! 

If I remember correctly, I think only mutators worked `onBlur`, validations showed up on form submit instead. If that's the case, email validation should work then

src/utils/validators.ts
drydenwilliams: Why is this an `async` function?

drydenwilliams: Also I would prefer to read it rather than it be "moderately readable". Just for new starters/debugging it's just way easier if it's something like: 

```
if (isUrl(value)) {
 return 'Invalid URL';
}

return 'Required';
```

src/utils/validators.ts
```diff
@@ -0,0 +1,35 @@
...
+const required = (value: any) => (value ? undefined : 'Required')
+
+const validateUrl = async (value: any) => {
+  return value ? (isUrl(value) ? undefined : 'Invalid url') : 'Required'
+}
+
+const validateEmail = (value: string) => {
+  console.log('isEmail', isEmail(value))
+
+  return value ? (isEmail(value) ? undefined : 'Invalid email') : 'Required'
```
drydenwilliams: Same for the point above. I know it's nice to be on one line, but I still find this confusing. Takes me a while to compute it rather than something like: 

```
if (isEmail(value)) {
  return something
}

return 'Required';
```

src/utils/validators.ts
```diff
@@ -0,0 +1,35 @@
...
+  const re = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
+  return re.test(String(email).toLowerCase())
+}
+
+/****************************************************************************
+ *            FORM MUTATORS
+ * **************************************************************************/
+
+const addProtocol = ([name], state, { changeValue }) => {
+  changeValue(state, name, (val: string) =>
```
drydenwilliams: Could we just do `://` and then it would resolve to either `http` or `https`? @BenGamma @chrismclarke

src/utils/validators.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+const validateEmail = (value: string) => {
+  console.log('isEmail', isEmail(value))
+
+  return value ? (isEmail(value) ? undefined : 'Invalid email') : 'Required'
+}
+
+const isEmail = (email: string) => {
+  const re = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
```
drydenwilliams: Please could you add the link where you got this from in a comment

lauvrenn: @BenGamma

src/utils/validators.ts
lauvrenn: I don't know if I understand you correctly here, but `http` resolves to `https` though

lauvrenn:
@BenGamma is this one up for merging yet?

src/utils/validators.ts
BenGamma: Yes an http site that allow https will redirect to it anyway, so no worries here 🤗

BenGamma:
This one needed a little bit of extra work :
- I fixed the issue with the onBlur method. Was tricky to get it working on both validators and mutators, because they were overriding the `onBlur()` method of each others. I had to create a `customOnBlur` to trigger both of them.
- Cleared the links in the profile page with a `mailto:` prefix for emails, display the discord ID of the user instead of "Discord" and link to the precious plastic discord ([you can't create link to discord profile yet](https://support.discordapp.com/hc/en-us/community/posts/360041519131-UserProfilesLinks))
- fix some icons display for social-media links
- fix the tests
Ready to go now 🚀

Issue #756, opened 2019-11-07
[bug] can't delete step images


BenGamma:
**Describe the bug**
if trying to remove a step image, it doesn't delete and the app crashes on dev mode. In production, the submit modal freeze. 

**To Reproduce**

1. Delete an image
2. Click on save
3. See submit modal freeze

**Expected behavior**
App shouldn't freeze on image delete.
If there is an error on submit you should have an error message displayed.

BenGamma:
fixed with #805

PR #805, opened 2019-11-28
Fixed howto submit after deleting step image

benjdelt:
Fix for #756 I found the problem. In howto.store.tsx, line 129, when an image is deleted, the object in the stepImages variable is set to null, then that array gets iterated on in uploadCollectionBatch() and the function tries to access properties on null. I added a filter to remove falsey values from the stepImages array. As I’m not very familiar with the codebase yet, I am open to suggestions as to solve this differently if there is a better way.

### DIFF
--- a/src/stores/Howto/howto.store.tsx
+++ b/src/stores/Howto/howto.store.tsx
@@ -126,10 +126,9 @@ export class HowtoStore extends ModuleStore {
     const stepsWithImgMeta: IHowtoStep[] = []
     for (const step of steps) {
       // determine any new images to upload
-      const stepImages = step.images as IConvertedFileMeta[]
-      if (!stepImages[0]) {
-        stepImages.shift()
-      }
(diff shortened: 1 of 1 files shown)


BenGamma:
Yep I like the check with the double exclamation point, it's a better approach than with the `shift()` method. If I understand well is to be sure that the value returned will be boolean ?

benjdelt:
> Yep I like the check with the double exclamation point, it's a better approach than with the `shift()` method. If I understand well is to be sure that the value returned will be boolean ?

Yes, the double exclamation point forces the value into a boolean and the filter function removes false values from the array, preventing the uploadCollectionBatch function to call properties on them. As every remaining image data is an object inside the array, they can never be falsey and they can never be filtered out. Only the deleted images will be filtered out.

benjdelt:
> Alright thanks for clarity, that seems like a good option 💪

Thanks, @BenGamma !

Issue #773, opened 2019-11-15
username with space in url


BenGamma:
At the moment we use the `userName` field to create the user url `/u/<userName>`.

We observe that some user have a username with space, which creates profile url like this :
https://community.preciousplastic.com/u/Plastiki%20Rafiki
https://community.preciousplastic.com/u/Recycle%20Rebuild

To avoid this behavior we will need to have two separate for the name :
- `userName` to use in the url `/u/<userName>`
- `name` to use in the profile page :
<img width="343" alt="Capture d’écran 2019-11-15 à 14 14 26" src="https://user-images.githubusercontent.com/4993215/68946054-4ca9f000-07b2-11ea-8e07-f5d52580e4dd.png">

This is the beavior used by instagram & twitter :
<img width="549" alt="Capture d’écran 2019-11-15 à 14 06 54" src="https://user-images.githubusercontent.com/4993215/68946543-8b8c7580-07b3-11ea-82ce-2a4cf56f74fa.png">
<img width="593" alt="Capture d’écran 2019-11-15 à 14 10 48" src="https://user-images.githubusercontent.com/4993215/68946552-8fb89300-07b3-11ea-855b-b8ab4c71bf93.png">

To migrate current user we could copy the value of `userName` field in new `name` field, and ask the user with a space to update their `userName` to a value without space. Or change it ourselves.

We probably need to add a field to the sign-up form with the `name` field.

We will have to check the `userName` field doesn't contain the following caracters :
- space
- uppercase
- ... to define with @ONEARMY/design

chrismclarke:
So just to be clear, why is this a problem? I followed the link you gave and the page loaded. I'm not opposed to having strict usernames, but I just want to know what we'd be trying to achieve with them?

davehakkens:
usually the username is different from the profile name.
Now many users insert their profile name.
Which ends up in very long usernames: 
**Plastikowa Rewolucja**
https://community.preciousplastic.com/u/Plastikowa%20Rewolucja
**Javier (Singular MARS Ltd)**
https://community.preciousplastic.com/u/Javier%20(Singular%20MARS%20Ltd)
**Inez UP!cycling**
https://community.preciousplastic.com/u/Inez%20UP!cycling

Which is not nice to share or use, but that's not the main issue.
I've been told this could cause bugs and results into mappins not linking properly. Resulting into "user not found" pages. 

[...]

PR #806, opened 2019-11-30
add dexie cache-busting methods

chrismclarke:
@BenGamma 

This should fix issues with caching data that has been deleted. 

**Done**
- Add methods to allow clearing of cached db. This is currently implemented within an error handler due to the complex nature of updating cached databases. All cached dbs have a version number, when adding a new version you are supposed to also keep the old version alongside notes to tell dexie how to upgrade. Whilst this is very useful to allow incremental upgrades and changes to database structures, it can quickly get messy if you have made numerous changes to the database over time and can't be sure what version everybody currently has in their web browsers. 

The error handler implemented catches the case when a newly versioned database is created but explicit guidance hasn't been set for how to upgrade any/all previous versions. It tells dexie to just delete the old database and start completely fresh with the new one (i.e. pull data cleanly from the server).

- Make explicit variable to set a `version` number for the database, written in date format _yyyymmdd_. Updating this number will force a purge of the cache next time the platform is loaded, pulling all new data from the database.

**Todo**
- Add better support for incremental updates. Whilst the above default behaviour is pretty reasonable (we tend not to update the cache version very often), it might be nice in the future to include optional support for incremental updates - such as a new version where only howTos are removed from the cache. This has been written as a TODO item in the `dexie.tsx` file, although I'm reluctant to create a full issue for it unless it comes up as something in the future if we need to update again.

### DIFF
--- a/src/stores/databaseV2/clients/dexie.tsx
+++ b/src/stores/databaseV2/clients/dexie.tsx
@@ -1,15 +1,23 @@
 import { IDBEndpoint, DBDoc } from 'src/models/common.models'
 import Dexie from 'dexie'
 import { DBQueryOptions, DBQueryWhereOptions, AbstractDBClient } from '../types'
-import { DB_API_VERSION, DB_QUERY_DEFAULTS } from '../utils/db.utils'
+import { DB_QUERY_DEFAULTS } from '../utils/db.utils'
 
-const db = new Dexie('OneArmyCache')
+/**
+ * Update the cache number either when making changes to db architecture
+ * or busting cache on db. This is used as the Dexie version number, see:
+ * https://dexie.org/docs/Tutorial/Design#database-versioning
+ */
+const DB_CACHE_NUMBER = 20191130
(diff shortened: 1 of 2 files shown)


src/stores/databaseV2/clients/dexie.tsx
```diff
@@ -1,15 +1,23 @@
...
-import { DB_API_VERSION, DB_QUERY_DEFAULTS } from '../utils/db.utils'
+import { DB_QUERY_DEFAULTS } from '../utils/db.utils'
 
-const db = new Dexie('OneArmyCache')
+/**
+ * Update the cache number either when making changes to db architecture
+ * or busting cache on db. This is used as the Dexie version number, see:
+ * https://dexie.org/docs/Tutorial/Design#database-versioning
+ */
+const DB_CACHE_NUMBER = 20191130
```
BenGamma: So if we do any changes in the database (i.e. will have to add a field on the user model to fix #773), we will change this variable to current date.

PR #809, opened 2019-12-02
Fix/807 idb incognito browsers

### LINKED ISSUE #807: [bug] can't interact with db on certain browsers.
**Describe the bug**
Some users experiences the impossibility to interact with the db. The error in the console is 
Known issue with :
- private browsing window on mozilla 70.0.1, on macOS
- for some users on windows 10 and same version of mozilla ( explained in #801 )
Error is the same in both cases : `uncaught exception: Object`. But when you open the map section, there is a warning in the console : `A mutation operation was attempted on a database that did not allow mutations.` 

**Expected behavior**
Databases should be accessible.

**Additional context**
Seems to be linked with indexedDB, and it's an old bug from mozilla's core. See https://bugzilla.mozilla.org/show_bug.cgi?id=781982

chrismclarke:
closes #807 

I've added a new search string that can be used to disable use of dexie, just add `?no-cache` to the url when first navigating. 

I've also improved the dexie error handler to identify when it's failing to load, and automatically refresh the page in no-cache mode. 

In addition I've fixed a minor bug from the firestore methods that showed up once in no-cache mode.

Tested and working with firefox incognito mode.

### DIFF
--- a/src/stores/databaseV2/clients/firestore.tsx
+++ b/src/stores/databaseV2/clients/firestore.tsx
@@ -19,7 +19,6 @@ export class FirestoreClient implements AbstractDBClient {
   }
 
   async setDoc(endpoint: IDBEndpoint, doc: DBDoc) {
-    console.log('setting doc', endpoint, doc._id)
     return db.doc(`${endpoint}/${doc._id}`).set(doc)
   }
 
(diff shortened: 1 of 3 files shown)


src/stores/databaseV2/clients/firestore.tsx
```diff
@@ -69,12 +68,14 @@ export class FirestoreClient implements AbstractDBClient {
...
     const { limit, orderBy, order, where } = query
-    const { field, operator, value } = where!
     const baseRef = db.collection(endpoint)
     const limitRef = limit ? baseRef.limit(limit) : baseRef
-    // if using where query ignore orderBy parameters to avoid need for composite indexes?
-    return where
-      ? limitRef.where(field, operator, value)
-      : limitRef.orderBy(field ? field : orderBy!, order!)
+    // if using where query ignore orderBy parameters to avoid need for composite indexes
+    if (where) {
```
BenGamma: on `no-cache` page, the order of the howto is different. Is it because of the indexDB problem with cache ?

BenGamma:
Thanks chris that's a good one ! I didn't expect a fix 💯 
See my comment below 👆

src/stores/databaseV2/clients/firestore.tsx
chrismclarke: The live db always orders by `_modified` to make sure the freshest data is retrieved. The above code changes haven't changed anything, the difference is that the cachedb (I think) has an extra sort method built into it to order by `_created`... but really we should just sort it in the mobx store instead

PR #812, opened 2019-12-02
v0.8.3

BenGamma:
### Added
- dexie cache busting method #806 
We now have a [version number for the cache db](https://github.com/ONEARMY/community-platform/blob/f742b5a15c8d800beecb65981a305d7582baed88/src/stores/databaseV2/clients/dexie.tsx#L11). Each time we'll make changes to the db structure we'll have to update this reference. By doing so, the cache in users local storage will be cleared.
- Academy section embed routing #808 
Academy routing used to be limited to  `/academy` because we use an iframe. You can now navigate in the academy section using academy full routing. Accessible in the [dev site](https://dev.onearmy.world/academy/) only.
- Fallback db for user using firefox incognito mode & other in browser no-cache settings #809 
### Fixed
- Delete individual step image #805 
- Lots of fixes & improvement for the validations of the contact links #788 :
    - fix issue with empty contact links saved
    - fix urls redirection with `mailto:` for emails
    - automatically add `http://` protocol for link to : social-medias, bazar, forum, website
    - improved validation for : email address, urls
    - link to Discord now leads to Discord invite link, and display username instead of "Discord". Because discord profile links [doesn't exist yet](https://support.discordapp.com/hc/en-us/community/posts/360041519131-UserProfilesLinks).
### Changed
- map cluster #800 

Thanks to our contributors @lauvrenn and @benjdelt 🖖 🦅


BenGamma:
@chrismclarke not enough PR sorry 🤷‍♂

PR #816, opened 2019-12-04
Added support to show pin in map via hash

alromh87:
With this you can create url's that will show details for a map pin, using #'_id'


BenGamma:
@alromh87 why did you closed it ?

alromh87:
Changed to a pull request for complete pin moderation function, so the hash link makes more sense in context

PR #821, opened 2019-12-10
Support for pin moderation

alromh87:
Moderation can have 4 different status as follows:
- 'draft'                 -> (Only for HowTos) Work in progress still not ready for moderation by admin, will be show only to user
- 'awaiting-moderation'   -> ready to be moderated, will be shown only to admin and owner
- 'rejected'              -> Not accepted, wont be shown to admins anymore, will be shown to user
- 'accepted'              -> Good, will be shown to all

New pins are shown to the admins in the map with distinctive icon
newPinsOnMap ForAdmins

Whitin pin popup admin can accept or reject pin
newPinToBeModeratedByAdmin

Pin status is reported to the pin owner on the map
RejectedPinShownToUser


src/pages/Maps/Content/View/Popup.tsx
```diff
@@ -102,6 +120,38 @@ export class Popup extends React.Component<IProps> {
             {shortDescription}
           </Text>
           <LastOnline>last active {lastActiveText} ago</LastOnline>
+          {pin.moderation !== 'accepted' && (
+            <Text
+              auxiliary
+              small
+              mb={2}
+              highlight
```
BenGamma: Here instead of adding 2 new props to the `Text` components that would only be used in that case, I would wrap the text component in a Box with the styles in a `sx` (style extended) prop like so :
```
         <Box
              sx={{
                background: theme.colors.yellow.base,
                padding: '7px',
                borderRadius: '5px',
                fontWeight: 'bold',
                border: '1px dashed',
              }}
            >
              <Text
                auxiliary
                small
                mb={2}
                color={
                  pin.moderation === 'rejected' ? theme.colors.red : 'inherit'
                }
              >
                {moderationStatus}
              </Text>
            </Box>
```
Same for the `critical` prop

alromh87: Don't you think it could be useful tho enhance the Text components if this is going to be reused for events and howto moderation?

Then it is not needed to put all that code everytime

 sx={{
                background: theme.colors.yellow.base,
                padding: '7px',
                borderRadius: '5px',
                fontWeight: 'bold',
                border: '1px dashed',
              }}
I will like to ask you to reconsider it. Let me know if there's something Im not seeing

src/stores/Maps/maps.store.ts
```diff
@@ -50,7 +52,16 @@ export class MapsStore extends ModuleStore {
...
     })
-
+    // HACK - ARH - 2019/12/09 filter unaccepted pins, should be done serverside
+    const activeUser = this.activeUser
+    const isAdmin = hasAdminRights(this.activeUser)
+    pins = pins.filter(p => {
+      return (
+        p.moderation === 'accepted' ||
+        (activeUser && p._id === activeUser.userName) ||
+        (isAdmin && p.moderation !== 'rejected')
```
BenGamma: I think for now it's ok if only filtered client side (no sensible data sended) but agree that could be better to do server side. Any thoughts on this file @chrismclarke ?

src/pages/Maps/Content/View/Popup.tsx
BenGamma: Yep I agree on the reusability side, it's just that the `Text` component should be used for text only.
You could create a shared component `Badge` in `src/components/` ?

src/pages/Maps/Maps.tsx
```diff
@@ -31,15 +33,24 @@ class MapsPageClass extends React.Component<IProps, IState> {
...
   public async componentDidMount() {
-    this.promptUserLocation()
     this.props.mapsStore.retrieveMapPins()
     this.props.mapsStore.retrievePinFilters()
+    if (!this.showPinFromURL()) {
+      this.promptUserLocation()
+    }
+  }
+
+  public async componentDidUpdate(prevProps) {
```
drydenwilliams: No `await` inside? Doesn't this need to be an `async`?

src/pages/Maps/Maps.tsx
```diff
@@ -64,9 +75,29 @@ class MapsPageClass extends React.Component<IProps, IState> {
...
       center: latlng as ILatLng,
       zoom: 8,
+      firstLoad: false,
     })
   }
 
+  private showPinFromURL() {
+    const pinId = this.props.location.hash.substr(1)
+    if (pinId.length > 0) {
+      this.props.mapsStore.getPin(pinId).then(pin => {
```
drydenwilliams: We're not passing the promise back so don't think when we need the `async` function when we call `showPinFromURL()`

src/pages/Maps/Maps.tsx
```diff
@@ -64,9 +75,29 @@ class MapsPageClass extends React.Component<IProps, IState> {
...
+        if (typeof pin !== 'undefined') {
+          this.props.mapsStore.setActivePin(pin)
+          if (this.state.firstLoad) {
+            this.setCenter(pin.location)
+          }
+        } else {
+          // Should we do something in case we couldn't find requested pin?
+        }
+      })
+      return true
```
drydenwilliams: We're use assuming the `this.props.mapsStore.getPin(pinId)` resolves ok?

src/utils/helpers.ts
```diff
@@ -56,11 +57,23 @@ export const isEmail = (email: string) => {
   return re.test(email)
 }
 
+export const hasAdminRights = (user?: IUser) => {
+  if (!user) {
+    return false
+  }
+  const roles = user.userRoles ? user.userRoles : []
```
drydenwilliams: Could just return the `return roles.includes('admin') || roles.includes('super-admin')`?

drydenwilliams: Also why is the `userRoles` optional? 🤔

alromh87:
Yes, done removing the hash after closing a pin

Will move the text props to a new component on the other PR

PR #825, opened 2019-12-12
Moderation for Events and Howtos

alromh87:
Builds on top of pin moderation adding support for Events and Howtos

**Events**
Admins can moderate events from list showing only accepted and unmoderated pins
Captura de pantalla de 2019-12-12 21-17-35

Moderation status is shown to owners
Captura de pantalla de 2019-12-12 21-17-45

**HowTos**
Admins are presented with HowTos ready to be moderated
Captura de pantalla de 2019-12-12 21-21-15

And can accepet or reject from within HowTo detail
Captura de pantalla de 2019-12-12 21-21-25

Owners can see moderation status
Captura de pantalla de 2019-12-12 21-21-07

Howtos have now the option to be saved to draft before requesting admin moderation
Captura de pantalla de 2019-12-12 21-21-40

PR #844, opened 2019-12-28
[wip] wrap up content moderation feature

BenGamma:
### This PR is a wip
It wrap up the work started in #821 & #825 

I updated the `DB_CACHE_NUMBER` because we are going to manually populate the `moderation` field with `accepted` value on the existing content (in production db) in order to keep on displaying it.
It may be better to handle this operation with scripting, we start to have quite a lot of entries in the map, and 58 howtos.
### TODO
- [x] fix the non display of moderator buttons even for super-admins
- [ ] handle the previous content (what I mentionned before)
- [x] update tests


tmhung88:
@BenGamma tests don't need any update. It seems your changes has introduced a bug. Tags in both how-to/event pages are empty and nothing is shown.

PR #856, opened 2020-01-06
Added discord webhooks and updated url for pin, now referencing own map

alromh87:
Rules for notifications are as follows:

Ready for moderation -> slack
Accepted by moderator -> discord

### DIFF
--- a/functions/src/Integrations/firebase-discord.ts
+++ b/functions/src/Integrations/firebase-discord.ts
@@ -5,13 +5,13 @@ import * as request from 'request'
 const SITE_URL = CONFIG.deployment.site_url
 // e.g. https://dev.onearmy.world or https://community.preciousplastic.com
 
-const SLACK_WEBHOOK_URL = CONFIG.integrations.slack_webhook
+const DISCORD_WEBHOOK_URL = CONFIG.integrations.discord_webhook
 
-export const notifyNewPin = functions.firestore
+export const notifyPinAccepted = functions.firestore
   .document('v2_mappins/{pinId}')
   .onWrite((change, context) => {
     const info = change.after.exists ? change.after.data() : null;
-    if(info === null || info.moderation !== 'awaiting-moderation'){
+    if(info === null || info.moderation !== 'accepted'){
       return
     }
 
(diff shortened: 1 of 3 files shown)


functions/src/Integrations/firebase-discord.ts
```diff
@@ -0,0 +1,98 @@
...
+import * as functions from 'firebase-functions'
+import * as request from 'request'
+
+const SITE_URL = CONFIG.deployment.site_url
+// e.g. https://dev.onearmy.world or https://community.preciousplastic.com
+
+const DISCORD_WEBHOOK_URL = CONFIG.integrations.discord_webhook
+
+export const notifyPinAccepted = functions.firestore
+  .document('v2_mappins/{pinId}')
```
BenGamma: Here you should use the new `v3_mappins` endpoint (we updgraded the db entries). Keeping v2 would result in no notifications.

functions/src/Integrations/firebase-discord.ts
```diff
@@ -0,0 +1,98 @@
...
+          console.error(err)
+          return
+        } else {
+          return res
+        }
+      },
+    )
+  })
+export const notifyHowToAccepted = functions.firestore
+  .document('v2_howtos/{id}')
```
BenGamma: Same here with endpoint, change to `v3_howtos`

functions/src/Integrations/firebase-discord.ts
```diff
@@ -0,0 +1,98 @@
...
+          console.error(err)
+          return
+        } else {
+          return res
+        }
+      },
+    )
+  })
+export const notifyEventAccepted = functions.firestore
+  .document('v2_events/{id}')
```
BenGamma: Same here with endpoint, change to `v3_events`

functions/src/Integrations/firebase-slack.ts
```diff
@@ -9,8 +9,12 @@ const SLACK_WEBHOOK_URL = CONFIG.integrations.slack_webhook
 
 export const notifyNewPin = functions.firestore
   .document('v2_mappins/{pinId}')
-  .onCreate((snapshot, context) => {
-    const info = snapshot.data()
+  .onWrite((change, context) => {
```
BenGamma: So now on slack notifications will be triggered on every content edit ? That's nice

BenGamma:
Was A quick one, I updated and merge in #866

chrismclarke:
Looks good, just a quick one @alromh87 , looks like we'll need to add `CONFIG.integrations.discord_webhook` to the backend. Does someone have a generated api key to do so with? Ideally we want to different ones, one that will publish notifications from the dev site and one from the live site. 

I can see you have access to the dev console so can set it yourself via terminal command:
```
firebase functions:config:set integrations.discord_webhook=My_Api_key
```
(assuming you have firebase installed globally, otherwise `npm i -g firebase-tools` first

For the live site I'll need to set it, so send me a message in slack if you have the key (otherwise I can request one from Dave). I'd say don't put it in a message here because it will accidentally be publicly exposed.

davehakkens:
replied in slack :)

PR #868, opened 2020-01-11
Functions/discord int fix

chrismclarke:
Fix discord integration
- change `request` package for `axios` (easier to handle response to post/get request)
- fix variable bindings/typos
- tidy code

Tested for updating map pin, howto and event on dev. config variable has been set on prod so should start working there also once in.

image

image

image

### DIFF
--- a/functions/src/Integrations/firebase-discord.ts
+++ b/functions/src/Integrations/firebase-discord.ts
@@ -1,6 +1,6 @@
 import { CONFIG } from '../config/config'
 import * as functions from 'firebase-functions'
-import * as request from 'request'
+import axios, { AxiosResponse, AxiosError } from 'axios'
 
 const SITE_URL = CONFIG.deployment.site_url
 // e.g. https://dev.onearmy.world or https://community.preciousplastic.com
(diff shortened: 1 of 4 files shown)


davehakkens:
looking neaat.. cant wait to use this one!
Is the Slack webhook also working?

chrismclarke:
I see plenty coming through the slack webhook so assume it's working as intended. Although if anybody notices any issues feel free to flag up and I'll take a look.

davehakkens:
huh into which channel? 
Because in the #community-notifications-dev I don't see any updates.
And in the #community-notifications i see updates. But I think they are still the old ones as it shows the old text version without the map url Location here (soon our own map)

chrismclarke:
Oh, I'm with you now... hmm, strange, can add to the list of things to investigate.

functions/src/Integrations/firebase-discord.ts
```diff
@@ -9,89 +9,61 @@ const DISCORD_WEBHOOK_URL = CONFIG.integrations.discord_webhook
...
-    const slug = info.slug
-    request.post(
-      DISCORD_WEBHOOK_URL,
-      {
-        json: {
-          content: `📓 Yeah! New How To "* ${title} *" by ${user}
+    const { _createdBy, title, slug } = info
+    await axios
+      .post(DISCORD_WEBHOOK_URL, {
+        content: `📓 Yeah! New How To "* ${title} *" by ${_createdBy}
```
BenGamma: I think here the message would be : `📓 Yeah! New How To **${title}** by *${_createdBy}*` without the " and with the correct font style markdown (**bold** with ** around how to title and _italic_ for creator name with one *)

functions/src/Integrations/firebase-discord.ts
```diff
@@ -9,89 +9,61 @@ const DISCORD_WEBHOOK_URL = CONFIG.integrations.discord_webhook
...
     const location = info.location.country
-    console.info(info)
-    request.post(
-      DISCORD_WEBHOOK_URL,
-      {
-        json: {
-          content: `📅 Jeej new event in *${location}* by ${user} posted here:
+    await axios
+      .post(DISCORD_WEBHOOK_URL, {
+        content: `📅 Jeej new event in *${location}* by ${user} posted here:
```
BenGamma: Same here for font style : `📅 Jeej new event in **${location}** by *${user}* posted here:`

BenGamma:
@chrismclarke shoudl I create a new PR for the text updates or you are going to push some updates anyway on this branch ?

chrismclarke:
This is already merged so I suggest updates go in a different PR. At the same time (or separately) you might want to make the required updates to the slack hook

Issue #864, opened 2020-01-08
[feature request] setup analytics


BenGamma:
_Updated 2020-01-20 CC_ 
Required for analytics:

- [x] Decide on analytics provider (use GA for now, but hopefully matomo in future, as per some discussion on here and on slack)
- [x] Add component binding/hooks (close with pr #869)
- [x] Add privacy policy/terms of use and opt-in/out mechanism (see #619)
- [ ] Add profile/data deletion options (currently on manual request)
- [x] Add data opt-in/out
- [ ] Add explainer to wiki

_End of Update_

**Is your feature request related to a problem? Please describe.**
We are currently not using any analytics solutions for the community-platform. Would be useful to collect some insight numbers.

**Describe the solution you'd like**
Implement a analytics tracking solution. Two options are :
- Google analytics.
Pros : 1. No server required 2. High granulatity level of analytics and data processing 
Cons : 1. make the page more heavy (loading the GA script is ~20kb per page) 2. it add a GA cookie, AND we are trying to avoid google 😈 
- Matomo.
Pros: 1. no script added if we use the [http tracking api](https://developer.matomo.org/api-reference/tracking-api). It makes the tracking very lightweight (only a GET request is made). 2. Total control over data. 
Cons : 1. Need our own server (can be a cheap 2$/month vps) 2.lower granularity of analytic

WIth any of these solutions we'll probably have to add some opt-in opt-out to tracking, to be gdpr compliant. More informations there https://matomo.org/blog/2018/04/how-to-make-matomo-gdpr-compliant-in-12-steps/

drydenwilliams:
Would be great to get some analytics on ASAP. I'm a big fan of GA as it's easy to add and give great granularity and events too. 

Seems there is quite a few different alternatives out there: 
https://www.leadfeeder.com/blog/google-analytics-alternatives/

Issue #722, opened 2019-10-29
GDPR regulation


BenGamma:
Align the platform with GDPR regulation.

chrismclarke:
Might be good if other people weigh in on this, but I'm naively assuming there isn't much we actually need to do at this stage. We're not currently using cookies to track/store any analytics, and the only user information we expose publically is username and profile picture (both of which users can choose to make non-identifiable). 

I can see a similar discussion on discourse here: https://meta.discourse.org/t/gdpr-compliance-when-using-public-user-profiles/95059/10

The only other thing I can think about is the right to remove/delete profile/data. I think there's open issues on this for some specific modules, but assuming it would be good to have a delete profile option somewhere.

I'm wondering if there is anyone we can reach out to in the community who might know more about this?

BenGamma:
I know @pmcp had some contact in that field, maybe you could make the link with us ?

chrismclarke:
merge into #619

Issue #619, opened 2019-10-07
[feature request] - Analytics/Cookie Policy


chrismclarke:
**Is your feature request related to a problem? Please describe.**
Discussion started in slack but better to capture here. Code previously included to test implementing google analytics, however this is no longer desirable. Want to determine best alternative including thoughts on:

1) Do we need to track any information? If so, why?
2) What are our options to capture this? (likely either hardcoding our own functions to update the db, or using a 3rd party analytics service)
3) What are the implications of (2)? (i.e. how would the community feel about this? what about GDPR? what additional steps would/should be necessary)

**Describe the solution you'd like**
To track a bare minimum of useful information (possibly just some non-identifiable (PII) overall usage and page view counts). Either using manual methods or non-google analytics provider (suggested Matomo). Allowing all data captured to also be publicly available and clear to users.

**Describe alternatives you've considered**
Using out-of-the-box solutions that typically capture a lot more than we would need, and adding a cookie banner to be GDPR compliant. 

**Additional context**
This is a good place where we can exhibit how we think data and consent should work, not just trying to be compliant with regulation but to promote an open/responsible way of working.

chrismclarke:
Following a bunch of chat on slack (mostly v4-core-dev channel), decided on minimal implementation to start, linked with login mechanism.

PR #869, opened 2020-01-13
Adding Google Analytics

drydenwilliams:
I decided to make a component to do all the analytics instead of using the Analytics store. It still uses `react-ga`. But I found the store to coupled to the frontend and backend and was confusing. So keep it simple and let the frontend and the `<Analytics />` component take care of the FE.

### DIFF
--- a/src/components/Analytics/index.tsx
+++ b/src/components/Analytics/index.tsx
@@ -1,0 +1,28 @@
+import React from 'react'
+import ReactGA from 'react-ga'
+import { GOOGLE_ANALYTICS_CONFIG } from 'src/config/config'
+import { RouteComponentProps, withRouter } from 'react-router-dom'
+
+class Analytics extends React.Component<RouteComponentProps> {
+  constructor(props: RouteComponentProps) {
+    super(props)
+
+    ReactGA.initialize(GOOGLE_ANALYTICS_CONFIG.trackingCode, { debug: true })
+  }
+
+  componentDidMount() {
+    this.sendPageView(this.props.history.location)
+    this.props.history.listen(this.sendPageView)
+  }
+
+  sendPageView(location: any) {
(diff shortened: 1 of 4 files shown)


chrismclarke:
Also, just realised that we still don't have anyone working on the privacy policy and opt-in/out GDPR requirements so we shouldn't merge into production site until that's done.

drydenwilliams:
Do we have a ticket for the privacy and GDPR work?

BenGamma:
> Do we have a ticket for the privacy and GDPR work?

I'm effraid not. I guess it would be a little banner in the bottom or top of the page mentionning that by browsing our site, they accept the tracking. Need to double check with the GDPR policy and with design team.
I will flag it to the team but for now we can consider it postpone to next release.

chrismclarke:
Turns out we already do! #722 and #619 both talk about this. I better close/merge these into a single issue to link to this PR also. @BenGamma , I can't see the issue you mentioned but if you point me in the right direction I can add it also

chrismclarke:
Ok, collated in #864

chrismclarke:
I think we should be about ready to go, although did you ever make the minor changes requested from the review @drydenwilliams ?

> I see you've only enabled for production, although it would be good to test on the dev site also. instead of checking `SITE === 'production' && GOOGLE_ANALYTICS_CONFIG.trackingCode` just check whether the code exists as this is only populated on staging and production branches.
> 
> Couple additional minor nits:
> 
> * can we decide a consistent naming convention for the tracking code/id. Google uses both terms but i think we should just pick one. Let me know if you want me to change the env variable to `TRACKING_CODE` or whether you will change to trackingId
> * call the component `GoogleAnalytics` in case we ever decide to change providers or offer multiple

chrismclarke:
Actually not to worry, I see they're there, approved

PR #870, opened 2020-01-13
Fix/v3 tests

chrismclarke:
Quite a big PR to fix/improve tests. Main changes:

## Done
- update bindings for new db structures
- add methods to automatically clean the db before/after tests
- add data to seed db from empty each run
- reduce timeouts to improve speed of tests
- add additonal timeout catches
- refactor firebase utils
- add back into ci pipeline
- various spec fixes and changes

## Notes
This turned into quite a tricky one due to the nature of cypress and the testing platform interact. Essentially cypress runs its own web-browser with the platform embedded as an iframe. The two communicate pretty well, although certain types of actions won't always be registered within the child if carried out in too quick succession on the platform (why cypress has it's custom chaining system). As such it's a bit of a balancing act  of timeouts and triggers.

In addition, the testing environment itself can prove problematic as not only do many of the tests rely on responses from outside sources (e.g. writing docs to firebase), but also mean that data populated in one test can confuse the next. To counteract this I've tried to add methods that clean both the test and server environments between tests, and add various timeouts and fallback mechanisms.

## Todo
- Add better tests for map specs
- Check for other missing tests
- Add tests for mobile
- Add support for parallel testing (ideally changing the database prefix to seed/clear just a subset of the database)
- Add more of the clearing/seeing methods to separate api endpoints (e.g. firebase functions) to make use of higher admin capabilities (e.g. creating/deleting batch auth accounts)

## Screenshots
image


chrismclarke:
Still working out a few bugs so should keep open until resolved

chrismclarke:
for whatever reason travis keeps running the same build even after I update code and submit, so will close here and open another full pr elsewhere tomorrow

PR #872, opened 2020-01-14
Fix/caching

chrismclarke:
Fixes a couple issues with caching

## Done
- Fix dexie query issue that prevented fetching only latest docs
- Add runtime caching support for the service worker to cache images downloaded for the map and howtos (or any other image extensions)
- Add a fallback for database queries which are otherwise server-first

## Todo
- Document how the main cache system works (for reference/interest of others) #873 
- Decide if any additional strategies are required for specific resource types (e.g. https://developers.google.com/web/tools/workbox/modules/workbox-strategies)

## Notes
There were various issues with CORS on image downloads, although I think they should now be handled with the service worker webpack config. It might be worth adding a note on the wiki to explain how to enable cors on firebase storage (as this is required).

Another weird quirk, only half of the images fully cache on load (possibly a limit of firebase request intervals or similar). They do cache eventually and from what I could see the images that don't are typically the smallest (<500B). Unlikely to be able to find a simple fix for this without refactoring our own image display components to execute fetch requests before passing (unless https://developers.google.com/web/tools/workbox/guides/advanced-recipes#warm_the_runtime_cache fixes it?)

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -557,8 +557,34 @@ module.exports = function(webpackEnv) {
       // the HTML & assets that are part of the Webpack build.
       isEnvProduction &&
         new WorkboxWebpackPlugin.GenerateSW({
+          cacheId: 'oa',
           clientsClaim: true,
           skipWaiting: true,
+          // NOTE 2020-01-14 CC - Add support to cache firebase storage and map tiles
+          runtimeCaching: [
+            {
+              urlPattern: /.*\.(?:png|gif|jpg|jpeg|webp|svg).*/gi,
+              handler: 'cacheFirst',
+              options: {
+                cacheName: 'oa-images',
+                fetchOptions: {
+                  credentials: 'same-origin',
+                  mode: 'cors',
+                },
+                backgroundSync: {
+                  name: 'oa-images-background',
+                  options: {
+                    maxRetentionTime: 60 * 24,
+                  },
+                },
+                expiration: {
+                  maxAgeSeconds: 60 * 24 * 7 * 30,
+                  maxEntries: 1000,
+                },
+              },
+            },
+          ],
+          // end update 2020-01-14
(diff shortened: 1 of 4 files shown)


chrismclarke:
> I wonder though how do I test this correctly : everything seems to work fine when I have a internet connexion, but as soon as I turn my wifi off and try to reload the page, the app crashes :

Yeah testing is a bit of a pain as by default the service worker isn't installed on localhost (would make developing a pain if changes weren't seen immediately).

The best way to test is create a local build and serve that via a web server (I usually do this on port 5000 as firebase also uses it when serving).

```
npm i -g serve
yarn build
serve -s build -p 5000
```

Now you'll see the full service worker (in debug mode as on localhost)
image


This should probably also be included in the docs todo #873

BenGamma:
Ok thanks for the instructions. Now no more crash of the app on the how-to list page.
I'm testing on firefox and one image of two seems to not being cached properly. See screenshot taken after a offline reload :

Capture d’écran 2020-01-15 à 15 26 01

_note : it's the same display with the rest of the list_

It seems like it's the whole how-to object is not available because when I click on the how-tos that doesn't have any cover image I can't load the howto page, whilst the ones with cover image, I can click on them and load the how-to page without crash. Are you experiencing the same ?

BenGamma:
Previous comment is not a crucial issue, I'll merge and add to current release as it improve & save a lot of bandwith already.
Offline mode was not in the scope and I consider it as extra feature. We'll upgrade it later on.

chrismclarke:
Just saw this, I think I mentioned in the PR that yeah I noticed this issue too - super weird, I'm guessing maybe something to do with max requests per second or something... but really don't know but would agree it's minor (it seems the images do cache eventually, and even if not the rest seems to be saving a fair amount of bandwidth)

PR #874, opened 2020-01-15
update messages posted for discord & slack webhooks

BenGamma:
Very little update on the messages posted by the webhooks, to fit the markdown used by each platform. Because yes they don't use the same 🤦‍♂


BenGamma:
Thad shouldn't happend since underscores are removed from usernames since the v3 upgrade. You can see that your username is now chrismclarke.

chrismclarke:
Also I'm guessing it would probably make sense to put the messages in their own file/function which can be pulled from both and any other future integrations. Also makes it easier if anyone ever wanted to make small adjustments, translations etc.

chrismclarke:
But no biggie

PR #880, opened 2020-01-20
Hotfix/cache all images

chrismclarke:
Small fix for issue outlined in pr #872, where for some reason only half the images cached.
Still unsure the root cause, but adding a second fallback cache seems to fix (see screenshot of all resources coming from cache).

image

Additionally added headers which will be set to anything in the firebase static folder (currently just icon and fonts), enabling cors support and some expiration.

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -563,8 +563,8 @@ module.exports = function(webpackEnv) {
           // NOTE 2020-01-14 CC - Add support to cache firebase storage and map tiles
           runtimeCaching: [
             {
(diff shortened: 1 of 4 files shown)


BenGamma:
Hey chris, thanks for working on a fix.
Could you give me more details on how to test this feature ?
Is it supposed to allow navigation even if network is down ? 
What I'm doing now to test is turn the wifi down when navigating on the how-to list and reload the page, but not sure it is the right way.

chrismclarke:
It's just an extension to #872 
See the discussion there for testing.
https://github.com/ONEARMY/community-platform/pull/872#issuecomment-574658876

You see your screenshot only has half the images cached... this should fix to make all of them cache

PR #881, opened 2020-01-20
Feat/privacy policy

### LINKED ISSUE #619: [feature request] - Analytics/Cookie Policy

chrismclarke:
closes #619 
Part of #864 

To go alongside pr #869

**Done**
- Add `/privacy` and `/terms` pages
- Add generated privacy policy and terms & conditions policy
- Add consent to sign-up with link to privacy and terms
- Updates sign-up test spec

**Todo**
Before merge into production/live it would be good notify PP core to check they're happy with how things are currently written. I've tried to make some small adjustments to fairly standard templates, but more more could be added. 

It would also be good to provide a permanent link to the pages in a header/footer (currently only viewable by direct link)

**Screenshots**
image

image

image


davehakkens:
nice one chris! Yeh looks like a good start!

PR #905, opened 2020-03-01
895 link username to profile

ginnie7:
Hello, I added linking to creator's profile. It works but the test probably expects plain text, not a link. Not sure how to adapt the test, any hint please?

### DIFF
--- a/cypress/integration/howto/read.spec.ts
+++ b/cypress/integration/howto/read.spec.ts
@@ -4,7 +4,6 @@ describe('[How To]', () => {
 
   describe('[List how-tos]', () => {
     const howtoUrl = '/how-to/make-glass-like-beams'
-    const howtoCreator = '/u/howto_creator'
     const coverFileRegex = /howto-beams-glass-0-3.jpg/
     beforeEach(() => {
       cy.visit('/how-to')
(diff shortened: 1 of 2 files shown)


BenGamma:
Hi @ginnie7 and thanks for the PR, looks good !
About the test have you seen [the section about testing in the wiki](https://github.com/ONEARMY/community-platform/wiki/Testing---Bugs) ? You'll find al the infos to test bugs locally.
After a quick look, it seems that the [first test](https://dashboard.cypress.io/projects/4s5zgo/runs/363/failures/1c898176-6b2b-426a-ae28-9ba088cb4780) is failing because you've added a test in the `[List How-tos]` section which is for the how-to list page, and there is no link to user profile on that page. The [second one](https://dashboard.cypress.io/projects/4s5zgo/runs/363/failures/b17bf277-cf89-4e94-bebb-03b7844423cf) is because you are not checking on the how-to page that the link to the user profile is present. So actually you just have to change the check you've added from `[List How-tos]` to `[Read a How-to]`, and the tests should pass nicely :v:

ginnie7:
Hi @BenGamma thank you for the hints! Looks like there was a problem with cypress reading the &nbsp space in the markup between "By " and the creator URL, so I replaced it by a normal space. Hope this is ok. Initially used &nbsp in case prettier causes a line break or something... But it looks ok now and tests are passing.
I'll read the wiki and cypress docs to learn the project better.

Issue #903, opened 2020-02-27
Remove double Discord and Slack Notifications 


davehakkens:
Description: 
Currently we get notifications on Slack and Discord and Slack using webhooks and Functions in Firbase. Method is great. But..
Problem:
1. **In Slack** we get multiple triggers per pin (image below). It might come because when people edit their profile. After every save it releases a notification. But not sure, it needs investigation.
2. **In Discord** we only want notifications for new pins, not edited (already existing ones). Currently when someone edits their pin or how-to, we get a notification in Slack. Approve it. And this triggers a Discord notification. This problem might be connected to the problem above. 

<img width="539" alt="afbeelding" src="https://user-images.githubusercontent.com/13672737/75468072-a70e3b80-59b2-11ea-8486-d5a0faac49ec.png">

alromh87:
Should we get notification after pin edition in slack, and just avoid discord notification when accepted? Or should the hole moderation should be avoided after first acceptance?

davehakkens:
This get complex. hehe
For us (in slack)  it's good to get a notification for edits of an accepted pin. Because we might need to re-approve. However we don not need to get a notification when someone edits their (not yet approved) unapproved pin

**So a user makes edit, we want a notification**
"⌛ waiting to be verified" + ✅ Notification
**If a users makes and edit, and another and another we really only need 1 notification**
"⌛ waiting to be verified" + "⌛ waiting to be verified" + "⌛ waiting to be verified"+ ✅ Notification
**If a user makes and edit, we approve, makes another edit we want notification**
"⌛ waiting to be verified" + ✅ Notification + "⌛ waiting to be verified"+ ✅ Notification

Does this still makes sense? :D

davehakkens:
And Discord really only has to come once.
When something is created, not editted (or re-approved)

alromh87:
Yep, makes sense, the slack part is done, will look into Discord functionality

PR #918, opened 2020-03-24
Bug903

alromh87:
First part of bug:

Avoid redundant notifications in slack when new pin needs moderation


BenGamma:
So if I understand well this is part of the fix for #903, the slack hook part. And if it works you'll adapt it to the discord hook ?

BenGamma:
Ok I merge so you can try on the staging env (https://dev.onearmy.world/)

PR #919, opened 2020-03-27
Improved how-to guidelines

davehakkens:
Added a few extra points + guidelines URL

### DIFF
--- a/src/pages/Howto/Content/Common/PostingGuidelines.tsx
+++ b/src/pages/Howto/Content/Common/PostingGuidelines.tsx
@@ -2,6 +2,8 @@ import * as React from 'react'
 import Flex from 'src/components/Flex'
 import Heading from 'src/components/Heading'
 import Text from 'src/components/Text'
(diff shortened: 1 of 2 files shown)


BenGamma:
I terms of code looks good to me, I just changed the link to a relative path so we could change the domain name if needed.
It's more for the text itself that I have little bits to say :
With your modifications it's like this :
Capture d’écran 2020-03-27 à 08 05 03

### Potential changes 1
We keep just one title, either "how does it work" or "how-to guidelines",
We use real emojis en only for the last point.
Capture d’écran 2020-03-27 à 08 06 57

### Potential changes 2
We keep just one title, either "how does it work" or "how-to guidelines",
We go full on emojis 🌮 🔢 🤖 
Capture d’écran 2020-03-27 à 08 04 29

What do you think ?

davehakkens:
I just copy pasted the text from Kat. thanks for checking.
I'd say full in emoji!

PR #922, opened 2020-03-28
Fix pr920 - clear indexDB local cache before automated test

BenGamma:
Since merge of #920 automated testings build were mysteriously failing, with various errors, always different.
That was linked to the [change of the IModerable interface](https://github.com/ONEARMY/community-platform/pull/920/files) in #920 and the fact that the IndexDB database interfaces cached in the automated browser used by cypress (in my case chromium) was different than the new ones.

In this PR I removed the need of modifiying the interface. We should modify interfaces only in major update cases because it will cause all our user to redownload the datas from db. With approximately 10 000 monthly visitors that will cost a lot of bandwith and more hosting cost too.
In this situation (update a webhook for internal use only), it doesn't appear worth to remove all caching.

Also I added an instruction to clear local indexDB cache at the beggining of the tests.

### DIFF
--- a/cypress/support/hooks.ts
+++ b/cypress/support/hooks.ts
@@ -9,6 +9,8 @@ import { Firestore } from './db/firebase'
  */
 
 before(() => {
+  indexedDB.deleteDatabase('OneArmyCache')
+  cy.clearLocalStorage()
   cy.wrap('Initialising Database').then({ timeout: 60000 }, doc => {
     // large initial timeout in case server slow to respond
     return new Cypress.Promise(async resolve => {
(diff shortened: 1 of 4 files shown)


chrismclarke:
I don't think the clearing should be necessary, Cypress does this automatically as mentioned:
https://docs.cypress.io/api/commands/clearlocalstorage.html#Syntax

But as long as things are working I'm not too worried either way.

BenGamma:
Yep it was clearing the localStorage but not the indexDB https://github.com/cypress-io/cypress/issues/1208

PR #928, opened 2020-03-30
Feat/#915 video embed

### LINKED ISSUE #915: Video Embed in how-to Step
## Dev Todo
- [x] Input field
- [x] Validation
- [x] Frontend component
- [x] Frontend case handling
- [x] Tests

### Description
Having the ability to embed a Youtube video instead of images as a step in the how-to
- Only Youtube is enough (no vimeo)
- Only 1 video per step
- Can't upload images + video
- Introduction (step 0) can't have video

### Page where it will be used
- how to create[ (example)](https://community-platform.webflow.io/u/create/how-to)
- how to single page ( [example](https://community-platform.webflow.io/u/how-to-article))

### Notes
Youtube has a few different URL structures. Share, Share with playlist, share on specific time shortlink etc. Below a few examples, all for the same video. We only need the basic structure. But people might copy/past another one. Can we make them all work, or should we ask users for one specific structure?

Same video, different urls:
1. https://www.youtube.com/watch?v=bsj6qHHLynk
2. https://youtu.be/bsj6qHHLynk
3. https://youtu.be/bsj6qHHLynk?t=60
4. https://www.youtube.com/watch?v=bsj6qHHLynk&list=PLE46712F10AA99DD4&index=1

### Mockup
video-create
video-frontend

chrismclarke:
closes #915

See issue for main implementation details, I also made a few minor additional changes:
- Image upload `isHovering` property was proving temperamental (wasn't showing the delete button correctly for me), so I found a simpler CSS selector method.
- Changed main form components for `PureComponent` - still seems to pass tests and renders changes less often (still 3x, but better than 10x previously and not really noticeable for performance)

### DIFF
--- a/cypress/fixtures/seed/howtos.json
+++ b/cypress/fixtures/seed/howtos.json
@@ -2214,6 +2214,14 @@
     "time": "< 1 week"
   },
   "gPpPDEvfNT9a6w5FWzaj": {
+    "moderation": "accepted",
+    "id": "me5Bq0wq5FdoJUY8gELN",
+    "_createdBy": "howto_creator",
+    "tags": {
+      "36hWyk3OckrLSH1ehdIE": true,
+      "Wk6RnHHFfKSiI71BlM8r": true
+    },
+    "title": "Make an interlocking brick",
     "steps": [
       {
         "images": [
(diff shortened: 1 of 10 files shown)


chrismclarke:
See there's a test failing, will fix shortly

PR #930, opened 2020-04-01
899 slideshow profile image

### LINKED ISSUE #899: Add Slideshow profile images
Description: Add a slideshow to the header image on user profiles. There is already some work done for this but needed some refinement and testing. @BenGamma do you know where/which branch the code is for the slideshow?
Location: User Profile 

- Max 4 images to upload.

afbeelding

BenGamma:
close #899

### DIFF
--- a/cypress/integration/settings.spec.ts
+++ b/cypress/integration/settings.spec.ts
@@ -358,8 +358,8 @@ describe('[Settings]', () => {
         country: 'United Kingdom',
         countryCode: 'gb',
         latlng: {
-          lat: 51.5073,
-          lng: -0.127647,
+          lat: 51.5156,
+          lng: -0.092,
         },
         name: 'City of London',
         postcode: 'EC1A',
--- a/cypress/integration/settings.spec.ts
+++ b/cypress/integration/settings.spec.ts
@@ -358,8 +358,8 @@ describe('[Settings]', () => {
         country: 'United Kingdom',
         countryCode: 'gb',
         latlng: {
-          lat: 51.5156,
-          lng: -0.092,
+          lat: 51.5073,
+          lng: -0.127647,
         },
         name: 'City of London',
         postcode: 'EC1A',
--- a/cypress/integration/settings.spec.ts
+++ b/cypress/integration/settings.spec.ts
@@ -50,7 +50,7 @@ describe('[Settings]', () => {
     cy.get('[data-cy=info-description')
       .clear()
       .type(info.description)
-    cy.get('[data-cy=cover-image]')
+    cy.get('[data-cy=cover_image-1]')
       .find(':file')
       .uploadFiles(info.coverImage)
   }
--- a/src/pages/Howto/Content/Common/Howto.form.tsx
+++ b/src/pages/Howto/Content/Common/Howto.form.tsx
@@ -83,7 +83,7 @@ const Label = styled.label`
 
 @inject('howtoStore')
 @observer
-export class HowtoForm extends React.PureComponent<IProps, IState> {
+export class HowtoForm extends React.Component<IProps, IState> {
   uploadRefs: { [key: string]: UploadedFile | null } = {}
   constructor(props: any) {
     super(props)
--- a/src/pages/Settings/SettingsPage.tsx
+++ b/src/pages/Settings/SettingsPage.tsx
@@ -100,14 +100,21 @@ export class UserSettings extends React.Component<IProps, IState> {
       isLocationSelected: true,
     })
   }
-  public onCoverImgChange(v: IConvertedFileMeta) {
+  public replaceAt(array, index, value) {
+    const ret = array.slice(0)
+    ret[index] = value
+    return ret
+  }
+  public onCoverImgChange(v: IConvertedFileMeta, i: number) {
     // TODO always use an array of coverImages (will do when allow multiple cover images)
-    const coverImagesArray: IConvertedFileMeta[] = []
-    coverImagesArray.push(v)
+    const coverImagesArray = this.state.initialFormValues.coverImages
+      ? this.state.initialFormValues.coverImages
+      : []
+    // coverImagesArray.push(v)
     this.setState({
       customFormValues: {
         ...this.state.customFormValues,
-        coverImages: coverImagesArray,
+        coverImages: this.replaceAt(coverImagesArray, i, v),
       },
     })
   }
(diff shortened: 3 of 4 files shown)


chrismclarke:
@BenGamma  -  The merge into master hasn't been reviewed, is there a reason for this?

BenGamma:
Since it was not a sensible code addition I merged without review yes, just with automated testing. If there is some part you noticed that should be refactored I can still update this PR

cypress/integration/settings.spec.ts
```diff
@@ -50,7 +50,7 @@ describe('[Settings]', () => {
     cy.get('[data-cy=info-description')
       .clear()
       .type(info.description)
-    cy.get('[data-cy=cover-image]')
+    cy.get('[data-cy=cover_image-1]')
```
chrismclarke, line 53: It would be good for `data-cy` refs to also match the variable in the form model where possible (easier to update later and clearer to cross-search). Would suggest same as `coverImages[0]`

src/pages/Howto/Content/Common/Howto.form.tsx
```diff
@@ -83,7 +83,7 @@ const Label = styled.label`
 
 @inject('howtoStore')
 @observer
-export class HowtoForm extends React.PureComponent<IProps, IState> {
+export class HowtoForm extends React.Component<IProps, IState> {
```
chrismclarke, line 86: Just to check, are things not working as intended with PureComponent? The forms currently re-render really inefficiently so if savings can be made it would be a good thing, either through use of pure components or `shouldComponentUpdate`

src/pages/Settings/SettingsPage.tsx
```diff
@@ -100,10 +101,40 @@ export class UserSettings extends React.Component<IProps, IState> {
...
-    coverImagesArray.push(v)
+  public replaceAt(array, index, value) {
+    const ret = array.slice(0)
+    ret[index] = value
+    return ret
+  }
+  public onCoverImgChange(v: IConvertedFileMeta, i: number) {
+    // Use existing array of images if exist & not empty or use empty array
+    let coverImagesArray =
+      Array.isArray(this.state.initialFormValues.coverImages) &&
```
chrismclarke, line 112: I'm guessing this is to make up for older cases where cover images are in a single json object format? If so please add a comment to explain why we need this check and a TODO item/git issue to migrate all profiles to a consistent format in the future (otherwise looking back this could be very confusing and messy).

src/pages/Settings/SettingsPage.tsx
```diff
@@ -100,10 +101,40 @@ export class UserSettings extends React.Component<IProps, IState> {
...
+      this.state.initialFormValues.coverImages.length
+        ? this.state.initialFormValues.coverImages
+        : []
+    // If value is null || undefined && coverImagesArray exist
+    // We want to remove the element from coverImagesArray at given index (delete image)
+    if (
+      (v === null || v === undefined) &&
+      (Array.isArray(coverImagesArray) && coverImagesArray.length)
```
chrismclarke, line 120: You've already performed this check earlier so I don't see why it needs repeat

src/pages/Settings/content/formSections/UserInfos.section.tsx
```diff
@@ -116,18 +122,76 @@ export class UserInfosSection extends React.Component<IProps, IState> {
...
             </Text>
-            <Box height="100px" width="150px" data-cy="cover-image">
+            <Box height="100px" width="150px" m="10px" data-cy="cover-image">
+              {/* <Field
+                  canDelete
+                  hasText={false}
+                  name={`${step}.images[0]`}
+                  src={images[0]}
+                  component={ImageInputField}
+                /> */}
```
chrismclarke, line 132: Unused code can probably be deleted

src/pages/Howto/Content/Common/Howto.form.tsx
BenGamma: I changed to a PureComponent to remove this warning

Capture d’écran 2020-04-11 à 23 29 09

I didn't know there was a performance issue with `Component` ?

src/pages/Settings/SettingsPage.tsx
BenGamma: Yes it's for older cases but no they were not in json format but already an array of images. That's why I'm double checking the length also.
Actually there was a first implementation a while ago of the multiple cover images upload feature, but at the time we had some troubles with the image upload component.
So I removed the ability to upload multiple but kept the array to make it easier when we'll have the image upload component ready.

src/pages/Settings/SettingsPage.tsx
BenGamma: That was fixed with #931

src/pages/Settings/content/formSections/UserInfos.section.tsx
BenGamma: Also removed with #931

BenGamma:
Thanks for review Chris. For my defense (see comments below), there was no change from json object to array, `coverImage` filed was already an array.
But yes totally agree on the review process, we'll be more chilled on the Merge button in the futur 😅

src/pages/Howto/Content/Common/Howto.form.tsx
chrismclarke: It's not necessarily a problem with component, more how component can be implemented. By default react components update whenever any props or state updates (even if using setState with the same values), which is both good (you don't get missed re-renders) and bad (quite often you get additional re-renders when not needed). Whilst building the platform we mostly used these a) because they just work, and b) because we didn't know much better. Particularly for some of the deeply nested form components and inefficient forms (which sometimes update all other fields whenever one changes), this turns out to be super inefficient (I've seen 20 renders when just one form input typed) and so good to slowly move away from. The ways to improve are either by adding a `componentShouldUpdate` method to better specify when re-renders should happen. or use pure components which automatically does a shallow comparison to check if props or state have changed before re-render. Pure components are usually the easiest quick fix, although have a slight risk if the data is more complex (e.g. deeply nested objects) then it could miss the re-render.

See more at: https://reactjs.org/docs/react-api.html#reactcomponent

Also, with React 16.6 theres also react `memo`, which handles similar to pure components but for functional components. https://reactjs.org/docs/react-api.html#reactmemo

src/pages/Settings/SettingsPage.tsx
chrismclarke: Ok, so if we've always used arrays then we only need to check if the field exists. I expect these 35 lines of code could be tidying to more like 10. But we can tidy up at another point along with ensuring that all profiles are actually in a consistent format to make the code easier to work with.

chrismclarke:
Yeah as I was going through I realised that actually what is needed is a bit of tidying up higher up the chain, e.g. we should probably ensure all user profiles have a consistent format before passing anything down to the profile edit component. Might try make another PR for that stuff in the future.

src/pages/Howto/Content/Common/Howto.form.tsx
BenGamma: Thanks for making it clearer !

PR #931, opened 2020-04-01
fix case for empty coverImage profile

BenGamma:
Fix a bug introduce with #930 when no cover image was uploaded, it will only add 1 cover image.


chrismclarke:
@BenGamma - I see this has been merged without review and with failing tests. Is the code 100% working as intended?

BenGamma:
These failing test are weird. I had to manually restart the build and they all passed https://dashboard.cypress.io/projects/4s5zgo/runs/422/specs
Tests were also 100% passing on my local machine.

PR #940, opened 2020-04-11
Make contact details madatory

tudi2d:
- Force at least 1 field to be shown

### DIFF
--- a/src/pages/Settings/SettingsPage.tsx
+++ b/src/pages/Settings/SettingsPage.tsx
@@ -61,6 +61,14 @@ export class UserSettings extends React.Component<IProps, IState> {
       user && user.profileType
         ? toJS(user)
         : { ...toJS(user), ...INITIAL_VALUES }
+    if (!initValues.links || (initValues.links && !initValues.links.length)) {
+      initValues.links = [
+        {
+          label: '',
+          url: '',
+        },
+      ]
+    }
     this.state = {
       initialFormValues: initValues,
       customFormValues: {},
--- a/src/pages/Settings/content/formSections/Fields/Link.field.tsx
+++ b/src/pages/Settings/content/formSections/Fields/Link.field.tsx
@@ -14,6 +14,7 @@ interface IProps {
   index: number
   mutators?: { [key: string]: (...args: any[]) => any }
   initialType?: string
+  hideDelete?: boolean
   onDelete: (index: number) => void
 }
 interface IState {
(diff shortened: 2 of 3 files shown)


src/pages/Settings/SettingsPage.tsx
```diff
@@ -61,6 +61,14 @@ export class UserSettings extends React.Component<IProps, IState> {
       user && user.profileType
         ? toJS(user)
         : { ...toJS(user), ...INITIAL_VALUES }
```
tudi2d, line 63: Why is the INITAL_VALUE only added with the spread operator when the user is not defined/ not loaded? The INITAL_VALUE already includes an empty link.

chrismclarke: Yeah I'm having some trouble with this too. I'll add a separate issue to do some housekeeping and tidy up.

src/pages/Settings/content/formSections/Fields/Link.field.tsx
```diff
@@ -106,12 +107,16 @@ class Link extends Component<IProps, IState> {
...
               this.setState({ linkType: v })
             }}
-            placeholder="type"
+            placeholder="Type"
             style={{ width: '160px', height: '40px', marginRight: '8px' }}
           />
-          <DeleteButton
-            sx={{ display: ['block', 'block', 'none'], height: '40px' }}
-          />
+          {!hideDelete ? (
```
chrismclarke, line 113: Instead of the empty ? : () format, you could just say `hideDelete && (...)`

src/pages/Settings/content/formSections/UserInfos.section.tsx
```diff
@@ -205,6 +205,7 @@ export class UserInfosSection extends React.Component<IProps, IState> {
                     }
                     link={name}
                     index={index}
+                    hideDelete={!index}
```
chrismclarke, line 208: if hideDelete only depends on the index, and we're passing the index do we actually need this variable at all?  Can it not just be based on the index in the component also...?

tudi2d: Oh. Yes, true. Thanks for mentioning that!

chrismclarke:
Thanks for the work on this, as it happens I made a big code tidy to the user settings module (or #942) and mostly included the code there, so closing this PR and marking issue as resolved.

PR #942, opened 2020-04-15
Code Tidying - Users, Profiles

### LINKED ISSUE #842: [bug] MOBILE - User profile settings
**Describe the bug**
When I change focus of my profile (for example: from member to workspace, but any combination applies) I am asked to confirm because my profile will be reset. If I select "Cancel" (to not change) the new focus I've just selected stays selected and looks like has been changed anyways.

chrismclarke:
closes #842 

@BenGamma,
Apologies in advance (!)
Whilst trying to review a few other PRs I took a dive into the general user/profile/settings code and found a few bits in need of tidying. There's quite a lot going on in the files because I kept coming across other minor issues that needed resolving (I ended up spending way longer on this than planned). Overall I found myself refactoring/reducing the code quite a lot (almost half), briefly outlined as:

`cypress/integration/settings.spec`
Not sure why but links are described by `type` whilst profile are described by `label`. I'd actually prefer using `type` to describe a link type (e.g. social-media), but as many profiles already use `label` I've just kept it common and updated the mocks.

`ImageInput` & `ImageInputField`
I think laziness when I first made this, I half supported multiple files and half didn't (I think you could allow a multiple file picker but it would only ever return the first file). I've corrected this and updated a few bindings accordingly. There was also various toggle/tracking properties that were a bit confusing and didn't always do anything. 

`Validation`
I’ve tried to move more methods to use the validators you made (like required) vs manually calculating and passing as props. The benefit to these methods is when the question is no longer rendered the validation disappears. But I think I may have missed a few so would be good to get a double check. I’ve also added improved array validation for the cover images and links and tried to add more validation messages where they hadn’t been included.

`misc/others`
Submit button now disabled when errors exist. Hovering on button shows short list of errors (mostly for debugging). Saving now has a callback in case save fails. Minor test adjustments (was surprised a couple of them were passing at all!). Various minor typescript and logs tidying. 

`IUser` & `IUserPP`
This again was a bit messy as there was quite a bit of overlap between the two and not always clear what fields a given user would actually had defined. I've done my best to ensure that all users will have the basic IUser fields provided (could do a future database update to fully clean, but for now it just merges the user on top of default fields to fill in any that don't exist), and then all users who have created a profile will have the rest of the IUserPP fields.

`settings/content/formSections/...`
A lot of these had `initialValues` props which were a bit misleading as they were also the full changing form values (not just initial), so renamed to `formValues`. There was also some complicated logic to keep account of users who didn't have correct fields, and also to keep separation between unsaved form fields and updated form fields. Most of this has now gone with the user tidying mentioned above. 
Also most components were passing the initial props instead of the forms own inner props, which are more accurate so they have been changed to `values`.

`formSections/Link.field`
This has been renamed to `ProfileLink` to not confuse with React's router links, and again some of the non-used code removed. 

`formSections/focus` and `workspace`
I’ve wrapped the component with `field` to make it easier to directly interact with the form and update values (without need for onInputChange style callbacks) – makes the code quite a bit simpler. I have a feeling that there will be many cases of confusing patterns related to final-form throughout the repo as quite simply there’s too many different ways to do things and the documentation is poor… but it’s slowly improving at least.

`Settings` (general)
There was some slightly complicated logic to clear part of user-profiles when changing profile type (to account for previous challenge of partial profiles). I've now set it to just keep all previously filled fields - that way if a profile changes from workspace to community-builder and back again they'll still have the information previously included about their workspace and we avoid deleting parts of profiles. Also removed some repetition in how fields are displayed and made it so mobx doesn’t try to track everything so strictly via a single `toJS` in top level (needed to avoid warnings on array push methods)

`CoverImages`
Converted from 4 standalone fields to a single field-array to allow simpler logic, with a few small helpers to ensure 4 fields exist and empty images aren't uploaded. Also a few more bits of general tidying/variable removal and proper value tracking for fieldnames/values

None of this is to add new features, just improving the general codebase which is important for building on in the future. I know a lot of the existing code had been written by you to handle lots of edge base scenarios (and definitely don’t envy you for that!), so I'd appreciate if you could find some time to test and see if things actually work as should – I’ve spent time to check/ensure tests pass but a manual check would also be appreciated, although am worried I might have lost some of the validation methods.

TODO
-	collection and expertise sections could be tidied in same way (replacing validation).
-	Ensure `validateFields:[]` appears on all fields (to prevent unnecessary validation checks)


src/mocks/Selectors.tsx
```diff
@@ -12,7 +12,7 @@ export const COM_TYPE_MOCKS = [
     label: 'website',
   },
   {
-    value: 'social-media',
+    value: 'social media',
```
BenGamma, line 15: why removing the dash here ? I think this `value` field populate the form `value` attribute.

chrismclarke: I thought so too but turns out everywhere else has it without a dash... I think the mocks was actually using it's own system that was different to the main repo

src/pages/Settings/SettingsPage.tsx
```diff
@@ -60,58 +59,88 @@ export class UserSettings extends React.Component<IProps, IState> {
...
+  /**
+   * Check for additional erros not caught by standard validation
+   * Return any errors as json object
+   */
+  public validateForm(v: IUserPP) {
+    const errors: any = {}
+    // must have at least 1 cover (awkard react final form array format)
+    if (!v.coverImages[0]) {
+      errors.coverImages = []
+      errors.coverImages[ARRAY_ERROR] = 'Must have at least one cover image'
```
BenGamma, line 110: Very good that you made the error handling more clear, I'm just not sure the messsage is displayed at any time to the user ?

chrismclarke:
Thanks for catching the flags, I had completely missed that one. I reviewed the flag code and realised it's probably a bit unnecessary to first convert the country code to a name, and then convert back for display (we can just keep country code). I think this might have been just to make the tests easier to manage. In any case, I've made that change, removed unnecessary code and refactored to its own component. I'm aware this will break the flags for existing users, but I think it's worth the simplification and easy enough for a user to just select from their profile again (or we can run a db update operation)

As for upload error images, yes I hadn't included the bindings, it's automatically included in the `meta` that's sent along with field data. Now included. I couldn't figure out how to get it to first check whether the field has been interacted with before showing a message, but I don't think that's a major issue.

The tests are still annoyingly a little bit inconsistent. I've noticed two issues, one being the validation sometimes the validation triggers just as something else is being clicked and can throw it off, the other is a (known) issue where some versions of cypress don't always type full inputs consistently (https://github.com/cypress-io/cypress/issues/5480). A couple future workarounds might be to avoid typing and instead just setting values on input boxes, or adjusting the default typing letter delay. For now I’ve added a plugin that automatically retries failed tests. I’ve also added support for parallel testing to try and bring overall test time down – kinda off track for this PR, but I’m deep enough down this rabbit hole that it’s hard to find a way out! (keep digging…)

In any case the most recent test pass ran successfully, so as long as you're happy with what I mentioned about country codes then should be good to merge and finally see the back of this PR!

src/utils/helpers.ts
```diff
@@ -120,22 +120,6 @@ export const isAllowToPin = (pin: IMapPin, user?: IUser) => {
   }
 }
 
-/************************************************************************
- *             Country code to country name converters
- ***********************************************************************/
-export const getCountryCode = (countryName: string | null | undefined) => {
```
BenGamma: The problem with removing these method is that is was used to display the default country when user have already selected an Icon. The lib react-flags-select [takes a country code](https://www.npmjs.com/package/react-flags-select#default-country) as value for the props `defaultCountry`.
Because now even when the user have selected a country, since the country name is injected and not the country code, the field never display the existing value.

src/utils/helpers.ts
```diff
@@ -120,22 +120,6 @@ export const isAllowToPin = (pin: IMapPin, user?: IUser) => {
...
-  const countryCode = Object.keys(countries).find(
-    key => countries[key] === countryName,
-  )
-  return countryCode ? countryCode.toUpperCase() : undefined
-}
-
-export const getCountryName = (countryCode: string) => {
```
BenGamma: This method though seems not used anymore, you were right to delete 👍

BenGamma:
Thanks for including the meta errors of cover images 👍 

About the test yeah I think may be good to separate the work on that side. Keep the work done here, but since it's already kind of a massive PR let's try to not add to it for our own sanity 😅 

Can you open separated issue(s) for what you've noticed on the cypress side of things ?

A last little edge case I just noticed with profile that I think is worth mentionning :
Select « I run a workspace » as profile type. Then Don’t select the workspace type straight away and add your map description and user infos. Come back to workspace type and select it. Then the save button is disable, saying the workspaceType have error.

src/utils/helpers.ts
chrismclarke: As I mentioned above, both `react-flags-select` and `react-flag-icon-css` expect country codes, so there really isn't any good reason that I can see to ever keep country names - particularly as these are somewhat language-specific and subject to change (just go speak to someone from Swaziland, or should I say the kingdom of Eswatini!). 

If an existing user with a named country goes to edit a profile, yes that will be passed as a prop but it just means no flag will be displayed (has a fallback method in case of no match). I think this is fine, better to introduce this change now then further along the line.  The user can either manually reselect their country which will populate the correct code, or we can run an operation of the db to automatically migrate.

chrismclarke:
Yeah I've set up a new branch for cypress stuff, although found myself needing to put a few things in here as otherwise the tests wouldn't pass (and can only see updated ci if making changes to a branch with an open PR). Will continue work, make a new PR at some point and outline any existing known issues. 

Good catch on the workspace type. This is because by default our validation only runs on blur (saves running everytime somebody types a letter into an input box), so any custom field handlers need to also trigger the `input.onBlur()` method after `input.onChange(...)`.  Have now fixed.

PR #951, opened 2020-04-19
hotfix - tsconfig minor tweaks

chrismclarke:
Latest build on master failing due to config errors with how `functions` shares some type definitions with the main `src`. Not really sure why this has happened now, but it's always been a bit of a fragile system so not hugely surprised.

This attempts fix by:
- update typescript version in functions to match src
- add project references tsconfig support
- add typing build script for project

In the future we can hopefully achieve a better/more-permanent fix via `lerna` and/or `yarn workspaces`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ functions/backup.json
 
 # production
 /build
+/lib
 
 # misc
 .DS_Store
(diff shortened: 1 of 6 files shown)


chrismclarke:
@BenGamma - As it's only config changes will merge without review and see if it fixes master. If not can always revert later.

chrismclarke:
Now working, although note that this does mean an additional lib folder is built alongside main build folder (minor issue, likely need to handle `noemit` correctly although possible issues discussed here: https://github.com/microsoft/TypeScript/issues/36917)

Issue #451, opened 2019-04-22
[optimisation] - code split main bundle with webpack


chrismclarke:
**Is your feature request related to a problem? Please describe.**
Currently the full app code is bundled together and loaded in a single file (still quite small, less than 1MB but likely to grow). Ideally we want it to be split with different pages as entrypoints. I think this is done automatically with create-react-app, so it might be worth investigating un-ejecting the app to use default build again (I think most of the reasons to eject are now deprecated, but will need to review and test).

chrismclarke:
Note, if doing so can also revisit current service worker behaviour if following issue resolved:
https://github.com/facebook/create-react-app/issues/5316

chrismclarke:
Another couple to consider:

https://www.styled-components.com/docs/tooling#babel-plugin
https://www.styled-components.com/docs/tooling#babel-macro
(allow custom css property on styled components)

https://github.com/timarney/react-app-rewired
https://github.com/withspectrum/react-app-rewire-styled-components
(cra overrides for working with things like styled components without eject)

chrismclarke:
Note, custom code currently added to ejected version to support skipWaiting (`config/webpack.config.js` L561), suggest waiting until good solution to above issue solved before any uneject

chrismclarke:
Update to skipWaiting, now possible through postMessage:
https://github.com/facebook/create-react-app/issues/5316#issuecomment-496292914

chrismclarke:
Note - whilst unejecting would be good to do some housekeeping.
Can use tools like [depcheck](https://www.npmjs.com/package/depcheck) to analyze unused dependencies

chrismclarke:
closing as we've been using lazy-loaded bundles for quite some time

PR #952, opened 2020-04-20
Refactor - CI Backend (+db prefixes)

### LINKED ISSUE #611: Define browser matrix [feature request]
It would be great to know what browsers we are going to support e.g. 

- Chrome - v-x-x
- Firefox - v-x-x
- Safari
- Internet Explorer

Once we have done this we can explore any polyfills needed and also code for older versions if needed. 

I raise this as we have used `object-fit` which is not supported on IE11 and looking at these stats from Africa (https://gs.statcounter.com/os-market-share/all/africa) Windows the most common OS.

---

From @chrismclarke point: 

> I would suggest opening an issue to include defining the platforms we would like to support, and if possible good ways to help automate testing on them

chrismclarke:
@BenGamma 
To follow-up on some of the work done and issues identified in #942, focuses on improvements to the automated testing and deployment system

## Main Changes
- Update cypress (should fix some of the input typing bugs)
- Add firefox browser testing support (closes #611)
- Simplify many of the scripts used in `package.json` to only include what is minimally required
- Add ci scripts directly to `.travis.yml`
- Tidy Travis scripts and caching (faster performance)
- Add parallel script functionality (now runs 4 machines in parallel, 2 for chrome testing and 2 for firefox. This roughly halves the total testing time whilst performing twice as many tests)
- DB prefix support: In order to allow tests to run in parallel some refactoring had to be done to organise how we reference database paths. We had a somewhat messy system where whenever wanting to make major changes we would manually prefix `v2` / `v3` across lots of places. Now there is a single variable populated throughout all src code (although functions still manually specified). This means when testing we can use one database but have `test1_events` , `test2_events` etc. so that the tests do not interfere with each other as they create and delete data.

Additionally have added minor revisions to testing info doc: https://github.com/ONEARMY/community-platform/wiki/E2E-Testing

## Additional Changes
- Minor fixes to tests incompatible with firefox (e.g. getting iframe contents for video embed) or that were working with old cypress but shouldn't have and broke (e.g. comparing uploaded images, said objects were the same despite meta changing such as downloadUrl populated on upload)
- Remove event title validation (as we should be allowing for events with the same name, e.g. recurring meetup, and as it was set to check uniqueness against howto slugs and not even events :p)
- Change a few existing environment variables and add script to make it easier to add variables to test or other environments (removes prior additions to custom webpack build config which will be required for whenever we revert back to the non-ejected cra scripts #451)
- Remove old user migration API, as it's not longer useful and potentially dangerous to have exposed (people could reset other's profiles).

## Breaking Changes
- Hopefully none :D

## Todo
No immediate outstanding issues, however a few flagged as good areas for development following recent work: https://github.com/ONEARMY/community-platform/issues/953


chrismclarke:
image
Example of ci-database, testing with different prefix

chrismclarke:
image
Example running the full test suite in parallel on firefox and chrome, 2 machines each

chrismclarke:
image
Example parallel processes in travis

BenGamma:
Thanks chris ! That seems like a needed ground work for v2 !
About the db prefix upgrade, something worth mentioning is that we will have to enforce the reloading of the cache for all our users (3201 today). Considering the how-to landing page is ~20Mb without cache and 8Mb with cache, better to be sure we do it just once.
I will have a closer look by the end of the month 👀

chrismclarke:
Yeah I've been thinking about that also, a couple specific points
1. 20MB is unzipped size, but as we use a server with Gzip it's actually about half of that, say 10MB (still huge I know)
2. The database is only a tiny part of this. If you look at the xhr for db requests it's actually only 1KB, so that is all that gets resent on a db update
image

The rest of the size come from a combination of core platform (~1MB) and loading images (~9MB). The core platform will get redownloaded after every new release and images only after a manual image cache wipe (so pretty much never). Seeing as we do still have quite a large number of new vs returning users then 10MB on new user first visit is still pretty significant (with return visits weighing around 1MB per month assuming we release that often - which feels very small for a frequent user, and around average weight for irregular user).

But I definitely think we can bring those numbers down and that's the plan for the next PRs (see plans in #953 and #956). Speficially, I expect we can have the biggest impact by:
- Preventing all howto images rendering for a new user. We can do this either by using a virtual-scroll component to only render as a user scrolls, or by adding a pagination system to show results 20 at a time. 
- Consider refactoring how we process thumbnails - currently we download the full cover image so that when a user opens a howto they don't have to download it again, but if most users don't open most photos then probably having a second generated thumbnail would be lighter weight.
- Break up core bundle to lazy load pages on demand
- Preventing howto/event modules loading when page not visited (change how stores are injected)

chrismclarke:
Tried to include some of the stuff mentioned here in a more-permanent write-up
https://github.com/ONEARMY/community-platform/wiki/Environmental-Impact

BenGamma:
Good call for the improvements of the testing workflow (I especially like the parallel testings and firefox compat 💪 )
I see there is a conflict with a cypress file, did you spotted it ?

chrismclarke:
Yeah the conflict was on purpose as I previously commented out some broken testing that this PR fixes. I think at this stage it should be good to go (tests are all passing) so I'm going to merge into master and can do some final testing there before release.

PR #954, opened 2020-04-20
Ci fail test

chrismclarke:
This is just to test that test failure correctly propagate to the main build. It will not be merged.


chrismclarke:
Fixed correct fail conditions and included with  #952. Closing.

PR #962, opened 2020-05-15
#493 - Add virtual scroll for HowtoList

### LINKED ISSUE #493: Virtual Scroll Component
### Description

When displaying large lists of data it would be more efficient to only render what is in the current viewport. Alternative would be to use infinite scroll to systematically query more data however this takes more of a technical implementation and doesn't hold much benefit (slightly smaller db requests but by no means significant, and could still be implemented on top)

### Page related

Will be used in :

- Events list (possibly also how-to)

### Example(s)

https://bvaughn.github.io/react-virtualized/#/components/List

### Build suggestion

Suggest using the package support by airbnb:
https://github.com/bvaughn/react-virtualized

tudi2d:
PR Type

PR Checklist

- [x] - Passes Tests

## Description

It adds a virtual scroll for the HowtoList to reduce initial rendering and only render the needed Howto´s for every visitor. It will render additional Howto´s while rendering, while removing Howto´s which are not in the viewport.

Therefor I added react-virtualized as discussed in the Issue. First I thought react-window would be good enough, but it does not support a Window scroll feature and there for react-virtualized was my choice here.

## Git Issues

Closes #493

### DIFF
--- a/src/pages/Howto/Content/HowtoList/HowtoList.tsx
+++ b/src/pages/Howto/Content/HowtoList/HowtoList.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react'
 import { Flex } from 'rebass'
+import { List, WindowScroller } from 'react-virtualized'
 // TODO add loader (and remove this material-ui dep)
 import { Link } from 'src/components/Links'
 import TagsSelect from 'src/components/Tags/TagsSelect'
@@ -13,6 +14,7 @@ import MoreContainer from 'src/components/MoreContainer/MoreContainer'
 import HowToCard from 'src/components/HowToCard/HowToCard'
 import Heading from 'src/components/Heading'
 import { Loader } from 'src/components/Loader'
+import themes from 'src/themes/styled.theme'
 
 interface InjectedProps {
   howtoStore?: HowtoStore
@@ -21,6 +23,11 @@ interface InjectedProps {
 interface IState {
   isLoading: boolean
 }
+
+const breakpointsInPX = themes.breakpoints.map(
+  // From em string to px number
+  breakpoint => Number(breakpoint.replace('em', '')) * 16,
+)
 
 // First we use the @inject decorator to bind to the howtoStore state
 @inject('howtoStore')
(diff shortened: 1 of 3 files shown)


src/pages/Howto/Content/HowtoList/HowtoList.tsx
```diff
@@ -22,6 +24,11 @@ interface IState {
   isLoading: boolean
 }
 
+const breakpointsInPX = themes.breakpoints.map(
```
tudi2d, line 27: I was not sure here how to access the theme properties the right was. Is there an alternative to this?

BenGamma: Yep that's it :)

BenGamma:
Hey @tudi2d thanks a lot for your work !
This feature was getting very needed as the number of how-tos on kept on growing pretty fast ! Also since we don't have a home page yet, users where landing on this page so this will save quite a lot of bandwidth 🌍 
It looks good from my side, I assign chris to be sure he double check since he was the one folling up on the issue.
Thanks again 💪

chrismclarke:
Thanks again for the work on this and sorry for my slow response, been a bit swamped for the past week. Will review this weekend and get back to you.

chrismclarke:
Apologies again for the slow reply to this. Took a while to understand what was going on (particularly why to divide up chunks instead of using masonry or grid, but then realised for use within the window scroller), but now it makes a lot of sense and works well. 

I've tidied up the code a bit, but as I can't add commits on top of your fork pr I'll merge this and then create another pr with the changes.

Thanks again

chrismclarke:
@tudi2d - We're starting to setup a bug bounty programme to show appreciation to developers who contribute to the platform. This issue was nominated as one to reward for May, so if you send me your Paypal details I can add you to the list for future payments. Either email me at chris.m.clarke@live.co.uk or find me on PP slack or discord as chrismclarke. 

The amounts are pretty small at the moment as we don't haven't much funds for the platform yet (so possibly not time to quit the day-job just yet!), but at least it's a start and hopefully a step in the right direction.

PR #969, opened 2020-06-01
Virtual scroll code tidying

chrismclarke:
PR Type

- Code Tidying

PR Checklist

- [x] - Passes Tests

## Description
Following #962 (thanks @tudi2d!), some minor code tidying
- swap getter function with regular (as it's used in multiple places it runs the code multiple times)
- simplify row/column calculation methods
- add overscanrowcount property (change default lookahead from 10 rows to 2)

## Screenshots/Videos
Probably deserves one as really neat how images now only load in as they scrolled past!

### DIFF
--- a/src/pages/Howto/Content/HowtoList/HowtoList.tsx
+++ b/src/pages/Howto/Content/HowtoList/HowtoList.tsx
@@ -1,20 +1,16 @@
 import * as React from 'react'
-import { Flex } from 'rebass'
-import { List, WindowScroller } from 'react-virtualized'
-// TODO add loader (and remove this material-ui dep)
+import { Flex, Box } from 'rebass'
 import { Link } from 'src/components/Links'
 import TagsSelect from 'src/components/Tags/TagsSelect'
-
 import { inject, observer } from 'mobx-react'
 import { HowtoStore } from 'src/stores/Howto/howto.store'
 import { Button } from 'src/components/Button'
-import { IHowtoDB } from 'src/models/howto.models'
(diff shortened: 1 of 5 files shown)


src/pages/Howto/Content/HowtoList/HowtoList.tsx
```diff
@@ -144,19 +129,17 @@ export class HowtoList extends React.Component<any, IState> {
...
-                    height={height}
                     width={window.innerWidth}
-                    isScrolling={isScrolling}
-                    onScroll={onChildScroll}
-                    scrollTop={scrollTop}
-                    rowCount={this.filteredHowtoChunks.length}
+                    rowCount={filteredHowtoRows.length}
                     rowHeight={410}
+                    overscanRowCount={2}
                     rowRenderer={data =>
```
BenGamma, line 139: Here is there a way we use the [onResize](https://github.com/bvaughn/react-virtualized/blob/master/docs/WindowScroller.md#prop-types) prop to update the width of the `filteredHowToRow` ?

chrismclarke: Thanks, yeah I've actually done that locally but forgot to push. I might add a couple more commits and re-request

chrismclarke:
Ok, so updated:
- Moved to own component to (hopefully) make more reusable in future and leave howto code tidier
- Add method to recalculate on window resize
- Fix max columns (now 3)
- Auto calculate heights based on row content
- Add typings

Known issues:
- Row height calculation seems to get the top/bottom padding slightly wrong. Still looks fine but would be good to fix in the future.

PR #970, opened 2020-06-02
fix profile double type and links bugs, tidy profile code

### LINKED ISSUE #818: [bug] Profile with "double type" after editing
I made first a sheetpress account. Then switched to member. Attached you can see what my profile says now.
unnamed

### LINKED ISSUE #968: [bug] - Profile links without full url (e.g. https://) don't link correctly
**Describe the bug**
When a profile enters a link (e.g. instagram.com/user) in their profile the link will only be appended to the current website (e.g. https://community.preciousplastic.com/u/instagram.com/user). This is the default behaviour of `<a href` tags (https://www.w3schools.com/html/html_links.asp), but ideally should be handled by the platform better.

**To Reproduce**
1. Go to your profile
2. Add a link without http:// or https:// at start
3. Save profile and try to follow link

**Expected behavior**
Urls should be re-written to link to correct external website, or correct format enforced when creating anywhere on site - preference for first option so that we don't have to rewrite existing links in the database

**Screenshots**
image

chrismclarke:
PR Type

PR Checklist

- [x - PR title descriptive (can be used in release notes)
- [x] - Passes Tests

## Description
- Remove additional workspace type text for regular profiles (closes #818)
- Refactor profile links to own component and remove code duplication (could be further tidied but small improvement for now)
- Add workaround for opening URLs without HTTP prefix (fixes #968)

## Git Issues

Closes #818
Closes #968

### DIFF
--- a/src/pages/Events/Content/EventsCreate/EventsCreate.tsx
+++ b/src/pages/Events/Content/EventsCreate/EventsCreate.tsx
@@ -17,7 +17,7 @@ import { IEvent, IEventFormInput } from 'src/models/events.models'
 import { LocationSearchField } from 'src/components/Form/LocationSearch.field'
 import styled from 'styled-components'
 import theme from 'src/themes/styled.theme'
-import { validateUrl, addProtocol, required } from 'src/utils/validators'
+import { validateUrl, addProtocolMutator, required } from 'src/utils/validators'
 import { Box } from 'rebass'
 import ElWithBeforeIcon from 'src/components/ElWithBeforeIcon'
 import IconHeaderEvents from 'src/assets/images/header-section/events-header-icon.svg'
@@ -94,7 +94,7 @@ export class EventsCreate extends React.Component<IProps, IState> {
         initialValues={formValues}
         mutators={{
           ...arrayMutators,
-          addProtocol,
+          addProtocolMutator,
         }}
         validateOnBlur
         render={({
@@ -236,7 +236,7 @@ export class EventsCreate extends React.Component<IProps, IState> {
                             component={InputField}
                             placeholder="URL to offsite link (Facebook, Meetup, etc)"
                             customOnBlur={e =>
-                              mutators.addProtocol(e.target.name)
+                              mutators.addProtocolMutator(e.target.name)
                             }
                           />
                         </Flex>
--- a/src/pages/Settings/SettingsPage.tsx
+++ b/src/pages/Settings/SettingsPage.tsx
@@ -20,7 +20,6 @@ import { UserMapPinSection } from './content/formSections/MapPin.section'
 import theme from 'src/themes/styled.theme'
 import INITIAL_VALUES from './Template'
 import { Box } from 'rebass'
-import { addProtocol } from 'src/utils/validators'
 import { Prompt } from 'react-router'
 import { toJS } from 'mobx'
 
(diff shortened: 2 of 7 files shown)


chrismclarke:
Sure thing, can take a quick look. Yeah I had remembered there being a better URL check before, so will be good to get that back in alongside the fix for existing links

chrismclarke:
OK, yeah I see the problem. It was part of the refactor, specifically the way the mutator was used previously isn't really compatible with array mutator used for form links. I've fixed it so that first the form now properly checks the validation methods (a state variable required wasn't being updated), and added a standalone format method to use directly in the input format function instead of form validators for the links. I also checked the only other place the protocol validator was used (events page) and that is still working fine without change.

So should be good to go assuming the latest commits passes tests. Will check back in tomorrow

TODO - test events

chrismclarke:
Great, thank for that. Yeah I jumped back and forth on whether directing to http or https site would be better, I still think it would be better to by default assume the site as https for a couple reasons:

1. Not all sites are setup to redirect from http to https if they have certificates installed. E.g. visiting http://apache.org/ will give the insecure version of https://apache.org/ in most browsers

2. More than half of the world runs on https and that includes all major social media outlets (likely where people will be linking to).

Let me know if there's another reason you had thought of though that I have missed, otherwise we can revert that one small change and merge.

Happy with the other additions though (I understand the warning given, ideally we should be make those form alterations at a higher level and passing through state but I tried and it was a bit messy and tricky to get working correctly...so I suggest just leaving for now).

BenGamma:
I agree on your points but I removed it because :

- Even if we consider https better & more secure we shouldn't change the user input (more than we already do).
- For social media they will be redirected to https anyway but you can also add your own website which is more likely not to be under https (small associations and non technical ppl don't have the reflex to activate https on their hosting like we do)

chrismclarke:
Ah, I understand the confusion now - yes we definitely don't want to be changing things a user has put (especially if it can't be changed back by them), but the protocol change only happens if there is nothing there - i.e. if they want their HTTP website, they can still type `http://...` and it will stay the same.

Actually even for personal websites HTTP is super rare these days, as most people like to use 3rd party web providers and all the largest ones (e.g. heroku, google firebase, netlify, zeit etc.) all provide free https as a default - so it's only really for sites running on their own personal web servers.  And again, this is still much less that people who will just be putting social media links. So I've reverted back to https and will merge

BenGamma:
Well I REVERTED BACK TO HTTP (that's a joke) 😁 

Sorry I'm reopening this branch because I've spotted a weird behavior in the link validation while doing a last check. See a screencast [here](http://recordit.co/Pphy7TNBU4)
It happend in the case you have an invalid url first, then the validation removes the last char of the input. But I don't get it because we are not striping any character, we are just adding from the existing value 🤔 

I've added a commit also to remove the url check on discord type, which are not links but discord id (mine for example is benj#2480)

I'm always surprised of the ability of this form errors to always end up in headbanging. But we're almost there 👊

BenGamma:
I moved the discussion to #979

chrismclarke:
hahahahaha! I wouldn't have been surprised. Whaat? that bug is super weird. Oh and good catch on the discord - I'm guessing we should actually link to the discord user page but that can be a job for another day if someone else views it as important.

chrismclarke:
I tried copying the steps exactly but can't recreate on my local browsers, so I think perhaps we can just put this down to a browser quirk and assume if it comes up again the user will be able to resolve themselves. Will merge the discord change

PR #979, opened 2020-06-09
Fix links validations

BenGamma:
PR Type

- [ x ] Bug fix (non-breaking change which fixes an issue)

PR Checklist

- [ ] - Passes Tests

## Description
- remove link validation for discord IDs
- As described in #970 thread, still one little bug to fix :
> I've spotted a weird behavior in the link validation while doing a last check. See a screencast [here](http://recordit.co/Pphy7TNBU4)
It happend in the case you have an invalid url first, then the validation removes the last char of the input. But I don't get it because we are not striping any character, we are just adding from the existing value 🤔

### DIFF
--- a/src/pages/Settings/content/formSections/Fields/Link.field.tsx
+++ b/src/pages/Settings/content/formSections/Fields/Link.field.tsx
@@ -48,8 +48,6 @@ export class ProfileLinkField extends Component<IProps, IState> {
   // e.g. only a username is given for a bazar link
   public formatLink(link: string) {
     switch (this.state.linkType) {
-      case 'discord':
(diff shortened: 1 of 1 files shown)


BenGamma:
The bug seems to come from the url validation `validate={value => this.validateDependingOnType(value)}`. But it's very weird, sometimes it append, sometimes not. Can you let me know if you can reproduce @chrismclarke ?

chrismclarke:
I tried copying the steps exactly but can't recreate on my local browsers, so I think perhaps we can just put this down to a browser quirk and assume if it comes up again the user will be able to resolve themselves. Will merge the discord change

PR #981, opened 2020-06-15
Refactor/dependencies

chrismclarke:
Follows #952  (will close/re-open after to see correct file diffs)

More core improvements, this time focused on our current dependency list

## Main Changes
- Add react-scripts. This is the original `create-react-template` that handles lots of build tooling behind the scenes. We had previously _ejected_ to have more custom control when there used to be no support for sass, service workers etc. but since then the tooling has come a long way and reverting back to the default scripts provides a lot of simplification and stabilisation improvements.
- Integrate react-rewired. This allows more flexibility and custom tooling with react scripts, specifically to add our custom improvements to the default service worker which had previously been nested deeply within workpack scripts (harder for devs to use)
- Upgrade react to latest version. v16.8 -> v16.13. This hadn't been easy to do previously because of managing lots of related dependencies in ejected app, but now a much smoother process. Should not have any breaking changes.
- Remove unused dependencies (either default included in react-scripts or previously added to the platform but no longer used)
- Significant net code reduction

## Additional changes
- Various package upgrades
- Swap tslint for eslint. Previously eslint didn't support typescript so we used a mix of both eslint (js/jsx) and tslint (ts/tsx) but now it does and is recommended a single solution (tslint is slowly being deprecated in favour or eslint).
- Handle additional linting rules and minor type fixes
- Continue to improve system that links different sets of dependencies together (e.g. functions with main src), ahead of eventual migration to lerna/yarn workspaces.

## Breaking Changes
The large change in dependencies will mean likely devs will need to delete their `node_modules` folder and run `yarn install` again.
## Recommended Testing
Have checked that main repo and functions build, as well as storybook and general tests. A few minor changes were made to component styles, however, these were only in terms of typing and for locationInput correcting a `noStyle` variable which wasn't correctly defined. A quick live test would be worth doing just to make sure things look generally as intended.

## Todo
- ~~Update wiki~~ 
- Refactor models to avoid import from rest of src (this makes cross-linking messier)
(likely to be done as part of modular refactor/lerna integration so not creating issue for now)
- Update pwa assets (manifest, logo, robots) (#982 )
- Test pwa support (e.g. app icon shortcut) (#982 )
- Include cypress in eslint (currently ignored) (not too important as only for tests and previously wasn't supported very well)
- Upgrade functions packages (not creating issue as there will always be updates)
- Upgrade further src packages (not creating issue as there will always be updates)


BenGamma:
Hey thanks for that one 💪 

Everything works fine on my machine too, there is just some linting warnings in my console but nothing critical (mainly unsused-var).

Special big up for the `yarn test` command, I was always mixing the 2 (between the app in test mode and open Cypress panel 😅)

PR #983, opened 2020-06-16
Add lazy loading for pages and stores

chrismclarke:
Draft PR - based on  #981, will re-open once merged for correct file changes 
Required merge of #1140 

PR Type

PR Checklist

- [x] - Passes Tests

## Description
- Add React.Lazy to all main to only DL required content on page load (required changing all pages from named exports to default exports)
- Add React.Lazy to howto create/edit subroutes
- Add lazy image component - even with the virtual dom it seems that elements off-screen still make an initial request for the image data, so lazy image component fixes this to reduce data consumption
- Add init functions to most modules to prevent automated initialisation on first load (all stores still injected together
- Speed up preload of fonts and preconnect to storage api
- move slick css import to module import (previously imported from index and blocked renderer)
- Rewrite service worker implementation
- Configure better custom service worker paths

## TODO (future)
- Add better documentation on caching strategies
- Decide if wanting to notify user of page updates
- It would also be good to separate out truly "common" stores that should appear everywhere (e.g. user, tags, stats) and module-specific stores. Module-specific should have their own provider injected into the module page container. (Note, may be some small added complications for cross-store like user profile calling map store, but hopefully can be resolved with better system of comoponent hierarchy or pubsub).
- Compress initial images lazy loaded

## Git Issues

_Work towards completion of #956_


public/index.html
```diff
@@ -38,7 +34,12 @@
 
       homescreen on Android. See https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/
     -->
-    <link rel="stylesheet" href="%PUBLIC_URL%/fonts/fonts.css" />
+    <link
+      rel="preload"
+      href="%PUBLIC_URL%/fonts/fonts.css"
+      as="style"
+      crossorigin
+    />
```
BenGamma, line 42: fonts are not loading on my local machine, could this be related ?

chrismclarke:
New service worker strategy added, some nerd stats (take 

**Before**
First howto page load: 17.7MB
howto page reload: 6kB

Page speed:
image

**After**
First howto page load: 5.7MB => **-68%**
howto page reload: 3.5kB => **-42%**
image

This isn't due to any content optimisation, just by not loading everything until needed (so a user visiting all the pages would still use the same amount of data, but a user just visiting and looking at a few things will see a lot less).

Of the initial 5.7MB around 0.5MB is the platform itself and the rest images lazy-loaded for preview in the first 12 or so howtos (depending what fits on the page and slightly below). These could be optimised by sending smaller versions.

chrismclarke:
@davehakkens 
I think this one is finally (!) ready for merge also. There shouldn't be any noticeable changes to the UI with the exception that because of a better implementation, we no longer have/need the banner that says 'site has been updated. please reload' (it will automatically reload the first time the user clicks anything that will take them to a new page, e.g. on a howto, and so shouldn't be very noticeable. It also updates more reliably).

The only question is whether you think we should have some sort of visual indicator to let people know the site has been updated (currently I hover over the pp logo to check the version)

davehakkens (approved):
haha almost a year after you started! But seems good to me

chrismclarke:
failed test was due to current flaky db teardown condition so assume fine. Merging

PR #1024, opened 2020-08-29
feat: Adds more time options to how-to form

### LINKED ISSUE #1023: [feature request] Add more time options to how-to form
Having the lowest time option to be `< 1 week` seems quite high

Add more options which are less than 1 week

Could have strict time differences like 0-1 hour, 1-5 hours, 5-10 hours & 10-1 day 1 day - 1 week but that would mean having to update the existing how-tos to have a different time frame than < 1 week

**Additional context**
For context we would suggest having the following additional options 
- < 1 hour
- < 5 hours
- < 10 hours
- < 1 day

darigovresearch:
PR Type

PR Checklist

- [x] - Passes Tests

## Description

Adds the following additional time options to the how-to form

- < 1 hour
- < 5 hours
- < 10 hours
- < 1 day

## Git Issues

Closes #1023 

## Screenshots/Videos

Pull_request_add_times_to_how_to

### DIFF
--- a/src/pages/Howto/Content/Common/FormSettings.tsx
+++ b/src/pages/Howto/Content/Common/FormSettings.tsx
@@ -1,5 +1,21 @@
 export const TIME_OPTIONS = [
   {
+    value: '< 1 hour',
+    label: '< 1 hour',
+  },
+  {
+    value: '< 5 hours',
+    label: '< 5 hours',
+  },
+  {
+    value: '< 10 hours',
+    label: '< 10 hours',
+  },
+  {
+    value: '< 1 day',
+    label: '< 1 day',
+  },
+  {
     value: '< 1 week',
     label: '< 1 week',
   },
--- a/src/pages/Howto/Content/Common/FormSettings.tsx
+++ b/src/pages/Howto/Content/Common/FormSettings.tsx
@@ -5,7 +5,6 @@ const makeEntry = (value: string, label?: string) => {
 export const TIME_OPTIONS = [
   makeEntry('< 1 hour'),
   makeEntry('< 5 hours'),
-  makeEntry('< 10 hours'),
   makeEntry('< 1 day'),
   makeEntry('< 1 week'),
   makeEntry('1-2 weeks'),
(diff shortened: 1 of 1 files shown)


src/pages/Howto/Content/Common/FormSettings.tsx
```diff
@@ -1,4 +1,20 @@
 export const TIME_OPTIONS = [
+  {
```
catx23, line 2: your could save space for equal key/value pairs by using a factory function, eg : 
´´´const makePair = (value: string, label?: string) => { return { value, label: label || value } }´´´

darigovresearch: @catx23 commit ef0e666 makes the suggested improvements. Is there anything else that needs to be resolved to merge this pull request?

src/pages/Howto/Content/Common/FormSettings.tsx
```diff
@@ -1,37 +1,21 @@
...
-  },
-  {
-    value: '3-4 weeks',
-    label: '3-4 weeks',
-  },
-  {
-    value: '1+ months',
-    label: '1+ months',
-  },
+  makePair('< 1 hour', '< 1 hour'),
```
catx23, line 6: hey, you don't need supply the label :-) makePair accepts an explict label but since it's the same as value, you can just write  makePair('< 1 hour'); // you could rename it to makeEntry to avoid confusion. 

As bonus, you could populate the arrays when really needed. If you'd do this somewhere in the component lifecycle, eg: componentWillMount you can save another good chunk CPU cycles as well as memory :-) 

When the browser loads the JS file for this app, it evaluates all globals and also module scoped variables as DIFFICULTY_OPTIONS .... Most of the time, users won't create HowTos .... I hope this makes sense.

darigovresearch: Commit b49c9b5 adjusts your first comment but we don't really understand your 2nd and 3rd paragraph as this framework is not really our specialty. This is now probably sufficient to resolve the original issue (#1024)  and keeps the code readable if we leave it in the same .tsx file for whoever comes next.

davehakkens:
thanks for the commit, and issue! 
If you remove the `<10 Hours` i'll add it to master. 
(This one is a bit confusing with `<a day`. Wheter to choose day (24H) or workday (8h)

darigovresearch:
@davehakkens you're very welcome! `<10 hours` has been removed. Yeah we agree that `< 1 day` can mean either 24hrs or a workday but since `1 day` is an upper bound for both, it takes into account both times.

Issue #989, opened 2020-06-22
[bug] Error message for password reset renders multiple times


jxofficial:
**Describe the bug**
Reset password error message flickers upon input change.

**To Reproduce**

1. Go to the login page.
2. Click on 'Lost password'
3. Type anything into the input fields
4. Error message flickers erratically.

**Expected behavior**
Error message should only render once and disappear upon timeout.

**Screenshots**
![animation flicker](https://user-images.githubusercontent.com/49927092/85322487-01e71800-b4f9-11ea-89da-2ac0a21d4592.gif)

Daniel-Davies:
Hello- I'm happy to take this one if you'd like!

davehakkens:
yes! :)

PR #1028, opened 2020-09-03
fixed jumping error message on login page- issue #989

Daniel-Davies:
PR Type

- [ X] Bug fix (non-breaking change which fixes an issue)

PR Checklist

- [ X] - Latest `master` branch merged
- [ X] - PR title descriptive (can be used in release notes)
- [ X] - Passes Tests

## Description

Issue '989 describes how the error message would re-render and therefore continually jump when  editing the user email textfield. In fact, the error was mainly with a negative offset on y that was being trigger from each re-render; I have checked all use cases of the component in question, and could not see why the negative offset was productive in any way; hence I removed it, which fixed the jumping. 

In addition, I noticed the timer wasn't being explicitly cleared. A one line change in the TextNotification component to clear the existing timer just before each new use case handles the situation in which a user can click "Lost Password?", changes their password, and clicks "Lost Password?" again (in which case, the old timer would have affected the delay on the latest click event).

## Git Issues

#989

### DIFF
--- a/src/components/Animations/FadeInOut.tsx
+++ b/src/components/Animations/FadeInOut.tsx
@@ -9,7 +9,7 @@ interface IState {
 }
 
 const AnimationContainer = posed.div({
-  enter: { y: -10, opacity: 1 },
+  enter: { opacity: 1 },
   exit: { y: 5, opacity: 0 },
 })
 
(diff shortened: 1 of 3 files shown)


BenGamma:
Thanks @Daniel-Davies ans sorry for the slow response.

I'm closing this PR because I opened #1046 with current merge from master

PR #1031, opened 2020-09-05
Fix #974 Extra infromation on Profile

### LINKED ISSUE #974: Extra information on Profile
### Description
In the profiles we need some additional information in a sidebar
**Map pin:** Show if profile has map pin (and [link to pin](https://community.preciousplastic.com/map#username). )
**How-to:** Show amount of how-to's created by this profile (all approved ones)
**Events:** Show amount of events created by this profile (all approved ones)

### Page related

Will be used in :
[Profiles](https://community.preciousplastic.com/u/precious-plastic)

### Mockup
profile-items-nopatreon_01
profile-items_explained_01

alromh87:
Shows information on user profile
- [x] Location with link to map pin
- [x] Accepted HowTos
- [x] Accepted Events

User stats are computed via firebase function triggered by database changes


alromh87:
This profiles show different validations:

- alromh87
  - Pin
  - Events
  - HowTos
- bentest12
  - Only Pin
- 7pgee0qxlzmdbxmfdtbe4hcdzom2
  - Only HowTos
- 4vd2ejopw1ftcunhj01o8vpbgq62
  - Only Event
- 1qsijzipovtkqrryfavkzc1dzd93
  - No Pins, events or howTos

As requested in #974

src/components/ElWithBeforeIcon/index.tsx
```diff
@@ -8,6 +8,7 @@ interface IProps {
   width?: string
   ticked?: boolean
   contain?: boolean
+  mr?: number
```
BenGamma, line 11: In that case it will be better to extend the props from rebass BoxProps directly, but I don't think you need to add it here, and I don't see it used anywhere else anyway ?

src/models/user.models.tsx
```diff
@@ -36,6 +36,7 @@ export interface IUser {
   country?: string | null
   location?: ILocation | null
   year?: ISODateString
+  _stats?: IUserStats
```
BenGamma, line 39: Here can you rename `stats` without underscore ? We keep the underscore variables to distinct the existing variables in a package or a library and the ones we add are without

BenGamma (changes requested):
If I understand well the counts for how-to and events will be triggered on creation or on edit right ?
So to have this information available we'll need to modify all the existing howtos and events ? I'm trying to understand how we are going to proceed for migration.
Other than that see my comments for small changes.
Thanks for the work !

src/components/ElWithBeforeIcon/index.tsx
alromh87: It was needed to be able to change the margin of the internal box, which had it fixed to 4, now it will default to 4(avoid braking other displays), but can be overwiten when usng component, used to make howto event display in one line

src/models/user.models.tsx
alromh87: Done :smile:

BenGamma:
> Functions are at [functions/src/Integrations/user-stats.ts ](https://github.com/ONEARMY/community-platform/commit/40b7deb8073677b2518906acb444bca19081df44), I gess I'm missing some config since they where not deployed.

As far as I know they will need to be deployed on firebase once merged with master.

> I already have script for the initial calculation and update of user.stats for migration, this would be executed just once, after that updates will be calculated based on individual howto/event data, for faster execution

Ok nice, can I see this script ?

BenGamma:
I added a commit so I had to create a new branch and new PR, discussion moved to #1040

alromh87:
https://github.com/ONEARMY/community-platform/pull/1040#issuecomment-692024726

PR #1035, opened 2020-09-06
Fix #964 "Create How-to" button for not-logged in

alromh87:
When not logged in create how to button will redirect to create account

Updated with changes from master, sends no error on build

PR #1036, opened 2020-09-06
Fix #964 "Create How-to" button for not-logged in

### LINKED ISSUE #964: "Create How-to" button for not-logged in
Hello,

Currently, as a not-logged-in user the user does not see the "Create How-to" button. 
image

I think this is a missed opportunity to engage with new users and ultimately achieve more signups and how-tos. I would propose to have the "Create How-to" button at any time and if the user is not logged in he/she gets redirected to https://community.preciousplastic.com/sign-up to create an account and eventually an how-to.

Hope it makes sense 🤙

alromh87:
When not logged in create how to button will redirect to create account

Updated with changes from master, sends no error on build


BenGamma:
Closing since duplicate of #1035

Issue #972, opened 2020-06-02
Add country flags at username


davehakkens:
### Description
Show the diversity of the worldwide community more, by adding locations flags to usernames. Based on the location selected by the user.

### Page related
Will be used in :
[How-to overview](https://community.preciousplastic.com/how-to) 
[Single How to](https://community.preciousplastic.com/how-to/add-interchangable-patterns-to-injection-moulds)

### Mockup
![profile-flag](https://user-images.githubusercontent.com/13672737/83517869-ef6d5600-a4d9-11ea-90d5-940574d67d2c.png)
![flag_howto](https://user-images.githubusercontent.com/13672737/83517876-f3997380-a4d9-11ea-8233-d52ac13d4048.jpg)

BenGamma:
This one is a bit in the same veine as #974. We'll probably have to add a field on the how-to model `creatorCountry` when howto is created, to be able to easily get this info.

Daniel-Davies:
Hello- I'd love to help out your platform and contribute to this issue, if possible. Could I take this issue? (Or would you recommend another?)

davehakkens:
Thanks for wanting to help @Daniel-Davies, however @alromh87 is currently working on this one. 
What do you like to work on/ what are your skills? We have quite some SEO improvement we need to make, optimise backend, and bugfixes.

We are working on some new features but are currently still in design stage.

Daniel-Davies:
Hello- I've had a look at other issues. Is #989 available?

alromh87:
@BenGamma @chrismclarke  I have some doubts about the cache for howto list:

Flag  **in howto list** gets updated if I edit the howto from the browser but not when i trigger from DB

Flag in Detail its updated every time (from browser or DB)

BenGamma:
@alromh87 yes it would be preferable to update from the browser. Otherwise we have to update the cache of all the user and that's a lot of data 😬 
Can you describe the procedure to update from the browser ? I'm not sure to understand

alromh87:
When a user is updates a howto, the fag gets correctly updated in the HowToList.

But if a user changes his location all of his how-tos should reflect the change, that its achived with a firestore function.
After change the new flag will display in how-to detail(edit, view) but not on how to list.

Im unsure if this has to do with my dev environment or some web cache

BenGamma:
> When a user is updates a howto, the fag gets correctly updated in the HowToList.

[...]

PR #1040, opened 2020-09-14
Feat/974 extra infos profile

### LINKED ISSUE #974: Extra information on Profile

BenGamma:
Closes #974 
Based on work in #1031


alromh87:
This is the script Im preparing form migration, I have been testing it from [Cloud Console](ttps://console.cloud.google.com/) since I have no permissions to execute https endpoint

It also contains code for migration for 972 Update all flags in HowTos

```
const functions = require('firebase-functions')
const request = require('request')

const admin = require('firebase-admin');
admin.initializeApp();
const db = admin.firestore();

// Change for prod:
// import { db } from '../Firebase/firestoreDB'


// Compute one user stats
async function computeUserStats(owner){
  console.log("Calculando para ", owner);

  stats = {};
  try {
    stats.howToCount = await db.collection('v3_howtos').where("_createdBy", "==", owner).where("moderation", "==", "accepted")
      .get()
      .then(querySnapshot => {
        let count = 0;
        querySnapshot.forEach(doc => {
          count++;
        });
//        console.log('Accepted howtos for', owner ,', count: ', count);
        return count
    });

    stats.eventCount = await db.collection('v3_events').where("_createdBy", "==", owner).where("moderation", "==", "accepted")
      .get()
      .then(querySnapshot => {
        let count = 0;
        querySnapshot.forEach(doc => {
          count++;
        });
//        console.log('Accepted events for', owner ,',count: ', count);
        return count
    });

    return stats
  } catch(e) { 
    console.log('Error compute:',e)
  }
  return null
}

// Compute all user stats
function computeAllUsersStats() {

  console.log('Computing all user stats ...')

  db.collection('v3_users')
//.where("_id", "==", "benjj")
  .get()
  .then(querySnapshot =>{
    return querySnapshot.forEach(user => {
      computeUserStats(user.id).then( stats => {
        console.log('Response @All users');
        console.log(stats);
        if(stats){
          user.ref.update({
            _stats: admin.firestore.FieldValue.delete(),
            stats: stats
          });
        }
        return null
      }).catch( e => {});
    })
  }).catch( e =>{
    console.log('Error query usuarios:', e)
    return null
  })
}

// Set all user flags
function updateAllFlags() {

  console.log('Updating all flags ...')

  db.collection('v3_users')
  .get()
  .then(querySnapshot =>{
    return querySnapshot.forEach(user => {
      const country = (user.data().location && user.data().location.countryCode) ? user.data().location.countryCode : (user.data().country ? user.data().country.toLowerCase() : '')
      if(country !== ''){

        console.log('Procesing', user.data()._id, 'flags');

        db.collection('v3_howtos').where("_createdBy", "==", user.data()._id)
        .get()
        .then(querySnapshot => {
          querySnapshot.forEach(doc => {
            console.log('updating doc', doc.data()._id, 'to', country);
            doc.ref.update({'_creatorCountry': country})
            .then( () => {
              console.log("Document successfully updated!");
              return null
            }).catch( error => {
              console.error("Error updating HowToCountry: ", error);
              return null
            });
          });
          return null
        }).catch(e => console.log('Error getting user howTos:', e));
      }else{
        console.log('No info for ', user.data()._id, 'flags');
      }
    })
  }).catch( e =>{
    console.log('Error query usuarios:', e)
    return null
  })
}

// Cant trigger from url, fix permissions
exports.batchUpdate = functions.https.onRequest(async (req, res) => {
  await computeAllUsersStats();
//  await updateAllFlags();
  res.status(200).send('Finished');
});

```

alromh87:
> This is the script Im preparing form migration, ...
> 

Updated the count functions to handle no previously calculated data:
- now for first accepted event/how-to after installation with will compute user stats [R38](https://github.com/ONEARMY/community-platform/pull/1040/commits/a84445621d4d9fe8d8468b6b7c48b96f10721ab4#diff-67b45feafd910733b7f7d84a23207d80R38)

alromh87:
> > This is the script Im preparing form migration, ...
> 
> Updated the count functions to handle no previously calculated data:
> 
> * now for first accepted event/how-to after installation trigger will compute user stats [R38](https://github.com/ONEARMY/community-platform/pull/1040/commits/a84445621d4d9fe8d8468b6b7c48b96f10721ab4#diff-67b45feafd910733b7f7d84a23207d80R38)

This makes the PR ready to merge no need to execute the migration script

BenGamma:
Thanks for updating the script !
I think it's better if we separate the part to update the flag on how-to (which is part of #972) from this PR, but since it's not part of the [user-stats.ts](https://github.com/ONEARMY/community-platform/blob/a84445621d4d9fe8d8468b6b7c48b96f10721ab4/functions/src/Integrations/user-stats.ts) file, to me it's good.
Other than that, it's good for me.
@chrismclarke can you double check the [user-stats.ts file](https://github.com/ONEARMY/community-platform/blob/a84445621d4d9fe8d8468b6b7c48b96f10721ab4/functions/src/Integrations/user-stats.ts) before merge ?

alromh87:
> Thanks for updating the script !
> I think it's better if we separate the part to update the flag on how-to (which is part of #972) from this PR, but since it's not part of the [user-stats.ts](https://github.com/ONEARMY/community-platform/blob/a84445621d4d9fe8d8468b6b7c48b96f10721ab4/functions/src/Integrations/user-stats.ts) file, to me it's good.
> Other than that, it's good for me.
> @chrismclarke can you double check the [user-stats.ts file](https://github.com/ONEARMY/community-platform/blob/a84445621d4d9fe8d8468b6b7c48b96f10721ab4/functions/src/Integrations/user-stats.ts) before merge ?

The files in PR are targeted only to this PR, the script posted [here](https://github.com/ONEARMY/community-platform/pull/1040#issuecomment-692024726) is only for reference and not included in PR.

BenGamma:
> The files in PR are targeted only to this PR, the script posted [here](https://github.com/ONEARMY/community-platform/pull/1040#issuecomment-692024726) is only for reference and not included in PR.

Yes I got that :)

functions/src/Integrations/user-stats.ts
```diff
@@ -0,0 +1,92 @@
+import * as functions from 'firebase-functions'
+import { db } from '../Firebase/firestoreDB'
+
+
+export const countHowTos = functions.firestore
+  .document('v3_howtos/{id}')
+  .onWrite(async (change, context) => {
```
chrismclarke, line 7: If I understand correctly we are just tracking how many `howTos` and events a particular user has created? I really like the use of the triggers to do this, although I think to keep things a bit simpler you could just use the `created` trigger to increment and `deleted` to decrement - assuming that if a doc has been created it will either eventually be approved or deleted. But if you want to keep this way it's unlikely to result in too many calls to the db so not a big issue.

functions/src/Integrations/user-stats.ts
```diff
@@ -0,0 +1,92 @@
...
+    return null
+  }
+
+  console.log('Update ', info._createdBy, ' ', target, ' delta: ', delta )
+
+  let user = await userStatsRef.doc(info._createdBy).get();
+  let stats = null;
+  if(user && user.exists){
+    if(!user.stats){
+      console.log('No previous stats for ', user.id ,' -> compute'),
```
chrismclarke, line 39: nice job to make sure we populate in case where empty. You could probably turn this into a one-liner (`let stats = user.stats || await computeUserStats`), but then you'd lose the logging which might be handy as we first test these so again no harm keeping

functions/src/Integrations/user-stats.ts
```diff
@@ -0,0 +1,92 @@
...
+      stats = await computeUserStats(info._createdBy);
+      console.log(user._id, ': ', stats);
+    }else{
+      stats = user.stats;
+      if (!(target in user.stats)) stats[target] = 0;
+    }
+    if(stats){
```
chrismclarke, line 46: So if the previous parts did there job we wouldn't need this extra condition, but again no harm having a fallback

functions/src/Integrations/user-stats.ts
```diff
@@ -0,0 +1,92 @@
...
+      user.ref.update({
+        stats: stats
+      }, {merge: true}).then(() =>{
+        return true
+      }).catch(e => {
+        console.log('Could not update user stats, error: ', e);
```
chrismclarke, line 52: We probably want to be identifying these types of errors in the platform, either via `console.error` or `throw new Error('could not update stats for....)`.

functions/src/Integrations/user-stats.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Compute one user stats
+async function computeUserStats(owner){
+  console.log("Calculando para ", owner);
+
+  stats = {};
+  try {
+    stats.howToCount = await db.collection('v3_howtos').where("_createdBy", "==", owner).where("moderation", "==", "accepted")
+      .get()
```
chrismclarke, line 68: Will need to check this, but there is a chance a new database index will be needed to run this. If so there will be an error in firebase and a link to automatically populate BUT we should instead populate manually instead via the `firestore.indexes.json` file so that the same rules get populated to production. Let me know if this happens and I can help resolve

src/models/user.models.tsx
```diff
@@ -51,6 +52,11 @@ interface IExternalLink {
     | 'instagram'
 }
 
+interface IUserStats {
+  howToCount: number
```
chrismclarke, line 56: Just a general thought here, but would it be better to keep the url/slugs of the howTos and events in an array so we can link to them directly instead of just the count? We can still display a count as the array length, just might be nice to have the extra info saved (understand if not a priority for others)

chrismclarke (approved):
Great to see a larger feature like this coming in, nice work on front and back end. I've given a few inline comments, but would be happy to still merge and review after we've tested a little on the dev site.

I'm guessing we might also want a one-off function that could be called from the API to populate the db with all the existing howtos/events.

functions/src/Integrations/user-stats.ts
alromh87: chenge is used in order to count only accepted howtos

functions/src/Integrations/user-stats.ts
alromh87: From my test, deploying the function directly to firestore, no error happend in firebase, guess it will be no need for new index.

functions/src/Integrations/user-stats.ts
alromh87: Good idea

alromh87 (approved):

src/models/user.models.tsx
alromh87: That would be as well possible, I would leave it as an option for now until there is such a need

functions/src/Integrations/user-stats.ts
alromh87: updated

alromh87:
Could this be merged?

PR #1053, opened 2020-10-02
Feat/974 Improve stats computation efficiency

alromh87:
PR Type

PR Checklist

- [x] - Passes Tests

## Description

Previous user stats algorithm was inefficent and would lead to unpredicted result due to being based on promisses, when querySnapshot already have *size* property.

To identify if full stats computation was needed user.stats was used instead of user.data().stats leading to inefficiently computing all stats every time.


chrismclarke (approved):
oh yeah, that makes sense, in terms of the time taken pretty negligible but nice to tidy up the code a bit.

PR #1055, opened 2020-10-12
Feat/user stats migration

chrismclarke:
## Breaking changes
Requires migration script run against production db before publishing live

## Description
Additional scripts and tidying to support user stats migration. 
### Main Changes:
- Refactor userStats to own folder (with consistent naming for exports)
- Add stricter typings
- Code-tidy stat count functions
- Add one-time migration script for initial stats population for all users, accounting for all events and howtos
- Change scripts to keep track of howto and event ids within stats object for future use

### Additional Changes
- Refactor db naming system to allow easier updates
- Remove deprecated backend functions
- Improve functions db methods
- Add to functions documentation
- Update tests to work with refactor

## Git Issues

_Closes #_

## Screenshots/Videos


src/models/user.models.tsx
```diff
@@ -53,8 +53,8 @@ interface IExternalLink {
 }
 
 interface IUserStats {
-  howToCount: number
-  eventCount: number
+  userCreatedHowtos: number
```
alromh87, line 56: Frontend would need to update accordingly

functions/src/index.ts
```diff
@@ -3,10 +3,9 @@ import { weeklyTasks, dailyTasks } from './exports/tasks'
 import { DH_Exports } from './DaveHakkensNL'
 import * as IntegrationsSlack from './Integrations/firebase-slack'
 import * as IntegrationsDiscord from './Integrations/firebase-discord'
-import * as UserTriggers from './Integrations/user-triggers'
```
alromh87, line 6: The purpose of this trigger is to update howTo's flag in case a user update his location, so if this behaviour is dessired  the trigger should be enabled.

This trigger could be used as well for handling when user requests username change, bt thats longer story.

alromh87 (changes requested):
I feel like the dessired rsult would not be achived with the changes as they are now.

functions/src/userStats/migration.ts
```diff
@@ -0,0 +1,86 @@
+import * as functions from 'firebase-functions'
+import { DB_PREFIX, IUserDB, IHowtoDB, IEventDB } from '../models'
+import { db } from '../Firebase/firestoreDB'
+
+/**
+ * One-off script to migrate legacy content to new format
+ * Once run this code will be deprecated, but retained in case
+ * it can be used in future migrations
+ */
+export const migrateUserStats = functions.https.onCall(
```
alromh87, line 10: Still unsure here but looks to me, from my testing, that calling the function this way migth lead to reaching firestore maximum execution time without producing any result, so next time is executed start from the begginig would be needed, again failing.

Because getting all events, howtos, users proccessing and updating and not keeping track of already updated  users.

functions/src/userStats/migration.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/****************************************************************
+ * Helper functions
+ ****************************************************************/
+/**
+ * Create a hashmap all all howtos with a count of users
+ */
+function calcUserHowtoCounts(howtos: IHowtoDB[]) {
+  const allHowTosByUser = {}
+  howtos.forEach(v => {
```
alromh87, line 54: Same as https://github.com/ONEARMY/community-platform/pull/1055/files/7928f26fcfdd5a8c0c08d8745e1ea8ba2e87f106#r503244379

alromh87 (changes requested):
Few more comments

functions/src/userStats/migration.ts
```diff
@@ -0,0 +1,86 @@
...
+    allHowTosByUser[createdBy]++
+  })
+  console.log('allHowTosByUser', allHowTosByUser)
+}
+/**
+ * Create a hashmap all events with a count of users
+ */
+function calcUserEventCounts(events: IEventDB[]) {
+  const allEventsByUser = {}
+  events.forEach(v => {
```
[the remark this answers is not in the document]
chrismclarke: Parallel here is fine as we're not running any async code (it's just taking an array and converting to an object). But yeah, otherwise choices are to use a `for-of` loop which works in series, or create a new array or map to record each op as a promise (e.g. `promises= someLoop.forEach(v=>promises.push(someAsyncFunc(v))` and use `await Promise.all(promises)` to carry out in parallel. The downside to the second approach is the function will fail if any one promise fails, however any individual promise that passed up until that point will still have be updated.

The exception to this is when writing db docs, firebase has an inbuilt method called `batch`, where you can run ops on up to 500 docs in parallel once per second.

functions/src/index.ts
chrismclarke: Ah that makes sense, and yeah could to have a place to record any side-effects for user changes. There's actually another (poorly named function) that listens for user updates also to automatically create a user backup (`FirebaseUserBackup`), so I'll probably merge these into a single function and document a bit better

functions/src/userStats/migration.ts
chrismclarke: Oh, I hadn't actually thought about timeout issues, but yes you are right that could happen. There is actually a way to manually increase timeouts for a specific function: https://firebase.google.com/docs/functions/manage-functions#set_timeout_and_memory_allocation, but nonetheless I'll do some testing and optimise for this.

functions/src/userStats/migration.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Create a hashmap all all howtos with a count of users
+ */
+function calcUserHowtoCounts(howtos: IHowtoDB[]) {
+  const allHowTosByUser = {}
+  howtos.forEach(v => {
+    const createdBy = v._createdBy || 'anonymous'
+    allHowTosByUser[createdBy] = allHowTosByUser[createdBy] || 0
+    allHowTosByUser[createdBy]++
+  })
+  console.log('allHowTosByUser', allHowTosByUser)
```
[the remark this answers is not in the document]
chrismclarke: Thanks!

chrismclarke:
OK, I think it should now be good to go. I've run against dev and seems to be working as expected, and did a test run on live (no db writes, just logging what would be changed) and seems reasonable. Output attached.
Thanks for your work on this @alromh87 .
@BenGamma - let me know when you want to do the next release and I can run against production and update functions.
[userMigrate_Live-1602563518371.log](https://github.com/ONEARMY/community-platform/files/5368948/userMigrate_Live-1602563518371.log)

functions/src/userUpdates/index.ts
```diff
@@ -0,0 +1,65 @@
...
+  const querySnapshot = await db
+    .collection(DB_ENDPOINTS.howtos)
+    .where('_createdBy', '==', userId)
+    .get()
+
+  if (querySnapshot) {
+    querySnapshot.forEach(doc => {
+      console.log('Updating howTo ', doc.data()._id, 'to', country)
+      doc.ref
+        .update({ creatorCountry: country })
```
alromh87, line 50: Adding `'_modified': new Date().toISOString()`  will show the new info on the frontend, thanks for the info Chris :wink:

alromh87 (changes requested):
LGTM

Just one change needed to update frontend howto list when user updates location.

I don't see the flag batch update though

BenGamma (approved):
Thanks a lot guys, I'm not a specialist in the backend optimisation but what I see looks good to me ! Special mention for updating the documentation as well 👍 
@chrismclarke I'll create a PR for the new release v1.3 to production branch. I'll leave it to you to press merge button and run the migration script once merged in production.

functions/src/userUpdates/index.ts
BenGamma: thanks @alromh87 👌

PR #1060, opened 2020-10-21
Add database archiving

chrismclarke:
PR Type

PR Checklist

- [ ] - Passes Tests

## Description
Deleting documents from the database is not a trivial issue due to the way docs are cached. If a document is deleted from the server there is no way for a user to know it doesn't exist unless they fetch updates for that specific doc on the server (which might happen when opening a howto for example, but not for the full list on first load as this is inefficient - the full docs would need to be retrieved as there is no way to just query the list of ids that exist within firestore)

Currently the database deletes documents by removing the content and simply marking a `deleted` field and updating the timestamp. This is good in that it will prompt all users to receive the update to that document, however there are still 2 problems with this approach:
1) If we ever want to restore a deleted document we can't as the data is wiped 
2) The database can quickly get  cluttered with document stubs which then need to be filtered out on every query

This update adds a new system, whereby deleted docs are moved from their existing collection (solving 2), and populated in an `_archived` collection with both metadata stubs and full data stored separately (solving 1). An additional query is added for users to receive updates from the collection, and delete their own cached documents accordingly (keeping solution to initial problem). Additional bindings are made for the 3 existing database systems (firestore, firebase realtime and dexie) to process the deletes.

## Git Issues

_Closes #531_

## Screenshots/Videos


chrismclarke:
Note, run failing due to legacy package.json version bug

BenGamma (approved):
Thanks @chrismclarke, yes this approach is better than the current and will help to better handle the deletion/archiving of content 💪 

Maybe as we discussed it could be good to have an other issue for further improvements regarding the deletion/archiving of content. Improvements that comes to my mind are :
- Allow to restore a specific version of a document (like the previous version of an how-to, since we keep an history of modifications)
- give the ability to an admin to delete a user/ map pin / content, by adding a delete button to them. This cases happend a lot and are not conviniently handled right now.

PR #1083, opened 2020-12-22
Prevent duplicate how-to title in edit

ivokh:
PR Type

PR Checklist

- [x] - Passes Tests

## Description

The slug of the how-to title was checked against existing how-to's when creating a new how-to but not when editing an how-to where it is also possible to edit the title (possibly into one that already exists).
These changes will also check the slug of the how-to against existing how-to's when editing an how-to.


chrismclarke (approved):
Thanks for catching this one, yep that makes sense to me and the solution looks good too.

Issue #766, opened 2019-11-11
show prompt only if form value changed


BenGamma:
In how-to form, show the <Prompt> warning when leaving the page without saving only if a change as be made to the form.
Should be doable using the `FormState` of react-final-form. Documentation [here](https://final-form.org/docs/final-form/types/FormState)

ivokh:
I started working on this issue so it can be assigned to me. Although I've been a software developer for a few years now I have no experience with React. I already ran into a few challenges where someone might have some ideas or suggestions for.

**Dirty state**
It's pretty easy to only show the confirmation prompt when the react-final-form is dirty. The challenge is the logic behind the dirty state. React-final-form will use a shallow equal by default to detect changes between the initial value and the current value. This will not work for example for the tags because the tags value is an object. This can be solved by giving an isEqual callback to the field with some custom logic for the comparison. For file uploads it's not obvious what comparison logic to use. I suggest to set it to dirty/not equal when a new file is selected.

**React Router Prompt**
The react router prompt component is used to display the confirmation alert. This only shows when the router is used to move to another page. The pp logo to go to the home page uses a href url and will not trigger the confirmation. The same applies for refreshing the page. I will have to look into the react router prompt to see if I can find a solution for this.

**Testing edit page**
I'm currently testing on the create page. Because I have no access to edit a how-to. Should I create a test how-to to use the edit page or is there some way to give my user permission to edit other how-to's?

**Automated tests**
I started looking onto the Cypress tests for this logic and I figured a way to test the confirmation alert. The question is how thoroughly we should test this logic? It's possible to validate the dirty stage of each field but that would result in a lot of tests.
I think it might be a good idea to test one field with an integration/Cypress test and maybe validate the dirty state of the other fields with unit tests?

**Adding the Lodash package**
For the comparison of the tags I added the [Lodash](https://lodash.com/) package which has an [isEqual](https://lodash.com/docs/4.17.15#isEqual) function that does a deep comparison. Are there any problems with adding this package to the project?

chrismclarke:
Thanks for taking the time to look at this and the depth of solution you've worked on here. I think the main issues you've addressed are more than enough, and we don't need to worry about going into 100% accuracy for things like images etc. 

For the react router, another option could perhaps be seeing if we can use the [componentWillUnmount](https://reactjs.org/docs/react-component.html#componentwillunmount) lifecycle event for the form component instead (although I'm not sure if this will trigger on page reload or not, could be worth checking).

[...]

PR #1086, opened 2020-12-30
#766 Show prompt only if form value changed

ivokh:
PR Type

PR Checklist

- [x] - Passes Tests

## Description

Only show the confirm dialog (prompt) when a value is changed inside the howto edit/create page

## Git Issues

#766

### DIFF
--- a/cypress/integration/howto/write.spec.ts
+++ b/cypress/integration/howto/write.spec.ts
@@ -186,6 +186,38 @@ describe('[How To]', () => {
       cy.visit('/how-to/create')
       cy.get('div').contains('Please login to access this page')
     })
+
+    it('[Warning on leaving page]', () => {
+      const stub = cy.stub()
+      stub.returns(false);
+      cy.on('window:confirm', stub)
+
+      cy.login('howto_creator@test.com', 'test1234')
+      cy.wait(2000)
+      cy.step('Access the create-how-to')
+      cy.get('[data-cy=create]').click()
+      cy.get('[data-cy=intro-title')
+        .clear()
+        .type('Create a how-to test')
+        .blur({ force: true })
+      cy.get('[data-cy=page-link][href*="/how-to"]').click().then(() => {
+        expect(stub.callCount).to.equal(1)
+        stub.resetHistory()
+      })
+      cy.url()
+        .should('match', /\/how-to\/create$/)
+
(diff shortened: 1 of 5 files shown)


src/pages/Howto/Content/Common/Comparisons.tsx
```diff
@@ -0,0 +1,44 @@
+import * as _ from 'lodash'
+import { ISelectedTags } from 'src/models/tags.model'
+
+// Functions used to give as callback to the isEqual prop of a form fields.
+// The isEqual callback is used to determine if a field is dirty.
+export const COMPARISONS = {
```
chrismclarke, line 6: Really nice to have these defined in a reusable way, will come in handy elsewhere for sure

cypress/integration/howto/write.spec.ts
```diff
@@ -186,6 +186,38 @@ describe('[How To]', () => {
       cy.visit('/how-to/create')
       cy.get('div').contains('Please login to access this page')
     })
+
+    it('[Warning on leaving page]', () => {
+      const stub = cy.stub()
+      stub.returns(false);
+      cy.on('window:confirm', stub)
```
chrismclarke, line 193: nice idea to stub this, wouldn't have thought of that!

src/pages/Howto/Content/Common/Howto.form.tsx
```diff
@@ -228,6 +251,7 @@ export class HowtoForm extends React.PureComponent<IProps, IState> {
                                     ? this.validateTitle(value)
                                     : false
                                 }
+                                isEqual={COMPARISONS.textInput}
```
chrismclarke, line 254: I feel like adding comparisons to every field might be a bit time-intensive to manage if this form were to change or if we want to apply the same to other forms (e.g. creating events). Although I'm guessing processing at the form level is then computationally quite a bit more intensive (checking every field any time anything changes).... So I think probably the best solution is using a better form library that has more native support for this, and for now we have the option to apply at the form level if we want.

chrismclarke (approved):
I added a few inline comments and a bit more context to the discussion on #766 , but overall I think the additions are great and my own testing seemed to work well. Merging.

https://user-images.githubusercontent.com/10515065/104279890-e70dce00-545f-11eb-9611-a2321257aae3.mp4

PR #1074, opened 2020-11-14
Useful button

tudi2d:
PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR Checklist

- [x] - Latest `master` branch merged
- [x] - PR title descriptive (can be used in release notes)
- [x] - Passes Tests

## Description

New "useful" button to give easy feedback. With this button there are 2 new fields added to the Firestore.

1. usefulCount: number (v3_howtos)
2. usefulHowTos: string[] (v3_users)

If a user marks a HowTo as useful for the first time the "usefulHowTos" field is added to the user profile and the "usefulCount"-field is added to the HowTo. This way everything should function as normal and there is no need of touching the database.

For not logged in users a Box is displayed with the same content as the Button, but - of course - not clickable.

Please test this for different edge cases as I´m not aware of all possible use cases and what rights a user can have.

## Git Issues

Closes #1072 

## Screenshots/Videos
image

zn1MpwwTui

PR #1094, opened 2021-01-31
Useful button feature (updated)

chrismclarke:
PR Type

PR Checklist

- [ ] - Passes Tests

## Description

Building from the work by @tudi2d  in #1074, update methods:

- move howto total vote counter to backend function to use utility increment feature (more reliable if multiple people vote at similar times, none of the votes will get lost and a user can't accidentally set the count to the wrong number)

-  change user list of voted howtos to object instead of array (easier lookups)

- remove numeric counter from description - this is annoying as it would be really nice to have, but the current db architecture just doesn't support well. Any time any user upvotes a howto that entire howto will be refreshed for all other users, quickly resulting in thousands of db operations. Ideally we will want to get users to manually check for updates when a howto is loaded (if desirable to see the number), or provide something like a monthly update which can be sync'd. But the number will still be tracked in the database (just the `_modified` field is not updated so it is not automatically sync'd to the user)

- various variable and function renaming

## Git Issues

_Closes #_

## Screenshots/Videos


chrismclarke:
Following discussion today, best to first try refactor so that live stats can be subscribed to when a user has the howto open.
This will likely work best by having a `stats` subcollection for each howto

chrismclarke:
Ok, now refactored to keep a separate set of stats for each howto that is updated when a user views a howto instead. Seems to work as intended (slight delay updating total as it has to update the database first)

https://user-images.githubusercontent.com/10515065/106558350-6ed37f00-64d8-11eb-8765-5dcf726ca555.mp4

tudi2d:
> [...] (slight delay updating total as it has to update the database first)

I like it - good job! :) Just a little question - is there a easy to change possibility to handle it not asynchronous, update the counter for the client & just handle potential update errors? This way the UX would feel a bit more direct for the user, no matter how bad the internet connection is, avoiding users to click the button multiple times, as it does not directly indicate a change.

mattia-io:
A more high-level question for this feature. Sorry 🤪

Do we only want Community Platform users to be able to **Useful** how-tos?

**Let me explain.** 
I think we see the Community Platform as the place for active users to interact. People that have a workspace and are recycling, collecting or connecting the community. I think we would like to try and keep simple members to a minimum as there is not much to do for them in there. So we could say that Community Platform traffic is made up of 60-70% non registered users? People viewing how-tos, browsing academy, checking the map. 

Following this logic, we could assume that at least 50% of people viewing how-tos are not a registered Community Platform user. And I think we're happy with them not being a registered user (are we @davehakkens). However, those users in my opinion should still be able to "Useful" a how-to as they're still absorbing the knowledge and are in a position to value a how-to useful or not. 

A solution to this (from my limited dev knowledge) could be to save "Useful" in the browser local storage? So people wouldn't have to create an account but still be able to "useful" how-tos.
A brilliant example working similarly is this https://colorhunt.co/ 

What do you guys think? @chrismclarke @davehakkens @tudi2d  
I hope I am not opening a pandora box 🙈

davehakkens:
Smooth example indeed. 
But not sure if we are happy with them not being registered, It's easier to misuse, you can upvote certian things with different browsers or devices which makes the whole mechanism oboslete. It's safer, at least in the beginning, to go for a less high number but all legit. So the number has a real value. Which is kinda what we intend useful to be (not just a quick like)
But I would either way to suggest to implement how it is. See how it goes and then decide if we need to open up the doors more.

+ we would also have more reasons to create an account in the future when comments will be enabled.
This is just a small first action you can do as a user, more features to come

chrismclarke:
Yeah, as Dave said without user accounts there's nothing we can do to stop people manipulating (just open a new private browsing window each time you want to upvote). Any measures to try and prevent this would be a lot more complex/invasive (e.g. storing IP addresses, which can still be worked around if someone really wanted to).

@tudi2d - definitely possible to manually override the number temporarily whilst waiting on the server updates. I was just too lazy to implement the logic (handle both cases where the number might go up or down, adding local state to the component to update the value (as it is passed from the store a prop the component can't directly edit so will need to first copy to local state).
If you were interested in doing this then definitely feel free, should be an extra 5-10 lines of code or so.

Issue #1088, opened 2021-01-16
[feature request] Search field for how-to page


darigovresearch:
**Is your feature request related to a problem? Please describe.**
A user in the discord mentioned that now there are many how-tos it would be helpful to have a search field to help find relevant how-tos a user is looking for, particularly when it is an old how-to

**Describe the solution you'd like**
A search field at the top of the page that would narrow down the visible how-tos from the database

**Describe alternatives you've considered**
Can use ctrl+f in the browser but does not always work as not all how-tos are loaded onto the page

**Additional context**
n/a

chrismclarke:
I really like this suggestion. There's a few different ways we could go about this:

1. Manual client-side search function that tries to match text from the input box against a few specific fields, such as title, description, and maybe howto steps. Pros: very quick to implement. Cons: search basic and harder to use elsewhere

2. Server-side search via algolia or elasticsearch. We use algolia in the platform already for map location lookup, so could use the existing account to also try add search against our own content. Pros: efficient search. Cons: not sure how well pricing scales

3. Client-side search via lib. I'd be leaning towards [fuse.js](https://fusejs.io/) for this, although there are plenty of [other options](https://www.npmtrends.com/fuse.js-vs-fuzzysearch-vs-lunr-vs-search-index-vs-elasticlunr-vs-js-search) also. Pros: efficient search. Cons: wouldn't work as well if we had very large datasets.

Personally I'd be leaning towards option (3), but open if anyone wants to take this on and explore any other options

danitrod:
I'd be willing to work on this, I think we can start with option 3, then eventually switch to server-side search if need be later on. 

About the layout of the input field, I think it could be similar to the filter by tags button, maybe with a magnifier icon aside, and could be positioned in 1. or 2. (before the filter button) like the image below. Any thoughts?

<img width="1437" alt="Captura de Tela 2021-02-04 às 11 12 08" src="https://user-images.githubusercontent.com/45438149/106905850-597a7280-66db-11eb-8175-83ff2f23094d.png">

chrismclarke:
Sounds great, yes I think for the howto's as there's only 100 or so for now and unlikely to grow too big too soon I think fuse.js will be a solid option for some time. We have a pretty similar layout already for events so I'd suggest starting with that and then if design team want some changes we can update.

![image](https://user-images.githubusercontent.com/10515065/106943204-fa682e00-66d9-11eb-8dbe-5d28eea686bb.png)

[...]

PR #1098, opened 2021-02-05
Add how-to client side searching

danitrod:
PR Type

PR Checklist

- [X] - Passes Tests

## Description

As suggested on #1088, adding client-side searching for how-tos using [Fuse](https://github.com/krisk/Fuse/). I added a `SearchInput` component that only renders the search field, and I did the searching logic on the howto store. After filtering the howtos by tags, if there is something typed in the search field the function will perform a Fuse search and return the search-sorted howtos.

On the first commit I also removed some warnings React was throwing for no apparent reason, screenshot below.

Also, I updated an integration test which was selecting the how-to tag selector by the input tag. It now selects the first input, as there are now two.

Note on Fuse - the current weights for the how-to search are:
```
0.5 - Title
0.3 - Description
0.15 - Steps titles
0.05 - Steps descriptions
```
although it looks pretty fast, adding the steps to the search significantly increased the search time, I think it is worth trying a few options and maybe consider not using the steps for now.

## Git Issues

_Closes #1088_

## Screenshots/Videos

demo
Demo

react-warnings
Warnings

### DIFF
--- a/cypress/integration/howto/read.spec.ts
+++ b/cypress/integration/howto/read.spec.ts
@@ -42,9 +42,7 @@ describe('[How To]', () => {
       cy.get('[data-cy=create]')
         .click()
         .url()
-        
     })
-   
   })
 
   describe('[Filter with Tag]', () => {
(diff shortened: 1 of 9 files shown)


chrismclarke (approved):
Looks really good from my side, appreciate the extra effort to update the tests and bits of code tidying. I also think the choice of howto search fields and weights seem like a pretty sensible starting point, but will wait and see what Dave and co think from the preview generated before merging.

Issue #1100, opened 2021-02-07
[feature request] - link howto search input with urls


chrismclarke:
**Is your feature request related to a problem? Please describe.**
As discussed in #1088, would be great to be able to share a link that has the search field already filled (e.g. all howtos written by a specific creator). 

**Describe the solution you'd like**
Use [react router query params](https://reactrouter.com/web/example/query-parameters), so that URLs are stored like `https://community.preciousplastic.com/how-to?search=plastmakers`, and on open will show the howto page with the search populated with `plastmakers`. 

**Describe alternatives you've considered**
We could consider a more advanced search (e.g. `/how-to?creator=plastmakers`), but for now the simple search would still be a really good start

**Additional context**
As URLs will be encoded will be good to consider how best to handle (e.g. /how-to?search=precious%20plastic should decode before populating search). Similarly, we will want the URL bar to update when a user searches so that they can copy/paste their search link.

I think @danitrod said he would be happy to take this one on also, although if you don't have the time feel free to let us know and I can add a help-wanted label.

danitrod:
Thanks @chrismclarke, I'll work on it during the week.

PR #1101, opened 2021-02-09
Add query params updating on search howtos

danitrod:
PR Type

PR Checklist

- [X] - Passes Tests

## Description

As discussed in #1100, adding the feature to update the URL query string when the search field is altered. I also implemented the same logic for tags selecting. Everything was done using the JavaScript URL API, so mapping spaces and special characters to encoded values is already working by default. 

Note: as tags are identified by the unique ID, the URL gets a little noisy with the params. If you guys think this is not desirable, I can revert this part.

## Git Issues

_Closes #1100_

## Screenshots/Videos

1100-search-url
(Demo is a bit slow cuz of my machine)

### DIFF
--- a/src/components/SearchInput/index.tsx
+++ b/src/components/SearchInput/index.tsx
@@ -4,6 +4,7 @@ import { Input } from '../Form/elements'
 interface IProps {
   placeholder: string
   onChange: (value: string) => void
+  value?: string
 }
 
 const style = {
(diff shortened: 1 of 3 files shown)


davehakkens:
Wauw super neat. At first I was a bit skeptical of adding search (often works shitty) but this works really smooth. 
Also nice to be able to combine tags + search. well done

chrismclarke:
Yeah, really nicely done. It actually almost feels too good! Like you don't expect searching for things to change so quick, and I really like the addition to combine with the tags - super useful for sure.

chrismclarke (approved):
Really nice implementation, thanks so much! I really like how it persists if navigating to a howto and the on return, good use of the store and URLs. Will merge now and plan to get into the live release this week.

PR #1102, opened 2021-02-14
Useful local state

### LINKED ISSUE #1099: [enhancment] Client-side update usefull state
**Is your feature request related to a problem?**

When a user clicks the Useful-button the client waits for the successful server side update without loading indication. As this can take a while due to firebase it feels as the "Click" did not cause any action.

Change it to handle the request not asynchronous, update the counter for the client & just handle potential update errors. This way the UX would feel a bit more direct for the user, no matter how bad the internet connection is, avoiding users to click the button multiple times.

**Additional context**

@tudi2d - definitely possible to manually override the number temporarily whilst waiting on the server updates. I was just too lazy to implement the logic (handle both cases where the number might go up or down, adding local state to the component to update the value (as it is passed from the store a prop the component can't directly edit so will need to first copy to local state).

_Originally posted by @chrismclarke in https://github.com/ONEARMY/community-platform/issues/1094#issuecomment-772653845_

tudi2d:
PR Type

PR Checklist

- [x] - Passes Tests

## Description

Add local state to update useful state & count of a HowTo to give a user direct feedback.

## Git Issues

Closes #1099


tudi2d:
@chrismclarke This should do it. One thing I encountered was, that `howToStats` was always `undefined`. Is this because there are no stats for the HowTo´s on the dev-server or is this an issue?

tudi2d:
Ok.... it is super weird. Why does the counter update anyway? It would be totally fine to load it only once & than just change the local state (so +1 & -1)

sfxKhuiizN

tudi2d:
Additional thing I thought of when seeing dave's post on slack - should it probably show the '0' instead of a blank? @chrismclarke

chrismclarke (changes requested):
Thanks for this. I think this makes sense to me, although would you be able to add a couple quick comments somewhere to explain why we are doing this for better future reference (i.e. why we have both the store state and local state, and why we don't always just read from the store)

Also, as I can't easily test, have you considered where the user is removing their useful vote? It looks like it works, but as I don't see any `-1` counters I'm confused by the implementation.

chrismclarke:
> @chrismclarke This should do it. One thing I encountered was, that `howToStats` was always `undefined`. Is this because there are no stats for the HowTo´s on the dev-server or is this an issue?

I think when a howto is first loaded the stats are by default set as undefined, and then if there are no stats found it will stay undefined, but yeah we probably need better distinction from after the fetch is made to know whether the count is actually 0, or it just hasn't fetched them yet

tudi2d:
@chrismclarke Sure, will change the request things & test the behavior locally again. I will find time during the weekend to continue working on it 🙂

davehakkens:
Like how snappy it feels!
But here I go from 0 to 2?
Or something in my caching?

https://user-images.githubusercontent.com/13672737/108629526-35978a80-7458-11eb-8fe2-7598f5e0fdb5.mov

tudi2d:
> 
> 
> Like how snappy it feels!
> But here I go from 0 to 2?

Yeah - it's still the old version (see github-bot above). Not sure how to trigger the deployment - thought it would happen automatically on a new commit.

tudi2d:
Nevermind - still broken... @chrismclarke Any idea? Could this be weird firebase things?

chrismclarke:
I think new deployment would be triggered, although as it's a production site possibly the old version would be cached (and the little update notification should appear). In any case I'll take a look now and try figure it out

tudi2d:
Thanks @chrismclarke ! 🙂

chrismclarke:
Can confirm, label needs removing and adding back.

I've just made a few updates, mostly to move everything into a new component to simplify things a bit (that way we can just render the component when stats have already been retrieved, instead of worrying about them getting updated after the component is rendered). 

Changes made in commit d9d2b6f5feb1955bfae801c6a2f71cdfaa2fdc1f

Seems to be working as intended for me, but good if you could also check @tudi2d

chrismclarke (approved):

chrismclarke:
actually, I think I accidentally overwrote the icon changes... will fix now

tudi2d:
Works better, but still not perfectly fine, right? It's really frustrating that this is so hard to get right somehow...

https://user-images.githubusercontent.com/8904624/108764414-da968e00-7552-11eb-83ca-a773fa7f79f6.mp4

chrismclarke:
Thanks for the demo, really useful. Super weird, maybe it's because the stats aren't cleared or howto reset when the back button is pressed? Will take a quick look now

chrismclarke:
https://user-images.githubusercontent.com/10515065/108767132-de9ebd80-750a-11eb-9d9b-9421a433ea19.mp4

Yeah I think it was just the store wasn't clearing old values before trying to load new. Hopefully this should fix it finally. The only case where it might be a bit inconsistent is if a user exits a howto and opens again really quickly (before the db has had a chance to update), but I think that's ok. 

Have also updated the pr preview to rebuild on new commits, so hopefully that should work for this update also.

tudi2d:
Thanks for helping me out here! Reviewed the new component you created & have to say that it makes more sense to keep that stuff separated 👍 Still having issues in the "Preview" version... 

image
---
image

chrismclarke:
It should now be building automatically, although yeah pretty hard to keep track of these things! I've also sent an invite to give you write access to the repo so you can also manage labels or issue assignment if you ever want to do so.

chrismclarke:
I think at this point we can just merge, and if any tweaks need making later we can do so then. Thanks for all the work on this!

chrismclarke:
Oh, and I also just realised you can check the original PR preview comment for updates, which is probably the fastest way to know if the preview has been rebuilt:
image

Issue #984, opened 2020-06-16
[bug] - User last active does not automatically update


chrismclarke:
**Describe the bug**
I think we created the field with the intention of updating it on login, but guessing forgot.

**To Reproduce**
Logout
Go to own map pin and see last active
Login
View map pin again (last active will not have changed)

**Expected behavior**
Likely will want a server-side or user store trigger that populates lastActive and modified on successful login. Server-side likely more reliable, but client side easier (could just add to existing `userSignedIn` field.

nahos:
Hey, Can you assign it to me. I will try fixing it.

nahos:
Btw I am not able to see my pin, it just shows my location but it doesn't show a  pin.

chrismclarke:
Sure thing, happy to have you work on this. 

Pins aren't enabled on all profile types so depending what you have selected in settings you may not have one (we don't have pins for members because there's just too many of them and don't represent physical locations as well as things like workspaces). If there seems to be some other problem then feel free to open another issue here with your username to help debug.

danitrod:
Hi @nahos, are you working on this? Looks like it's been a while, I'd like to work on it if you're not.

chrismclarke:
Reassigned. Thanks @danitrod!

This should hopefully be fairly straight-forward, the userstore has a `userSignedIn` method which can be used to also update the last active once the user profile has been successfully loaded. 

The only complication would be with the `functions/src/userUpdates/backupUser.ts` function, which by default creates a backup of the user whenever the profile is updated - which we probably don't want to do if the only thing that is changing is the last active time. But it seems that function is not currently active (not sure why, but it would otherwise be called in the `handleUserUpdates`), so probably would be good enough to just add a note to update it in the future if it becomes enabled again.

PR #1105, opened 2021-02-23
Fix user last active updating

danitrod:
PR Type

PR Checklist

- [ ] - Passes Tests

## Description

A login now updates the user's last active status to the current time, as discussed in #984. The update is done in the `userSignedIn` method in the store. I tested with a custom pin and the last active info is updating correctly now.
Also, in order to not backup a user every time he logs in, I added a shallow comparator on a previous and updated user, to check if `lastActive` is the only field updated. This hasn't been tested though as I'm not sure how to test the backup function and see the backup data. If the test is desirable I'd be happy to test it, just will need some directions.

## Git Issues

_Closes #984_

### DIFF
--- a/functions/src/userUpdates/backupUser.ts
+++ b/functions/src/userUpdates/backupUser.ts
@@ -1,4 +1,21 @@
-import { DBDoc, IDBDocChange } from '../models'
+import { IDBDocChange } from '../models'
+import { IUserDB } from 'src/models/user.models'
+
+/** Helper function to check if the only field changed is lastActive
+ * (updates on login), in which case we will not want
+ * to backup the user again.
+ */
+const hasUserDataUpdated = (
+  prevUser: IUserDB,
+  updatedUser: IUserDB,
+): boolean => {
+  return Object.keys(prevUser).some(
+    key =>
+      key !== '_modified' &&
+      key !== '_lastActive' &&
+      prevUser[key] !== updatedUser[key],
+  )
+}
 
 /**
  * Automatically create user revision on update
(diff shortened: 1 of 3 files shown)


chrismclarke:
Thanks for adding this, I've been a bit busy recently but will do my best to review and respond properly about the backup functions this week.

functions/src/userUpdates/backupUser.ts
```diff
@@ -1,4 +1,21 @@
-import { DBDoc, IDBDocChange } from '../models'
+import { IDBDocChange } from '../models'
+import { IUserDB } from 'src/models/user.models'
```
chrismclarke, line 2: Due to a limitation with how the functions folder is organised, imports from the local 'src' folder won't actually be recognised on build (the way typescript resolves paths is fiddly at best, there's quite a bit of tooling out there which is better for working in these monorepo structures such as lerna or yarn workspaces (particularly with newer yarn 2.x), but haven't got round to updating.

So for now can you just change it to a relative import (`../models`)

chrismclarke (changes requested):
Looks great, just one small import to fix which I've added a comment for.

chrismclarke:
Also, if you're updating this PR could you also include the code to run the user backup function? (I'm not sure why it was ever removed, I'm guessing by mistake during some code refactoring). 

It can be called from `functions/src/userUpdates/index.ts` within the `handleUserUpdates` trigger, i.e.
```
export const handleUserUpdates = functions.firestore
  .document(`${DB_ENDPOINTS.users}/{id}`)
  .onUpdate(async (change, context) => {
    await processCountryUpdates(change)
    await backupUser(change)            <-------------------------- Call here to run on user updates
  })
```  
I quickly revisited the current implementation for testing these backend functions. It still could do with a lot of improvement although I'm reluctant to spend too much time on this as I'm expecting/hoping that over the next year we will be able to replace firebase entirely in the project. I've sent an invite so you can receive edit access to the testing firebase site, and there is some general notes in `functions\README.md`

More specifically it involves installing and logging in to the firebase cli, running the start script from the functions folder, and then making manual updates to the emulated database to see changes (e.g. I created a user document and tried updating different fields to check if the revision would be made)

image

Issue #1080, opened 2020-12-11
Research Module - Listing


chrismclarke:
### Mockup

Requires #1079

![image](https://user-images.githubusercontent.com/10515065/101934855-07ade400-3b93-11eb-8366-d5cb612f8e69.png)

### Example(s)

If you have any examples of existing similar components, add the link here. It can be a link to a code snippet and/or to a page of a projet/website where we can interact with the similar component.

danitrod:
Can I start working on it? I see the store and page are already setup

davehakkens:
from my end yes! 
But not sure if you need some input from @chrismclarke?

danitrod:
i think the implementation seems pretty straight forward - not sure if there's any prior work done before I should start though

davehakkens:
the orignal plan was for @alromh87 to take it one. But don't think that ever took off.

chrismclarke:
Thanks for moving forward on this, yeah from me happy to see any additions

PR #1110, opened 2021-03-07
Add initial research module layout

danitrod:
PR Type

PR Checklist

- [ ] - Passes Tests

## Description
Adding the first layout for research module listing, details and updates. Also finished setting up the folder structure and mock data for researches.

Still missing:
- insert the DBDoc metadata into a newly created update
- add forms for creating a research, creating an update, and editing a research
- setting up permissions for deleting/editing a research
- add tests for the new pages

## Git Issues
_Closes #1080 and closes #1079_ - probably should create some more issues for the above points

## Screenshots/Videos
research-module


davehakkens:
niceee!

src/components/Research/ResearchListItem.tsx
```diff
@@ -0,0 +1,71 @@
...
+import theme from 'src/themes/styled.theme'
+import styled from 'styled-components'
+
+interface IProps {
+  item: IResearch.ItemDB
+}
+
+const GridBox = styled(Box)`
+  display: grid;
+  grid-template-columns: 2fr 1fr 1fr;
```
chrismclarke, line 18: personally I like css grid and media queries, but in case you weren't aware there's an added option included in rebass components to provide widths for the theme breakpoints (https://rebassjs.org/layout). You'll see in a lot of the howto components code like `<Flex px={4} py={4} flexDirection={'column'} width={[1, 1, 1 / 2]}>` where the width array corresponds to sizes at the various theme breakpoints (at least I think that's what it does! I try to stay away from frontend where I can)

src/pages/Howto/Content/Howto/Step/Step.tsx
```diff
@@ -1,13 +1,13 @@
...
 import ReactPlayer from 'react-player'
-import { IHowtoStep } from 'src/models/howto.models'
 import { Box } from 'rebass'
 import Flex from 'src/components/Flex'
 import Heading from 'src/components/Heading'
+import ImageGallery from 'src/components/ImageGallery'
 import Text from 'src/components/Text'
+import { IHowtoStep } from 'src/models/howto.models'
 import { IUploadedFileMeta } from 'src/stores/storage'
-import ImageGallery from './ImageGallery'
```
chrismclarke, line 10: thanks for refactoring this into a shared component, makes a lot of sense to do so. Feel free to make any other changes you see as sensible for tidying up component structures or sharing code between components

src/pages/Research/Content/ResearchItemDetail.tsx
```diff
@@ -1,26 +1,75 @@
...
 import * as React from 'react'
-import Text from 'src/components/Text'
-import { ResearchStoreContext } from '../research.store'
+import { useHistory } from 'react-router'
+import { Box, Flex } from 'rebass'
+import { Button } from 'src/components/Button'
+import { MOCK_UPDATES } from 'src/mocks/research.mocks'
+import { NotFoundPage } from 'src/pages/NotFound/NotFound'
+import { ResearchStoreContext } from 'src/stores/Research/research.store'
+import { Loader } from '../../../components/Loader/index'
```
chrismclarke, line 9: is this also just the `src/components` folder? if so probably easier to import it that way

src/stores/Research/research.store.tsx
```diff
@@ -18,21 +20,48 @@ export class ResearchStore extends ModuleStore {
...
+
+  public addUpdate(item: IResearch.ItemDB, update: IResearch.Update) {
+    const updatedItem = {
+      ...toJS(item),
+      updates: [
+        ...toJS(item.updates),
+        {
+          ...update,
+          // TODO - insert metadata into the new update
+          _id: Math.random().toString(),
```
chrismclarke, line 52: As it happens this should be handled automatically in `src\stores\databaseV2\index.tsx` - basically all db writes get automatically populated with standard _created, _modified_, _id metadata so individual stores don't need to worry about it.

chrismclarke: Oh, also if you ever did want to use Math.random to generate a random string you can also include a radix in the string method. E.g. `Math.random().toString(36)` is pretty common, as base36 would use 10 digits (0-9) and 26 letters of alphabet (a-z), so you get an output like `0.pwqjb68tqqr`. Take the substring from digit 2 and you get a quick unique id `pwqjb68tqqr`. There's still a few limitations with this however, as strings won't always be uniform length if ending in 0s (which would just be omitted) - small discussion on this if you scroll down here: https://stackoverflow.com/questions/1349404/generate-random-string-characters-in-javascript

chrismclarke:
What's there looks really good to me. For the metadata you actually don't need to worry as that's all handled by automatically by the common db methods mentioned in the comment above. So you can just remove the code to manually add. Otherwise from a dev side I'm happy to merge and continue to build from here.

Assuming you're happy with the design implementation so far @davehakkens? (can follow-up next steps in separate issue)

src/stores/Research/research.store.tsx
danitrod: Thank you for the reviews! And nice tip on the `Math.random().toString()` usage - wasn't used to that. A question here, I did try to insert the updates without the metadata, the meta fields weren't automatically populated by the DB though, I think it's because the updates are an array inside of the research object which is in fact populated with the metadata. Do we have to store the updates in a separate collection and reference them from the research object?

src/components/Research/ResearchListItem.tsx
danitrod: nice touch! i was kinda fixed on the grid idea since i'm used to work with bootstrap

danitrod: nice touch! i was kinda fixed on the grid idea cause i used to work with bootstrap a lot

src/pages/Research/Content/ResearchItemDetail.tsx
danitrod: yep, using auto-import is tricky sometimes

danitrod:
Added a couple of changes for the suggestions, the only issue is with the update DBDoc metadata populating - as I responded to your comment the db isn't populating it automatically when concatenating a new raw update to the array :(

src/components/Research/ResearchListItem.tsx
chrismclarke: Yeah it's a kind of weird niche that these rebass components are trying to fill. I see the benefit if you're trying to keep consistency across all screens/components, but like most js and css libraries/frameworks if you know what you are doing as a developer you can usually achieve the same outcomes using more vanilla-style approaches. I don't think we're super opinionated, as long as there is a reasonable level of consistency or standard approaches so that someone else can make changes in the future if required, happy with whatever way devs find themselves working best.

danitrod:
took the liberty to add a `Mod: Research 🔬` label for PRs and issues

src/stores/Research/research.store.tsx
chrismclarke: Oh, that makes a lot more sense, I had been thinking about individual docs and not data stored within the research object array. I think we can go about it either way, either creating a new subcollection doc each time there is an update, or just storing everything in a single array in the research object itself. If we store things in the object itself then we will want to manually add some metadata as you started to do. Happy for you to proceed however you think best and we can always decide if we want to refactor later

chrismclarke:
If you really wanted to get fancy you could also

> took the liberty to add a `Mod: Research 🔬` label for PRs and issues

Nice! Was gonna ask if you could also add it to the pr-labeller action that automatically tags things based on the code that changes but can see you've already done it!

PR #1116, opened 2021-03-19
Research Module - create/edit

danitrod:
PR Type

PR Checklist

- [X] - Passes Tests

## Description

Adding create/edit research forms, and create research update form. A couple of notes:

- when accessing the URL to directly edit an existing research or add an update to it (/research/\<slug\>/edit or /research/\<slug\>/create-update), the store takes a while to validate if a user is logged in or not. I found no way for now to assure the user store has tried loading the user, so I added a forced timeout of 3 seconds to load the user before the code makes any decisions (visible in lines 21-26 of [CreateUpdate](https://github.com/ONEARMY/community-platform/pull/1116/files#diff-f37151495cbb1be377817e61d8517ce278cd5c379ae8b91e36a3ce0541f00e60)). Its probably best if we can take the "triedToLogin" information from the user store directly.
- I implemented moderation for the researches but couldn't test accepting/rejecting a research as I do not have an admin account. 

## Git Issues

_Closes #1111_

## Screenshots/Videos

research-forms


davehakkens:
wauw again a super nice pull request! 💪 

Not a coder myself, so can't comment to much on code, i'll leave that to @chrismclarke.
Few remarks from testing it.

1. Image upload doesn't work here.
2. I get a message when saving: "you have unsaved changes, are you sure to leave the page"  (but it looks like you get this as well)
3. Notifications is not a must in the beginning. We would start using the module internally first for a while with a selected group of people. No need for verification or moderation.
4. I started a research and added a few steps. When I changed the title of my research all the steps dissapeared
5. In how-to's we have the original step + all the following steps in one page. (so you can edit all of the steps in one place and also move certain steps up or down) Looks like you splitted this up, each update having its own "page". Any specific reason for this? And how could you edit an update?

danitrod:
Thank you for testing it and for the valuable feedback!

1. not reallly sure why that happens, but trying in dev mode it works ok. I think its probably something related to preview mode, as I just tried to create a new how-to with images in the preview and the images don't load as well. Maybe @chrismclarke has some insight?
2. yeah, that's kinda weird, ill try to understand what's happening and fix it
3. cool, no worries then, ill just make any new research be published as 'accepted'
4. definitely a bug, forgot about the updates on editing a research - will fix
5. I was thinking the UI for the research should be split for the initial intro and each following update, as the updates will be normally created one by one - i might have misunderstood, is it supposed to work like this? About editing an update, its not implemented for now but i can surely work on it, the only question is if they should be edited separately or edited like a howto, all in a single page.

I'll take a look at those bugs and commit fixes soon.

davehakkens:
5. Yeh i'm up for trying it the way you made it (seperated). Seems more clear and perhaps you don't want to move steps around anyway since its all chronological timebased anyway. Would be good to add that "edit update" though.

davehakkens:
SUPER neat. it passes all my tests 💪

src/components/EventCard/EventCard.tsx
```diff
@@ -32,7 +32,7 @@ export const EventCard = (props: IProps) => (
     data-eventId={props.event._id}
   >
     {props.event.moderation !== 'accepted' && (
-      <ModerationStatusText event={props.event} top={'0px'} />
+      <ModerationStatusText moderable={props.event} kind="event" top={'0px'} />
```
chrismclarke, line 35: Good to see this refactored into a common component, I don't really understand what the word `moderable` means, but I can see you inherited the code from others so definitely not blaming! 

Would you have any suggestion for a word that makes more sense? maybe `moderatedContent` or similar? Similarly we could change the word `kind` to something a bit more specific (e.g. `contentType`) to make things clearer for future devs. 

But these are minor issues so I might just add a small comment myself to tidy in the future.

src/pages/Research/Content/Common/Research.form.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+  React.useEffect(() => {
+    if (store.researchUploadStatus.Complete) {
+      window.removeEventListener('beforeunload', beforeUnload, false)
+    }
+  }, [store.researchUploadStatus.Complete])
+
+  React.useEffect(() => {
+    if (submissionHandler.shouldSubmit) {
+      const form = document.getElementById('researchForm')
```
chrismclarke, line 73: I think react probably prefers when people use refs (https://reactjs.org/docs/refs-and-the-dom.html) instead of accessing dom directly, but I'm not aware of any issues doing it the vanilla way (the id is defined well enough to avoid multiple elements with the same id).
A bit more here: https://stackoverflow.com/questions/37273876/reactjs-this-refs-vs-document-getelementbyid/37274379#:~:text=In%20general%2C%20refs%20is%20better,rest%20of%20your%20react%20code.&text=And%20as%20%40Isuru%20points%20out,multiple%20inputs%20with%20same%20ID.

src/pages/Research/Content/EditUpdate/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+  React.useEffect(() => {
+    ;(async () => {
+      let loggedInUser = store.activeUser
+      if (!loggedInUser) {
+        // TODO - handle the case where user is still loading
+        await new Promise<void>(resolve =>
+          setTimeout(() => {
+            loggedInUser = store.activeUser
+            resolve()
+          }, 3000),
```
chrismclarke, line 44: I think this is something you mentioned in your comments - agreed it's a pain to know whether there is no user or if they just haven't finished login initialisation after direct nav / page reload. This workaround is fine for now, I think there would probably be some better way long term to handle more generally but I'll create a separate issue for that

chrismclarke (approved):
There's a huge amount here, which is really appreciated but also a bit tricky to review in depth. Overall the code looks really solid (I've added a few minor comments), so happy to merge. I think we can do some manual testing to check if everything is behaving as expected, and just make minor issues to fix anything as we come across it. 

As the majority of the code is nicely encapsulated within the research module we can test most from that page, but for reference @davehakkens - there's also a few updates included to re-organise/re-use some of the content from howtos and events (specifically moderation status) - so will be good to double check that is still working as expected too.

src/components/EventCard/EventCard.tsx
danitrod: Agreed, I was struggling to find a good name for this but definitely `moderatedContent` and `contentType` would be more describing names. Should I open a PR for this?

src/pages/Research/Content/Common/Research.form.tsx
danitrod: I think I was adapting some of the logic from the HowTo form, which used this approach. Definitely a future improvement to not access the DOM directly.

davehakkens:
Got this message in Slack today. Are they related?

afbeelding

PR #1129, opened 2021-04-16
Research module live, but only for specific users

danitrod:
PR Type

PR Checklist

- [X] - Passes Tests

## Description

With this update, the research module page becomes visible in the production site, but only for users that have the `beta-tester` role. I added a `requiredRole` attribute to the `IPageMeta`, so pages using this attribute will have a rendered link in the menu only for users who have the required role.
Also added an admin page to manage beta tester users.

Also, minor changes to clarify prop names from my last PR (https://github.com/ONEARMY/community-platform/pull/1116#discussion_r607033140).

## Git Issues

_Closes #1118_

### DIFF
--- a/src/components/EventCard/EventCard.tsx
+++ b/src/components/EventCard/EventCard.tsx
@@ -32,7 +32,11 @@ export const EventCard = (props: IProps) => (
     data-eventId={props.event._id}
   >
     {props.event.moderation !== 'accepted' && (
-      <ModerationStatusText moderable={props.event} kind="event" top={'0px'} />
+      <ModerationStatusText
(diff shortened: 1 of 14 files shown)


chrismclarke (approved):
Looks good to me, I tested logging in and out with beta-tester users and non-beta-tester user and the research page link was gone and it would not let me directly navigate either.

I found one small error thrown for a case when there were no updates returned (still trying to access the length) so I added an extra commit to catch that one. Also I wasn't expecting the admin page to add beta testers but really nice to have so thanks for adding.

Overall looks great to me, thanks again for the contributions!

PR #1140, opened 2021-05-03
Core Updates - React v17, Yarn v2

### LINKED ISSUE #1124: [bug] - Pages requiring login don't support reload
**Describe the bug**
As mentioned in #116 - various pages such as howto editing or the admin dashboard require the user to be logged in (and sometimes also have specific user roles). The problem is, when a page is reloaded (or navigated to directly) there is a delay between the page loading and the user being logged in (even when the login is 'persisted' between page refreshes, there is a delay whilst the profile is loaded via firebase auth). This usually means that the user is kicked out the page.

There are a few workarounds for different pages, either by letting the page be accessible to everyone and just not displaying the content until the user is verified (no route guard), or by using timeouts to wait an arbitrary amount of time before kicking (research module)

**To Reproduce**

1. Create a howto
2. Go to edit
3. Refresh the page

**Expected behavior**
The system should wait to see if a previous user already exists and is waiting login before kicking

**Additional context**
Currently we have a single auth listener that listens to auth state changes in the user store, so we only know about login operation after it's been completed, not as it's starting. We should check to see if there's any other way of knowing that a potential login interaction has started or is availble, and exposing some sort of auth ready state that can be used to distinguish between there being no user persisted vs login pending. There might be a hacky solution checking the cookies/localstorage that firebase auth uses to persist this data, or maybe a better solution in their docs somewhere

### LINKED ISSUE #1141: Update core to React 17 and React scripts 4
Improvements to the service worker and caching strategy in #983 require more flexibility with service workers currently allowed with workback generate sw and react-rewired scripts. The latest cra release (4) adds much better support for custom service workers and so makes sense to update and then remove the custom react-rewired implementation to use a more standard approach.

There's also issues with circle-ci and the default webpack-terser-plugin that should be resolved by an update.

At the same time it makes sense to update a few other core dependencies, such as react 16->17 (no major feature updates but should provide smoother future updates) and related packages.

- Core updates
- General testing

Could further build out custom implementation, but doesn't really make much sense as will have to upgrade at some point eventually

**Additional context**
React scripts mentions some breaking changes to review during update (https://github.com/facebook/create-react-app/blob/master/CHANGELOG.md)

chrismclarke:
PR Checklist

PR Type

## Breaking Changes
Yarn update will invalidate current node_modules, so requres developers to delete their local `node_modules` folder and reinstall via `yarn install`

## Description
Update core platform dependencies to keep in line with recent feature and security updates

- Update yarn to v2, which should be faster for running locally, take up less disk space for node_modules and provide a smoother pathway to later split the repo into workspaces
- Upgrade react-scripts to version 4 and react to version 17
- Remove custom service worker integration (now supported natively in react-scripts 4)

Additional migrations required
- Type definitions to support eslint updates
- Minor typings fixes in app
- Remove/refactor react imports no longer required
- Fix forms broken by updates
- Circleci caching

Additional code changes
- Refactored auth routes to help with testing and to close #1124
- Refactored form unsaved changes dialog to standalone component for re-use

## TODO
- New service worker will need a bit more testing and refinement (planned in #983)
- Remove destructured namespace imports via `npx react-codemod update-react-imports` script (optional, might make sense to turn more into functional components during process)
- Storybook will likely need upgrade and testing (few known issues with yarn 2 but most seem resolved as of last year)

## Git Issues

Closes #1141 
Closes #1124 (see 43fb4cb)

## Screenshots/Videos

---

## What happens next?


chrismclarke:
All tests passing on circleci. Travis would need some updates for caching etc., but as we plan to remove shortly assume fine. So overall I think it should be good to merge.

@davehakkens - the one minor frontend change I made was to fix #1124 (direct nav to a page that requires login). I made a simple change so that instead of redirecting a non-logged in user (which is everyone on first load because of the delay to logging in) it simply shows a message instead. 

If a particular role is required it shows the role needed, or for everything else it just says to login first - is this ok for now?
E.g. linking to the research page shows:
image
(and will update if user logs in)

image
Example trying to access a howto edit page before login (note, if logged in and not an admin or the howto owner you'll still be kicked out)

PR #1156, opened 2021-05-18
Change display name to username

davehakkens:
We often mix username with a display name. This caused some confusement between users. This should clear it up

PR Checklist

PR Type

## Description
It aligns the wording from Displayname/username/profilename

## What happens next?


chrismclarke (approved):
Cool, I saw one test was failing after the update because the error message was hardcoded into the tests. Instead I've moved the message text to a shared `messages.ts` file which is read from both for this message and the test. If we make more changes in the future I'll try to also add to this file.

PR #1169, opened 2021-06-08
#724 Update react-pose(deprecated) to framer-motion

### LINKED ISSUE #724: Update react-pose (deprecated) to framer
React-pose [has been deprecated](https://github.com/Popmotion/popmotion/tree/master/packages/react-pose) for [framer motion](https://github.com/framer/motion).

Do a search of `react-pose` in the project and update where needed.

j-hertzog:
PR Checklist

PR Type

## Description

This PR replaces all instances of react-pose, a [deprecated](https://popmotion.io/blog/20200115-pose-is-deprecated/) animation library for a library called framer-motion.

## Git Issues

Closes #724

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "final-form-arrays": "^3.0.2",
     "final-form-calculate": "^1.3.2",
     "firebase": "^8.6.3",
+    "framer-motion": "^4.1.17",
     "fuse.js": "^6.4.6",
     "is-url": "^1.2.4",
     "leaflet": "^1.5.1",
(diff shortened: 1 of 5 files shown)


chrismclarke (approved):
Thanks for the PR, had no idea pose was deprecated so really good to have the update included. Also thanks for going through all the existing components to migrate, super helpful. 

In terms of frontend code I don't think we use the animations in many places, but all the ones I could find I tested (mobile menu open, profile save, howto save) and all seemed to be working in the same way so that's great.

Also in case of interest @davehakkens, if we do want to think about including more animations in the future then the new library @j-hertzog  has updated us to is called Framer Motion - you can see some examples in the docs here: https://www.framer.com/docs/examples/

Might be of interest to the design team

PR #1188, opened 2021-07-14
fix map pin detail

chrismclarke:
PR Checklist

PR Type

## Description

When mappins load they should retrieve details from db (including verified badge), except some legacy mappins already have details hardcoded and so were skipping this step. This PR fixes the behaviour (force always to look at db and ignore hardcoded data), as well as some minor code tidying. 

## Git Issues

Closes #

## Screenshots/Videos

image

---

## What happens next?


davehakkens:
nice

davehakkens (approved):
thanks, works here! :)

PR #1205, opened 2021-08-12
feat: verified icon in usernames

### LINKED ISSUE #1163: Display verified icon in username
### Description

If a user is verified it should show the icon next to the username

### Page related

How to overview
How to page
Events
Research overview (optional)
Research page (optional)

### Mockup
verified_events
verified
verified_home

### Life example
events: https://community-platform.webflow.io/u/events
map: https://community-platform.webflow.io/u/map
how-to: https://community-platform.webflow.io/u/how-to-article-no-comments

danitrod:
PR Checklist

PR Type

## Description

Adding verified icon to usernames in howto listings, howto pages, howto comments, events, research listings and research pages.

The approach chosen was to create a `fetchAllVerifiedUsers` method in the `userStore`, which queries the database for all users who meet the `badges.verified == 1` condition and stores the list in a `verifiedUsers` prop. Based on that list, each component that displays usernames basically checks if the username is on the list to display a badge along.

Note: I changed the value type of `badges.verified` from `boolean` to `number`, having the number be 0 or 1. The reason is that boolean was causing problems with Dexie as booleans are not indexable (ref: https://stackoverflow.com/a/56661425/10736715). I added a comment explaining this in the model definition.

## Git Issues

Closes #1163


danitrod:
I see tests are failing, will have a look at that later today

danitrod:
Alright, looks like all set now!

src/pages/Howto/Content/Howto/Howto.tsx
```diff
@@ -126,6 +129,9 @@ export class Howto extends React.Component<
         <>
           <HowtoDescription
             howto={activeHowto}
+            verified={this.injected.userStore.verifiedUsers.some(
+              user => user.userName === this.store.activeHowto?._createdBy,
+            )}
```
thisislawatts, line 134: **question:** Can we move the verified state to the HowTo document rather than perform this look up?

thisislawatts:
@danitrod I missed the original conversation around this feature.

Did we consider how `fetchAllVerifiedUsers` may affect the charges incurred by the Project? I understand that Firebase charges for each read operation on a record.

So this change would a new query of N records to the following routes:

1. Events listing
1. Howto listing 
1. An individual howto page 
1. Research listing
1. Individual research page

Things I either don't know or understand:

1. FireBase caching means we can make repeated queries to the DB without incurring additional charges
2. There are only a low number of verified users so this is not a significant issue
3. This approach has already been approved with full awareness of the cost implications

src/pages/Howto/Content/Howto/Howto.tsx
danitrod: I don't recall any specific reason why I did this, it's been a while but maybe I just didn't think it through enough. If we move this logic to the store and have the verified state in every HowTo doc, it would certainly be a lot more efficient and only perform the search once 😅

danitrod:
> @danitrod I missed the original conversation around this feature.
> 
> Did we consider how `fetchAllVerifiedUsers` may affect the charges incurred by the Project? I understand that Firebase charges for each read operation on a record.
> 
> So this change would a new query of N records to the following routes:
> 
>     1. Events listing
> 
>     2. Howto listing
> 
>     3. An individual howto page
> 
>     4. Research listing
> 
>     5. Individual research page
> 
> 
> Things I either don't know or understand:
> 
>     1. FireBase caching means we can make repeated queries to the DB without incurring additional charges
> 
>     2. There are only a low number of verified users so this is not a significant issue
> 
>     3. This approach has already been approved with full awareness of the cost implications

I don't have the basis to answer that right now as well, would have to study a bit more on how Firebase works. I don't remember if I talked with Chris about the implications on charges for this logic. Sorry, been some rough few months over here. Hopefully at the beginning of next year I'll be able to dedicate some more time into the project and help figuring this out.

thisislawatts:
Thanks for the speedy response @danitrod, of course look after yourself first. This discussion can wait 🤗

chrismclarke:
I have a good feeling this will finally be the week to unblock this PR!
As discussed on the calls and slack message last week, the blocking issue on this was how best to keep a shortlist of all user badges, as querying across all users is both tricky and inefficient. This is a common problem so decided best to come up with a general solution the make it easier for storing similar aggregations, via an _aggregations_ db endpoint and set of backend functions

This just leaves a few small tasks

- [ ] Add db bindings for aggregations system
- [ ] Resolve conflicts (mostly just minor import issues)
- [ ] Small refactor to point to aggregations endpoint
- [ ] Test and release

I'm working on the DB bindings now and will link the relevant PR when open

PR #1207, opened 2021-08-18
Add local backend testing environment

### LINKED ISSUE #1203: [feature request] Add local development database emulators
When testing things like triggered functions (e.g. when a user updates their location automatically update their map pin), the workflow isn't ideal as we can only test after the functions have been deployed on the master branch, there is no way to revert changes if implemented badly, and can incur billing costs if lots of changes made. 

Adding local firebase emulators should make this process much smoother to run entirely in a local environment, without billing and able to rollback changes.

Provide integration an option to use the updated [Firebase emulators suite](https://firebase.google.com/docs/emulator-suite) when developing locally. This should probably be opt-in as requires a bit of extra technical setup (e.g. having java installed) so good to keep existing workflow for regular developers also.

If we tried to fast-track planned database migrations there could be other options, but unlikely to happen in short term so this is a good interim solution

**Additional context**

chrismclarke:
PR Checklist

PR Type

## Description

Adds a new suite of local emulators so that the platform backend can be tested locally. This will make it much, much easier to test new and create new backend functions, without the need for users to be invited as admins to the staging project and having to build and deploy to the staging server each change made. It also provides a way to reverse changes made when testing, which previously was not easily possible.

Also included is a set of custom scripts to load backups from the production site into the emulator environment, so that scripts can be tested against real data (no more deploy-and-hope-for-the-best! 😂 )

These emulators do not replace the standard workflow for developers wanting to use staging and preview sites, but provide a handy alternative for those wanting to work more closely with backend functions. 

## Breaking Changes
While testing it was noticed that deleting documents (howtos, mappins, events etc.) triggered the functions to send slack/discord notifications. 

Both these integrations have been updated to only trigger on specific events (`created` for notifying slack moderators of new content and `updated` for notifying discord users of moderation accepted), which I think should be the expected behaviour, although they should be tested more properly when merged to master and deployed on the dev site.

## Git Issues

Closes #1203 

## Screenshots/Videos

Link to added [Firebase Emulators documentation](https://github.com/ONEARMY/community-platform/blob/2a0a82d3a697e3e1d63e1072abce3d6dbb5db6a4/packages/documentation/docs/Backend%20Development/firebase-emulators.md)

Local emulator dashboard - running with a cleaned/testing copy of main site data
image

---

## What happens next?


chrismclarke:
Should probably fix these tests.... 😅 
But hopefully should be working for other to start testing out locally

patrycjapraczyk:
Hi @chrismclarke . I am wondering if it matters what account you log into with firebase or if you need to be granted some permissions to view the local database.
When running `npm run start:emulated` I got the following error message:
image
Then when trying to fix this by running `firebase use default` I got:
image
I was logged in to firebase with my personal Google account.

chrismclarke:
That's interesting, I know you need to be logged in to use resources on the firebase server but had hoped they would skip that when running locally. I tried logging out myself and running the functions start script and did receive a warning message:
```
> yarn workspace functions start
!  emulators: You are not currently authenticated so some features may not work correctly. Please run firebase login to authenticate the CLI.
```
But the rest of the scripts did seem to run at least. Although I'm running on windows, and when I retried on Linux I saw the same error as you, so guessing that's a bug with the firebase CLI (really good to know).

I just tried manually specifying which project to use as part of the start script to see if that would at least prevent firebase from trying to access the server to see the list of projects available, and for me it actually seemed to work to start the emulators without being logged in on Linux.

I've just pushed a small update with that one line of fix, let me know if it works for you.

If not then I can just invite you to the project and update the notes to say you must be logged in (although much nicer if all devs can just use without an invite)

chrismclarke:
Although now I seem to be getting another issue on Linux when the emulators try loading the default seed data (it seems to be getting file paths a bit messed up, e.g. mixing up forward and backward slashes for files like `data/emulated/firestore_export/all_namespaces\all_kinds\all_namespaces_all_kinds.export_metadata`.

It might be because I created the seed data on a windows machine - if you can let me know if you get the same issue then that would be really useful to know, and I'll see if I can find a way to fix.

chrismclarke:
OK, that was a pretty deep rabbit-hole! Long story short Firebase only lets the cli work with authenticated users because despite the emulators not requiring auth they call various external APIs whilst running commands which do require authentication.

As a workaround I've added some default credentials that it will attempt to sign in via an automated service account (so no firebase login required). In order to use them you will just need to pull the updated commits and run `yarn install` again. 

I've also updated the default seed data to (hopefully) work on Linux, so you will also need to run the `yarn workspace functions run emulator:seed` function again

Let me know if it's working for you with these changes, if not I can dive back down the hole again! 😅

chrismclarke:
Well that's one step in the right direction at least! Sorry things still aren't working though @patrycjapraczyk 

The error seems to suggest that it's now failing whilst trying to compile the backend functions, which actually makes sense. Basically the functions imports code from the main repo src folder (namely models imported in `functions/src/models.ts`), but for whatever reason it doesn’t seem to be able to detect it properly. 

I've just tried checking out the repo clean in Linux and can see the same error so at least I should be able to troubleshoot it from there and will let you know when I get it figured out 🤔

chrismclarke:
OK I think I've worked it out - the type definitions in the src folder have to first be compiled so that they can be imported into the functions (essentially an extra build step). I had included that step as part of the functions build scripts but had forgot to include when trying to run locally. I've just updated the scripts to (hopefully) automate this step a bit better, and at least now on my Linux machine things seem to be running correctly

image

Hopefully this will work for you too (just sync the latest commits), but if not then feel free to share any new error messages and possibly the version of node and yarn you are running (`node --version` `yarn --version`) in case there are any issues with that

patrycjapraczyk:
@chrismclarke success! It works pretty well, thank you for your help. I can view the Firestore as in the screenshot:
image

chrismclarke:
That's great to hear @patrycjapraczyk , thanks so much with your help with all of this. Will merge into the main codebase now and can hopefully start to use to develop the new notifications module!

TinkeringAlways:
Wondering if anyone has had success on a windows machine? Getting this error:
image

chrismclarke:
@TinkeringAlways 
oh, that look like my bad! Yes I did get things working on a windows machine but I think I likely have the [concurrently package](https://www.npmjs.com/package/concurrently) installed globally so that it can be called directly like that (really it should be called from the installed `node_modules` folder instead. 

Will make a quick fix shortly, but in the meantime you can run `npm i -g concurrently` as a temporary workaround.

Oh, and quickly unrelated, I know you had mentioned you were planning to send me an email to follow-up on a few things. I haven't received anything yet - just in case you have the wrong address and still need to get in touch you can send to `chris.m.clarke@live.co.uk` or message me if you are on slack

TinkeringAlways:
@chrismclarke 
So, the first instruction did allow for the site to build and deploy in emulated mode. However the local instance of Firebase was not accessible from port 4001 on localhost. I did try the second instruction, and it didn't seem to have any effect after going through the terminal output.

Here is the term output before the second change.

image

chrismclarke:
@TinkeringAlways - looks like it's having an issue with handling folder path names (the subfolder you have cloned the repo to contains spaces which is throwing an error). The quick/lazy solution would probably be to move everything to a more top-level folder (e.g. on my machine I use `C:/apps` for my repos) - you get the added bonus of not syncing everything to google drive which is probably quite a large/lot of files once the 3rd party dependencies have installed in the node_modules folder and the `.git` folder is extracted.

But longer term would be good to know how to write code that works on windows even with complex path names.

TinkeringAlways:
@chrismclarke - Ok, I set up a new directory at C:/scratch. And have success deploying the server instance again on Node. 

However, now there is another issue with the Firebase emulator. I will look into it more tomorrow, but for now here is the term output.

image

chrismclarke:
hmm, yeah for some reason it doesn't seem to be loading the set of temporary credentials that give access to the testing project. Basically even though we're running the emulator locally, firebase still expects the user to be authenticated against a real project to allow it to make some specific API requests. To avoid the need of manually adding every developer to one of our firebase projects I added a service account with limited permissions, and a script that attempts to authenticate using that. But given the error message says `please verify project precious-plastics-v4-dev exists and you have access` it seems that this piece of the puzzle is currently not working for you...

Can you check that you have both firebase cli installed (`firebase --version`) and java installed (`java -version`)? Docs with links to installers can be found in `packages\documentation\docs\Backend Development\firebase-emulators.md`

If this still doesn't work I'll just invite you manually to the project so that you don't need to use the temporary credentials.

TinkeringAlways:
The tooling is there...
image

Want me to create an issue for this?

PR #1208, opened 2021-08-26
Fix lost transparency on PNG images

### LINKED ISSUE #933: [bug] - PNG images lose transparency (background turns black)
**Describe the bug**
Uploaded pngs seem to have a black background instead of transparent.

**To Reproduce**
Add a png with transparency to any how-to upload

**Expected behavior**
Keep transparency. If converting to JPG (which I don't think happens) then probably default to white background.

**Screenshots**
image

iconfinder_wuhan-coronavirus-virus-outbreak-02_5730675
Original

mef:
PR Checklist

PR Type

## Description

This Pull request fixes the bug #933, where PNG images get a black background instead of transparency.

This is done by replacing `image-compression` dependency by `browser-image-compress`.

It also does a small update to commands listed in CONTRIBUTING.md.

N.B.:

* images previously uploaded are not modified. These should keep their black background, if they have one.
* the fixed version does not keep as much metadata as previous one. It is not clear to the developer whether this metadata was ever used or stored anywhere:
  * `startSize`: original image size
  * `endSize`: compressed image size
  * `compressionPercent`

## Git Issues

Closes #933

## Screenshots/Videos

Before:

---before

After:

after

## What happens next?


chrismclarke:
image
Looks awesome to me!

CONTRIBUTING.md
```diff
@@ -51,16 +51,13 @@ You can find useful links to learn more about these technologies [in the resourc
   `yarn start`
 
 - Run the end-to-end tests
-  `yarn cy:test`
```
chrismclarke, line 54: Thanks for tidying these up!

I know we don't really have a good test for checking the images, I'll just quickly add a png to one of the existing specs for now to make sure the filetype is at least handled and maybe in the future if we add visual difference testing it will pick up if there's any more problems down the road.

src/components/ImageInput/ImageConverter.tsx
```diff
@@ -39,8 +39,7 @@ export class ImageConverter extends React.Component<IProps, IState> {
     super(props)
     this.state = {
       compressionOptions: {
-        maxWidth: imageSizes.normal,
-        quality: 0.75,
```
chrismclarke, line 43: I can see https://www.npmjs.com/package/browser-image-compression does have an option for `initialQuality` so it might be good to keep in case we ever want to change it

I'll add a quick commit for that now

chrismclarke (approved):
Thanks for this @mef , it looks great!
Makes sense to me that we leave existing images unchanged (people can re-upload if they want to fix).

Yes I think that's sensible to remove the metadata, we used to have a small feature where you could pick the quality size you wanted (e.g. low, normal, high) and it would popup with a message saying `you just saved 72% file size!` or similar, but now we just use default settings for everyone so no longer required.

PR #1212, opened 2021-09-18
Fix comment flag storing

### LINKED ISSUE #1211: [bug] Country flag not showing in comments
**Describe the bug**
The country flag of a user is not showing in the comments.
[(example bottom of page)](https://community.preciousplastic.com/how-to/plastic-medals--free-injection-mould-design)

**Expected behavior**
Flag should be shown like in this issue #1127 
afbeelding

**Screenshots**
Possible problem mentioned by @tudi2d on Slack
Could this be the reason? (Image from the user page) For the `addComment` in the store it is only `creatorCountry: user.country ? user.country.toLowerCase() : null` saved to the backend.

afbeelding

danitrod:
PR Checklist

PR Type

## Description

Adds the `user.location.countryCode`, when `user.country` isn't available, to comments. Hopefully this will fix the issue of some comments not displaying the flags.

## Git Issues

Closes #1211

### DIFF
--- a/src/stores/Howto/howto.store.tsx
+++ b/src/stores/Howto/howto.store.tsx
@@ -151,7 +151,11 @@ export class HowtoStore extends ModuleStore {
           _created: new Date().toISOString(),
           _creatorId: user._id,
           creatorName: user.userName,
-          creatorCountry: user.country ? user.country.toLowerCase() : null,
+          creatorCountry: user.country
+            ? user.country.toLowerCase()
+            : user.location
+            ? user.location.countryCode
+            : null,
           text: comment,
         }
 


src/stores/Howto/howto.store.tsx
```diff
@@ -151,7 +151,11 @@ export class HowtoStore extends ModuleStore {
           _created: new Date().toISOString(),
           _creatorId: user._id,
           creatorName: user.userName,
-          creatorCountry: user.country ? user.country.toLowerCase() : null,
+          creatorCountry: user.country
+            ? user.country.toLowerCase()
+            : user.location
+            ? user.location.countryCode
+            : null,
```
thisislawatts, line 158: **suggestion (non-blocking):** This looks great @danitrod, what do you think about using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) instead of nested ternary operators? It is subjective, but I feel that this format is easier to read.

Also note that the suggestion adds `.toLowerCase()` for the nested location.countryCode property to ensure consistency. 

```suggestion
          creatorCountry:
            user.country?.toLowerCase() ||
            user.location?.countryCode?.toLowerCase() ||
            null
```

danitrod: Great suggestion @thisislawatts! Definitely makes the code more readable and scalable, commiting

PR #1221, opened 2021-10-03
fix: yarn install script

thisislawatts:
PR Checklist

PR Type

## Description

Generate a new yarn lock file as I noticed other PRs raised against this repo fail at dependency install step currently.

Not to sure about the _why_ here, I don't work with Yarn too much currently. From what I understand it looks like the current immutable lock file can not be used (🤯) and instead requires a new lock file to be generated. So I have done just that without any meaningful context on how this will affect things as a _lot_ of dependencies have changed.

Let's see what Circle thinks…

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5,31 +5,31 @@ __metadata:
   version: 4
   cacheKey: 7
 
-"@algolia/autocomplete-core@npm:1.2.1":
-  version: 1.2.1
-  resolution: "@algolia/autocomplete-core@npm:1.2.1"
+"@algolia/autocomplete-core@npm:1.2.2":
+  version: 1.2.2
+  resolution: "@algolia/autocomplete-core@npm:1.2.2"
   dependencies:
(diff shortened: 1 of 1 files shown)


chrismclarke (approved):
Thanks for adding this - yeah yarn lock files can be a bit fiddly, we tend not to pin dependencies too strongly so every now and then packages appear that don't download correctly (particularly some of the larger ones like cypress). But at least we've come a long way since the days we had both yarn and npm lock files fighting it out in the repo :D

chrismclarke:
Looks like circleci still a bit unhappy unfortunately - I remember having the same issue a while back and took quite a bit of fiddling with different dependencies/versions to resolve. I'll try update a few other core dependencies and see if that fixes things

chrismclarke:
OK, looks like this PR is passing the build stages at least:
https://github.com/ONEARMY/community-platform/pull/1223

Thanks again for flagging the issue and getting us on the road to a fix!

PR #1225, opened 2021-10-03
feat: adds redirect for 404 how tos

thisislawatts:
PR Checklist

PR Type

## Description

There are many reasons a how to may not be available
on the server. Such as:

* Deleted :fire:
* Renamed :pencil2:

In an effort to make the experience less of a dead
end for the visitor we redirect them to the search
page.

This current implementation takes place on the client
side after attempting to load a how-to entry from
the backend data store.

## Git Issues

Reference: https://github.com/ONEARMY/community-platform/issues/1168


cypress/integration/howto/read.spec.ts
```diff
@@ -195,4 +197,18 @@ describe('[How To]', () => {
         .should('include', `${specificHowtoUrl}/edit`)
     })
   })
+
+  describe('[Fail to find a How-to]', () => {
+    const uuid = crypto.randomBytes(16).toString('hex')
+    const howToNotFoundUrl = `/how-to/this-how-to-does-not-exist-${uuid}`
```
thisislawatts, line 203: 🤔 Perhaps unneccessary as I _think_ the datastore is completely reset between test runs so there are unlikely to be any collisions due to leaky state between test runs.

chrismclarke: It should do (cypress by default clears localstorage and we have manual methods to clear indexeddb), we also use a custom prefix in case of tests running in parallel. That being said, no real harm to include

src/pages/Howto/Content/Howto/Howto.tsx
```diff
@@ -158,7 +157,18 @@ export class Howto extends React.Component<
...
+      return isLoading ? (
+        <Loader />
+      ) : (
+        <Redirect
+          to={{
+            pathname: '/how-to',
+            search:
+              '?search=' +
+              (this.props?.match?.params?.slug).replace(/\-/gi, ' '),
+          }}
```
chrismclarke, line 169: Tested this out and really like the functionality, nice idea to also replace the dashes and make it look more human-readable

chrismclarke (approved):
Thanks for the contribution, really nice addition to have.
From my side I really like this and think it works well. I'll quickly check in with Dave before merging to see if he has any strong opinions whether we should also display some sort of message (hopefully not, as that could get a tiny bit messier assuming we would need to pass an extra URL param to let the page know it was redirected to after failed search or some other means)

chrismclarke:
Just to quickly note, as mentioned on the call would be good to have some sort of message to show to let users know that the howto they were searching for couldn't be found. 

Feel free to draft/design any first pass at this that you want and once the PR is updated tag Dave in review for feedback on the design and if you have any concerns about the code feel free to tag me for re-review.

chrismclarke (changes requested):
Just to quickly note, as mentioned on the call would be good to have some sort of message to show to let users know that the howto they were searching for couldn't be found.

Feel free to draft/design any first pass at this that you want and once the PR is updated tag Dave in review for feedback on the design and if you have any concerns about the code feel free to tag me for re-review.

thisislawatts:
@davehakkens see below for a preview of how the search page looks if you arrive via a 404 how to link. I've changed the heading and introduced a sub header as an extra prompt for the user. Let me know what you think:

Screenshot 2021-10-04 at 21 55 55

davehakkens (approved):
What a beautiful artwork @thisislawatts! 
But yeh, looks neat :)

chrismclarke (changes requested):
One small issue I noticed, is that the source does not reset after navigation, and so the message stays there even after the user opens a howto and returns to a clean howto page (source no longer in URL). So maybe we need to move the check to the list page itself onComponentDidMount/useEffect hook, or just add something to the store to reset when setting an active howto?

**When navigating back the URL changes to just be /howto but the message still exists - would be good to revert to standard page message**

https://user-images.githubusercontent.com/10515065/135929045-6fcb55d4-a2e9-4b8b-bd7f-d2941025ff34.mp4

thisislawatts:
Thanks @chrismclarke, very helpful! The persistence of state is an existing behaviour which my change makes more visible 🤔 

Try the following (desktop device):
1. Perform a search on https://community.preciousplastic.com/how-to, for example 🔍 `table lamp`
2. Navigate to [Events](https://community.preciousplastic.com/events) using the top menu
3. Click back to the How to listing, the original search query remains in the box.

Is that deliberate and designed behaviour?

chrismclarke:
Oh, I see what you mean, the search param disappears but the box and search contents persist. No, none of this was active design choices and I don't think has been noticed up till now. I'm guessing probably easiest if we merge this in and create another issue to address? 

Unless there was any quick fix you wanted to apply - I'm assuming the expected functionality should be to take the URL bar as main source of truth, so that if navigating back via browser back button search will be retained in URL and UI, but if navigating by menu bar or other triggered methods URL and page will be clean state. But not sure how quick a fix this would be so happy to write-up in another issue if easier to manage that way

thisislawatts:
Yes, I am still gaining context on the MobX module and state management, I found this interesting article entitled, [How to decouple state and UI (a.k.a. you don’t need componentWillMount)](https://hackernoon.com/how-to-decouple-state-and-ui-a-k-a-you-dont-need-componentwillmount-cc90b787aa37), from 2016.

Interesting, but definitely not straight forward to implement if it's even a valid approach. My experience with SPA of this complexity is limited. So I am here to learn.

thisislawatts:
In the meantime I will take a look at a hotfix

chrismclarke:
Yeah no worries at all. I know the existing codebase is a bit messy around these things (we have lots of different lifecycle hooks used in different ways, some of which are now deprecated (e.g. componentWillMount vs componentDidMount vs UseEffect vs unsafe_componentWillReceiveProps etc.)), and mobx itself has had quite a few breaking changes moving through v4 -> v5 etc. 

So yeah, as long as you're having fun basically! And if there's any particular pain points you come across feel free to reach out on slack and I can let you know if I can share any insights around it

thisislawatts:
Useful context @chrismclarke, I've pushed a fix which uses the `UNSAFE_componentWillMount` hook for the time being but I think there's an opportunity to a refactor at a later date.

chrismclarke:
Nice - makes sense to me for now. There's still the issue that the search box contents persist across navigation (unrelated to this pr) so can just do the extra bit of refactoring when we turn to address that

PR #1226, opened 2021-10-03
chore: push circle docker image to latest node 12 lts

thisislawatts:
PR Checklist

PR Type

## Description

1. Bumps the CircleCI node image to the latest LTS release. Although I don't yet have enough understanding on how that may align with the Firebase deployment environment. 
  1a. Adds a `.node-version` file to the repo which allows various developer tools such as [fnm](https://github.com/Schniz/fnm) and [others](https://stackoverflow.com/a/62978089) to load the correct version. I appreciate that the documentation currently mentions _nvm_ but maybe that is something open for discussion 👼🏻.


chrismclarke (approved):
Makes sense to me, I think most developers are likely only going to be using node to serve the frontend react app so can keep recommendations light (I included mention of node 12 in the docs, but as 14 is now the LTS should probably also ensure it works. If using nvm we could also add a `.nvmrc` file, but up till now at least I haven't heard from people having specific node version issues).

For backend/ci it might also be good to start looking into 14 as security support for 12 is due to end next year (https://endoflife.date/nodejs), but equally happy for now to just bump to latest stable release.

Firebase functions configure node runtime from the `engines` field in the local `package.json` - it's currently set to 12, not sure if it supports pinning an exact version or not (docs here: https://firebase.google.com/docs/functions/manage-functions#set_nodejs_version)

thisislawatts:
@chrismclarke I think it would be great to aim for 16, but deliberately go via 14 as an intermediate step. This will give us the best ROI and ensure we've got the longest possible support window for the project.

PR #1231, opened 2021-10-05
[feature] Show display name on map pins, instead of the username

### LINKED ISSUE #1210: [feature request] show display name on map pins, instead of the username
My "username" doesn't quite match the workspace, so this is confusing both on the map and in the how-tos. See this:
image.
Now when I click on that name, I get to the profile page, which doesn't say the name anymore, but rather the name of the workspace, or "Display Name" as it's called in the settings
image

Replace most/all instances of the user name with the "Display Name". Most importantly in the "popups" for pins on the map, but maybe also in the thumbnail things for how-tos: 
image

Another and maybe better, but also probably much more involved and complicated solution would be to split users and map pins, so one user can maintain multiple map pins, with their own names. This would be much more complicated to implement I'd assume.

**Additional context**
I did some digging but could find any other issues matching this, but some that are related, like #1200 and also #773.

tiefpunkt:
PR Checklist

PR Type

## Description

Instead of showing the username in map pins and their popup, these changes use the Display Name instead, as discussed in #1210

To achieve this, I added the displayName field to IMapPinDetail, and use that displayName in map popup.

I hope I got all the spots where I needed to change something, not much of a JS developer, and I never worked with React before 🙈 

Side note: do you want to join hacktoberfest?

## Git Issues

Closes #1210

## Screenshots/Videos
image

image

---

## What happens next?


davehakkens:
nice one. Works good here!

thisislawatts (approved):

PR #1232, opened 2021-10-05
Missing build dependencies & fixing node-sass for macOS Big Sur / Node 16

tiefpunkt:
PR Checklist

PR Type

## Description

I was having trouble building and testing on macOS BigSur / Node 16. The `node-sass` package wouldn't build, and what I found was that it's related to Node 16 ([StackOverflow thread](https://stackoverflow.com/questions/67241196/error-no-template-named-remove-cv-t-in-namespace-std-did-you-mean-remove)). Updating it fixed this issue.

I was also unable to run `yarn lint` due to missing `eslint`, and got an error when committing due to missing `eslint-plugin-react`.

## Git Issues

_none_

## Screenshots/Videos

---

## What happens next?


thisislawatts:
@tiefpunkt the project is currently [pinned to v12 of Node](https://github.com/ONEARMY/community-platform/blob/master/.node-version)

Have you got a tool on your machine to manage multiple Node versions? Currently I am using [fnm](https://github.com/Schniz/fnm) to do this as it can be configured to switch versions automatically as you move between projects in the terminal.

tiefpunkt:
@thisislawatts oh, didn't know that, haven't noticed that file either 🙈 Thanks for the hint

No, I don't have any tooling like that, never really do any Node.js programming unless to fix open source stuff that bothers me 😅 I'll certainly look into that now.

But at least the `yarn lint` dependencies wouldn't be impacted by the Node version, though, right?

thisislawatts:
@tiefpunkt I will mark this PR as closed in favour of #1233. Thanks again for taking the time and care to raise your original PR ✨

chrismclarke:
> Please take a look at `eslint-plugin-react` as well, which was required for some pre-commit hook apparently. Or is that also a node 16 vs 12 thing?

Thanks for investigating this. It should be installed as a child dependency of react-scripts, and you're right it's triggered from the `lint-staged` precommit hook which just runs eslint (so nice to have eslint properly defined in there!). The eslint config applies `plugin:react/recommended`  which is what calls the `eslint-plugin-react`. 

All of this should be pretty much the same as the boilerplate from [Create React App](https://create-react-app.dev/docs/setting-up-your-editor) with some extra config for prettier, although I think since it was first put in there have been a few manual updates so a chance it's not perfectly in sync.  

But it looks like perhaps that repo doesn't fully support node 16 yet (`https://github.com/facebook/create-react-app/issues/11037` ), so I'd say probably better to use 12 (if possible), and we can create an issue to start a move towards 14

PR #1242, opened 2021-10-09
feat: adds theming to support multiple instances of the platform

thisislawatts:
PR Checklist

PR Type

## Description

Eventually the community-platform service will be installed as multiple instances to help grow communities around fixing fashion and sustainable living. This PR represents a step towards that goal by moving the **styling** out of individual components to a higher level abstraction represented by the type `PlatformTheme`.

* Adds a theme switcher to the `DevSiteHeader` component, which is mainly for demo purposes to demonstrate how the site can be themed for different instances. 

For the reviewer:

* 😬 Sorry for the many files being changed in this PR, a lot of this is small and repetitive changes so that components fetch their style from the Theme store rather than a hardcoded style file.


thisislawatts:
@davehakkens Updated to add a variant for use on Project Kamp ✨ 

Screenshot 2021-11-20 at 16 13 06

package.json
```diff
@@ -50,7 +50,6 @@
   "resolutions": {
     "__NOTE__": "2021-05-28 storybook clash: https://github.com/storybookjs/storybook/issues/6505",
     "webpack": "4.44.2",
-    "styled-components": "^4",
```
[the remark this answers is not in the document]
chrismclarke: I think that might have been related to storybook back when all package installs were hoisted to top-level, but as we now have `nmHoistingLimits: workspaces` I expect fine to remove

src/components/Avatar/index.tsx
```diff
@@ -1,6 +1,6 @@
 import { Component } from 'react'
 
-import { Image, ImageProps } from 'rebass'
+import { Image, ImageProps } from 'rebass/styled-components'
```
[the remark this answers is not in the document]
chrismclarke: I remember at one time the lead frontend (Benj) had wanted to move to emotion, but as we had only just moved from material -> styled components seemed like one step too far. Would definitely be outside scope of the PR to make a switch so seems like a sensible option until such a time if we did decide (npm trends still has styled-components way more popular but I have no strong opinion on the matter otherwise)

src/assets/images/themes/unbranded/logo.svg
```diff
@@ -0,0 +1 @@
+<svg width="160" height="149" xmlns="http://www.w3.org/2000/svg"><g stroke="#F82F03" stroke-width="14.12" fill="none" fill-rule="evenodd" stroke-linecap="round" stroke-linejoin="round"></g></svg>
```
chrismclarke, line 1: Really like idea of having an `unbranded` theme! Although definitely voting unbranded => wacky design a la 00s myspace

src/components/DevSiteHeader/DevSiteHeader.tsx
```diff
@@ -1,64 +1,83 @@
...
-  </>
-)
+      )}
+    </>
+  )
+})
 // we have 2 different dev sites, only show this component when on one and provide select
 const devSites = [
-  { value: 'localhost', label: 'Dev' },
+  { value: 'dev_site', label: 'Development' },
```
chrismclarke, line 80: Thanks for fixing this, was wondering why I hadn't seen it recently :S

src/pages/Howto/Content/HowtoList/HowToCard.tsx
```diff
@@ -71,4 +73,14 @@ export const HowToCard = (props: IProps) => (
   </Flex>
 )
 
+function formatCountryToISOCode(str) {
+  if (str === 'united kingdom') return 'gb'
```
chrismclarke, line 77: Is this to fix the issues seen in dev console or related to #1211?
In either way I don't seem harm in it, but am confused as to why there is such a short/explicit list of countries?

src/pages/Research/Content/Common/Research.form.tsx
```diff
@@ -51,7 +51,7 @@ const beforeUnload = function(e) {
 }
 
 const ResearchForm = observer((props: IProps) => {
-  const store = useResearchStore()
+  const researchStore = useResearchStore()
```
chrismclarke, line 54: nice to be clear, thanks for this!

chrismclarke (approved):
Thanks for the huge amount of work that's gone into this, I can see how extracting as much of the hardcoded styles/text as you have really goes a long way to making the site re-usable for other sites.

A couple specific points of feedback:
1. Having to inject the themeProvider into every component feels a bit awkward, particularly when there's added boilerplate of having to define injected props or pass via commonStores. I know this is the pattern already inherited from the mobx setup, but I'm wondering if there might be more elegant solutions possible using either a `useContext` hook or better yet a custom modified/adapted `useTheme` hook like in some of the examples here: https://styled-components.com/docs/advanced

I wouldn't go as far as to say it's worth refactoring anything at this stage, but might be worth testing so we can provide as an option for new components moving forward.

2. Each of the current themes are pretty large in size, but contain mostly identical information. I'm expect this is mostly for the purpose of getting things to work quickly now and that they'll want freedom to diverge more in the future, but equally I think there's probably a lot of 'core' theming that's unlikely to change much and might worth separating out (e.g. spacing, font sizes etc.). Better yet splitting the theme across layout, typography, colors etc. might be useful. In either case I'd probably recommend we come up with a system that allows one theme to inherit from another (e.g. the core unbranded), and making changes on top instead of writing from scratch. 

This again I'd recommend as a future task, as what's there currently is still really nice to have and a good base to work from.

3. Given that this PR touches pretty much every component in the codebase and we have no form of automated visual regression testing, I'd strongly suggest/request we have a deep manual review like you've already started @davehakkens - aware that really we want to test everything (from main pages, hovering stuff, clicking buttons, seeing popups while uploading a howto etc.). 

Otherwise from the code side I'm happy to approve, overall the changes are a really nice addition to have and unlock a lot of exciting things for the future!!! Thanks again!

Looking forward to getting this merged once we get design sign-off from @davehakkens and can get the tests to pass

chrismclarke:
As for the tests - I've just checked out and tried running locally. There's quite a lot of issues with some of the props inheritance (trying to use `auxilary` text type but props are undefined) and mobx also unhappy about some of the nested structures (possibly related to `https://github.com/styled-components/styled-components/issues/3117`)

I think this is gonna take a deeper dive to untangle than I have time for right now, so I'll wait to hear back from @thisislawatts in case any of this is resolved in #1300 and if not we can perhaps troubleshoot on slack if useful

thisislawatts:
Closing in favour of #1300, leaving this branch hanging around for a little longer for reference.

PR #1243, opened 2021-10-10
chore: looking at you Cypress 🥸

thisislawatts:
PR Checklist

PR Type

## Description

Cypress looks to be broken in Firefox. Pushing an empty commit to take a closer look and see what can be done about this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -303,8 +303,6 @@ jobs:
     # build matrix will run 4 parallel builds handled by cypress, so don't need to specify more here
     parallelism: 1
     parameters:
-      CI_NODE:
-        type: integer
       CI_BROWSER:
         type: string
     steps:
@@ -328,8 +326,6 @@ jobs:
           command: npm run test ci prod
           environment:
             CI_BROWSER: << parameters.CI_BROWSER >>
-            CI_NODE: << parameters.CI_NODE >>
-            CI_GROUP: 2x-<< parameters.CI_BROWSER >>
 
 ######################################################################################################
 #  Workflows - Collections of jobs to define overall processes
@@ -348,12 +344,11 @@ workflows:
               BUILD_ENV:
                 ['FORCE_COLOR=1 REACT_APP_SITE_VARIANT=test-ci']
       - test_e2e:
-          name: e2e-<< matrix.CI_BROWSER >>-<< matrix.CI_NODE >>
+          name: e2e-<< matrix.CI_BROWSER >>
           requires:
             - 'build_test'
           matrix:
             parameters:
-              CI_NODE: [1, 2]
               CI_BROWSER: ['chrome', 'firefox']
       #---------------------- Dev Build-Deploy ----------------------
       - build:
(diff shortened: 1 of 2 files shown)


thisislawatts:
Based on this documentation:
> During parallelization mode or when grouping runs, Cypress will wait for a specified amount of time before completing the test run in case any more relevant work remains. This is to compensate for various scenarios where CI machines could be backed-up in a queue.
> This waiting period is called the run completion delay 
> https://docs.cypress.io/guides/guides/parallelization#Run-completion-delay

I believe the issue is that some of our Cypress tests fail to run because the CI machine has not contacted Cypress within the required 60seconds. For example see here, where the Firefox job takes >1 minute to boot:
https://app.circleci.com/pipelines/github/ONEARMY/community-platform/486/workflows/eea0a138-672c-4739-b733-3a5204898080

We _could_ look further into this and perhaps debug the CI build times. I don't understand why it is reinstalling all of the yarn dependencies again as this has already been done previously. However on our current [Circle CI plan](https://circleci.com/pricing) we can only run one job at a time. So parallelisation isn't actually providing any value here.

chrismclarke:
We definitely do want the parallelism, circleCI still allows 4 steps to run in parallel on free tier so that's what splits tests across 2xchrome and 2xfirefox in parallel and greatly reduces test time.

Also timeout shouldn't be an issue as the scripts that trigger the test runner (`packages\cypress\scripts\start.ts`) wait for the platform to be responsive before starting the test suite (this timeout was a major issue previously however, hence the scripts)

I think the issue I can see in those runs is that since removing parallel runs we can no longer attach to the same cypress build id (as that is expected for parallel runs)

chrismclarke:
As for the yarn packages reinstalling - this is needed as each of the parallel runs occur in their own isolated environment and so need dependencies installed. In theory we could install as a step earlier and pass the installed state to each runner, but storing all that as a layer between steps is actually slower than just reinstalling (node-modules folders are a pain to share). Usually the process is quick because it has a cache to read from, however that becomes invalidated after each package.json update and only recreated after successful test run

chrismclarke:
But agreed cypress can be a bit of a pain when it's not playing nice!

thisislawatts:
@chrismclarke aha, that is really useful, maybe I don't correctly understand the difference between jobs and steps. Although it looks like the account is currently only running one job:
Screenshot 2021-10-10 at 17 48 13

Could be we've crossed the threshold for free credits so have been throttled down to one job at a time? 🤔

chrismclarke:
image
I even tried turning the setting off and on again (the ultimate fix). But still no joy... Yeah I'll have to reach out to them, sorry it's such a pain!

chrismclarke:
OK, I've made a support ticket. Also strange because looking through docs and even if we were only receiving free tier credits they are supposed to reset sunday at 12UTC (and the dashboard shows us well within the free credit tier also). 

If we need a temporary workaround it would be to disable parallelism (as you previously did) but also to remove the `CIRCLE_WORKFLOW_ID` from the `buildId` in `packages\cypress\scripts\start.ts` to prevent cypress expecting parallel builds. 

But I'd probably say for now can just accept tests might be broken for a few days and prioritise fix whenever next ready for a release (and run tests locally as required until then)

PR #1252, opened 2021-10-24
test: introduce test infrastructure to functions

thisislawatts:
PR Checklist

PR Type

## Description

The aim of this change is to introduce the infrastructure so 
that future contributors to the `functions` workspace can 
be asked to include tests with minimal overhead.

A test has been added to demonstrate how functions triggered
by changes within the FireStore data can be validated. 

If you want to run the tests locally you can do so using
the [firebase tools](https://www.npmjs.com/package/firebase-tools)

```bash
firebase emulators:exec --project default "jest ."
```

```bash
firebase emulators:exec --project default "jest . --watch"
```


thisislawatts:
These tests are going to be quite expensive to introduce, and I am not sure about the value they provide.
A recent run had this job taking 4m47s:
https://app.circleci.com/pipelines/github/ONEARMY/community-platform/978/workflows/52883866-0c64-4c64-aabd-85350205de25/jobs/4162

chrismclarke (approved):
I tried locally and working really nicely, great to have this to build on. 

I had one minor issue with node_modules failing install, but bumping the version of firebase-tools seems to fix; going through the release notes there's no breaking changes of note and has a few benefits too (remove log4js, add support for node_16 functions) so I assume that should all be fine.

Otherwise happy for whatever minor adjustments are needed on CI and feel free to merge.

chrismclarke:
As for install time, I'm guessing the quickest win will be using yarn workspaces focus install instead of the generic install called in the setup_repo script (not sure if there's a way to pass arg in circle_ci or if we'd just need to create multiple templates)

The issue of multiple jest is indeed confusing (particularly because react hides the version behind react-scripts so a few calls to `yarn why` required to figure out what version of types should be used or not), .

thisislawatts:
@chrismclarke Sorry there should have a Blocked or In Progress label on this PR. The `Cloud Function test` is not yet configured as part of the entire pipeline so there are no dependencies. It is also not gated by the linting step which means that we have this expensive 7m test running for every job on the pipeline.

chrismclarke:
Oh, whoops. Well if any consolation been really handy to have for the backend functions work I've been looking into. I'm done for the day now but if you want me to look into how to better optimise or review any other changes tomorrow just let me know

PR #1261, opened 2021-11-13
feat: introduce logging library

thisislawatts:
PR Checklist

PR Type

## Description

Before the addition of this tool all logs
within the primary client side application
are being done via console.log. This is
okay for local development but makes it
harder to investigate issues affecting other
people as we are not currently piping this data
to anywhere.

Pino is a performant log tool that will
allow us to configure additional destinations for
these logs. We could easily aggregate logs data to
a 3rd party tool based on the configuration of a specific
deployment.

For the time being this logger will be hardcoded to output
logs to browser console. This is to avoid any disruption to
expected dev behaviour.


src/components/FirebaseFileUploader/FirebaseFileUploader.tsx
```diff
@@ -81,7 +82,7 @@ export class FirebaseFileUploader extends React.Component<IProps, IState> {
 
   public deleteUploads = async (fileInfo?: any) => {
     // WiP
-    console.log('TODO - handle delete', fileInfo)
+    logger.warn('TODO - handle delete', fileInfo)
```
thisislawatts, line 85: This should is being flagged as a warn because this case has not been implemented. If/when we introduce aggregated logs we want to be able to easily see how often this is impacting our users.

chrismclarke (approved):
I've never used Pino before but used Winston a lot so looks similar, again makes a lot of sense to have (I think currently we have a messy prod workaround of rebinding `console.log = ()=>null`, so nice to be able to have different levels instead).

I know this is going to cause a lot of conflicts but probably better to get it merged in sooner and resolve conflicts on other PRs than wait around for a clear backlog. 

A few future ideas that might be nice to build on from this:
- query param to allow enabling console logs in prod environment (e.g. `?log_level=debug) 
- sentry or any other monitoring tool integration with logger here instead of core code
- Setting up a few more sensible log levels (e.g. debug, info, warning, error)
- General tidying/removal of unnecessary logs (I expect there's a lot of legacy ones)
- Possibly tightening lint rules to warn/error on console.log statements (`no-console`)

PR #1271, opened 2021-11-25
fix: remove unhelpful logging

thisislawatts:
PR Checklist

PR Type

## Description

The Sentry alert that is triggered as a result
of this log is not actionable nor does
anyone seem to concerned about. So we should
probably remove it ¯\_(ツ)_/¯

### DIFF
--- a/src/common/errors.ts
+++ b/src/common/errors.ts
@@ -1,12 +1,6 @@
 import * as Sentry from '@sentry/react'
 import { SENTRY_CONFIG } from '../config/config'
 
-export const logToSentry = {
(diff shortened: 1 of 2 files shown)


chrismclarke (approved):
Oh yeah, this is related to a legacy bug where sometimes a user's profile would get wiped accidentally because it thought there wasn't a profile when there actually was (I think something with querying for user email vs id after some breaking changes were brought in to how profiles were created). I haven't from anyone complaining about the issue in at least 6 months so assume whatever fixes worked and we don't need.

As for Sentry more generally.... nice to have the option I guess??(!)

PR #1274, opened 2021-11-27
fix: regression in background colour

thisislawatts:
PR Checklist

PR Type

## Description

A recent change moved the colour definition
out of the SVG into a CSS file using CSS masks. We missed
the usage of the SVG within the mobile navigation.

## Screenshots/Videos

Before:

The decorative SVG is black making the text illegible

After: 

The decorative SVG uses the CSS colour

### DIFF
--- a/src/components/ProfileModal/ProfileModal.tsx
+++ b/src/components/ProfileModal/ProfileModal.tsx
@@ -104,7 +104,7 @@ export class ProfileModal extends React.Component<IProps> {
               to={window.location.pathname}
               data-cy="menu-Logout"
             >
-              <Flex color="rgba(27,27,27,0.5)">Log out</Flex>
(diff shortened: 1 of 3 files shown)


src/components/ProfileModal/ProfileModal.tsx
```diff
@@ -104,7 +104,7 @@ export class ProfileModal extends React.Component<IProps> {
               to={window.location.pathname}
               data-cy="menu-Logout"
             >
-              <Flex color="rgba(27,27,27,0.5)">Log out</Flex>
```
thisislawatts, line 107: Removing the light grey colour here to improve the legibility by making the contrast between background and foreground much more significant.

davehakkens (approved):

PR #1289, opened 2021-12-04
chore: adds project kamp production target

thisislawatts:
PR Checklist

PR Type

## Description

Extends the CI pipeline so that it deploys to the Project Kamp production instance as well as community.preciousplastic.com each time we make a Release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -371,7 +371,7 @@ workflows:
             - 'test_e2e'
           <<: *filter_only_master
       - deploy:
-          name: deploy_dev
+          name: "Deploy: dev.onearmy.world"
           requires:
             - build_dev
           <<: *filter_only_master
@@ -381,7 +381,7 @@ workflows:
             - circle-ci-slack-context
             - community-platform-dev
       - deploy:
-          name: "Deploy: DEV Project Kamp"
+          name: "Deploy: dev.community.projectkamp.com"
           requires:
             - build_dev
           <<: *filter_only_master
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -371,7 +371,7 @@ workflows:
             - 'test_e2e'
           <<: *filter_only_master
       - deploy:
-          name: deploy_dev
+          name: "Deploy: dev.onearmy.world"
```
thisislawatts, line 374: Update name to explicitly state _where_ the deployed code will be visible. This does couple the CI pipeline to our deployed environments however as domains are updated rarely I feel that the clarity it provides makes it worthwhile.

davehakkens:
Not sure if this is a good moment for it or worth the complexity to change.
Just throwing it here:

The original dev site is running on [dev.onearmy.world.](https://dev.onearmy.world) 
As we developed One Army we changed our main domain name to [ onearmy.earth](https://www.onearmy.earth)

So the .world domain for the dev-site is a bit off from the rest. 
Two more logical options for the dev-site would be:
1. dev.community.preciousplastic.com (if this is the dev site of PP community)
2. dev.platform.onearmy.earth (if this is the main test site for all instances)

thisislawatts:
@davehakkens I totally agree and think that it would be worth adjusting these URLs at a later date. More immediately I think it is really helpful to decode what we mean when we say `dev_site`, because it is not obvious that dev.onearmy.world is the development version of what will become https://community.preciousplastic.com.

thisislawatts:
But I don't think that should be a blocker to merging this PR ;) Something for future versions of us to sort out.

davehakkens:
hehe noted!

davehakkens (approved):

PR #1249, opened 2021-10-17
feat: adds global one army footer

thisislawatts:
PR Checklist

- [ ] - Latest `master` branch merged
- [X] - PR title descriptive (can be used in release notes)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

Work in progress, this component should not be visible on all pages, as it doesn't work with:
* The Map
* The Academy content

Some work here needs to be done so that it correctly renders depending on the page URL. It does that at the moment only after a **full page load** , see how it will not be added to [the homepage](https://onearmy-next--pr1249-feat-adds-global-one-wjhffkfz.web.app/academy/intro.html) but if you load an [individual howto](https://onearmy-next--pr1249-feat-adds-global-one-wjhffkfz.web.app/how-to/recycle-plastic-waste-into-3d-printing-filament) it will be at the bottom. You may need to refresh the page to force the conditional render logic to trigger. 

I believe the higher level App wrapper introduced as part of https://github.com/ONEARMY/community-platform/pull/1242, may be helpful here. So consider this blocked for the time being.

PR #1293, opened 2021-12-04
chore: fetch version number from git sha

thisislawatts:
PR Checklist

PR Type

## Description

At the moment the version is taken from the `package.json` version field. This is manually updated during the release process. However we don't want to rely on such a manual process for our development environments, no deployment to a development environment needs to be mapped to a releaseable artefact.

Development environments are intended to represent the HEAD state of our default branch. So instead of using a value that looks like a [semantic version](https://semver.org/), we can load a SHA value which references a specific point in the project history.  

Hovering over the logo will give you something like this instead:

Screenshot 2021-12-04 at 15 16 12

**Note:** This will only affect the development environments and not the production instances. The production instances will continue to use the more familiar `1.6.x` format.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -266,9 +266,12 @@ jobs:
           command: |
             echo 'export REACT_APP_BRANCH=${CIRCLE_BRANCH}' >> $BASH_ENV
             echo 'export NODE_OPTIONS=--max-old-space-size=4096' >> $BASH_ENV
+            echo 'export REACT_APP_PROJECT_VERSION=$(git rev-parse HEAD)'
       - run:
           name: Check environment variables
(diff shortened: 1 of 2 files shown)


thisislawatts:
Ideally we could include a clickable link through to the GitHub commit that relates to that SHA but we don't really have a globally available location where we could add that.

It may be a good thing to add into a global footer, as proposed here:
https://github.com/ONEARMY/community-platform/pull/1249

chrismclarke (approved):
Sounds good to me, could also add a link to the devsite header (not opposed to having visible version number somewhere, I think there used to be a footer but not sure what happened to it so why we just have logo hover currently)

PR #1304, opened 2021-12-11
feat: conditional module loading

thisislawatts:
PR Checklist

PR Type

## Description

Introduces a mechanism for selectively loading modules based on environmental variables. This can be done with a comma separated list of modules defined on the `REACT_APP_SUPPORTED_MODULES` environment variable.

This check takes priority over any role based permission checks. 

The available modules are defined as:

```ts
export enum MODULE {
    CORE = 'core', // This is enabled on all installations
    HOWTO = 'howto',
    MAP = 'map',
    EVENTS = 'events',
    RESEARCH = 'research',
    ACADEMY = 'academy',
    USER = 'user',
}
```

### DIFF
--- a/src/modules/index.ts
+++ b/src/modules/index.ts
@@ -10,7 +10,6 @@ export enum MODULE {
 
 export function getSupportedModules(): MODULE[] {
     const envModules: string[] = process?.env?.REACT_APP_SUPPORTED_MODULES?.split(',').map(s => s.trim()) || []
-    console.log({ envModules });
     return [MODULE.CORE]
         .concat(
             Object.values(MODULE)
--- a/src/modules/index.ts
+++ b/src/modules/index.ts
@@ -1,0 +1,22 @@
+export enum MODULE {
+    CORE = 'core',
+    HOWTO = 'howto',
+    MAP = 'map',
+    EVENTS = 'events',
+    RESEARCH = 'research',
+    ACADEMY = 'academy',
+    USER = 'user',
+}
+
+export function getSupportedModules(): MODULE[] {
+    const envModules: string[] = process?.env?.REACT_APP_SUPPORTED_MODULES?.split(',').map(s => s.trim()) || []
+    console.log({ envModules });
+    return [MODULE.CORE]
+        .concat(
+            Object.values(MODULE)
+                .filter(module => envModules.includes(module)));
+}
+
+export function isModuleSupported(MODULE): boolean {
+    return getSupportedModules().includes(MODULE);
+}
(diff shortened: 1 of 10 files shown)


src/pages/PageList.tsx
```diff
@@ -31,7 +31,14 @@ const ForgotPasswordMessagePage = lazy(() =>
...
 const TermsPolicy = lazy(() => import('./policy/terms'))
 
+export function getAvailablePageList(supportedModules: MODULE[]): IPageMeta[] {
+  return COMMUNITY_PAGES.filter(pageItem =>
+    supportedModules.includes(pageItem.moduleName),
+  )
+}
+
 export interface IPageMeta {
+  moduleName: MODULE
```
thisislawatts, line 41: Every page should be associated with a module so that we can gate it based on whether it is available to the current installation.

src/pages/Research/index.tsx
```diff
@@ -1,9 +1,10 @@
 import { Route } from 'react-router'
+import { MODULE } from 'src/modules'
 import {
   ResearchStore,
   ResearchStoreContext,
 } from 'src/stores/Research/research.store'
-import { IPageMeta } from '../PageList'
+import type { IPageMeta } from '../PageList'
```
thisislawatts, line 7: Introduce `import type` convention to mark interfaces which are not required during runtime. This ensures the compiler can remove it entirely during the build process. 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

Perhaps in the future this is something we can look to include in our linting configuration. 
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md#enforces-consistent-usage-of-type-imports-consistent-type-imports

src/modules/index.ts
```diff
@@ -0,0 +1,21 @@
...
+    const envModules: string[] = process?.env?.REACT_APP_SUPPORTED_MODULES?.split(',').map(s => s.trim()) || []
+    return [MODULE.CORE]
+        .concat(
+            Object.values(MODULE)
+                .filter(module => envModules.includes(module)));
+}
+
+export function isModuleSupported(MODULE): boolean {
+    return getSupportedModules().includes(MODULE);
+}
```
thisislawatts, line 21: The new `src/modules` provides a new entry point for all things module orientated. In this instance to high level helper methods for working with the module entity. 

I am unsure whether we would ultimately refactor to move all modules under this directory, for example:

```
src/
    modules/
        Research/
        Howto/
```

davehakkens (approved):
lets see 🤞

src/modules/index.ts
chrismclarke: I really like the direction this is headed in, I'd definitely vote for eventually moving every module to within these module folders

PR #1305, opened 2021-12-12
feat: move configuration outside of core bundle

thisislawatts:
PR Checklist

PR Type

## Description

A full build of the application is needed for each target installation
because the credentials needed by Firebase, Algolia, Sentry etc are
loaded out of the environmental variables.

Builds require multiple minutes to run and this work is duplicative.
Recently we had to upgrade the CircleCI resources used to sidestep
performance issues around the build step.

Duplicate builds mean that the code which we deploy to our production
environment is not the same as the code under test.

**Note:** All environmental variables are publicly accessible in the current
implementation. It is an accepted limitation of this pattern:
https://create-react-app.dev/docs/adding-custom-environment-variables

This implementation adds the installation variables directly into the generated index.html with a light weight script
+ Variables will be instantly available to the application
+ Minimal refactor needed to load variables from window object
- Complicates build step, potentially moving more logic out of the create-react-app boilerplate
white_check_mark


.circleci/config.yml
```diff
@@ -352,12 +356,9 @@ workflows:
       #---------------------- Test ----------------------
       # Note - when calling test we also let the test script handle building as it injects random variables for seeding the DB
       - build:
-          name: build_test
+          name: Build Application
```
thisislawatts, line 359: Switching to more human friendly naming convention, explicitly stating that this is now the build step for the entire application.

.circleci/config.yml
```diff
@@ -278,13 +284,14 @@ jobs:
             echo REACT_APP_BRANCH=$REACT_APP_BRANCH NODE_OPTIONS=$NODE_OPTIONS
             echo $REACT_APP_PROJECT_VERSION
       - run:
-          command: << parameters.BUILD_ENV >> npm run build
+          command: yarn build
```
thisislawatts, line 287: We use `yarn`, let's use `yarn`!

davehakkens (approved):

PR #1306, opened 2021-12-12
feat: increase supported length of text

thisislawatts:
PR Checklist

PR Type

## Description

Increases the number of characters that can be provided for the initial description and individual steps of a Howto article. 

Originally raised as part of:
https://feedback.preciousplastic.com/suggestions/202757/longer-character-limit-for-howto-steps

### DIFF
--- a/src/pages/Howto/constants.ts
+++ b/src/pages/Howto/constants.ts
@@ -1,0 +1,1 @@
+export const HOWTO_MAX_LENGTH = 1000;
--- a/src/pages/Howto/constants.ts
+++ b/src/pages/Howto/constants.ts
@@ -1,1 +1,2 @@
 export const HOWTO_MAX_LENGTH = 1000;
+export const HOWTO_TITLE_MAX_LENGTH = 50;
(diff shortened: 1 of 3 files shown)


src/pages/Howto/constants.ts
```diff
@@ -0,0 +1 @@
+export const HOWTO_MAX_LENGTH = 1000;
```
thisislawatts, line 1: Introduces a top level entry point for containing information about the `Howto` content type.

tudi2d:
I like this - would love to keep it as a pattern for the other pages! Also: We use constants for the `max-length` of the title in the `HowTo` Page as well as the title & text of each `HowToStep`. Do you think it make sense to add those constants to the `constants.ts` in this PR as well? :)

thisislawatts:
Good catch @tudi2d. I have pushed the commit; https://github.com/ONEARMY/community-platform/pull/1306/commits/85d7237fdc70547bf45f3d9ac1f3d3fae7b72c6d which also standardises the title length between the root Howto article and each of its nested step.s

chrismclarke (approved):
All seems super sensible and nice to have. Agree moving out `constants` to own file makes a lot of sense and we can slowly add to it over time.

If I were to future-nitpick I'd probably recommend:
1) A single export to reduce amount of namespacing at the individual name level and more consistency across modules.
2)  Trying to start all constants with a clearer descriptive (e.g. change `MAX_LENGTH` to `CONTENT_MAX_LENGTH` (like you've done for title). I.E.

```
export const HOWTO_CONSTANTS = {
  CONTENT_MAX_LENGTH: 1000;
  TITLE_MAX_LENGTH: 50;
}
```
(or provide as a default export and allow the rest of module to consume under whatever namespacing seems most sensible)

But that's mostly just personal preference and what's here is already miles better than the current system so very happy to merge.

PR #1320, opened 2021-12-16
test: require Unit tests for e2e

thisislawatts:
PR Checklist

PR Type

## Description

The unit tests are very quick but reliable
if they fail we do not have a candidate which
can progress to our slower and more expensive
e2e test suite.

The main benefit of this change is to save spending
CircleCI credits running e2e tests.


chrismclarke (approved):
We have unit tests?? Haha, wow that's awesome, a very legacy TODO right there. Makes complete sense to be initial part of deployment pipeline before e2e, hopefully over time as we add to these we might find some of the e2e cases no longer really required and that can speed things up too.

PR #1337, opened 2021-12-20
fix: service worker cache headers

thisislawatts:
PR Checklist

PR Type

## Description

Over the last month we've shipped multiple
releases to production, at a far higher
cadence then recent months.

A side effect of this has been observing
multiple "odd" behaviour on the production
deployments which can not be recreated locally.

The primary candidate for causing these issues
is the service worker. The service worker is a
built in component of create react app (CRA).

Interestingly in the CRA documentation when discussing
servic workers they mention:
> they (service workers) can make debugging deployments more challenging.

So as a result I have been hunting out for more information around
other folks encountering this behaviour.

An interesting thread here around the combination of
service workers AND firebase:
https://github.com/facebook/create-react-app/issues/2440#issuecomment-305592301

Potentially related, is all of the _ChunkLoadError_ we are seeing from the production environment.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -40,6 +40,15 @@
             "value": "*"
           }
         ]
+      },
+      {
+        "source": "/service-worker.js",
+        "headers": [
+          {
+            "key": "Cache-Control",
+            "value": "no-cache"
+          }
+        ]
       }
     ]
   },


chrismclarke:
Yeah another painfully tricky area/issue. I think the core issue likely stems from how best to manage service worker lifecycles, lazy-loading and cache-busting in live environments.

As far as my understanding:

When we make a deployment all the code needed to run the version will be code-split and cache-busted, so instead of one giant js chunk we will have around 40 or so chunks with names including a hash of their content (e.g. `37.fa7c6d12.chunk.js`). The chunk will never change unless the content in it also changes.

When loading the platform the index page will determine which chunks are needed for initial render (or features required for the specific page the user is in). You'll see the script tags in the index file once built, e.g. 
```
<script src="/static/js/13.e1ba0722.chunk.js"></script>
<script src="/static/js/main.c59b6458.chunk.js"></script>
```

Ordinarily, this is all fine because the page will just request from the server the same way all scripts are imported, but issues may arise if a service worker has been registered and is intercepting requests between server and client. For the initial example, the service worker will cache all chunk files and index file, so that when the user returns to any page they have previously visited they will not have to re-download any of the JS (allowing the site to reasonably function offline and helping reduce overall bandwidth). 

When we create a new version with changes to core files the index chunks will change, and so the service worker will need to replace its cached index file with the new index file and serve content from there. This is handled by having two service workers running in parallel, `SW1` which has a cache of all the old site files and index, and `SW2` which will immediately cache the new index file and continue to cache other chunks as requested. As the page can only be controlled by one service worker at a time, generally the way things work out is:

1 - SW1 continues to run the site until SW2 indicates it's ready
2 - SW2 loads in the background and caches the new index
3 - SW2 sends a message to SW1 to let it know it can take over
4 - SW1 hands over control and disappears forever

A few friction points can occur, namely:
1->2 - If SW1 needs to download more chunks because user navigates to a page they have previously not used the chunk no longer exists on the server (might be the chunk error seen)
2->3 - If SW2 has invalid configuration it will never reach this stage and will require a bit of cache-clearing/window close and reload for SW1 to know that it needs to keep running and fetch the updated index.html itself (I think)
3->4 - This is manually handled from the codebase. Some sites keep this process simple by just immediately refreshing the page, or if using a static site waiting for the user to navigate, other SPAs try to be less obtrusive by popping up a toast saying 'the page has updates, reload' or similar, and some (including ours) tries to wait until the user is performing some sort of navigation event and using as the chance to handover control without need for a refresh.

Testing these things can be a bit tricky, the best option I've come up with so far is creating production build locally and then serving via a local webserver (e.g. `yarn build; npx serve build`)

chrismclarke:
So what I assume the included fix is trying to do is replace steps 1 and 2 by ensuring the browser always first checks for service worker updates before loading (although I'm not 100% sure). The service worker itself is actually quite big (around 100kb) so ideally we don't want to fetch this too often, although as I understand specifying no-cache just means a validation request will be sent which I think is basically just a HEAD request and super small so hopefully won't cause too much inconvenience

chrismclarke (approved):

PR #1344, opened 2021-12-23
feat: introduce customised description

thisislawatts:
PR Checklist

PR Type

## Description

Adds a custom description used by Search engines and social networks when sharing links to the site.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -52,6 +52,7 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>Precious Plastic Community</title>
+    <meta name="description" content="A series of tools for the Precious Plastic community to collaborate around the world. Connect, share and meet each other to tackle plastic waste.">
     <script id="CommunityPlatform">
       /**
        * The following object has environment specific
--- a/scripts/post-cra-build.ts
+++ b/scripts/post-cra-build.ts
@@ -63,6 +63,14 @@ if (platformTheme) {
     console.log(`Applying theme: ${platformTheme}`);
     console.log(`Copying src/assets/theme/${platformTheme}/public to build/`);
     fsExtra.copySync('../src/assets/images/themes/' + platformTheme + '/public', '../build');
+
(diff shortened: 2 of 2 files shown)


public/index.html
```diff
@@ -52,6 +52,7 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>Precious Plastic Community</title>
+    <meta name="description" content="A series of tools for the Precious Plastic community to collaborate around the world. Connect, share and meet each other to tackle plastic waste.">
     <script id="CommunityPlatform">
```
thisislawatts, line 56: Introduces missing attribute.

davehakkens (approved):

scripts/post-cra-build.ts
```diff
@@ -63,6 +63,14 @@ if (platformTheme) {
     console.log(`Applying theme: ${platformTheme}`);
     console.log(`Copying src/assets/theme/${platformTheme}/public to build/`);
     fsExtra.copySync('../src/assets/images/themes/' + platformTheme + '/public', '../build');
+
+    const siteDescription = platformTheme === 'precious-plastic'
```
chrismclarke, line 67: Might work better as a switch-statement with default if we ever decide to launch more versions (or just config file)

thisislawatts: @chrismclarke, totally agree. I was going to import the PlatformTheme file but unfortunately the tsconfig differences between the root and scripts directory do not play nicely.

chrismclarke: ahh... yes. tsconfig. The good thing about the scripts is in theory, because all the code is our own we should have a bit more freedom/flexibility on the module system. But I expect a handful of packages we use have strong preferences which could conflict, particularly if we want to keep compatibility with node 12 (thinking es-modules)...

Maybe one for the new year, see if we can just get things working node-16 and make the switch. Or next time we need to do any work on this I can probably find a workaround (seem to spend half my life doing that these days!)

PR #1348, opened 2021-12-23
test: increase timeout

thisislawatts:
PR Checklist

PR Type

## Description

This test has been flaking a little bit in Circle
recently. Looking at the resource usage the small
Docker resource is running at 100% CPU. We have
also been introducing _more_ unit tests into the
platform.

This is an experiment to see if this stabilises
the tests, but it may be we need to consider either:

* Increasing the resource for running tests
* Dig into jest configuration inherited from create
react app to identify optimisations.

### DIFF
--- a/src/pages/Research/research.routes.test.tsx
+++ b/src/pages/Research/research.routes.test.tsx
@@ -1,4 +1,4 @@
-import '@testing-library/jest-dom';
+import '@testing-library/jest-dom'
 import ResearchRoutes from './research.routes'
 import { render, waitFor, cleanup } from '@testing-library/react'
 import { MemoryRouter } from 'react-router'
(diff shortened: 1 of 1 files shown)


chrismclarke:
I'm not sure I fully understand why this would reduce flakiness (I think default timeout is 5s so not sure what reducing to 2s will do...)

Could it be worth trying either `--maxWorkers` or `--maxconcurrencynum` to reduce CPU load?
https://jestjs.io/docs/cli#--maxconcurrencynum

chrismclarke:
Oh, ignore me, I'm thinking of the test suite as a whole and not the waitFor method (can see that's 1s default, so yeah happy to try longer)

chrismclarke (approved):
Makes sense

PR #1356, opened 2021-12-28
feat: adds logflare drain

thisislawatts:
PR Checklist

PR Type

## Description

Adds a transport to the pino logger library to
forward your messages to Logflare.

This can be verified locally by adding a `.env` file
with the following values:

```
REACT_APP_LOGFLARE_KEY=<LOG_FLARE_API_KEY>
REACT_APP_LOGFLARE_SOURCE=<LOGFLARE_SOURCE_ID>
```

This is a trial with Logflare, we can easily switch
to another provider at a later date.

See attached screen grab for an example of how a message appears in [Logflare](https://logflare.app/guides/overview)
Screenshot 2021-12-28 at 21 02 47

ℹ️ Before merging this change ensure that environmental variables `REACT_APP_LOGFLARE_KEY` and `REACT_APP_LOGFLARE_SOURCE` have been added for all of the instances we wish to add logging to.

### DIFF
--- a/src/logger/index.ts
+++ b/src/logger/index.ts
@@ -1,6 +1,28 @@
 import Logger from 'pino'
+import { createPinoBrowserSend, createWriteStream } from 'pino-logflare'
 import { getConfigirationOption } from 'src/config/config';
 
 const logLevel = getConfigirationOption('REACT_APP_LOG_LEVEL', 'warn');
 
-export const logger = Logger({ browser: { asObject: false }, level: logLevel })
+let loggerInstance;
+
+if (getConfigirationOption('REACT_APP_LOGFLARE_KEY', '')) {
+    const logflareConfiguration = {
+        apiKey: getConfigirationOption('REACT_APP_LOGFLARE_KEY', ''),
+        sourceToken: getConfigirationOption('REACT_APP_LOGFLARE_SOURCE', '')
+    };
+
+    loggerInstance = Logger({
+        browser: {
+            transmit: {
+                send: createPinoBrowserSend(logflareConfiguration)
+            }
+        },
+        level: logLevel
+    }, createWriteStream(logflareConfiguration))
+} else {
+    loggerInstance = Logger({ browser: { asObject: false }, level: logLevel });
+}
+
+
+export const logger = loggerInstance;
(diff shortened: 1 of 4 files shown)


src/components/LocationSearch/LocationSearch.tsx
```diff
@@ -105,14 +105,18 @@ export class LocationSearch extends React.Component<IProps, IState> {
...
     })
     // need to manually trigger an input event as this is what algolia places uses to pick up change
     const event = new Event('input', { bubbles: true })
     this.placesInputRef.current.dispatchEvent(event)
 
+    logger.info('LocationSearch input changed', {
+      locationSearchVariant: this.props?.trackingCategory || 'LocationSearch',
+      value,
+    })
+
```
thisislawatts, line 119: Introduced a log sibling to the ReactGA.event. This is pretty much only to validate the approach, although I am curious to see if this helps us catch events which will not reach Google Analytics due to users with ad blockers.

chrismclarke: Nice to have a test case for sure

src/logger/index.ts
```diff
@@ -1,6 +1,28 @@
...
+import { createPinoBrowserSend, createWriteStream } from 'pino-logflare'
 import { getConfigirationOption } from 'src/config/config';
 
-const logLevel = getConfigirationOption('REACT_APP_LOG_LEVEL', 'warn');
+const logLevel = getConfigirationOption('REACT_APP_LOG_LEVEL', 'info');
 
-export const logger = Logger({ browser: { asObject: false }, level: logLevel })
+let loggerInstance;
+
+if (getConfigirationOption('REACT_APP_LOGFLARE_KEY', '')) {
```
chrismclarke, line 9: So as I understand we can still merge and then just update env later as required (the if statement will catch if no configuration provided?)... If yes I'd say good to go

chrismclarke (approved):
Nice to include, particularly interesting timing given that logflare has just been acquired by supabase who we plan to migrate our DB to (so mid-term should be a nice shared ecosystem) - and assume until then we can probably qualify for a free tier somewhere. 

I expect there should be a reasonable migration process that we'll just need to do as/when required, also good that they're committed to making it fully open source so that fits well with the overall model.

https://supabase.com/blog/2021/12/02/supabase-acquires-logflare#supabase--logs

PR #1361, opened 2021-12-29
feat: upgrade project to node16

thisislawatts:
PR Checklist

PR Type

## Description

Currently the project runs on Node 12, this is in maintenance mode until April 2022. 

Upgrade node version across the project to run with the latest version of Node 16, which will be supported until April 2024.


package.json
```diff
@@ -166,7 +166,7 @@
     "idb": "^6.0.0",
     "lint-staged": "^8.1.5",
     "mobx-react-devtools": "^6.0.3",
-    "node-sass": "^4.14.1",
+    "node-sass": "6",
```
thisislawatts, line 169: The latest version of `node-sass` is version 7, howevr we are tied to 6.x due to the current version of react-flag-select.

thisislawatts: This also may impact #1370 1370

chrismclarke (approved):
Wow, I'm shocked everything just works... But that's great!

Yeah really nice to be back on stable, and not have to trawl through legacy versions to figure out what the 'latest' version of node 12 is.

I can see one final reference to node 12 in the docs which I can quickly update but then I think we're good to go. 

As we don't impose a node engine in package.json I'm pretty sure it will still let users carry on using whatever (which I think is fine for now), but might be worth a quick mention in the next dev meeting in case anyone is unaware.

PR #1364, opened 2021-12-30
feat: introduce security rules

thisislawatts:
PR Checklist

PR Type

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?


.circleci/config.yml
```diff
@@ -376,16 +382,28 @@ workflows:
...
+      - test_security_rules:
+          name: 'Security rules validation'
+          filters:
+            branches:
+              ignore:
+                - production
       - test_unit:
           name: 'Unit tests'
+          requires:
+            - "Security rules validation"
```
thisislawatts, line 398: **Issue (blocking)**: This will be removed before merge. This is just a hack to prevent too many jobs being triggered on CircleCI

chrismclarke (commented):
Really great to have some progress on this one, and not just for the inbox full of security warnings. I like being able to include tests explicitly for them and handy to have the utility functions you've created.

A couple thoughts on the rules part:

1) Given that we already have a strongly typed list of db endpoints, would it make sense to try and generate these rules programmatically? The benefit would be avoiding have them accidentally be out of sync with platform (hopefully caught by tests), and making it easier to make changes. The downside would be harder to make more fine-grained changes?

2) A few of the collections have subcollections that might need explicit inclusion. I think currently we have `v3_users/{slug}/revisions` (automated tracking of user revisions, not currently exposed in frontend so no rules required) and `v3_howtos/{slug}/stats` which will require read access (unless covered by the default read block?)

3) I'm guessing similar create/update rules will be needed for most endpoints. I'm not sure if there's any way to create a rule group or equivalent (anyone can read, only signedIn + owner can update). Might again be a case for programmatic inclusion.

4) We might need to add an extra condition for admin overrides. I think there's a legacy code snippet to do this in the old `firestore.rules.WIP` file but not sure if still valid under v2 rule spec

chrismclarke:
But to (hopefully) keep this moving forward and get merged I would suggest we first address any changes needed in points 2,3,4 above and try get working in production before making things more complex with programmatic rules

thisislawatts:
Thanks for the feedback @chrismclarke I will pick this up during the February sprint. I was thinking about focusing on Security for my contributions next month.

chrismclarke:
Nice! I think it's possibly the oldest outstanding issue (and has 1 or 2 others linked) so very much a nice to have. As well as securing DB the same rule syntax can be used for storage objects which could be another future consideration.

The other option of course would be to reduce the number of explicit db calls from the frontend app, and instead introduce more firebase functions to do CRUD db and storage tasks, and add middleware to handle security concern - however rules is a faster win in my opinion and worth having even if we did migrate in the future

thisislawatts:
@chrismclarke I was thinking about moving towards more DB CRUD operations happening behind cloud functions but would that break the real-time behaviour of firebase? Or would retain that due to subscriptions it has to the underlying collections?

chrismclarke:
> I was thinking about moving towards more DB CRUD operations happening behind cloud functions but would that break the real-time behaviour of firebase? Or would retain that due to subscriptions it has to the underlying collections?

I'd be a fan of more crud. Subscriptions are mostly just a means to stream db reads and could still be kept (with write/update moved to functions), or even removed entirely with more functionality moved to page containers instead of stores to know when to check for data updates. 

Particularly it would be nice to add a layer of abstraction to support easier change of db platform in the future (or whenever we get round to prioritising)

PR #1393, opened 2022-01-18
WIP: Maps page Unit tests (Issue #1378)

kfern:
PR Checklist

- [ X] - Latest `master` branch merged
- [ X] - PR title descriptive (can be used in release notes)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ X] Developer experience (improves developer workflows for contributing to the project)

## Description

Unit tests is a "legacy todo".  This PR is a starting point for a new unit tested Maps page. 

## Git Issues

#1378 

---

## What happens next?

Thanks for the contribution! We try to make sure all PRs are reviewed ahead of a monthly dev call (first Monday of the month, open to all!).

If the PR is working as intended it'll be merged and included in the next platform release, if not changes will be requested and re-reviewed once updated.

If you need more immediate feedback you can try reaching out on slack in the `platform-dev` channel.

PR #1394, opened 2022-01-19
feat: add support for experimental module

chrismclarke:
PR Checklist

PR Type

## Description

Currently the process of adding a new module is somewhat tricky, as involves adding a main pagelist entry, updating supported modules enum, and figuring out how to include in local/test builds whilst omitting from production.

This PR proposes a single `/experimental` page route to be configured only to run in local/dev context, where any other modules can be included. Initial test case for potential changes to maps module proposed in #1393

This will allow new modules to be added in the same way we would later import into the main page list, but without modifying the core files until ready to do so.

## Review Notes
- I'm not sure the best way to handle only allowing the pages on dev/local mode. I'm assuming that all prod releases uses a hardcoded `REACT_APP_SUPPORTED_MODULES` variable, but if not then it might be accidentally included.

- It appears my prettier code formatted has picked up a few additional changes. I'm assuming these are correct and older versions just didn't run the updated pre-commit hooks, but if not then I can look into making local changes to try fix

## Git Issues

Closes #

## Screenshots/Videos

image

image

---

## What happens next?

### DIFF
--- a/src/modules/index.ts
+++ b/src/modules/index.ts
@@ -1,23 +1,30 @@
-import { getConfigirationOption } from "src/config/config";
+export * from './experimental'
+import { getConfigirationOption } from 'src/config/config'
 
 export enum MODULE {
-    CORE = 'core', // This is enabled on all installations
-    HOWTO = 'howto',
-    MAP = 'map',
-    EVENTS = 'events',
-    RESEARCH = 'research',
-    ACADEMY = 'academy',
-    USER = 'user',
+  CORE = 'core', // This is enabled on all installations
+  HOWTO = 'howto',
+  MAP = 'map',
+  EVENTS = 'events',
+  RESEARCH = 'research',
+  ACADEMY = 'academy',
+  USER = 'user',
+  EXPERIMENTAL = 'experimental',
 }
 
 export function getSupportedModules(): MODULE[] {
-    const envModules: string[] = getConfigirationOption('REACT_APP_SUPPORTED_MODULES', 'howto,map,events,research,academy,user').split(',').map(s => s.trim()) || []
-    return [MODULE.CORE]
-        .concat(
-            Object.values(MODULE)
-                .filter(module => envModules.includes(module)));
+  const envModules: string[] =
+    getConfigirationOption(
+      'REACT_APP_SUPPORTED_MODULES',
+      'howto,map,events,research,academy,user,experimental',
+    )
+      .split(',')
+      .map(s => s.trim()) || []
+  return [MODULE.CORE].concat(
+    Object.values(MODULE).filter(module => envModules.includes(module)),
+  )
 }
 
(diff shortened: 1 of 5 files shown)


src/modules/index.ts
```diff
@@ -1,23 +1,30 @@
...
 
 export enum MODULE {
-    CORE = 'core', // This is enabled on all installations
-    HOWTO = 'howto',
-    MAP = 'map',
-    EVENTS = 'events',
-    RESEARCH = 'research',
-    ACADEMY = 'academy',
-    USER = 'user',
+  CORE = 'core', // This is enabled on all installations
```
chrismclarke, line 5: local prettier code formatter seems to have replaced 4 spaces with 2 spaces (or tabs?). I'm not sure which should be correct, although hoping this is as it passed the pre-commit prettier hook (if not I can look into the hook)

src/modules/index.ts
```diff
@@ -1,23 +1,30 @@
...
-    const envModules: string[] = getConfigirationOption('REACT_APP_SUPPORTED_MODULES', 'howto,map,events,research,academy,user').split(',').map(s => s.trim()) || []
-    return [MODULE.CORE]
-        .concat(
-            Object.values(MODULE)
-                .filter(module => envModules.includes(module)));
+  const envModules: string[] =
+    getConfigirationOption(
+      'REACT_APP_SUPPORTED_MODULES',
+      'howto,map,events,research,academy,user,experimental',
+    )
```
chrismclarke, line 20: I've included in the default list as I want the page to appear locally, although is there a better way to do so?

thisislawatts:
I feel that some of the complexity you've described here around modules could be resolved through documentation. When I addded the loading mechanism it was more or less a one man PR frenzy which is not typically a good way for well documented features to be introduced 🙈  

I would prefer to see documentation covering an easy to reason with and configure system. This would be instead of introducing too much functionality intended only to be used on a local/dev environment. For certain tasks it is useful to have a local development environment that is as close as possible to production and managing that within configuration seems preferable to more complex application code.

chrismclarke:
> I feel that some of the complexity you've described here around modules could be resolved through documentation. When I addded the loading mechanism it was more or less a one man PR frenzy which is not typically a good way for well documented features to be introduced 🙈

Fair - although to your credit it didn't take too long to figure out what was going on to build from it.

> I would prefer to see documentation covering an easy to reason with and configure system. This would be instead of introducing too much functionality intended only to be used on a local/dev environment. For certain tasks it is useful to have a local development environment that is as close as possible to production and managing that within configuration seems preferable to more complex application code.

Drafted docs in #1410. I think still a few tweaks will be required to make it easier to enable/disable certain modules in different environments (notably pr preview/staging as not all reviewers/commenters will want to check out the repo and run on localhost), but can figure that one out when working on the map update testing integration

Issue #1378, opened 2022-01-09
[feature request] RFC: Unit tests


kfern:
**Is your feature request related to a problem? Please describe.**

Unit testing is a "legacy todo".

If we like to have it we must do a big refactor because the components are not isolated. They are using the store inside it.

For example this function 
https://github.com/ONEARMY/community-platform/blob/1535808b3951c88b2a86d06d3fb3c5ef14c2fdbd/src/pages/Maps/Content/Controls/GroupingFilterDesktop.tsx#L75
do calls to mapsStore.getPinsNumberByFilterType before render for each option. 
PS: This behavior may be related with #1257

**Describe the solution you'd like**
I'm going to use the Maps page as a proof of concept for a new unit tested version.

Because a line of code is better than 1000 words, I have prepared a temporary repository at https://github.com/kfern/wip-maps-page. I'm not solving every use cases but It can be a starting point.

I'm using checkboxs in this poc because @khanacademy/react-multi-select looks like it's dead. The last commit was on May 28, 2019 and does not support typescript. 

Feel free to close this issue if it's not a good idea.

thisislawatts:
Thanks for raising this @kfern 🙇🏻 I will take a look at this over the next weekend.

Very interested in improving the test coverage on this project!

thisislawatts:
Thanks for the patience @kfern, I got some time this morning to get the demo repo up and running. 🤩 

I think this is a really great step forward, and I appreciate you putting into code some of the issues I have experienced whilst working with the stores over the last few months. 

I am curious why you opted to spin up a standalone repo rather than working within a branch of the `community-platform` codebase? 

Given the performance issues some users are experiencing on the map page I do think we need to take another look at the current implementation. 

How can I help you to get this raised as a PR against the primary repo?

kfern:
@thisislawatts I'm glad you like it :-D

I created a new repository because we can go further with integration and e2e testing using GitHub actions, and I will need a "lab" to test.

I am using updated versions of leaflet related components and they are not compatible with the actual dependencies. Could you align both? I could downgrade if you prefer.

I also encounter problems passing lint

thisislawatts:
@kfern what do you think is the smallest step we could take to move the project towards the proposal you have outlined? 

I had missed the GitHub workflows. At the moment, I would prefer to keep Circle CI as the primary location for verifying the project. We should be able to make this change at a later date but I do not think it should coupled with this refactor.

kfern:
[...]

PR #1395, opened 2022-01-19
chore: remove empty function lint rule

chrismclarke:
PR Checklist

PR Type

## Description

As raised in #1378, the lint rule `@typescript-eslint/no-empty-function` can be a bit annoying when doing things like stubbing/mocking tests. It is possible to work around by adding a disable statement (already been done twice previously), however probably makes sense to just remove it as most use cases are intentional as opposed to true code-quality issues.

An alternative would be to modify to just provide warning, however as these will be thrown as errors during CI build (and for reason above) I think best to just remove.

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/src/stores/common/__mocks__/module.store.ts
+++ b/src/stores/common/__mocks__/module.store.ts
@@ -1,11 +1,9 @@
 export class ModuleStore {
-    db = {
-        collection: jest.fn().mockReturnThis(),
-        doc: jest.fn().mockReturnThis(),
-        set: jest.fn().mockReturnThis(),
-    };
+  db = {
+    collection: jest.fn().mockReturnThis(),
+    doc: jest.fn().mockReturnThis(),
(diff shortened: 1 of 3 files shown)


src/stores/common/__mocks__/module.store.ts
```diff
@@ -1,11 +1,9 @@
...
-        set: jest.fn().mockReturnThis(),
-    };
+  db = {
+    collection: jest.fn().mockReturnThis(),
+    doc: jest.fn().mockReturnThis(),
+    set: jest.fn().mockReturnThis(),
+  }
 
-    // eslint-disable-next-line @typescript-eslint/no-empty-function
-    constructor() {
```
chrismclarke, line 9: remove previous workaround (plus some auto-formatting)

thisislawatts (approved):

PR #1290, opened 2021-12-04
chore: add linting for commit messages

thisislawatts:
PR Checklist

- [x] - Latest `master` branch merged
- [x] - PR title descriptive (can be used in release notes)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [X] Developer experience (targets the way that contributing to the project works)

## Description

As part of the `commit-msg` hook validate the stucture of the commit message to ensure that it is well formed. This change will allow us to parse the commit messages programmatically to generate release notes. 

More information:
* https://github.com/conventional-changelog/commitlint/#what-is-commitlint

In addition:

* Updates husky to the latest version 1.x => 7.x
* Moves [lintstaged](https://www.npmjs.com/package/lint-staged) configuration to dedicated file. 

This has no impact on the user facing software and only affects the developer workflow. 

[Why Use Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.2/#why-use-conventional-commits)

- Automatically generating CHANGELOGs.
- Automatically determining a semantic version bump (based on the types of commits landed).
- Communicating the nature of changes to teammates, the public, and other stakeholders.
- Triggering build and publish processes.
- Making it easier for people to contribute to your projects, by allowing them to explore a more structured commit history.

PR #1397, opened 2022-01-20
Chore/extend linting of commits

chrismclarke:
PR Checklist

PR Type

## Description

PR to fix issues with proposed PR #1290, and extend to include cli and custom configs. Additional changes:

- add interactive prompt option. I originally tried with `@commitlint/prompt-cli` but didn't like it, so instead added `@commitlint/cz-commitlint` which works with the `commitizen` package

- convert js config file to ts (because why not)

-  add custom formatter to give more flexibility on output messages shown

## Git Issues

Closes #

## Screenshots/Videos
_Formatter output 1_
image

_Formatter output 2_
image

_Default output format_
image

_Interactive output_
image

---

## What happens next?


commitlint.config.ts
```diff
@@ -0,0 +1,124 @@
...
+      max: 'upper %d chars',
+      min: '%d chars at least',
+      emptyWarning: 'can not be empty',
+      upperLimitWarning: 'over limit',
+      lowerLimitWarning: 'below limit',
+    },
+
+    questions: {
+      type: {
+        description: "Select the type of change that you're committing:",
```
chrismclarke, line 26: I don't think these need to be made explicit (copy/paste from https://commitlint.js.org/#/reference-prompt?id=prompt),
So more for reference if we want to make changes to interactive prompts

thisislawatts:
I think that the Circle CI step may need updating to align with the changes you've made in this branch? It's actually a helpful demonstration of another facet of this work. 

Git hooks are entirely optional for local developers and introducing a fail fast `Lint` job helps us to provide that feedback quickly to contributors who have opted out of commit hooks on their local machine. It also protects our CI from running jobs for PRs fail to meet the commit style.

.husky/common.sh
```diff
@@ -0,0 +1,12 @@
...
+# https://github.com/commitizen/cz-cli/issues/627
+
+command_exists () {
+  command -v "$1" >/dev/null 2>&1
+}
+
+# Workaround for Windows 10, Git Bash and Yarn
+if command_exists winpty && test -t 1; then
+  exec < /dev/tty
+fi
```
thisislawatts, line 12: Great to catch these Windows requirements 🙏🏻

thisislawatts:
@chrismclarke I also pushed a commit into this branch to verify that the CI job fails:
https://app.circleci.com/pipelines/github/ONEARMY/community-platform/1181/workflows/55bf62f6-e5c8-4a9d-a10d-be657d3a6e9c/jobs/5393

thisislawatts (approved):

chrismclarke:
Looks good to me - I'm guessing that if someone tries to make a pr that contains commits that don't meet style we can just squash-merge to clean up.

I removed `pinst` as realised that only applies if publishing package to npm. I kept finding conflicting info on whether yarn postinstall hooks actually run correctly, and the verdict I reached is yes they do (but only if package.json actually changed, and also any output gets swallowed up. testing via `"postinstall": "husky install > husky.txt",` gave correct output

chrismclarke:
PR failing and node_module install, possibly due to legacy dependency issue (see #1409). I'm not 100% sure if it actually should be installing the full dependencies to run the check and if so whether there's a way to ignore certain errors...

PR #1404, opened 2022-01-25
chore: BREAKING CHANGE remove dev api

chrismclarke:
PR Checklist

PR Type

## Description

The dev API is a set of hacky/temporary functions used for working with local emulator. As they are not intended for use in production commenting out until we have better means of separating. 

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/functions/src/dev/index.ts
+++ b/functions/src/dev/index.ts
@@ -21,9 +21,9 @@ app.use(function(req, res, next) {
   else res.status(403).send(`Dev api methods can only be accessed on localhost:4002. Host: [${host}]`)
 })
 
-// app.get('/', (req, res) => res.status(200).send('Dev Api Working'))
+app.get('/', (req, res) => res.status(200).send('Dev Api Working'))
 
-// app.post('/seed-clean', (req, res) => cleanSeedData(req, res))
+app.post('/seed-clean', (req, res) => cleanSeedData(req, res))
 
 export = functions.https.onRequest(app as any)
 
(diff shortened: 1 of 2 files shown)


thisislawatts:
**issue (blocking)**: as an alternative to commenting out the code, how about we only conditional export this function from our functions/index.ts.

I found this promising answer on Stack Overflow referencing an emulator specific environmental variable but I have not as yet been able to find any documentation supporting it's existence. 

https://stackoverflow.com/a/65254135

The other benefit of this approach is that it should also ignore it as part of our build/deploy process. 

Using the commented code approach will be a little harder to work with as it requires a developer to convert all of the correct comments and then pollutes the git stage, meaning that this could be inadvertently reverted as part of some other work.

chrismclarke:
> **issue (blocking)**: as an alternative to commenting out the code, how about we only conditional export this function from our functions/index.ts.

That's a great idea, hadn't even crossed my mind that was possible. 

I've updated the export code to conditionally export, note this will cause a **BREAKING CHANGE** as the existing function will need to be manually deleted as firebase does not like removing functions without user confirmation.

I want to check that the conditional export actually works, so would suggest merging into master as-is, hoping for a failure in the staging deploy, handling manual removal and redeploying and then updating other projects as required. 

Let me know if that makes sense @thisislawatts

thisislawatts:
Sounds great, let's do it @chrismclarke

chrismclarke:
For ref, does appear to be working locally at least. Toggling the condition adds/removes the API endpoint from the emulator
image

thisislawatts (approved):

PR #1429, opened 2022-01-30
chore: remove node-linker

thisislawatts:
PR Checklist

PR Type

## Description

`nodeLinker=node-modules` sets the yarn resolution to use the old Yarn 1 way of doing things. The documentation does recommend this as a workaround required for specific packages, such as `react-native`.  We should better align with the Yarn Way(TM)

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,7 +1,5 @@
 # nmHoistingLimits: workspaces
 
-nodeLinker: node-modules
-
 plugins:
   - path: .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
     spec: "@yarnpkg/plugin-workspace-tools"


chrismclarke:
Definitely node-linker should go, I would hope/expect this is being overridden by the hoisting limits anyways. That being said, it might be good to just define the hoisting limits in the top-level `.yarnrc.yml` instead of individual package jsons. 

When yarn 2 first came out I was more keen on keeping installs lighter and leaving yarn to optimise how it saw fit, although at this point I'm probably more tempted to just have all workspaces install their own dependencies and have a bit of file duplication if required. There's constant conflicts between react, webpack etc. across the different toolings (e.g. docusaurus, frontend react app, storybook) I think it's just easier that way. 

I also remember testing support for pnp a long time ago and nothing was working, but times might have changed too. The other option of course is zero-installs where compressed versions of all modules are included in git, although I have a feeling our node_modules might be a bit heavy to be worth it (current cache sits at around 230MB, which I think is still within git limits but would make for very slow checkout).

chrismclarke:
Scratch that previous comment. I just realised that nodeLinker is what forces use of node_modules instead of pnp. PnP has issues with dependencies that aren't directly included (i.e. everything masked by create-react-app), so would likely require eject to use

I tried on a fresh cra repo and came across this problem: `https://github.com/facebook/create-react-app/issues/11793`

On the plus side, looks set to be fixed with react 5.0.1

chrismclarke:
_30 minutes later_

My brief conclusion is that if we want to remove node-linker we will need to eject react-scripts (just so many issues), and ideally get most other packages up to a reasonably recent version first also. Particularly it may be worth getting react-scripts updated from 4->5, as that way when we eject we will have a slightly nicer webpack 5 workspace (instead of webpack 4) and various babel upgrades likely useful for working with our updated typescript.

From there we can plot a way forward to drop nodeLinker and reconsider any other changes to overall build system

chrismclarke:
I'm going to drop my work for now, but if it's something you find yourself interested in taking up at any point feel free

thisislawatts:
Thanks for the extra set of eyes on this. To me this underlines the value in ejecting from CRA, the time between v4 and v5 releases has led to so many stale dependencies it is hard to engage with. 

I also think that with our use of yarn, workspaces and Typescript we have deviated from the "standard" user scenario

chrismclarke:
> Thanks for the extra set of eyes on this. To me this underlines the value in ejecting from CRA, the time between v4 and v5 releases has led to so many stale dependencies it is hard to engage with.
> 
> I also think that with our use of yarn, workspaces and Typescript we have deviated from the "standard" user scenario

Definitely agree, I ended up doing a quick test run of upgrade and eject and most things went smooth (few additional upgrades/migrations required which I'll start making a couple other smaller PRs for). I think we can make some quick progress focusing on a few core updates and then ejecting, and after that have a bit more flexibility for looking into existing webpack changes or alternative build systems.

At the same time as researching this I came across various bits looking at pnp - it seems to have been a low-priority issue for most repos to add support (e.g. firebase, mobx, react) and so I would consider for now that to be non-standard and probably worth sticking with node-modules until we have everything else working a bit nicer.

PR #1439, opened 2022-02-01
chore(deps): fix typescript configuration

thisislawatts:
PR Checklist

PR Type


chrismclarke (approved):
So if I understand things correctly, this PR actually removes the dependence on precompiling frontend type definitions by doing 2 things:

1. Change imports that reference the compiled type folder (`one-army-community-platform/lib/models`) for `../../src/models`

The knock on of this is that now the functions webpack will need to know how to compile the types for the referenced model files (and by association any other frontend platform code caught up in the imports... so a fair chunk of the platform). This would usually be a bit problematic, however you realised we can basically achieve this by

2. Adding additional `include` code to the functions tsconfig that references the additional type-definition files that are required to build the frontend.

I think this is a really great solution, and actually should be able to go a couple steps further. Namely:

A) We should be able to replace `../../src` for just `onearmy-community-platform/src/models`, as both point to the same files thanks to the yarn workspace resolution (and might look a bit tidier)

B) We can (finally) remove the janky project references used to link the tools together (so everything from `paths` in the functions tsconfig can go.

C) We can also remove all the functions code that calls for precompiling (as it now just does it on demand), the build script can just be `yarn build`. 

Of course we can hold off making these changes until what is here is tested as working, but overall I'm looks great to me! The best test will be to see if it correctly deploys the functions code when merged to master

thisislawatts:
@chrismclarke, reverting the `functions/src/models.ts` to reference `one-army-community-platform` instead of a relative path did not work :( 

However local tests show that your suggested improvements, **B** and **C** can be included. 

At this point I am bias towards unblocking our release process rather than fixing this behaviour.

chrismclarke:
That's still awesome news, some good steps to tidying some of the messiest parts of the codebase.

Now let's break everything again upgrading to CRA5 and then ejecting

PR #1446, opened 2022-02-02
Feat/db aggregations

chrismclarke:
PR Checklist

PR Type

## Description

- Add a new `aggregations` table to the database
- Add methods to automatically seed and ongoingly update a user badges aggregation (for use in #1205)

## Git Issues

Closes #

## Screenshots/Videos
Demo running in emulator - initial seed on first update and then ongoing changes

https://user-images.githubusercontent.com/10515065/152216805-13799524-5612-40e2-992a-243afa3e7204.mp4

---

## What happens next?


functions/src/Utils/data.utils.ts
```diff
@@ -3,10 +3,23 @@
...
+    chunkedArrays.push(chunk)
+  }
+  return chunkedArrays
+}
+
+/**
+ * Compare two json objects nested one level deep for equality (key order not important)
+ * Adapted from https://stackoverflow.com/a/52323412/5693245
+ * */
+export function objectsAreEqualShallow(obj1: any = {}, obj2: any = {}) {
```
thisislawatts, line 18: **question:** Can we introduce a unit test to validate the behaviour of this method?

davehakkens:
Good to see this coming 🔥🔥

chrismclarke:
@danitrod - I think this should be able to support what you are trying to do with the verified everywhere feature. This PR will add a single doc that can be queried by any of the stores and used as a single list of all users with badges. Let me know if it makes sense to you

chrismclarke:
> Good to see this coming 🔥🔥

Haha, only been, what, 3, 4 months for less than 100 lines of code😳
Here's hoping at least it works!

functions/src/Utils/data.utils.ts
chrismclarke: I think unit tests are still an outstanding PR? At least I don't see anything obvious I can run/adapt right now so perhaps a future TODO? (I started reviewing the PR last week but got stuck, so planning to revisit this week)

chrismclarke: The obvious failure is passing any falsy non-objects (e.g. `"",null,undefined` etc.) will assert as equal, but I felt easier to include the default here instead of relying on code to ensure all objects exist before running. I.e. if before and after snapshots are both undefined makes sense not assert as equal, and if one is undefined but the other nested json makes sense to identify as different (instead of just breaking)

chrismclarke:
@thisislawatts - The backend test code is now merged so I've added a couple quick unit tests for the data util functions.

I don't really know what's best when it comes to writing unit tests, I've tried to just highlight the main pass/fail cases but if you have any suggestions how to better organise just let me know and I can edit. Otherwise hopefully good to merge

chrismclarke:
I'm going to temporarily close this as I realised a few extra changes that would be useful to help with aggregating howto useful votes at the same time. Will re-open when ready

davehakkens:
haha so close!

chrismclarke:
image

PR #1204, opened 2021-08-08
Add useful to howto cards / add howto stats doc

danitrod:
PR Checklist

- [X] - Latest `master` branch merged
- [X] - PR title descriptive (can be used in release notes)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Description

As discussed on last meeting, changing the way howto stats (useful count) are currently stored to enable fetching all the stats to show when listing all howtos. The approach I chose was to add a `stats` doc, which is a map with the key being howto ID's and the value being their stats, inside the `v3_howto` collection. This way, even with useful counts updating quickly everywhere, we only need to refetch one document from the database to have the latest updated stats.

**Important notes**

1. As this PR introduces a new format for storing howto stats, we will need to migrate existing useful counts to the new format in order to not lose our data, so some additional work is needed (maybe creating a one-time running migration function).
2. This PR changes how the back-end updates the useful count, so preview probably won't work until the back-end is updated (which would be a breaking change so not sure how/when to do that).
3. Also, since this includes a back-end change, wasn't able to test locally yet (#1203), so for now I was testing manipulating the dev Firestore from console.

## Git Issues

Closes #1174

## Screenshots/Videos

Screen Shot 2021-08-07 at 20 57 52

Issue #1198, opened 2021-07-29
Add verified icon to map-pin


davehakkens:
### Description

To show and promote our verified workspaces more we add a verified icon on the map. The verified icon can be applied to all types of profiles. `Machine shop, community point, sheetpress workspace, extrusion workspace etc.`
If the profile is verified -> show blue checkmark
If the profile is not verified -> show no blue checkmark

### Assets

The Verified user data is stored in the V3_users>username>badges>verified
The icon can be found [here](https://github.com/ONEARMY/community-platform/blob/master/src/assets/icons/icon-verified-badge.svg)

### Page related

Will be used in :https://community.preciousplastic.com/map

- Related page name (#1197)

### Mockup

![afbeelding](https://user-images.githubusercontent.com/13672737/127539262-0696c7c1-e77c-4ca7-83a5-1e258da602a9.png)

chrismclarke:
See my comments in https://github.com/ONEARMY/community-platform/issues/1163
Basically this is a little bit tricker given that the badge info is stored in a different place to the map info, but once we sort that out should be quick to implement. 

If anyone wanted to start just testing getting the size and positioning on the frontend looking right (e.g. assume all pins verified for first test) then definitely welcome to take the issue on and make a start while we sort out the backend on the side.

TinkeringAlways:
Hi, taking this on as a first issue to help out with. Starting to work out the frontend like you suggested!

davehakkens:
cool! Whats your mail? then I'll invite you to our slack

TinkeringAlways:
alwaystinkering@outlook.com

TinkeringAlways:
Can I get access to the dev database?

davehakkens:
Just double checking, you got access to the dev database right @TinkeringAlways ?

TinkeringAlways:
Not yet, @chrismclarke could you add me?

davehakkens:
I've added you @TinkeringAlways. Hope thats ok @chrismclarke.
You are up for taking on the issue, should I assign you?

TinkeringAlways:
Yes please

TinkeringAlways:
@davehakkens could you also add "andrew@tinkerlabs.dev" for the Firebase dev access. The @outlook.com is a personal email. Thanks

TinkeringAlways:
That worked

thisislawatts:
[...]

Issue #1197, opened 2021-07-29
Add verified filter to the map


davehakkens:
## Component infos
We need a filter on the map to filter our verified workspace. It would work the same as the existing dropdown filter`(workspace, extrusion, sheetpress, community point etc)` 
- When selected it will show verified workspaces. 

### Assets
The Verified user data is stored in the V3_users>username>badges>verified
The icon can be found [here](https://github.com/ONEARMY/community-platform/blob/master/src/assets/icons/icon-verified-badge.svg) And it uses our aggregation
### Page related

Will be used in :https://community.preciousplastic.com/map

### Mockup
![map-dropdown-verified-icon](https://user-images.githubusercontent.com/13672737/127496318-6df09f65-a6e7-467c-bf24-8d279886707b.jpg)

chrismclarke:
See my comments #1198 and #1163
If anyone wants to get started on the frontend then should be ready to go on that, but will need a few small updates on the backend to get the filtering logic working correctly.

davehakkens:
Sorry, missed this comment @elvezpablo. Are you still good to finish it? Then i'll assign

thisislawatts:
:wave: Hey @elvezpablo, I wanted to check in on this issue and see if you've had a chance to work on it?

We try to make sure any PRs are reviewed ahead of a monthly dev call (which is the first Monday of each month), so even if you've only got an early stage draft PR it could be a good starting point for discussion 😁

Thanks again for offering to help with this issue ✨

davehakkens:
> I've started working on this issue and have been able to add the verified filter icon and label to the map. I need to run the integration tests and ensure the functionality is working as expected (minus the back end changes.). Could I get this ticket assigned to me? Thanks!

Hi @elvezpablo. FInal check in. let me know within 2 days if you want to finish this issue or share the code you made.
If we don't here anything no problem but then we will forward it to someone else.

kfern:
@davehakkens Can I get to work on this?

davehakkens:
> @davehakkens Can I get to work on this?

Great. yes! :)

kfern:
We need to make big changes to the map page components to close this issue.

Can we use an approach like [RFC: Unit Tests #1378](https://github.com/ONEARMY/community-platform/issues/1378)?

I can do it in the maps page component without any breaking changes.

@chrismclarke @thisislawatts What do you think?

chrismclarke:
[...]

Issue #1174, opened 2021-06-12
Add useful to howto


davehakkens:
### Description

Show the amount of useful for each how-to one the overview page. 
Will also be used with dropdown filter from #1172 

### Page related

Will be used in :
- [How to Overview](https://community.preciousplastic.com/how-to)

### Mockup

![useful](https://user-images.githubusercontent.com/13672737/121788744-9a255380-cbc7-11eb-92c6-098bd7453b74.jpg)

chrismclarke:
So this would actually end up being harder than it looks, due to the way the database is structured and read-optimisations work. Bascially the total votes for a howto is stored inside a sub-collection document of the howto, and by default only read when a user opens the howto. So the naive way to get all totals would be to read all the individual documents every time the page loads, which is a really expensive operation to do (assuming 20k monthly users and 100 or so howtos, this feature alone would probably cost around 2 euro per month in billing costs).

So an alternative method would probably rely on creating a single `stats` or possibly `data_feed` document or similar somewhere, where all the totals are collated and read as a single document (and updated via a cloud function), costing more like 2c per month. Or we try to fast-track the process of moving to a different database setup...

davehakkens:
ah I see the problem. Bummer..
Hope we can find a good way 🤞

danitrod:
@chrismclarke question, couldn't we save the useful stats as a prop in the actual howto docs, so they are loaded together with the other info from howtos?

chrismclarke:
I think we covered this on the call, but just to keep a written record. 

We decided that keeping the info in the individual doc itself would result in too many read updates on the main howto list page (we would need to retrieve all howtos every time the page loads - currently we avoid this by reading once, caching, and only reading again if updated). Instead we can just keep a single document with the stats for all howtos that can be read each time the page is loaded.

PR #1450, opened 2022-02-03
Feat: db aggregations alt

chrismclarke:
PR Checklist

PR Type

## Description
- Add general methods to seed and update aggregation values from collection docs (e.g. list of all users with badges)
- Add aggregation to collate all user badges from user profiles in a single `user_badges` aggregation doc
- Add aggregation all user voted howtos in a single `user_votedUsefulHowtos` aggregation doc
- Add additional data utils and unit tests to calculate deep diffs between objects

## Additional Notes

Alternative to #1446, addressing a few specific issues:

- Realised that more generic methods required if wanting to provide additional aggregations (existing votedUseful count needs refactor and a few other potential fields to aggregate). 
- The previous implementation would run aggregators on every change (not just changes to targetted field), and depending on individual aggregation-level code to check if the change actually had any knock-ons for the aggregation
- Only shallow changes were supported, but useful counts require deep diff so have added
- Previous implementation required individual subscribers for each aggregator, so would mean multiple function triggers to every user change event, which could get messy and have a lot of code duplication (also harder to delete functions)

Also, we already have functions that aggregate useful counts (within howto stats items). Ideally these methods will replace, however recommend leaving both in place for now and possibly link in #1204 which will require the aggregation to display on the howto list page.

Once working would also be good to add some documentation (but this PR is already a bit hefty and blocking on other issues so suggest as another future todo)

## Git Issues

Closes #

## Screenshots/Videos

Demo of user badge aggregation. Handles creating initial seed when first badge update received, and ongoing keeps list of all relevant badge info

https://user-images.githubusercontent.com/10515065/152414006-eca201e1-1128-4995-be5f-b7f4adb5dcb9.mp4

Demo of howto vote aggregation. Handles initial seed when any vote changed, and then aggregating as a counter by howto id (different to badges which aggregate values by user id)

https://user-images.githubusercontent.com/10515065/152415201-3d0f250a-a2d4-4ef0-a030-b1cda895d5e3.mp4

## What happens next?


functions/src/aggregations/user.aggregations.ts
```diff
@@ -0,0 +1,69 @@
+import * as functions from 'firebase-functions'
+import * as admin from 'firebase-admin'
+import { compareObjectDiffs } from '../Utils/data.utils'
+import { DB_ENDPOINTS, IUserDB } from '../models'
+import { handleDBAggregations, IAggregation } from './common.aggregations'
+
+interface IUserAggregation extends IAggregation {
+  field: keyof IUserDB
+}
+
```
chrismclarke, line 10: This is the current proposed syntax. For any aggregator that simply wants to copy value to top-level aggregation all that is required is `field` and `aggregationField`. For more complex aggregations overrides are exposed for core methods

functions/src/aggregations/user.aggregations.ts
```diff
@@ -0,0 +1,69 @@
...
+        const counterChange = change.after === true ? 1 : -1
+        await aggregator.aggregationRef.update({
+          [howto_id]: admin.firestore.FieldValue.increment(counterChange),
+        })
+      }
+    },
+  },
+]
+
+/** Watch changes to all user docs and apply aggregations */
```
chrismclarke, line 45: A single function will listen for any changes to users and handle all listed aggregations

functions/src/aggregations/common.aggregations.ts
```diff
@@ -0,0 +1,117 @@
...
+  //   aggregationType: 'value' | 'count'
+  handleAggregationUpdate?: (handler: AggregationHandler) => Promise<any>
+  handleAggregationSeed?: (handler: AggregationHandler) => Promise<any>
+}
+
+/**
+ * Takes triggered database change and determines which (if any) aggregations
+ * to update, and runs via an aggregation update handler
+ */
+export async function handleDBAggregations(
```
chrismclarke, line 26: This is the main exposed method that can be called by any observer to handle aggregation changes (e.g. changes to user, howto or event collections). It handles checking for specific field-level changes and triggering aggregation as required

functions/src/aggregations/common.aggregations.ts
```diff
@@ -0,0 +1,117 @@
...
+    for (const aggregation of aggregations) {
+      if (aggregation.field === fieldname) {
+        const update = changedFields[fieldname]
+        await new AggregationHandler(aggregation, dbChange, update).run()
+      }
+    }
+  }
+}
+
+export class AggregationHandler {
```
chrismclarke, line 46: The code that handles the actual aggregation functions is contained in a class (for easier variable sharing and exposing override options).

functions/src/Utils/data.utils.ts
```diff
@@ -3,10 +3,74 @@
...
-}
\ No newline at end of file
+  })
+  return diffs
+}
+
+/**
+ * Compare two values (including arrays and json objects), and
+ * assert if deeply equal (recursively for arrays and json objects)
+ */
```
chrismclarke, line 44: We could import a package like lodash that has more advanced equality operators, although as we're only dealing with primative data types that come from the DB I'm hoping my own implementation should cover most of our use cases.

functions/src/aggregations/common.aggregations.ts
```diff
@@ -0,0 +1,117 @@
...
+    return this.aggregationRef.update({
+      [doc_id]: valid ? value : admin.firestore.FieldValue.delete(),
+    })
+  }
+
+  /**
+   * Query all documents under aggregation watch, retrieve those with entry for
+   * watched field and batch update all aggregation entries
+   */
+  private async seed() {
```
chrismclarke, line 97: the seed function will trigger anytime the target aggregation field does not exist. This provides us with an initial migration the first time a document is populated, and also a means to re-calculate/alter logic by updating function and then just deleting the database doc (or targeting a different field name)

functions/src/aggregations/common.aggregations.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+export async function handleDBAggregations(
+  dbChange: functions.Change<functions.firestore.QueryDocumentSnapshot>,
+  aggregations: IAggregation[],
+) {
+  const { before, after } = dbChange
+  const changedFields = compareObjectDiffs(before.data(), after.data())
+  if (Object.keys(changedFields).length === 0) {
+    // changed detected by firestore but not locally
+    console.warn('change missed', before.data(), after.data())
```
chrismclarke, line 34: not sure where/how to pipe these logs... possibly could just throw as error as no further processing will be done anyway

src/stores/databaseV2/clients/dexie.tsx
```diff
@@ -9,7 +9,7 @@ import { DB_ENDPOINTS } from '../endpoints'
  * or busting cache on db. This is used as the Dexie version number, see:
  * https://dexie.org/docs/Tutorial/Design#database-versioning
  */
-const DB_CACHE_NUMBER = 20210130
+const DB_CACHE_NUMBER = 20220202
```
chrismclarke, line 12: I don't think it's strictly required to update the dexie cache number as adding a new table for aggregations isn't a breaking change, but also no harm in doing so

functions/src/Utils/data.utils.ts
```diff
@@ -3,10 +3,74 @@
  * length provided by @param chunksize
```
thisislawatts, line 3: **question (non-blocking):** What is the value of adding the `.utils.ts` suffix to a file nested within the Utils directory?

functions/src/Utils/data.utils.ts
thisislawatts: Any reason **not** to use lodash here? We're in a backend context so importing a 3rd party lib does not introduce any performance penalties?

functions/src/aggregations/user.aggregations.ts
thisislawatts: **question:** Given this is the primary entry point for the functionality can we add test coverage at this point?

functions/src/aggregations/user.aggregations.ts
```diff
@@ -0,0 +1,69 @@
...
+exports.default = functions.firestore
+  .document(`${DB_ENDPOINTS.users}/{id}`)
+  .onUpdate(change => {
+    return handleDBAggregations(change, userAggregations)
```
thisislawatts, line 49: **praise:** ✨ This is a great, a simple entry point hooking into the onUpdate method from Firestore.

functions/src/aggregations/common.aggregations.ts
thisislawatts: **question:** Given this is the primary entry point for the functionality can we add test coverage at this point?

functions/src/aggregations/common.aggregations.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  constructor(
+    public aggregation: IAggregation,
+    public dbChange: IDBChange,
+    public valueChange: { before: any; after: any },
+  ) {
+    this.collectionRef = dbChange.after.ref.parent
+    // apply function overrides
+    if (aggregation.handleAggregationSeed) {
+      this.seed = () => aggregation.handleAggregationSeed(this)
```
thisislawatts, line 60: **question(blocking):** Why are we overriding private methods?

It seems like the Abstract class could provide a useful pattern for us here:
https://levelup.gitconnected.com/design-patterns-with-typescript-interfaces-vs-abstract-classes-b6aab6e2ad21

I think that rather than introducing default behaviours we should enforce on the `IAggregation` interface that _handleAggregationUpdate_ and _handleAggregation_ are required behaviours.

This would also simplify the interface as we would no longer need field or aggregrationField as that logic would then be internal to the Aggregation.

Or potentially I am missing some benefit of those being present

thisislawatts (changes requested):

functions/src/aggregations/common.aggregations.ts
thisislawatts: Would this ever happen? This seems like firestore would be breaking the documented contract of the event: https://firebase.google.com/docs/reference/functions/providers_firestore.documentbuilder.html#onupdate

thisislawatts:
Thanks for the videos, very cool to see it all in action! 🎬 

So if I understand correctly the schema is:

```
user_badges: {
  [username]: {
    [badgeType]: boolean
  }
}
```

My understanding of the original ask was for [a list of the verified users](https://github.com/ONEARMY/community-platform/pull/1205). So whilst this aggregation does some of that work it will still require some additional computation on the client. We could instead switch to using flat arrays which would make the data easier to parse.

```
// Example schema:
user_badges: {
  [badgeType]: username[]
}

// Example of data structure
user_badges: {
  verified: ['precious-plastic']
}
```

This would then mean the implementation of [fetchAllVerifiedUsers](https://github.com/ONEARMY/community-platform/pull/1205/files#diff-777873ae5fb42cece21e6af1347582045f2955fd0ae23244773becf8673d3a8dR256-R261
) could be simplified.

Looking through the PR atm we are repeatedly [performing this kind of evaluation](https://github.com/ONEARMY/community-platform/pull/1205/files#diff-5a9f32ae0ff4d8d701d7f20f8a5e42fa0bd99bcb06e79e190b5fc606942d3525R192-R194) like:

```
// Current implementation in PR
verifiedUsers.some(
  user => user.userName === data._createdBy,
)

// Using proposed data structure
verifiedUsers.includes(data._createdBy);
```

functions/src/Utils/data.utils.ts
chrismclarke: Personal preference - a name like `data.ts` is pretty generic and really easy to confuse when importing or if multiple code tabs open. 

Would probably make sense at some point to decide on naming conventions, e.g. [backstage adr003](https://github.com/backstage/backstage/blob/master/docs/architecture-decisions/adr003-avoid-default-exports.md)

functions/src/Utils/data.utils.ts
chrismclarke: I'm typically reluctant on using 3rd party libs when all you need is a dozen or so lines of code - there's the ongoing overhead of ensuring the package is pinned at a compatible version, small performance hit whilst running (many of the methods introduce large number of additional checks and polyfills for methods that are now supported in node), as well as a bit of bloat depending on whether you've configured it well for tree-shaking or not. You're right that in the backend this is not so significant, I expect once containerised any of our function code pales in comparison to what is thrown in by firebase

I would say if we have any other use cases for it, it's not a bad library to have (pretty good chance one of our deps will have it installed anyways). I don't have any myself, but happy to add it pre-emptively if you think that's better to do.

functions/src/aggregations/common.aggregations.ts
chrismclarke: Do you mean integration to check that the emulators correctly trigger the function on change, or via mock data?

functions/src/aggregations/common.aggregations.ts
chrismclarke: It purely depends on whether the code I've written to detect diffs functions the same as the code firebase has to detect diffs. Deep-diving how firebase works is a horrible experience (done before, pledged never again) so I think makes sense to keep as a means to catch issues.

functions/src/aggregations/common.aggregations.ts
thisislawatts: I think invoking this function with mock data probably gives us enough confidence.

Ideally we would treat the entire thing as a black box and update the document and then observe the changes being aggregated. However I think the dependency on the emulator would this too brittle and more expensive.

functions/src/aggregations/common.aggregations.ts
chrismclarke: The methods should be public (at least that's how I understand typescript - things are assumed public unless private specified) - I don't think it is possible to overwrite a private method

The reason for the default is that I anticipate many of our aggregation cases being the same - take a collection, pick a field, summarise for all docs in the collection. I originally wanted to create a couple different types of aggregator (e.g. one that summarises a field, possibly one for multiple fields, one that returns a count of values etc.), but as the scope of this was already getting huge and exact use-cases hard to pin down I decided to leave things fairly basic with room to override.

The howto useful count provides such a unique case because the data is stored in user profiles, not in the howto itself (and unpicking that pattern again goes far beyond the scope of this PR). So we essentially need to aggregate across users, but then refactor the results to collate by howto id instead of the default user id. I didn't want to turn this into a generic method as I'm not sure it's a pattern we really want to be encouraging, so for now it's just a custom implementation on top of core aggregation methods.

As for abstract classes more generally, I really don't like them. For every article you read saying to use them, you will find another saying not to. My main reason for trying to avoid, is that for a new developer coming from javascript/typescript background, the odds of them having used before and understanding are pretty low - and so it adds an extra layer of terminology to get accustomed to. We do use them for the frontend db code as a fairly specific use-case for providing consistent db adaptors across sdks, but even that I'm not super convinced on.

functions/src/aggregations/common.aggregations.ts
thisislawatts: Sorry, let me clarify. If I am following the code flow correctly this method is only invoked via:

```
exports.default = functions.firestore
  .document(`${DB_ENDPOINTS.users}/{id}`)
  .onUpdate(change => {
    return handleDBAggregations(change, userAggregations)   
  });
```

The [documentation](https://firebase.google.com/docs/functions/firestore-events#trigger_a_function_when_a_document_is_updated) around `onUpdate` suggests that a document has changed. So why are also checking that a document change has occurred?

chrismclarke: Oh, because the aggregations aren't interested to a document change, but instead a specific field within the document. So whilst onchange will be triggered to any user profile change (e.g. their name), we only want to trigger the voting aggregation when the relevant user doc field changes. Firestore doesn't provide any means of listening to field changes, only doc changes.

Although, all this talk about change has made me realise that really I don't want the onUpdate listener anyway because that doesn't get triggered on create/delete events, so may have to go for onChange instead

https://firebase.google.com/docs/functions/firestore-events

functions/src/aggregations/common.aggregations.ts
thisislawatts: My preference is to avoid early abstractions. At the moment we have two aggregations in place yet we have decided that one of them represents the what should be the  _default_ behaviour, I am not sure that we have enough data to come to that conclusion at the moment.

chrismclarke:
> So whilst this aggregation does some of that work it will still require some additional computation on the client. We could instead switch to using flat arrays which would make the data easier to parse.
> 
> ```
> // Example schema:
> user_badges: {
>   [badgeType]: username[]
> }
> 
> // Example of data structure
> user_badges: {
>   verified: ['precious-plastic']
> }
> ```

Right now we have 3 open issues related to verified users: #1163, #1197, #1198, all which require some means of checking if one or more users is verified, and various issues related to aggregation in other ways including #1174, #1154. So whilst the flat arrays does definitely make a lot of sense for a list of all users with the verified badge, there are already some others (and likely many more in the future) that will still require some level of processing. So I'm inclined that we start with a system that replicates data in as standard a way as possible (i.e.. `users.user_id.badges` has the same data as `aggregations.user_badges.user_id`) and build out more methods from there (like the custom aggregator used for voted howtos).

thisislawatts:
I'm not sure I see the advantage of replicating data structures across multiple locations. As any schema change would then need to be made in the following places:

1. The User model
2. The Aggregations
3. Consumers of the Aggregations

My original suggestion around a flat array, was based on an assumption that each aggregation is being used to answer a single question, almost the equivalent of a key value store rather than creating a subset of User data elsewhere. This excerpt around MongoDB's aggregations has better words than me:

> When working with a database management system, any time you want to retrieve data from the database you must execute an operation known as a query. However, queries only return the data that already exists in the database. In order to analyze your data to find patterns or other information about the data — rather than the data itself — you’ll often need to perform another kind of operation known as an aggregation.
> https://www.digitalocean.com/community/tutorials/how-to-use-aggregations-in-mongodb#understanding-aggregation-pipelines

However I don't want to be blocking so many Issues 😅

thisislawatts (approved):

PR #1367, opened 2022-01-02
feat: notifications

patrycjapraczyk:
PR Checklist

- [x] - Latest `master` branch merged
- [x] - PR title descriptive (can be used in release notes)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

This PR added notifications to user profile. They are triggered when another user votes 'useful' or comments on their how-to. The notifications are shown under the thunderbolt icon in a pop-up. The user can 'clear notifications', which marks them as 'read' and removes them from the pop-up.

## Git Issues

Closes [#1199](https://github.com/ONEARMY/community-platform/issues/1199)

## Screenshots/Videos

No notifications:

Desktop:
image

Mobile:
image

There are notifications:

Desktop:
image

Mobile:
image

Issue #1199, opened 2021-07-31
Add notifications


davehakkens:
### Description
After introducing comments we need notifications so users can see they have a message to their content. There will be a small icon in the menubar that shows notifications. We start rolling out notifications in the how-to with two triggers

### 2 Triggers for notifications
**- New comment on a how-to** (if someone posts a comment on a how-to the creator of the how-to receives a notification)
**- New useful on a how-to** (if someone gives a useful to a how-to the creator of the how-to receives a notification) 

# Behaviour

## If there is no notification
- The icon is grey
- On click it says "no notifications"

https://user-images.githubusercontent.com/13672737/127753619-8d7cc8b2-27d5-4723-a480-614ad24e1c2a.mov

## If there are notifications
- The icon changes color.
- New notifications are shown chronologically on date (latest on top)
- The box has a fixed size> if the amount exceeds, scroll
- Notification messages contain URLs to visit the user or how-to
- "Clear notifications" removes all and turns the icon back to grey

https://user-images.githubusercontent.com/13672737/127753685-2636b7e5-1c3a-44ef-aa2d-889dd3f88de9.mov

### Example Live
💻 Live test "no notifications" [here](https://community-platform.webflow.io/u/how-to)
💻 Live test "notifications" [here](https://community-platform.webflow.io/u/how-to-with-notification)

chrismclarke:
Really nice feature to have, we would definitely handle most of this with backend triggered functions so likely will want to add a better testing setup for these before too much work can be done on this. I'll hopefully have that update live at some point this month however, so if anyone does want to take the issue on I'd be more than happy to help support.

patrycjapraczyk:
@chrismclarke could I try to help with this once the testing environment is ready?

chrismclarke:
That would be great! I'm mostly there with the testing environment, just need to find another half day to finish up (hopefully by the end of this week). Will add another message in this thread when ready

chrismclarke:
@patrycjapraczyk - The PR to implement a better system for testing backend functions is now open 🎉 : https://github.com/ONEARMY/community-platform/pull/1207 

Would you be willing to help with testing it at some point over the next week or so? It would be great to have another developer try testing to see if they can also get the emulators working locally and start being able to write their own test functions. This will also be an additional bounty task, because it really helps us make sure the code and instructions are good enough before merging.

The rough steps needed would be:
- Checkout the branch with the emulators code in (`feat/dev-emulators`)
[...]

PR #1464, opened 2022-02-08
feat: adds notifications functionality for howto useful votes and comments

thisislawatts:
PR Checklist

PR Type

## Description

A PR duplicating the work from #1367 with a cleaner `git` history after some fun and games. 

This PR added notifications to user profile. They are triggered when another user votes 'useful' or comments on their how-to. The notifications are shown under the thunderbolt icon in a pop-up. The user can 'clear notifications', which marks them as 'read' and removes them from the pop-up.

✨ As discussed during the recent dev call we will roll this out initially to our `beta-testers` for feedback. 

## Git Issues

Closes [#1199](https://github.com/ONEARMY/community-platform/issues/1199)

## Screenshots/Videos

No notifications:

Desktop:
image

Mobile:
image

There are notifications:

Desktop:
image

Mobile:
image


thisislawatts:
I think the issue is tied to wrapping the component in `beta-tester`

packages/cypress/src/fixtures/seed/users.json
```diff
@@ -1,4 +1,28 @@
 {
+  "notification_howto_author": {
```
chrismclarke, line 2: Are these fixtures used? I couldn't find a reference to them anywhere else (tests look like they're using the event_reader user) but could have missed

packages/cypress/src/integration/notifications.spec.ts
```diff
@@ -0,0 +1,127 @@
+import { DbCollectionName } from '../utils/test-utils'
+
+
+describe('[Notifications]', () => {
+
+    it('[are not generated when the howTo author is triggering notification]', () => {
```
chrismclarke, line 6: really nice to have this as an added catch

packages/cypress/src/integration/notifications.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+        cy.visit('how-to')
+        cy.login('howto_reader@test.com', 'test1234');
+        cy.visit('/how-to/testing-testing')
+        cy.wait(2000)
+        cy.get('[data-cy="comments-form"]')
+            .type('some sample comment')
+        cy.wait(2000)
+        cy.get('[data-cy="comment-submit"]')
+            .click()
+        cy.wait(7000)
```
chrismclarke, line 65: That's quite a long arbitrary wait. Might be easier to just wait until the comment appears in the frontend but not a huge issue (particularly if we later have tests that either call an API or emulator instead)

src/components/Notifications/NotificationItem.tsx
```diff
@@ -0,0 +1,106 @@
...
+      width="100%"
+      data-cy="notification"
+      style={{
+        fontSize: '12px', marginBottom: '10px',
+        paddingBottom: '10px', borderBottom: '1px solid #c7c7c7',
+        fontFamily: "Inter, sans-serif"
+      }}
+    >
+      {
+        type === "howto_useful" ?
```
chrismclarke, line 31: Specifying just `howto_useful` or _other_ might be a bit limiting if we want to extend on it. An alternative would be to either wrap in a logic block (e.g. `type === "howto_useful && (<flex>...</flex>)` ... `type === "howto_comment && (<flex>...</flex>)` or use a switch statement (where we also have a default).

But for now at least it covers what we need so can always just update as required later

src/components/Notifications/NotificationList.tsx
```diff
@@ -0,0 +1,20 @@
+import { NotificationItem } from 'src/components/Notifications/NotificationItem'
+import { INotification } from 'src/models'
+
+
+interface IProps {
+    notifications?: INotification[]
```
chrismclarke, line 6: I think given that this is the notification list it's reasonable to say notifications area required param and lose the conditional logic below (instead including the check higher up where implemented). But also no real issue having here either (just a matter of personal preference)

src/models/user.models.tsx
```diff
@@ -71,3 +72,17 @@ interface IUserStats {
 export type IUserDB = IUser & DBDoc
 
 export type UserRole = 'super-admin' | 'subscriber' | 'admin' | 'beta-tester'
+
+export interface INotification {
+  _id: string,
+  _created: string,
+  triggeredBy: {
```
chrismclarke, line 79: nice that we can track both of these to save lookup, although it may cause an issue if a user changes their name (can't remember if we have implemented that yet or not).  But I can't think of a reasonable alternative for now so say let's keep as is

src/pages/common/Header/Header.tsx
```diff
@@ -79,6 +100,15 @@ export class Header extends Component<IProps> {
 
   render() {
     const menu = this.injected.mobileMenuStore
+    const user = this.injected.userStore.user
+    const areThereNotifications = Boolean(
```
chrismclarke, line 104: It might be better to move this check to inside the notificationsIcon component (instead of passing as a prop). This helps keep the Header component a bit cleaner and avoids recalc every re-render on the header. You could instead just pass something like `user?.notifications || []` so that by default the icon still has an empty array, and run the verification check from there.

If this is an easy fix I think probably worth doing, or if you don't have the time I can also try making the changes (just let me know)

src/pages/common/Header/Menu/Notifications/NotificationsDesktop.tsx
```diff
@@ -0,0 +1,73 @@
...
+        {user ? (
+          <div data-cy="notifications-desktop">
+            <NotificationsIcon
+              onCLick={() => this.toggleNotificationsModal()}
+              isMobileMenuActive={false}
+              areThereNotifications={areThereNotifications}
+            />
+            <Flex>
+              {showNotificationsModal && (
+                <Foco onClickOutside={() => this.toggleNotificationsModal()}>
```
chrismclarke, line 61: I haven't used `Foco` before, looks handy!

src/pages/common/Header/Menu/Notifications/NotificationsModal.tsx
```diff
@@ -0,0 +1,100 @@
...
+  get injected() {
+    return this.props as IInjectedProps
+  }
+
+
+  render() {
+    const user = this.injected.userStore.user;
+    const notifications = user?.notifications?.
+      filter(notification => !notification.read).
+      sort((a, b) => new Date(b._created).getTime() -  new Date(a._created).getTime());
```
chrismclarke, line 63: I think as the dates are already formatted as isoString you can just compare the text (e.g. `b._created > a._created ? 1 : -1`), but converting to dates works too

src/stores/User/user.store.ts
```diff
@@ -284,8 +290,114 @@ export class UserStore extends ModuleStore {
...
+        } 
+
+        const lookup = await this.db
+          .collection<IUserPP>(COLLECTION_NAME)
+          .getWhere('userName', '==', username)
+
+        const user = lookup[0];
+        
+        const updatedUser: IUser = {
+          ...toJS(user),
```
chrismclarke, line 324: manipulating mobx objects outside of components can be a bit of a pain. I think what you've done with `toJS` seems reasonable for now - the other option would be to use things like `@computed` to automatically generate notifications as a separate entity (might be easier to manipulate), or firebase has some inbuilt utility methods to append items to arrays instead of having to pass back a full user object.

But given things are working I think fine to keep how they are for now.

src/stores/User/user.store.ts
```diff
@@ -284,8 +290,114 @@ export class UserStore extends ModuleStore {
...
+      }
+
+    } catch (err) {
+      console.error(err)
+      throw new Error(err)
+    }
+  }
+
+  @action
+  public async markAllNotificationsRead() {
```
chrismclarke, line 344: Really nice to have this option

chrismclarke (approved):
So much really nice stuff here, I know notifications are a feature people have been wanting for quite some time. I particularly appreciate how you've gone ahead and added tests, type definitions, plenty of components including mobile/desktop etc.

I've made a few minor comments, but nothing that should be seen as blocking (if the time is there to make minor fixes great, if not no worries we can iterate on further in the future)

If I were to suggest larger changes for the future these would probably be:
- Using a new store + db endpoint/subcollection for managing notifications (might be easier than managing as part of user)
- Use mobx computed properties to make it easier to handle read/unread notifications (avoid having to filter during render etc.)
- Generalise a bit to not just support howtos but general notifications (e.g. map pin accepted)

But this is more just food for thought/future iteration, and not to take away from a super cool new feature to have for now!

packages/cypress/src/fixtures/seed/users.json
patrycjapraczyk: Nah, they are not being used because I couldn't get the users I declared there to log in. I guess we could remove them or try to fix them, so that you use them in the tests.

packages/cypress/src/integration/notifications.spec.ts
patrycjapraczyk: Agreed. A bit messy on my side here but I didn't know all the cypress functions in my defense.

src/models/user.models.tsx
patrycjapraczyk: We need the name to display it in the notification box. I guess we could also try to retrieve the user name by its id. Do you think this'd be better?

packages/cypress/src/fixtures/seed/users.json
chrismclarke: That's what I thought - I expect larger changes will be needed to fix (I can't remember myself how the seed worked, pretty sure there's duplicate date in multiple places to manage thought). I think fine to just leave for now then for whoever is next working on tests

src/pages/common/Header/Menu/Notifications/NotificationsDesktop.tsx
patrycjapraczyk: Me neither. This was used for user profile modal as well. There's actually some problems with this. The modal can only be closed if you click outside it within the header but it doesn't work when you click outside anywhere. Same problem applies to profile modal.
I tried using something like this instead of Foco as well

`const ref = React.createRef<HTMLDivElement>();
    const handleClickOutside = (event) => {
      if (ref.current && !ref.current.contains(event.target)) {
        console.log("outside")
      }
    };
    document.addEventListener('click', handleClickOutside, true);`
    
   It seems like **document** in this case is only Header component.

src/models/user.models.tsx
chrismclarke: No I think what you've done is our best option for now. I know firebase has some additional methods to try to help with references to other documents, but I don't fully understand them and am not sure how much we want to lock ourselves into using them. I think it is likely to be very little impact also, as by the time a user changes their name there is a good chance the notification will have already been read/the person reading isn't likely to care very much.

src/pages/common/Header/Header.tsx
chrismclarke: We decided to merge anyway (didn't make sense to block things for this one small issue), so I would say less of a priority for now (just next time if anyone is working on this code could be a change worth making)

patrycjapraczyk: Sure, I guess once I get a new job I could have a look at these improvements. Thanks for the feedback!

src/pages/common/Header/Menu/Notifications/NotificationsDesktop.tsx
chrismclarke: oh, that seems to be a bit of a design flaw with foco! oh well, for now I think we can consider it a feature and wait to see if anyone complains. I'm guessing the eventual fix would be manually adding a click event listener to the main dom on modal and removing once closed, but probably easier to solve with a better general modal component than any specific implementation.

src/pages/common/Header/Header.tsx
chrismclarke: No rush or pressure, always nice to have a bit of time for things to settle before making lots more changes. Good luck with the job search and interviews!

src/pages/common/Header/Menu/Notifications/NotificationsDesktop.tsx
patrycjapraczyk: I tried using a manual event listener and also some other libraries and there was the same behaviour --> clicking outside working only for the header. I'm thinking it's because of the way we're loading these components but that's just guessing.

PR #1474, opened 2022-02-15
chore: move classes into their own files

thisislawatts:
PR Checklist

PR Type

## Description

* Adds an eslint rule to check for this behaviour
* Switch to interface for `AbstractDBClient` which better reflects
its usage
* Refactor AbstractDatabase to align with Abstract class conventions

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -46,6 +46,7 @@
     "react/prop-types": "off",
     "react/no-unknown-property": "warn",
     "@typescript-eslint/no-var-requires": "off",
-    "@typescript-eslint/no-namespace": "off"
+    "@typescript-eslint/no-namespace": "off",
+    "max-classes-per-file": "error"
   }
 }
--- a/src/stores/databaseV2/AbstractDatabase.ts
+++ b/src/stores/databaseV2/AbstractDatabase.ts
@@ -1,0 +1,36 @@
+import { CollectionReference } from "./CollectionReference";
+import { DBEndpoint } from "./endpoints";
+import { DBClients } from "./types";
+
+export abstract class AbstractDatabase {
+  private _clients: DBClients;
+
+  /**
+   * Gets a `CollectionReference` instance that refers to the collection at
+   * the specified path.
+   *
+   * @param clients pass a `DBClients` object to specify what to use as
+   * server and cache dbs. Uses a default configuration if not specified
+   */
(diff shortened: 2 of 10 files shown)


chrismclarke (commented):
Just checking, do you think anything requires testing or specific review?

I think if I understand correctly this PR replaces what was a an interface masquerading as a class (with accompanying .d.ts definition) for an actual abstract class?

I think at the time of creation the reason for doing so was just to copy how firebase managed code itself, but I expect a lot has changed since then (e.g. firebase now supporting tree shaking which I'm sure has involved a bunch of refactoring). But it makes pretty intuitive sense to me to use abstract classes for abstract classes.

So if there's any actual code you want me to look at/test let me know, otherwise I think it's mostly just refactoring so feel free to go ahead and merge if you've already tested and it is working.

thisislawatts:
@chrismclarke, this should be a non functional change. I _think_ that interfaces may not have been available in Typescript when this code was originally introduced.

My expectation is that tests protect us entirely. Aside from a human validating that, a single class per file is preferable and eyeballing the file moves I defer to the automated tests to protect us.

chrismclarke:
> @chrismclarke, this should be a non functional change. I think that interfaces may not have been available in Typescript when this code was originally introduced.

That's what I assumed. I'm pretty sure there were other reasons (alongside copying firebase maybe to do with not being able to re-implement private methods or similar), but these would have been purely functional (i.e. not able to do something wanted) and so given that this code works totally happy with the change. 

> My expectation is that tests protect us entirely. Aside from a human validating that, a single class per file is preferable and eyeballing the file moves I defer to the automated tests to protect us.

I've definitely used multiple classes in files before, sometimes I'd say very lazily/not great practice (e.g. when using class-based components if wanting to keep a few smaller ones together, so good to encourage the change), and other times for some reasonable level of utility (e.g. subroutines in a larger workflow) but never in such a way that separating causes harm (and there's always disable flags if last resort). 

So yeah, happy to go ahead with this. It's a shame we're using a json eslintrc file as might be nice to link to PR (so in future it's more clear to see it's a considered change and not just default that can be disabled when next seen as _inconvenient_)... unless it's one of those magic jsons where you can?

Agreed on the automated tests - one nice thing at least about tests that link up with an actual production-style environment.

chrismclarke (approved):

PR #1491, opened 2022-02-19
feat: add useful count to howto card

### LINKED ISSUE #1174: Add useful to howto

chrismclarke:
Author: danitrod <danitrod@outlook.com>

PR Checklist

PR Type

## Description

Rebase of #1204 to fix commit lint
Review changes:
- add howto stats aggregation bindings
- replace previous howto stats with aggregation

## Review Notes
This throws out most of the old system for collating total counters as now handled by more general aggregations method (and the counts seemed to be drifting off wildly - worst case I could see one howto with 32 upvotes from 6 users)

A couple minor caveats/notes
- The counter would appear on top of tags in cases where the tags overflowed, and so I've wrapped the tags to spread across multiple lines instead. This means some howtos are longer than others

- When liking a howto there is a delay before the aggregation is updated on the server (usually 1-2s), and also the update is not reflected on the main list page until after page refresh. We can improve on this in the future although it will require more general improvements to aggregations and so better left out of this specific pr

- I've made it so the counter does not appear on the list page when there are 0 votes (not sure if preferred or not)

## Git Issues

Closes #1174

## Screenshots/Videos

Video showing longer tags (as mentioned) and slower update of counter

https://user-images.githubusercontent.com/10515065/154826339-90b9954c-9006-435e-af85-c99892a16605.mp4

---

## What happens next?

### DIFF
--- a/src/pages/Howto/Content/Howto/Howto.tsx
+++ b/src/pages/Howto/Content/Howto/Howto.tsx
@@ -134,7 +134,9 @@ export class Howto extends React.Component<
             verified={
               this.injected.userStore.verifiedUsers[activeHowto._createdBy]
             }
-            votedUsefulCount={this.store.howtoStats?.votedUsefulCount}
+            votedUsefulCount={
+              this.injected.userStore.userVotedHowtos[activeHowto._id]
+            }
             loggedInUser={loggedInUser}
             needsModeration={this.store.needsModeration(activeHowto)}
             userVotedUseful={this.store.userVotedActiveHowToUseful}
--- a/src/pages/Howto/Content/HowtoList/HowtoList.tsx
+++ b/src/pages/Howto/Content/HowtoList/HowtoList.tsx
@@ -92,6 +92,17 @@ export class HowtoList extends React.Component<any, IState> {
     this.props.howtoStore.updateSearchValue('')
   }
 
+  componentDidMount() {
+    /**
+     * Currently the `userVotedHowtos` property is only
+     * populated by the constructor of UserStore.
+     *
+     * To ensure the value is updated check the store
+     * each time the component is mounted.
+     */
+    this.props.userStore?.loadUserHowtoVotes()
+  }
+
   public render() {
     const { verifiedUsers } = this.injected.userStore
     const {
--- a/src/pages/Howto/Content/HowtoList/HowtoList.tsx
+++ b/src/pages/Howto/Content/HowtoList/HowtoList.tsx
@@ -191,11 +191,14 @@ export class HowtoList extends React.Component<any, IState> {
             >
               <VirtualizedFlex
                 data={filteredHowtos}
-                renderItem={(data: IHowtoDB) => (
+                renderItem={(howto: IHowtoDB) => (
                   <Box px={4} py={4}>
                     <HowToCard
-                      howto={data}
-                      verified={verifiedUsers?.[data._createdBy]}
+                      howto={howto}
+                      verified={verifiedUsers?.[howto._createdBy]}
+                      votedUsefulCount={
+                        this.injected.userStore.userVotedHowtos[howto._id]
+                      }
                     />
                   </Box>
                 )}
--- a/functions/src/models.ts
+++ b/functions/src/models.ts
@@ -6,11 +6,10 @@ import type {
   IDBEndpoint,
   IEventDB,
   IHowtoDB,
-  IHowtoStats,
   IUserDB,
   IMapPin,
 } from 'one-army-community-platform/src/models'
-export { DBDoc, IDBEndpoint, IEventDB, IHowtoDB, IHowtoStats, IUserDB, IMapPin }
+export { DBDoc, IDBEndpoint, IEventDB, IHowtoDB, IUserDB, IMapPin }
 
 import { generateDBEndpoints } from 'oa-shared'
 export const DB_ENDPOINTS = generateDBEndpoints()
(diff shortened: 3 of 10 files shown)


src/pages/Howto/Content/Howto/Howto.tsx
```diff
@@ -134,7 +134,9 @@ export class Howto extends React.Component<
             verified={
               this.injected.userStore.verifiedUsers[activeHowto._createdBy]
             }
-            votedUsefulCount={this.store.howtoStats?.votedUsefulCount}
+            votedUsefulCount={
+              this.injected.userStore.userVotedHowtos[activeHowto._id]
```
thisislawatts, line 138: **nitpick (non-blocking):** Should we coerce this to a number to protect against malformed data? 

```suggestion
              parseInt(this.injected.userStore.userVotedHowtos[activeHowto._id]|| 0, 10)
```

thisislawatts (approved):

davehakkens:
Small note: 
Looks like the `useful` button behavior changed after this update.

**Original**
By default the button is black. When you hover or give useful it becomes colored and saves this state. So next time you visit you can see you already used the useful button for that how-to.

**In this PR**
After using, it adds the point but it seems to go back to the default black state. 
So you don't see if you've already used the useful button which is very confusing
video to illustrate
https://user-images.githubusercontent.com/13672737/154836504-35a65e5a-d53f-4922-8827-f7b6e9b39d4f.mov

thisislawatts:
@davehakkens, this is mentioned in the original PR description. Do you think this is an issue that blocks merging this feature or something we can iterate on in future releases? 

> * When liking a howto there is a delay before the aggregation is updated on the server (usually 1-2s), and also the update is not reflected on the main list page until after page refresh. We can improve on this in the future although it will require more general improvements to aggregations and so better left out of this specific pr

thisislawatts:
@chrismclarke and @davehakkens, I have pushed a naive fix to ensure the count is updated when a user takes the following steps:
1. Loads howto listing
2. Clicks through to how to and either votes useful OR removes their useful vote
3. Clicks back to howto listing.

davehakkens:
Ah I guess I misread that description. Thought it was more about the count being a bit delayed. 
Because here it looks that you will never see which how-to's you marked as useful. Since they always show a black star (not colored). 

Not sure how complex it is to change. But I would see that as quite an important behaviour, otherwise the `useful` feels a bit buggy.

davehakkens:
> I have pushed a naive fix to ensure the count is updated when a user takes the following steps:

@thisislawatts I think after this change some `0 `showed up? - I don't recall seeing them before
useful_zero

thisislawatts:
🦅 👁️ @davehakkens Investigating the isssue around Useful button not accurately showing when the current user has voted _Useful_

thisislawatts:
https://github.com/ONEARMY/community-platform/pull/1491/commits/484fe02cb0d198d762431df1e30d0a00d82798e7 Should fix it :sparkles: @davehakkens, take it for a spin when you get a chance.

davehakkens:
> [484fe02](https://github.com/ONEARMY/community-platform/commit/484fe02cb0d198d762431df1e30d0a00d82798e7) Should fix it ✨ @davehakkens, take it for a spin when you get a chance.

Thanks for looking into it. Took a spin but looks like it's still black here by default..

https://user-images.githubusercontent.com/13672737/154861858-31e51752-10df-4412-a5ff-fdf56b475ece.mp4

chrismclarke:
There could easily be something I overlooked at some point too. If you're happy applying fixes @thisislawatts I'm cool with whatever, if not I can update next week

src/pages/Howto/Content/Howto/Howto.tsx
chrismclarke: I think I made a similar correction one level down the chain to distinguish but no harm having here for sure. Given that the data is coming from an aggregation function I don't really think we need to worry about type coersion

chrismclarke:
@davehakkens - I just the updated fix myself too and working for me, so there's a chance the preview link either hadn't been updated for you or some weird caching. Would you be able to quickly try again? If same maybe try an incognito window?

src/pages/Howto/Content/HowtoList/HowtoList.tsx
```diff
@@ -92,6 +92,17 @@ export class HowtoList extends React.Component<any, IState> {
...
 
+  componentDidMount() {
+    /**
+     * Currently the `userVotedHowtos` property is only
+     * populated by the constructor of UserStore.
+     *
+     * To ensure the value is updated check the store
+     * each time the component is mounted.
+     */
+    this.props.userStore?.loadUserHowtoVotes()
```
chrismclarke, line 103: seems like a sensible fix to me for now

davehakkens:
> @davehakkens - I just the updated fix myself too and working for me, so there's a chance the preview link either hadn't been updated for you or some weird caching. Would you be able to quickly try again? If same maybe try an incognito window?

Tried Chrome, Firefox, Safari incognito. They all show a black star here...
Are we sure we are talking about the same thing? 😅

But If it works for you guys happy to merge and I'll continue to test in dev.

chrismclarke:
Super weird, I should probably try investigate more but might be better saved for another day.  For reference here's what I see @davehakkens 

https://user-images.githubusercontent.com/10515065/154881859-8353cc2a-25d2-4549-b1af-17de7dd1b12d.mp4

davehakkens:
> Super weird, I should probably try investigate more but might be better saved for another day. For reference here's what I see @davehakkens
> Precious.Plastic.Community.2.mp4

Thanks for making the video. That looks good, seem like we are talking about the same thing!
Strange indeed. I'll keep on polking some people to test on their devices..

PR #1492, opened 2022-02-20
feat: extract components out to standalone package

thisislawatts:
PR Checklist

PR Type

## Description

**The Problem:**

At the moment our components are stored
within the root of the yarn workspace. This
works well for the existing usage but limits
the portability of our UI components. Whilst
researching server side rendering I attempted
to load a component into a server-less function
I encountered numerous issues tied to the
combination of TypeScript and Yarn workspace.

**Proposed Solution:**

Extract all core and reusable components into
a new `oa-components` workspace located under
the packages directory.

This package could then be consumed by the primary
application AND additional applications, for example
a standalone serverless function to generate
server side rendered documents.

The external interface for this package would be:

1. Components imported via `import {Button} from 'oa-components'`
2. A storybook build published for external reference.

The scope of this PR demonstrates the approach and moves a single component
`Button` to be loaded from the new `oa-components` library.

**Advantages:**

1. Simplify the project configuration and better align with established patterns of Yarn Workspaces & Typescript.
2. Support additional usecases for UI components
3. Lower barrier to entry for designers/contributors looking to make improvements and fix bugs in UI components.
4. Encourage a separation between the data layer and UI components, which could also simplify the testing of components.
5. Move our storybook configuration closer to the components which it aims to provide documentation for.
6. Allow us to apply more structured coding standards for components.

**Disadvantages:**

1. Complicates contribution for folks working across backend logic, which would remain in the project root, whilst other changes would be required to made in `oa-components`. Similar to the division between the primary application and our serverless functions.
2. We would need to define a criteria of what constitutes a reusable component vs those stateful singletons which only many sense within the root application.


chrismclarke (commented):
I'd say in principle I'm in favour of doing this and the implementation makes sense to me, although I'm aware that I'm a bigger fan of monorepos than most and also have done very little work on the frontend components. I'd be interested in seeing whether this impacts much on developer workflow (e.g. making tweaks to components whilst trying to preview from within the full app)

I think probably worth quickly discussing on the maintainer call any potential knock-ons and required next steps to see if worth moving ahead with

thisislawatts:
Totally agree @chrismclarke, there's no rush on getting this change applied. Would be good to give this time to gather feedback from the wider contributor community.

tsconfig.json
```diff
@@ -41,6 +48,12 @@
     "functions",
     "src/utils/UserMigration",
     "src/setupProxy.js",
-    "src/assets/userReportInit.js"
+    "src/assets/userReportInit.js",
+    "packages/components"
+  ],
+  "references": [
+    {
+      "path": "./packages/components"
```
chrismclarke, line 56: (non-blocking) - do we need this reference? given that the components are being provided via node_modules (from yarn workspace) I don't think this should be required but could be wrong

packages/components/src/Button/Button.tsx
```diff
@@ -1,44 +1,43 @@
...
   Button as RebassButton,
   ButtonProps as RebassButtonProps,
 } from 'rebass/styled-components'
-import Text from '../Text'
 import styled from 'styled-components'
 
 // extend to allow any default button props (e.g. onClick) to also be passed
 export interface IBtnProps extends React.ButtonHTMLAttributes<HTMLElement> {
-  icon?: keyof IGlyphs
+  icon?: any
```
chrismclarke, line 10: (non-blocking) - shame to lose the type-checking/autocomplete, but assume that is temporary for now as the components can't import from the main repo and we can hopefully add back once icon migrated over

packages/components/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "@babel/core": "^7.14.3",
+    "@mdx-js/react": "^1.6.22",
+    "@react-theming/storybook-addon": "^1.1.5",
+    "@storybook/addon-actions": "^6.4.19",
+    "@storybook/addon-docs": "^6.4.19",
+    "@storybook/addon-essentials": "^6.4.19",
+    "@storybook/addon-knobs": "^6.2.9",
+    "@storybook/addon-links": "^6.4.19",
+    "@storybook/preset-create-react-app": "^4.0.1",
+    "@storybook/react": "^6.4.19",
```
chrismclarke, line 28: I had upgrading storybook on a backlog list so great to see you've managed to achieve it here. Smart move to keep the addon-knobs pinned lower to give us more time to handle the deprecation/refactor required from 6.4

packages/components/.storybook/main.js
```diff
@@ -4,7 +4,10 @@ const paths = {
 }
 
 module.exports = {
-  stories: ['../../../src/**/*.stories.@(js|jsx|ts|tsx|mdx)'],
+  stories: [
+    '../../../src/**/*.stories.@(js|jsx|ts|tsx|mdx)',
+    '../../components/src/**/*.stories.@(js|jsx|ts|tsx|mdx)',
```
chrismclarke, line 9: really nice to be able to use both systems in tandem for a bit whilst we organise

package.json
```diff
@@ -15,12 +15,13 @@
   "private": true,
   "main": "lib/index.js",
   "scripts": {
-    "start": "cross-env FAST_REFRESH=false react-scripts start",
+    "start": "yarn build:components && cross-env FAST_REFRESH=false yarn react-scripts start",
```
chrismclarke, line 18: A bit tedious for now, but hopefully once we upgrade and eject we can add better build configuration to handle with the same script. As mentioned we will lose live-reload on the components, but that is easy enough to fix if it becomes a chore (replacing the build:components with a serve:components script that can run tsc-watch or similar)

chrismclarke (approved):
I've finally had a bit more time to look at this properly and I really like it, the separation makes a lot of sense and all working well locally for me. I agree with limitations you've raised, but think these are all things we can iterate on and having this core in place puts us in a much better position to do so.

I'll approve here and merge once we can get tests to pass. I've also added an umbrella issue (#1532) to link back and track future work as required, but overall really nice addition, thanks for taking the time to do all of this!

PR #1515, opened 2022-03-07
Feat: aggregations store

chrismclarke:
PR Checklist

PR Type

## Description

Creates a common aggregations store that can be used to populate aggregation data (previously required explicit db calls)

The methods try to safeguard against multiple subscriptions being called against the same aggregation endpoint, and also try to find a balance between keeping aggregation data fresh (subscribing to specific endpoints and streaming updates) vs leaving open longer than necessary (excessive requests and memory leaks).

This improves on previous implementation by:
- Avoiding direct db calls from components/stores
- Reducing number of calls made to keep data fresh (e.g. component mount hook)
- Correctly update data in cases where function side-effects take time (e.g. going back to howto list after vote, but before vote count updated)

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/src/stores/Aggregations/aggregations.store.tsx
+++ b/src/stores/Aggregations/aggregations.store.tsx
@@ -12,8 +12,9 @@ const AGGREGATION_DOC_IDS = [
   'users_verified',
 ] as const
 
-// Utility type to generate from list of aggregation docs ids
+// Utility types generated from list of aggregation docs ids
 type IAggregationId = typeof AGGREGATION_DOC_IDS[number]
+type IAggregations = { [aggregationId in IAggregationId]: any }
 
 /** Aggregation subscriptions default close after 5 minutes */
 const DEFAULT_TIMEOUT = 1000 * 60 * 5
@@ -24,9 +25,7 @@ const DEFAULT_TIMEOUT = 1000 * 60 * 5
  */
 export class AggregationsStore {
   /** Observable list of all aggregations by id */
-  @observable aggregations: {
-    [aggregationId in IAggregationId]: any
-  }
+  @observable aggregations: IAggregations
 
   private db: DatabaseV2
   private subscriptions$: { [aggregationId: string]: Subscription } = {}
@@ -54,7 +53,7 @@ export class AggregationsStore {
    * Subscribe to updates to a particularly aggregation. Includes a default timeout so that subscriptions are not
    * left hanging too much longer than required, but at the same time are not repeatedly created to re-fetch same data
    */
-  updateAggregation(
+  public updateAggregation(
     aggregationId: IAggregationId,
     timeoutDuration = DEFAULT_TIMEOUT,
   ) {
--- a/src/stores/Aggregations/aggregations.store.tsx
+++ b/src/stores/Aggregations/aggregations.store.tsx
@@ -62,7 +62,11 @@ export class AggregationsStore {
         .collection('aggregations')
         .doc(aggregationId)
         .stream()
-        .subscribe(value => this.updateAggregationValue(aggregationId, value))
+        .subscribe(value => {
+          if (value) {
+            this.updateAggregationValue(aggregationId, value)
+          }
+        })
       this.subscriptions$[aggregationId] = subscription
     }
     this.setSubscriptionTimeout(aggregationId, timeoutDuration)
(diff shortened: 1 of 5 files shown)


src/stores/Aggregations/aggregations.store.tsx
```diff
@@ -0,0 +1,96 @@
...
+/** Aggregation subscriptions default close after 5 minutes */
+const DEFAULT_TIMEOUT = 1000 * 60 * 5
+
+/**
+ * The aggregation store provides access to various aggregation endpoints, created as a means to cheaply
+ * retrieve data collated across entire collections as single documents
+ */
+export class AggregationsStore {
+  /** Observable list of all aggregations by id */
+  @observable aggregations: {
```
chrismclarke, line 27: This is the main observable object that other components/stores can use to subscribe to lists like `users_verified`

src/stores/Aggregations/aggregations.store.tsx
```diff
@@ -0,0 +1,96 @@
...
+  @action
+  private updateAggregationValue(aggregationId: IAggregationId, value: any) {
+    this.aggregations[aggregationId] = value
+  }
+
+  /**
+   * Subscribe to updates to a particularly aggregation. Includes a default timeout so that subscriptions are not
+   * left hanging too much longer than required, but at the same time are not repeatedly created to re-fetch same data
+   */
+  updateAggregation(
```
chrismclarke, line 57: This is the other public method for prompting update for an endpoint - it uses a subscriptions tracker and timeouts to prevent repeated subscription to the same endpoint

src/stores/User/user.store.ts
```diff
@@ -32,12 +32,15 @@ export class UserStore extends ModuleStore {
   @observable
   public updateStatus: IUserUpdateStatus = getInitialUpdateStatus()
 
-  /** A list of all the verified users, to display verified icons where needed */
-  @observable
-  public verifiedUsers: { [user_id: string]: boolean } = {}
+  // redirect calls for verifiedUsers to the aggregation store list
+  @computed get verifiedUsers(): { [user_id: string]: boolean } {
```
chrismclarke, line 36: Interim solution to prevent PR growing too big - in upcoming PR plan to refactor so that a shared component can handle check and display of verified user badge instead of repeated calls to injected stores

chrismclarke:
(will take a look at broken tests now)

thisislawatts (approved):

src/stores/User/user.store.ts
```diff
@@ -53,9 +56,12 @@ export class UserStore extends ModuleStore {
...
     this._listenToAuthStateChanges()
-    // load aggregations - TODO - likely better in a separate aggregations store
-    this.loadVerifiedUsers()
-    this.loadUserHowtoVotes()
+    // update howto votes and verified users on intial load
+    // use timeout to ensure aggregation store initialised
+    setTimeout(() => {
+      this.loadUserHowtoVotes()
+      this.loadVerifiedUsers()
+    }, 50)
```
thisislawatts, line 64: **question:** Is the `50` significant here? I am worried this may be vulnerable to JS execution times differing between devices causing this to break. Is there a way to explicitly await on the aggregation store being available?

thisislawatts: > The issue is with the way in which we need some stores to access others (and provide a common root store that all register themselves in), however accessing from the constructor methods don't work because it's only after constructor complete that each store is available in root store.

This to me seems like a largely knot which needs to be unpicked and I worry a little that we're not making the situation any better by introducing a **new** module but I don't have anything more constructive to suggest at this moment in time.

These dependencies between the stores are hard to reason with and also complicate testing these components in isolation.

chrismclarke: Agreed, I do find it a bit ridiculous that injecting all stores within all stores is still what mobx recommends for these situations (https://mobx.js.org/defining-data-stores.html#combining-multiple-stores)

The current saving grace is that with the advent of provider/hooks in React another option is enabled for accessing things from different places without lots of messy injection and prop passing (well, still messy but less so from a syntax point of view). This is what the `useCommonStores` hook provides, and what I'm slowly refactoring more components to use. The only limitation is that this only applies to rendered functional components, so will always need the store-specific approach if sharing across stores. 

But yes, one for the _thorny_ list for sure - definitely could be improved on but I think that will have to happen independent of aggregation work as would quickly become blocking on some roadmap-critical features (namely admin panel)

PR #1533, opened 2022-03-10
Feat/admin dashboard

chrismclarke:
PR Checklist

PR Type

## Description

- Add draft pending content approvals page to admin panel

## Git Issues

Closes #

## Screenshots/Videos

https://user-images.githubusercontent.com/10515065/157598958-ad756a45-95b5-40c8-9dc0-5ce2a6defce2.mp4

---

## What happens next?


src/modules/admin/pages/admin-approvals.tsx
```diff
@@ -1,10 +1,175 @@
 import { observer } from 'mobx-react'
```
thisislawatts, line 1: **question:** Why are the files in this directory using a different naming convetion?

Elsewhere in the application we use [`PascalCase`](https://www.theserverside.com/definition/Pascal-case) but in this module we are using [`kebab-case`](https://www.theserverside.com/definition/Kebab-case)

I think it would be worth standardising, I also found [a eslint rule](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/filename-case.md) we could introduce to automate checking for this.

What do you think?

src/modules/admin/pages/admin-approvals.tsx
```diff
@@ -1,10 +1,175 @@
...
+      const pendingApprovals = await getPendingApprovals(endpoint, db)
+      approvals[endpoint] = pendingApprovals
+    }
+    console.log('allPending', approvals)
+    setPending(approvals)
+  }, [])
+
+  useEffect(() => {
+    getApprovals()
+  }, [])
```
thisislawatts, line 68: **question:** Do we need `useCallback` here or could we inline the getApproval method inside of useEffect?

Otherwise, if we do want to keep the code extracted, based on what I read in [this post](https://devtrium.com/posts/async-functions-useeffect#what-if-you-need-to-extract-the-function-outside-useeffect), we should add `[getApprovals]` as a dependency of the useEffect method.

I have no strong opinions here, looking to improve my understanding of best practices around React hooks.

thisislawatts (changes requested):

src/modules/admin/pages/admin-approvals.tsx
chrismclarke: nice catch, likely fallout from whatever other projects I was working on that day...

I expect any lint rule will fail initially and involve a small amount of refactoring (we've never actively decided on one convention or another, just by coincidence/copy-paste-kismet that most files tend to have similar conventions), but no harm running a quick check to see how much work would be involved

src/modules/admin/pages/admin-approvals.tsx
chrismclarke: could be inlined, although initial dev the function was a lot messier so nicer to have outside. I think it's a reasonable contender for future refactoring so probably makes sense to keep workable.

I struggled myself to get to grips with how to use async functions with hooks, by default I always write class components but felt like worth trying to figure out other ways seeing as it's what seems to be promoted a bit more. 

From what I understand about the dependencies object, the list is mostly used to decide when to potentially trigger/retrigger evaluation. I never want to trigger the re-evaluation (just hook into the mount lifecycle) so they've been purposefully omitted. I think at one point during dev I found the query being duplicated and removing fixed.

The react FAQs discusses this a little bit more here: https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies

Which might make more sense with the context of: https://dmitripavlutin.com/react-useeffect-explanation/

src/modules/admin/pages/admin-approvals.tsx
chrismclarke: tidied locally and added #1616 generally

thisislawatts (approved):

PR #1536, opened 2022-03-11
chore: refactor to enforce functional components

thisislawatts:
PR Checklist

PR Type


packages/components/src/Icon/svgs.tsx
```diff
@@ -6,10 +6,8 @@ const imgStyle = {
...
 }
 
-const iconMap = {
+export const iconMap = {
   loading: <img alt="icon" style={imgStyle} src={loadingSVG} />,
   star: <img alt="icon" style={imgStyle} src={starSVG} />,
   starActive: <img alt="icon" style={imgStyle} src={starActiveSVG} />,
 }
-
-export default iconMap
```
thisislawatts, line 15: Removing default exports

chrismclarke (approved):
I definitely prefer named exports so feels like it makes sense for the components package at least. If I remember, there was a time back-when that default exports were incompatible with some specific react-wrapper features like `withRouter` or `withStore` , however I would hope that's still not an issue and definitely shouldn't be a problem for the components package

PR #1538, opened 2022-03-13
fix: increase caching for static JS resources

thisislawatts:
PR Checklist

PR Type

## Description

After deploying a new iteration of the site a user
can hit an error state due to the service worker
attempting to load a JS file which no longer exists
on the Firebase hosting. This results in a 
broken white page. 

The default caching is 1 hour. However as the JS
bundles include a md5 string they are unique so can
be cached indefinitely. 

The intent of this change is to reduce the number of times
people encounter this white page of death. As each deployment
removes all previous files from the hosting, the aim 
here is for users to be able to load from their own cache.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -42,6 +42,19 @@
         ]
       },
       {
+        "source": "/static/**/*.js",
+        "headers": [
+          {
+            "key": "Cache-Control",
+            "value": "public, max-age=31536000"
+          },
+          {
+            "key": "Access-Control-Allow-Origin",
+            "value": "*"
+          }
+        ]
+      },
+      {
         "source": "/service-worker.js",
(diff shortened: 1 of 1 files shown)


chrismclarke (approved):
Oooh, interesting. I know that the workbox service worker is set to automatically use a cache-first approach on these assets
```
registerRoute(new RegExp('/static/'), new CacheFirst())
```
_src\service-worker.ts_

but I hadn't stopped to think what happens if the cache-control max-age is invalidated by firebase. So if this works it will be a really nice fix to have. 

The other option would possibly be to invalidate the index file faster, as that is the file that stores references to the no-longer-available js files.... But let's try this one out first

PR #1545, opened 2022-03-17
feat(howtos): add sorting by most useful

### LINKED ISSUE #1172: Add dropdown filter for most useful
### Description
Add a dropdown filter on the how-to overview page to sort on most useful.
- Default is latest.
- Ideally it works in combination with the other filters. So if you select a tag or search. You can also filter on "latest" or "most usefull"
- Will make more sense when #1174 is implemented

### Page related

Will be used in :https://community.preciousplastic.com/how-to

### Mockup

afbeelding

###Live version
https://community-platform.webflow.io/u/how-to

danitrod:
PR Checklist

PR Type

## Description

Adds the ability to sort Howto cards by most useful count.

## Git Issues

Closes #1172

## Screenshots/Videos

Screen Shot 2022-03-17 at 01 01 08

### DIFF
--- a/src/components/SortSelect/index.tsx
+++ b/src/components/SortSelect/index.tsx
@@ -1,0 +1,75 @@
+import { inject, observer } from 'mobx-react'
+import { Component } from 'react'
+import Select from 'react-select'
+import { OptionsType, ValueType } from 'react-select/lib/types'
+import { HowtoStore } from 'src/stores/Howto/howto.store'
+import { DropdownIndicator } from '../DropdownIndicator'
+import { FieldContainer } from '../Form/elements'
+import { FilterStyles } from '../Form/Select.field'
+
+interface ISortOption {
+  value: string
+  label: string
+}
+
+interface IProps {
+  usefulCounts: { [key: string]: number }
+}
+
+interface InjectedProps extends IProps {
+  howtoStore: HowtoStore
+}
+
+interface IState {
+  value: ValueType<ISortOption> | null
+}
+
+const sortingOptions: OptionsType<ISortOption> = ['Latest', 'Most useful'].map(
+  label => ({
+    label,
+    value: label.toLowerCase(),
+  }),
+)
+
+@inject('howtoStore')
+@observer
+class SortSelect extends Component<IProps, IState> {
+  constructor(props: any) {
+    super(props)
+    this.state = { value: null }
+  }
+
+  get injected() {
+    return this.props as InjectedProps
+  }
+
+  get store() {
+    return this.injected.howtoStore
+  }
+
+  onSortValueChange(value: ISortOption) {
+    this.setState({ value })
+    if (value.value === 'most useful') {
+      this.injected.howtoStore.sortHowtosByUsefulCount(this.props.usefulCounts)
+    } else {
+      this.injected.howtoStore.sortHowtosByLatest()
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


davehakkens:
awesome! works very nice here @danitrod 

I did notice I can use the "sort by most useful" when i'm filtering with tags. But tot when I search.
Would it be possible to be able to "search" and then "sort by most useful"?

Or complex to add?

src/components/SortSelect/index.tsx
```diff
@@ -0,0 +1,75 @@
+import { inject, observer } from 'mobx-react'
```
thisislawatts, line 1: **issue (blocking):** We are inlining the `HowtoStore` into this component, which explicitly couples it to the Howto context. Where as the idea for components under `src/components` is that they are suitable for reuse in multiple places. 

I see the following options available here:
1. Move this component so it sits under `pages/Howto`
2. Refactor so the component to emit an event which can be defined at the call site. https://reactjs.org/docs/handling-events.html. You would also want to move `sortingOptions` out to passed through as a prop.

thisislawatts:
Thanks for raising this PR @danitrod, this is a really "Useful" thing to have added to the site 😉 

Will take a look now 👀

thisislawatts (changes requested):

src/components/SortSelect/index.tsx
chrismclarke: I think were you to refactor as a pure component you basically would just be left with the select box and a bunch of props/events passed back from it, so I'd say option (1) sounds most sensible to me

danitrod: Yeah agree that if we go with (2) we would be left with a redefinition of the Select component. So refactoring to go in `pages/Howto`.

danitrod:
@davehakkens I think it does get a little more complicated to sort when searching because we are using fuzzy matching, and the search matching order currently has priority over the sorting priority. Which does make sense if you think about it: if you search for "machine" you'd want the card that matches the best with this word, not the card that has something similar, like "macaroon", but has more useful counts.

An alternative would be swapping to strict search, or giving priority to the sorting even though that could mean poor search matches.

thisislawatts:
Kicking off the tests again @danitrod ⌛

davehakkens:
> @davehakkens I think it does get a little more complicated to sort when searching because we are using fuzzy matching, and the search matching order currently has priority over sorting. Which does make sense if you think about it: if you search for "machine" you'd want the card that matches the best with this word, not the card that has something similar, like "macaroon", but has more useful counts.
> 
> An alternative would be swapping to strict search, or giving priority to the sorting even though that could mean poor search matches.
> 
> Edit: we could also add a "best match" option to the sort dropdown, which would be the default (I like this one)

Thanks for explaining. Clear. Let's skip that one.
The "best match" sounds like a nice option though.
If it's easy to add squeeze it in. If not we can make another issue for it.
Let me know :)

thisislawatts (approved):

thisislawatts:
@danitrod let me know if you want to add the _Best match_ approach to this branch or raise separately. I am happy to merge this PR as it is and iterate!

danitrod:
Sorry for the delay, I do think it is better to separate the `Best match` approach into another issue, as we would have to restructure quite a bit of how the sorting/filtering works.

PR #1547, opened 2022-03-17
chore: reduce react icon bundle size

thisislawatts:
PR Checklist

PR Type

## Description

It seems that although we are only using
a limited selection of icons from this package
we are seeing the react-icons bundle at 2.61Mb

Switching to use @react-icons/all-files instead
which seems to offer better tree shaking.

The linked GitHub issue thread suggests that
webpack _should_ be able to optimise the bundle
size. However I think owing to our dependency
on an outdated CRA we are not getting those benefits.

https://github.com/react-icons/react-icons/issues/154

Before: 

Screenshot 2022-03-17 at 22 24 05

After:

Screenshot 2022-03-17 at 22 57 58

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "terser-webpack-plugin": "4.2.3"
   },
   "dependencies": {
+    "@react-icons/all-files": "^4.1.0",
     "@sentry/react": "^6.15.0",
     "@uppy/core": "^2.1.4",
     "@uppy/dashboard": "^2.1.3",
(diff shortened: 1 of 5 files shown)


chrismclarke (approved):
Nice catch, a bit tedious that the tree-shaking doesn't happen automatically but the fix looks good to me. I don't _think_ we have any automated tests that would catch broken/missing icons, but I very quick manual test looks like at least some are there so that's good enough for me!

I think there's another tedious task of at some point figuring out how many of them we still actually use (I'm guessing probably only around half or so), but definitely one for a later day

PR #1569, opened 2022-03-25
chore(deps): adds prop-types

thisislawatts:
PR Checklist

PR Type

## Description

Aims to resolve the following issues reported by yarn during installation:

```
➤ YN0002: │ one-army-community-platform@workspace:. doesn't provide prop-types (p25e4b), requested by react-ga
➤ YN0002: │ one-army-community-platform@workspace:. doesn't provide prop-types (p89ff4), requested by react-hamburger-menu
➤ YN0002: │ one-army-community-platform@workspace:. doesn't provide prop-types (p1f2ea), requested by react-table
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
     "oa-shared": "1.0.0",
     "pino": "^7.2.0",
     "pino-logflare": "^0.3.12",
+    "prop-types": "15",
     "pubsub-js": "^1.7.0",
     "react": "^17.0.2",
     "react-datepicker": "^2.9.6",
--- a/yarn.lock
+++ b/yarn.lock
@@ -23264,6 +23264,7 @@ fsevents@^1.2.7:
     pino: ^7.2.0
     pino-logflare: ^0.3.12
     prettier: ^1.16.4
+    prop-types: 15
     pubsub-js: ^1.7.0
     react: ^17.0.2
     react-datepicker: ^2.9.6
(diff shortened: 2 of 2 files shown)


chrismclarke (changes requested):
So I don't think this is actually an issue that warrants fixing (unless you've seen it breaking something somewhere).

The problem is that we are using a lot of 3rd party libs, many of which (rightly) declare peer-dependencies. I.e. not every react plugin wants to declare a named version of react as a core dependency, but assumes the user installing the plugin has some version that they will be using.  For the specific case of `prop-types`, that's shipped as a dependency of `@types/react` (along with 15-20 other packages we use), so it's definitely in our node_modules and  really no need for us to explicitly declare it unless we are making direct reference in our code base (which I don't think we are).

This becomes more of an issue when using `pnp` as it expects flatter structures than the `node_module` resolution we have setup, but even in that case you can define in the `.yarnrc` default resolution for specific peer dependencies. 

If you're hoping to suppress some of the log outputs, a config I use in the `.yarnrc` for another project is:

```
# Ignore some common warnings https://yarnpkg.com/advanced/error-codes
logFilters:
  - code: "YN0076" # INCOMPATIBLE_ARCHITECTURE
    level: "discard"
  - code: "YN0060" # INCOMPATIBLE_PEER_DEPENDENCY
    level: "discard"
  - code: "YN0002" # MISSING_PEER_DEPENDENCY
    level: "discard"
```
The first filter ignores warnings caused by modules that ship both linux and windows build scripts (only the relevant is executed so the existence of the other really doesn't matter), the second mostly just comes from outdated packages that don't know whether they support new deps in use or not, and the last is the issue mentioned above.

Let me know if that sounds at all useful, or if you've found any specific errors thrown from the missing prop-types dep.

thisislawatts:
@chrismclarke no specific errors that I see being caused by it currently. Just attempting to reduce the number of warnings emitted by yarn. 

However I do not work with many yarn projects so I defer to your experience here. It seems odd to suppress their warnings but if it's baked into their configuration then that seems like an acceptable path forward. 

Can we bring this into version control so that all developers get a shared experience?

thisislawatts:
Ah yes, I see that we have .yarnrc.yml under git ✨

chrismclarke:
> It seems odd to suppress their warnings but if it's baked into their configuration then that seems like an acceptable path forward.

It's the same with npm installs, javascript ecosystems are a big sprawling mess, so many tools of varying code quality, pinning dependencies in different ways, varying update frequency and scope, quirks of different operating systems etc. Still, I'd take it over a `requirements.txt` file any day

PR #1572, opened 2022-03-27
feat: add country flag to research username

### LINKED ISSUE #1406: Add country flag to research username
### Description

Everywhere on the platform we show username with country flag (comments/how-to) This was never integrated in the research module. That is an issue. Not cool.

### Page related

Will be used in :

- [Research page](https://community.preciousplastic.com/research/make-things-from-ocean-plastic) where it should come
- This layout already used in [how-to's ](https://community.preciousplastic.com/how-to/make-christmas-ornaments-from-upcycled-bottle-caps)
- Previous PR introducing the flag https://github.com/ONEARMY/community-platform/pull/1033

### Mockup
Add flag research

jojudre:
PR Checklist

PR Type

## Description

add country flag to research username

## Git Issues

Closes #1406

## Screenshots/Videos

Снимок экрана 2022-03-27 в 23 20 57

---

## What happens next?


jojudre:
Hi! 
my account don't have a location/country, can't check case when user upload new research, to understand if flag is visible

davehakkens:
nicee! If you add yourself to the map you get a flag! 
afbeelding

jojudre:
@davehakkens can't save pin (
Снимок экрана 2022-03-27 в 23 59 10

davehakkens:
Did you fill in all fields (including contact details) if not you cant save

jojudre:
@davehakkens everything works correctly )

src/pages/Research/Content/ResearchItemDetail.tsx
```diff
@@ -27,7 +27,7 @@ const ResearchItemDetail = observer((props: IProps) => {
   }
 
   React.useEffect(() => {
-    ;(async () => {
+    (async () => {
```
thisislawatts, line 30: **praise:** Nice spot, thanks for tidying up! ✨

src/stores/Research/research.store.tsx
```diff
@@ -294,6 +294,17 @@ export class ResearchStore extends ModuleStore {
...
+          (values._createdBy && values._createdBy === user.userName) ||
+          !values._createdBy
+            ? user.location
+              ? user.location.countryCode
+              : user.country
+              ? user.country.toLowerCase()
+              : ''
+            : values.creatorCountry
+            ? values.creatorCountry
+            : '',
```
thisislawatts, line 307: **issue(blocking):** There are quite a few different logical flows here somewhat hidden by the shorthand ternary style.

Have you considered extracting this logic out to a help method which could then be put under test? At the moment writing tests for the stores is non trivial due to their dependencies. But if you introduced a new standalone file it should be straight forward enough to test the various scenarios handled by this code.

thisislawatts (changes requested):

src/stores/Research/research.store.tsx
jojudre: i took this logic from howTo store

jojudre: i have a question, what will happen when user adds map pin? every howtos, comment and research which was created by him should be update, do we have this implementation?

thisislawatts: It looks like there are some mechanisms in place to ensure data is updated, using serverless functions to make updates after the User object has been changed:
https://github.com/ONEARMY/community-platform/blob/master/functions/src/userUpdates/index.ts#L20

thisislawatts: At the moment it is only making updates to Howto Articles and Howto Comments

thisislawatts: > i took this logic from howTo store

This sounds like a very good reason to extract it out to a shared function and place it under test. Let me know if I can provide any guidance on adding a new test file. I think this is a helpful example of a lightweight unit test:
https://github.com/ONEARMY/community-platform/blob/master/src/utils/formatters.test.ts

jojudre: @thisislawatts thanks for helping! ^.^  i will try to make updates for research like in example for howto

PR #1583, opened 2022-04-01
chore: prevent redirect of static assets

chrismclarke:
PR Checklist

PR Type

## Description

The default behaviour in the platform hosting is to redirect any URLs that don't match an exact file to the main index.html. This is needed for us to handle spa routing for urls like `/howto` within the main app (via the react-router), because there is not /howto.html page). A weird caveat of this is fetching of static resources (like the main build js and CSS files), which firebase defaults to not redirecting IF AND ONLY IF the file exists. So any requests for JS/CSS that does not exist in the server is instead sent a copy of the index.html page, which will be stored in cache and throw an error when accessed. In production this can sometimes mean the user sees a blank white screen and the app won't load until the cache has had time to sort itself out (usually requiring the tab to be closed and reopened)

A larger question this opens is why does the site want to serve JS/CSS that no longer exists. I'm not sure the exact reason, but likely linked to a combination of imperfect caching of the main index.html file and required assets, and our general code-splitting strategy.

This is a small patch to try and prevent firebase from redirecting static assets to the main index.html file. It doesn't address the reason why this is happening, but hopefully should at least stop html being cached as JS (!)

## Review Notes
I've added a preview link, ideally it should show that trying to load a static file does not exist, e.g. `/static/js/doesnotexist.js`, results in a 404 page instead of redirect to home index like on the main site.

## Git Issues

Closes #

## Screenshots/Videos

Serving locally, missing static file returns 404 response
image

Existing site, missing static shows page redirect (to custom 404 page)
image

---

## What happens next?

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -9,13 +9,7 @@
       },
       {
         "source": "/static/**",
-        "destination": "/static/**",
-        "headers": [
-          {
-            "key": "Access-Control-Allow-Origin",
-            "value": "*"
-          }
-        ]
+        "destination": "/static/**"
       },
       {
         "source": "**",
--- a/firebase.json
+++ b/firebase.json
@@ -33,7 +33,7 @@
         ]
       },
       {
-        "source": "**/*.@(jpg|jpeg|gif|png)",
+        "source": "**/*.@(jpg|jpeg|gif|png|svg)",
         "headers": [
           {
             "key": "Cache-Control",
@@ -46,7 +46,7 @@
         ]
       },
       {
-        "source": "static/**/*.js",
+        "source": "static/**/*.@(js|css|map)",
         "headers": [
           {
             "key": "Cache-Control",
--- a/firebase.json
+++ b/firebase.json
@@ -8,6 +8,16 @@
         "function": "api"
       },
       {
+        "source": "/static/**",
+        "destination": "/static/**",
+        "headers": [
+          {
+            "key": "Access-Control-Allow-Origin",
+            "value": "*"
+          }
+        ]
+      },
+      {
         "source": "**",
         "destination": "/index.html",
         "headers": [
--- a/src/service-worker.ts
+++ b/src/service-worker.ts
@@ -35,7 +35,23 @@ precacheAndRoute(coreManifest)
 
 // static assets are already cachebusted with their file names so can just serve cacheFirst
 // for same origin (https://developers.google.com/web/tools/workbox/modules/workbox-routing)
-registerRoute(new RegExp('/static/'), new CacheFirst())
+registerRoute(
+  new RegExp('/static/'),
+  new CacheFirst({
+    plugins: [
+      // Allow static assets to be cached for up to 1 month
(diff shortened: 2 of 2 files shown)


firebase.json
```diff
@@ -7,6 +7,16 @@
         "source": "/api",
         "function": "api"
       },
+      {
+        "source": "/static/**",
```
chrismclarke, line 11: Adding this rule should mean it takes precedence over the default fallback (below), preventing static files from being redirected. Alternatively we could try modify the `**` rule to exclude `/static/` , but I'm not sure how well the glob pattern matching respects more complex negation patterns (e.g. `!(static)/**`)

https://firebase.google.com/docs/hosting/full-config#glob_pattern_matching

thisislawatts: @chrismclarke do we need the `headers` attribute defined on this rule? 

I am wondering if we can remove it so as not to introduce unnecessary configuration. This way we can defer to the `headers` rules list which sits at the same level as the hosting rules list.

chrismclarke: Really interesting catch. Short answer I think you should definitely be right, no reason to redefine what is already clearly defined in the regular hosting behaviour (so will remove and check live deploy preview still populates correct headers). Long answer/question is it makes me wonder what order the rules are respected for the other headers? I'd hope it starts with more specific `headers` block and then merges (but not overwrite) anything found in the `redirect` block, but might be worth a quick double-check (will do so now)

chrismclarke:
@thisislawatts 
I've made the update mentioned in the review comments. One thing I realised is that actually most of what is being set in headers is actually completely ignored because the service worker has its own set of rules that determine how long assets are cached for.

I don't think this actually let to the undesired behaviours we've been seeing (still place most the blame on the index redirect for static files), but it definitely does complicate/confuse. Whilst trying to dive more into it I've made a couple minor changes to the workbox sw config:

- Add an expiration for static files. This is set at a hefty 365 days given that all the assets are already cache-busted from their naming, so something like `a3a3sddef.js` we have little reason to try and purge. Whilst the core module is changed pretty much every update, some of the lazy-loaded feature modules aren't anywhere near as often so makes sense to me to have reasonably large. Again, I don't think this fixes any particular issues apart from perhaps small possibility if falling back to browser defaults maybe files could be evicted too early.

- Add a separate cache for map tiles. This doesn't do much but I became quickly aware that the (arbitrary) 1000 image limit on the image cache is quickly used up with map tiles. We could remove the limit although as I don't have anything sensible to suggest changing it to I think fine to leave for now

- Add SVG, CSS and map files to firebase hosting cache-control headers. Not sure why these were omitted, I expect small oversight (not that the headers do much with caching, but could be more useful if moving to ssr in the future)

The more I dig into the codebase the more I see need for a bigger revisit at some point, but hopefully for now this can at least help address the initial issue

firebase.json
chrismclarke: (update, the headers are passed as expected but all actually ignored because the service worker config likes to do its own thing instead...)

thisislawatts (approved):

PR #1608, opened 2022-04-11
fix: component props

chrismclarke:
PR Checklist

PR Type

## Description

Testing locally seems to be logging a lot of errors about invalid props. From what I can understand we use custom props like `<Text medium>` to generate styles, styled components/emotion should strip this out when compiling to underlying html element (e.g. a `<p>` or `<span>` tag, however as our Text components sits on top of another styled component (theme-ui text) this doesn't happen.

It may be this has always been the case and only recently has React started caring, or possibly the change from styled-components to emotion has lost some of the prop-stripping ability. In any case I've added a manual list of props to not pass for some of our most common components (which seem to be logging the most errors).

## Git Issues

Closes #

## Screenshots/Videos

image

---

## What happens next?

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -53,7 +53,11 @@ const translateY = (props: IBtnProps) =>
 
 export type BtnProps = IBtnProps & ThemeUiButtonProps
 
-const BaseButton = styled(ThemeUiButton)`
+const BaseButton = styled(ThemeUiButton, {
+  // avoid passing custom props
+  shouldForwardProp: (prop: keyof IBtnProps) =>
+    !['translateY', 'small', 'medium', 'large'].includes(prop),
+})`
   ${translateY}
   ${small}
   ${medium}
--- a/src/components/Text/index.tsx
+++ b/src/components/Text/index.tsx
@@ -1,5 +1,4 @@
-import { forwardRef } from 'react'
-import { Text as ThemeUiText, TextProps as ThemeUiTextProps } from 'theme-ui'
+import { Text as ThemeUiText } from 'theme-ui'
 import theme from '../../themes/styled.theme'
 import styled from '@emotion/styled'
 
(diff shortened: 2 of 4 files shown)


packages/components/src/Button/Button.tsx
```diff
@@ -53,7 +53,11 @@ const translateY = (props: IBtnProps) =>
 
 export type BtnProps = IBtnProps & ThemeUiButtonProps
 
-const BaseButton = styled(ThemeUiButton)`
+const BaseButton = styled(ThemeUiButton, {
+  // avoid passing custom props
+  shouldForwardProp: (prop: keyof IBtnProps) =>
+    !['translateY', 'small', 'medium', 'large'].includes(prop),
```
chrismclarke, line 59: remove the dynamic style generator props after generating styles (below)

src/components/Text/index.tsx
```diff
@@ -30,28 +28,28 @@ export interface ITextProps {
...
 }
 
-export const uppercase = (props) =>
+export const uppercase = (props: ITextProps) =>
   props.uppercase
-    ? {
+    ? ({
         textTransform: 'uppercase',
-      }
+      } as const)
```
chrismclarke, line 35: use `as const` to keep typescript happy (otherwise thinks passing the `textTransform` a `string` instead of specific, allowed value

src/pages/common/Layout/Main.tsx
```diff
@@ -9,32 +9,36 @@ interface ILayoutProps {
...
-          : {
-              flexDirection: 'column',
-              width: '100%',
-            }
-      }
-    >
-      {props.children}
+const Main = (props: IProps) => {
+  // avoid passing custom props
+  const { ignoreMaxWidth, customStyles, ...rest } = props
```
chrismclarke, line 14: As this `Main` component isn't wrapping a styled component include manual extraction of custom props to not pass to underlying Flex

thisislawatts (approved):

PR #1620, opened 2022-04-27
Feat/admin profile edit

chrismclarke:
PR Checklist

PR Type

## Description

- Add an admin-protected auth route that allows admins to edit a user profile page as the user
- Refactor passed shared AdminContact component to specific UserAdmin profile component
- Add edit button to user profile (viewable by admins)
- Add demo users to seed data to make it easier to add tests against specific user roles
- Add user profile test specs

## Review Notes
I corrected two issues whilst developing where first once updated the existing user would have their profile replaced with the user they were trying to edit (if going back to own settings page), and the second where the updated profile would have their information merged with the editing user. 

I haven't noticed any other issues but would be good to have a quick functional test in case there's anything I missed (see video below for testing I have done so far).

## Git Issues

Closes #

## Screenshots/Videos

Updated edit button on profile
image

Demo video

https://user-images.githubusercontent.com/10515065/165475401-8bb3f0bd-19db-4cdd-bd07-de45a317620f.mp4

---

## What happens next?


src/pages/User/content/SpaceProfile.tsx
```diff
@@ -35,14 +35,14 @@ import type { IConvertedFileMeta } from 'src/types'
...
 import UserContactAndLinks from './UserContactAndLinks'
+import { UserAdmin } from './UserAdmin'
 
 interface IBackgroundImageProps {
   bgImg: string
 }
 
 interface IProps {
   user: IUserPP
-  adminButton?: JSX.Element
```
chrismclarke, line 45: Previously admin-only buttons were passed as props, now just included in the profile render

packages/cypress/src/fixtures/seed/users.json
```diff
@@ -1,4 +1,55 @@
 {
+  "subscriber": {
+    "_authID": "a0dFrGVJTlQUA9BqH0QmnQM6flX2",
```
chrismclarke, line 3: The same users have been copied over from the mockAuthUsers in shared. Ideally at some point in the future we may want to do a better refactor and import direct, but beyond the scope of this PR

packages/cypress/src/integration/profile.spec.ts
```diff
@@ -0,0 +1,62 @@
+import USERS from '../fixtures/seed/users.json'
+import { UserMenuItem } from '../support/commands'
+
+describe('[Profile]', () => {
+  beforeEach(() => {
+    cy.visit('/')
+  })
+
+  const subscriber = USERS.subscriber
```
chrismclarke, line 9: Most of the other test specs use data copied as strings, but here we are directly pulling in from the seed json which feels quite a bit nicer. It may be worth tidying up other specs in a similar way in the future

src/pages/Settings/SettingsPage.tsx
```diff
@@ -27,7 +27,10 @@ import { toJS } from 'mobx'
 import { isModuleSupported, MODULE } from 'src/modules'
 import { logger } from 'src/logger'
```
chrismclarke, line 28: This file appears to show a much larger diff than there should be, so recommend viewing with whitespace ignored (either using the settings dropdown in file view mode, or appending `&w=1` to the url

src/pages/User/User.tsx
```diff
@@ -23,6 +25,14 @@ class UserPageRoutes extends React.Component<IProps, any> {
             <UserPage {...props} key={props.match.params.id} />
           )}
         />
+        {/* Allow admins to edit via userSettings as target user */}
+        <AuthRoute
+          roleRequired="admin"
+          path="/u/:id/admin"
```
thisislawatts, line 31: **suggestion:** Consider naming this route `/u/:id/edit`, this would align with the naming convention we have when user's edit a Research or Howto Resource. I do not think we want to explicitly encode the _role_ into the URL structure.

For example we may introduce a Community Manager role in the future, who are permitted to edit User Profiles but nothing more. In that the scenario it would become confusing to think of access to `/u/:id/admin` for users without the Admin role.

thisislawatts (approved):

src/pages/User/User.tsx
davehakkens: This would make sense to me as well @chrismclarke . 👆
Let me know what you think. After that happy to merge.

chrismclarke: That's actually how I originally specified the url, but then changed to /admin as it seemed weird that a user wouldn't be able to edit their own profile via the /edit link (that's currently handled by /settings - which is also super confusing because your settings are just your profile...)

But let's take a step towards consistency, I think /edit makes a lot of sense and we can just decide in the future if we also want to remove the /settings link and just have profile /edit (would be a bit of extra code tidying probably beyond the scope of this PR).

chrismclarke: Updated and merging

PR #1621, opened 2022-04-27
chore: introduce component scaffold

thisislawatts:
PR Checklist

PR Type

## Description

Two changes introduced as part of this PR:

1. Introduce new `yarn new-component MyNewComponent` to quickly scaffold out the files you need when creating a new component.
2. Restore theme support in Storybook so that is easier to review components using the various Platform styles.

Screenshot showing Button component with UI toggle to make it easy to jump between the different Themes.

Screenshot 2022-04-27 at 16 17 14


packages/components/package.json
```diff
@@ -6,12 +6,13 @@
...
-    "start": "start-storybook -p 6006",
-    "format": "prettier --write '**/*.{tsx,mdx,md}'",
+    "start": "start-storybook -p 6006 --loglevel verbose",
+    "format": "prettier --write '**/*.{ts,tsx,mdx,md}'",
     "build:sb": "yarn build && build-storybook",
     "build": "tsc",
     "dev": "tsc --watch",
-    "lint:code": "eslint . --ext .js,.jsx,.ts,.tsx src --color"
+    "lint:code": "eslint . --ext .js,.jsx,.ts,.tsx src --color",
+    "new-component": "ts-node scripts/new-component.ts"
```
chrismclarke, line 15: `ts-node` isn't currently a dependency so you will need to add (otherwise assumes global install). I'll add a quick commit as I don't know how that fancy `suggest change` thing works...

packages/components/README.md
```diff
@@ -24,3 +24,15 @@ cd ./packages/components
 yarn install
 yarn start
 ```
+
+## Creating a new Component
```
chrismclarke, line 28: We should probably migrate all this to the main docs at some point, but still a bonus to have here for now

packages/components/scripts/new-component.ts
```diff
@@ -0,0 +1,52 @@
...
+if (!componentName) {
+  console.log(`⚠️ Component name is required. Example usage:
+$ yarn new-component ComponentName`)
+  process.exit(1)
+}
+
+try {
+  mkdirSync(resolve(__dirname, `../src/${componentName}`))
+} catch (e) {
+  console.log(`${componentName} already exists`)
```
chrismclarke, line 23: Good idea to catch this

packages/components/package.json
thisislawatts: Good catch, thanks!

chrismclarke (approved):
I really like the idea of using templates to speed up the dev workflow, particularly given how much of this is fairly common boilerplate. I tried adding my own test component and all worked smoothly on the component front. It didn't live-reload as I might have expected in the story, but assume that might be a missing concurrent script to run both storybook and the tsc watch server (?). In any case I'd say good to go from here and we can debug the rest later.

As for the theming side - I don't really understand what did/didn't work with theming, but happy to hear someone else does!

chrismclarke:
(live reload was down to poor choice of component names on my part, now working as expected)

PR #1629, opened 2022-04-27
fix: documentation deps

chrismclarke:
PR Checklist

PR Type

## Description

It appears that the updates made to docusarus have unwanted knock-ons across the rest of the platform. This is due to the fact it is based in react and has no hoisting limits imposed, so will override similar packages used at main platform level.

Longer term likely we will need some better means of isolating the different react-based platforms as conflicts seem pretty common across docusaurus, storybook and the main platform src

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?


packages/documentation/docs/Backend Development/firebase-emulators-docker.md
```diff
@@ -61,7 +61,7 @@ You can see the version of data used in the command line output, e.g. data expor
 
 ### User Login
 
-By default the image comes preloaded with user auth accounts as found in [shared\mocks\auth-users.ts](../../../../shared/mocks/auth-users.ts). This means you can login as any of these users, e.g.
+By default the image comes preloaded with user auth accounts as found in [shared\mocks\auth-users.ts](https://github.com/ONEARMY/community-platform/tree/master/shared/mocks/auth-users.ts). This means you can login as any of these users, e.g.
```
chrismclarke, line 65: docs changes aren't part of the fix, just was in my workspace so made sense to commit here

package.json
```diff
@@ -151,7 +151,7 @@
     "@types/jest": "^26.0.15",
     "@types/node": "16",
     "@types/pubsub-js": "^1.5.18",
-    "@types/react": "^16.9.53",
+    "@types/react": "17.0.44",
```
chrismclarke, line 154: The type update appears to flag a lot more errors (non-breaking, mostly cases where render functions return html instead of a react fragment which seems to still be handled fine but is technically an error). Hopefully react 18 has some better handling for these

.circleci/config.yml
```diff
@@ -457,7 +457,7 @@ workflows:
           name: 'Deploy: fixing-fashion-dev.web.app'
           requires:
             - test_e2e
-          # <<: *filter_only_master
+          <<: *filter_only_master
```
chrismclarke, line 460: assumed this was missing from the config (saw attempted deployment to fixing fashion on regular prs)

package.json
chrismclarke: (nvm - breaks too many things so reverting back)

.circleci/config.yml
thisislawatts: This makes sense and would be good to fix asap, for that reason I would move this commit into a standalone PR to decouple it from the other changes being made here.

thisislawatts: Raised: https://github.com/ONEARMY/community-platform/pull/1630

chrismclarke: Thanks for doing that, I'll do the same for the docs changes and likely close this in favour of opening something cleaner

chrismclarke:
closing in favour of small fixes elsewhere

PR #1663, opened 2022-05-14
feat: drop Text component in favour of Text component

thisislawatts:
PR Checklist

PR Type

## Description

Removes our own `Text` component in favour of directly using the `Text` component from [theme-ui](https://theme-ui.com/). Our implementation was a thin wrapper that introduced a few attributes like `txtCenter` which seems like it would be better to set via the `sx` attribute.

This introduces changes like the following:

```tsx
// Before
<Text txtCenter>I am text</Text>

// After
<Text sx={{textAlign: 'center'}}>I am text</Text>
```
 
 There are no user facing changes here, the intention here is to simplify the experience when building UI components. Before there were two possible options for including `Text` now there is only one.


davehakkens (commented):
looks good here!

thisislawatts:
@davehakkens Okay to approve these changes?

davehakkens (approved):

chrismclarke:
Thanks for tidying this, I definitely waas not a fan of the layers of minor wrappers (anything we need to configure with text should be themeable anyways, and also was leading to a bunch of knock-ons passing props like `center` down which react complains about). I can't remember if we fixed the prop-drilling issue or just avoided react upgrade, but in either case this will definitely help

PR #1680, opened 2022-05-18
fix: remove service-worker caching for index (WHITE SCREEN OF DEATH 🔥👻)

thisislawatts:
PR Checklist

PR Type

## Description

The service worker should not cache the index.html page as
this is the entry point to our service which contains
the manifest for which JS files to load.

After each deployment, the old JS files are removed from the
server. As a result the index.html loaded from cache by the service
worker can include references to JS files which no longer exist.

Screenshot 2022-05-18 at 19 51 00

Q. How can I verify that a stale HTML file was the root cause?
A. I checked the `REACT_APP_PROJECT_VERSION` variable injected during build time and observed that the sha 0a2475d054a016d5ed3642c6abc299978c9ace12 was _several_ deployments behind the head when compared with GitHub.


thisislawatts:
@chrismclarke I welcome your perspective here! I am still young in the ways of Service Worker(TM)

chrismclarke:
Haha, yeah I might need to revisit this one next week. It's a perennial issue, I've been forever subscribed to the thread `https://github.com/facebook/create-react-app/issues/5316` and saw a couple new posts over the weekend with some more ideas.

My other thought would be to look at including a specific within the error boundary (although not sure if the error boundary would load if main chunk failing...)

thisislawatts:
That would be great @chrismclarke, it would be great to see away with this screen of doom.

davehakkens:
Sounded interesting! 
But I do think we got the "Report Feedback" page more often, as a comment from users.

PR #1687, opened 2022-05-23
1617-add-download-count-howto

AlfonsoGhislieri:
PR Checklist

PR Type

## Description

I have added a total downloads counter feature. When the download file button is clicked this increments the total_downloads property of the howto model and also updates state inside of the FileInfo component in order to re-render the newly incremented value. 

 - [x] Implemented the logic to not allow users to repeatedly click the button and inflate the downloads counter. I used local storage with a 12 hour cooldown to not let users spam click download. 
  - [x] Styling fixed

BUG: I am also aware that since older howto's don't have a total_downloads property then they currently just show 'downloads' without a value. I didn't implement a conditional check because I was thinking we would update all howto's to have the total_download property if they have an attached file? Or is this feature not going to be retroactive and only affect newly created howto's?

Potential issue with using LocalStorage: a malicious user could find remove the cooldown manually and spam download.

### DIFF
--- a/src/components/FileInfo/FileInfo.tsx
+++ b/src/components/FileInfo/FileInfo.tsx
@@ -4,17 +4,11 @@ import type { IUploadedFileMeta } from 'src/stores/storage'
 import { FileDetails } from './FileDetails'
 import { ExternalLink } from 'oa-components'
 import styled from '@emotion/styled'
-import { useCommonStores } from 'src/index'
 
 interface IProps {
   file: File | IUploadedFileMeta | null
   allowDownload?: boolean
-  howToID?: string
-  setFileDownloadCount?(val: number | undefined): void
-}
-
-interface localStorageExpiry {
-  expiry: number
+  handleClick?: () => Promise<void>
 }
 
 const FileContainer = styled.a`
@@ -24,10 +18,8 @@ const FileContainer = styled.a`
 export const FileInfo: React.FC<IProps> = ({
   file,
   allowDownload,
-  howToID,
-  setFileDownloadCount,
+  handleClick,
 }) => {
-  const { stores } = useCommonStores()
   const meta = file as IUploadedFileMeta
   const size = file ? bytesToSize(file.size) : '0'
 
@@ -35,49 +27,15 @@ export const FileInfo: React.FC<IProps> = ({
     return null
   }
 
-  // not sure where to store these functions, if here is ok?
-  const setCooldownWithExpiry = () => {
-    const now: Date = new Date()
-    const twelveHoursInSeconds: number = 12 * 3600
-    const expiry = { expiry: now.getTime() + twelveHoursInSeconds }
-    localStorage.setItem('downloadCooldown', JSON.stringify(expiry))
-  }
-
-  const checkForExpiry = () => {
-    const downloadCooldown: string | null =
-      localStorage.getItem('downloadCooldown')
-
-    if (typeof downloadCooldown === 'string') {
-      const downloadCooldownParsed: localStorageExpiry =
-        JSON.parse(downloadCooldown)
-
-      const now: Date = new Date()
-      // remove localstorage item if its timestamp is expired
-      if (now.getTime() > downloadCooldownParsed.expiry) {
-        localStorage.removeItem('downloadCooldown')
-      } else {
-        return true
-      }
-    }
-  }
-
-  const handleClick = async () => {
-    if (howToID && setFileDownloadCount && !checkForExpiry()) {
-      const updatedDownloadCount =
-        await stores.howtoStore.incrementDownloadCount(howToID)
-      setFileDownloadCount(updatedDownloadCount)
-      setCooldownWithExpiry()
-    }
-  }
-
   return (
     <>
-      {allowDownload && meta.downloadUrl ? (
+      {allowDownload && meta.downloadUrl && handleClick ? (
         <ExternalLink
           m={1}
           onClick={() => handleClick()}
           href={meta.downloadUrl}
           download={file.name}
+          a
           style={{ width: '300px' }}
         >
           <FileDetails file={file} glyph="download-cloud" size={size} />
--- a/src/components/FileInfo/FileInfo.tsx
+++ b/src/components/FileInfo/FileInfo.tsx
@@ -11,10 +11,6 @@ interface IProps {
   handleClick?: () => Promise<void>
 }
 
-const FileContainer = styled.a`
-  width: 300px;
-  margin: 3px 3px;
-`
 export const FileInfo: React.FC<IProps> = ({
   file,
   allowDownload,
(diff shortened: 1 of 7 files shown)


davehakkens:
nice one! Local storage with cooldown sounds good to me, but not an expert in the technical side.

>  I was thinking we would update all howto's to have the total_download property if they have an attached file?
That sounds good, we want it retroactive. Looking forward to test.

AlfonsoGhislieri:
Just pushed the newest commit which has the cooldown implemented, should all be good to go.

AlfonsoGhislieri:
Will sort out the styling! And sure happy to work on that after this is all finished up 💯

src/components/FileInfo/FileInfo.tsx
```diff
@@ -3,28 +3,77 @@ import { bytesToSize } from '../ImageInput/ImageInput'
...
+}) => {
+  const { stores } = useCommonStores()
   const meta = file as IUploadedFileMeta
   const size = file ? bytesToSize(file.size) : '0'
 
   if (!file) {
     return null
   }
 
+  // not sure where to store these functions, if here is ok?
```
thisislawatts, line 37: ✨ A good question. I would prefer to see this behaviour lift out of FileInfo and instead be implemented within the `HowtoDescription` component.

The FileInfo component should be generic and available for reuse in multiple places across the system. Introducing the `howToID` into the props is a signal that it knows to much about the outside world. 

Instead I think we should extend the `FileInfo` component so that it supports an `onClick` function provided via a prop and then we can lift all of our Howto specific logic out to HowtoDescription. What do you think?

src/components/FileInfo/FileInfo.tsx
```diff
@@ -3,28 +3,77 @@ import { bytesToSize } from '../ImageInput/ImageInput'
...
+  const setCooldownWithExpiry = () => {
+    const now: Date = new Date()
+    const twelveHoursInSeconds: number = 12 * 3600
+    const expiry = { expiry: now.getTime() + twelveHoursInSeconds }
+    localStorage.setItem('downloadCooldown', JSON.stringify(expiry))
+  }
+
+  const checkForExpiry = () => {
```
thisislawatts, line 45: **suggestion:** Consider refactoring this method so that it does one thing only or renaming it to better communicate the behaviour. At the moment it is checking whether the method is expired **and** then removing expired items. 

There are three possible flows through this function but there is only instance where we explicitly return a value.

thisislawatts (changes requested):

src/components/FileInfo/FileInfo.tsx
AlfonsoGhislieri: Will definitely refactor, going to move everything onto `howToDescription`  and refactor in there!

AlfonsoGhislieri: @thisislawatts Ok well what started as a minor refactor has turned out into me redoing most of the logic to make sure everything would work as intended and is future proof. Local storage now has an array that stores objects with a download cooldown for specific download cooldowns. 

I pushed everything up like we discussed and out of `fileInfo.` I also created a `utils` file for `howToDescription` to avoid bloating the actual original file too much since I had to create a lot of extra functions for the new implementation.

Let me know what you think!

src/stores/Howto/howto.store.tsx
```diff
@@ -334,6 +350,7 @@ export class HowtoStore extends ModuleStore {
         steps: processedSteps,
         fileLink: values.fileLink ?? '',
         files: processedFiles,
+        total_downloads: processedFiles ? 0 : undefined,
```
thisislawatts, line 353: **question:** Why would we want `total_downloads` to be undefined? We only show this indicator if there are files available to download.

thisislawatts (changes requested):

thisislawatts:
Thanks for the work @AlfonsoGhislieri looking really good! I have added a few comments/suggestions but otherwise I think we're almost there. I have noticed that some of this GitHub actions have been failing without reason recently, will need to dig a little further into those.

src/stores/Howto/howto.store.tsx
AlfonsoGhislieri: Well the anwser to this is pretty simple and silly, when I started working on the issue I was still rather new to using typescript and I think that was just the easiest way to get it to work from the get go. I never revisted it after... 

First solution I could think is to just add to the total_downloads property to the howto model only if a file exists in the model.

https://github.com/ONEARMY/community-platform/pull/1687/commits/fb12a0712b44b23b408bfda5fd19c1a13ded0978#diff-a2437a58b8d602deb7c8d0b921e348a6c2c561c7af93bcaeb0a13833b0e9541eR365

AlfonsoGhislieri:
@thisislawatts thanks for all the suggestions! Should all be implemented and now the old howto's are fully functional too! 👍

src/pages/Howto/Content/Howto/HowtoDescription/HowtoDescription.tsx
```diff
@@ -245,8 +302,23 @@ export default class HowtoDescription extends PureComponent<IProps> {
                   allowDownload
                   file={file}
                   key={file ? file.name : `file-${index}`}
+                  handleClick={this.handleClick}
```
thisislawatts, line 305: **issue:** Note that there are another flow here which need to be included when the download is an external resource:

```
{howto.fileLink ? <DownloadExternal link={howto.fileLink} /> : ''}
```

Apologies for missing this in the previous review

src/components/FileInfo/FileInfo.tsx
```diff
@@ -19,9 +24,10 @@ export const FileInfo: React.FC<IProps> = ({ file, allowDownload }) => {
 
   return (
     <>
-      {allowDownload && meta.downloadUrl ? (
+      {allowDownload && meta.downloadUrl && handleClick ? (
```
thisislawatts, line 27: **Issue:** This conditional check has become stricter, however `FileInfo` is a generic and reusable component so there are instances which will not pass handleClick. This represents a breaking change for those usecases.

Instead we should consider removing the conditional check change and making the onClick binding stricter, like `onClick={() => handleClick && handleClick()}`, this will only execute an onClick method if one is passed through as a property.

thisislawatts (changes requested):

src/pages/Howto/Content/Howto/HowtoDescription/HowtoDescription.tsx
AlfonsoGhislieri: No worries, will add that in straight away! Didn't realize it had been added already!

src/components/FileInfo/FileInfo.tsx
AlfonsoGhislieri: Makes sense! Often forget to keep components reusable my bad..

thisislawatts (approved):

PR #1699, opened 2022-05-30
feat: drop duplicate heading component

thisislawatts:
PR Checklist

PR Type

## Description

The aim here is to drop our own Heading component
and instead reuse the primitive provided by `theme-ui`

The significant change here is that we will be adopting a
semantically meaningful `h2` tag in our pages. This is a
really positive step forward for our pages in terms of providing
structured content to better improve accessibility for humans and
robots (SEO)

The current `Heading` component has three variants, small,medium,large.
The large variant is currently unused in the system so is not being
replaced at this time. Reviewing all instances of <Heading> the medium
variant is used most often, so for this reason we shall adopt this as
the Default style. The only place where the variant is not specified is
within the admin pages so there are no user facing implications of
changing this style.

This commit also introduces a Story which captures the default and small
variant heading styles.

Rendering of story


davehakkens (approved):
Nice one! I expected to find at least one but I literally cant see any visual change. 👍

PR #1701, opened 2022-05-31
feat: consolidate inputs

thisislawatts:
- fix: extend theme to define heading styles
- feat: remove duplicate heading component
- feat: migrate to documented components

PR Checklist

PR Type

## Description

Refactor our input elements to:
1. Extract them to our components package, with a single component per file
2. Introduce documentation for the extracted components
3. Use primitives from theme-ui over our own implementation.
4. Drop 3rd party date-picker in favour of the standard browser implementation for the date

**Before:** with 3rd party date picker

  **After:** implementation in MacOS Chrome


davehakkens:
Date inputfield looks good.
But seems like all input fields became round?

afbeelding

Same for search bar on how-to 
afbeelding

Comments
afbeelding

Profile
afbeelding

thisislawatts:
@davehakkens were you running this locally? If so which command did you use to start the service?

davehakkens:
Yarn start, firefox

thisislawatts:
@davehakkens, spotted the issue. The form components are now loading their styling from the theme, this had been done for Precious Plastic but not the Fixing Fashion instance.

davehakkens (approved):
Well done!

chrismclarke:
Really nice to keep removing these custom component wrappers and also start documenting what we do have. 

I assume the issue of styling has been resolved for now, but if moving forward we do find a bit more inconsistency than we like there are tools like https://github.com/merelinguist/styled-preflight (built off normalise.css) which might be able to resolve some minor issues (although I think more with safari vs other browsers which I know we don't do much explicit testing for anyways).

PR #1726, opened 2022-06-05
chore: reorganize components

thisislawatts:
PR Checklist

PR Type

## Description

- move VirtualizedFlex out to howto section
- move GoogleAnalytics out to global component
- move ExternalEmbed into Academy
- move EventCard into pages/Events
- drop DiscordLink as it is not a valid component
- move DevSiteHeader out to global element
- move howto categories under Howto **section**


src/pages/Events/Content/EventsList/EventsList.tsx
```diff
@@ -4,7 +4,7 @@ import { Button } from 'oa-components'
 import { Flex, Box, Link } from 'theme-ui'
 import MoreContainer from 'src/components/MoreContainer/MoreContainer'
 import { Heading } from 'theme-ui'
-import EventCard from 'src/components/EventCard/EventCard'
+import EventCard from 'src/pages/Events/EventCard/EventCard'
```
chrismclarke, line 7: nit (non-blocking) - depending on the answer to comment below this could possibly be a relative import

src/pages/Events/EventCard/EventCard.tsx
```diff
@@ -3,9 +3,9 @@ import { Card, Text, Flex, Link } from 'theme-ui'
 import { Button } from 'oa-components'
 import TagDisplay from 'src/components/Tags/TagDisplay/TagDisplay'
 import { FlagIconEvents } from 'oa-components'
-import type { IEvent } from '../../models/events.models'
+import type { IEvent } from '../../../models/events.models'
 import { getMonth, getDay, capitalizeFirstLetter } from 'src/utils/helpers'
-import { VerifiedUserBadge } from '../VerifiedUserBadge/VerifiedUserBadge'
+import { VerifiedUserBadge } from '../../../components/VerifiedUserBadge/VerifiedUserBadge'
```
chrismclarke, line 8: nit (non-blocking) - I'm guessing VSCode has updated imports following the convention it started with (e.g. relative or absolute), although that starts to look a bit ugly for deeply nested imports so might be nice to push for a bit more consistency

I'm guessing the best way to be consistent would be to say always use absolute imports (and possibly get a lint rule to help) as I'm guessing that will also make life easier when refactoring to external components package, but I don't have anything against using relative imports for things like page-specific components (albeit a bit harder to automate checks for)

Any thoughts or preferences @thisislawatts ?

thisislawatts: My preference here would be remove the absolute imports for relative, but that is only due to a lack of familiarity with code bases that use absolute import paths. 

However it's not something I had been thinking about. The point you raise about consistency and automated linting is a good one. I think it would be worth raising an Issue to capture this item.

chrismclarke: Fair enough - yeah it's a quirk of working originally with webpack systems but now also typescript supports out of the bag with project references so that you can just say 'src' and at build time it will replace with whatever relative path. 

In any case given the mess I'm starting to make in other branches let's get this merged in and create a future issue

PR #1735, opened 2022-06-06
chore: remove legacy app-engine code

chrismclarke:
PR Checklist

PR Type

## Description

The way we currently handle taking backups on the PP site is through a weekly task defined in functions\src\exports\tasks.ts and triggered using google cloud pubsub. The pubsub integration was manually setup right at the beginning of the project, and would not be automatically managed for new deployments

The only reason we did things in this way was because cloud functions previously never had support for defining cron triggers, so we had created the functions-cron folder which created a google app-engine instance (which does have cron) to spin-up and publish a cloud pub-sub message on a weekly and daily basis which the functions could be triggered by. In order to trigger on a different project we would need to deploy new app engine instances and add the pubsub events manually through the cloud console page.

Firebase finally added support for scheduled functions a bit over a year ago so actually all this system can finally be thrown out and replaced by much simpler functions (that would also get copied over to any other instance we deploy to). 

Changes:

- Remove legacy functions-cron code and readme notes
- Refactor existing pubsub daily and weekly functions to use scheduled functions
- Rename folder containing scheduled functions to be called `scheduled` (previously `exports`)

## Review Notes
The only real way to test is to deploy to staging and check at least the daily tick is triggered successfully (scheduled for 2am UTC daily). 

There is a chance deployment will throw an error if specific apis need to be enabled on google cloud, e.g. [pubsub API](https://console.cloud.google.com/apis/library/pubsub.googleapis.com), or if the existing firebase service account needs additional permissions (there have been quite a few changes over the years so hard to know for sure what is/isn't included by default or within our current setups) 

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?


functions/src/scheduled/tasks.ts
```diff
@@ -0,0 +1,25 @@
...
+Scripts executed on a cron schedule, such as daily or weekly
+https://firebase.google.com/docs/functions/schedule-functions
+************************************************************************************/
+
+import * as functions from 'firebase-functions'
+import { BackupDatabase } from '../Firebase/databaseBackup'
+import * as FirebaseSync from '../Firebase/firebaseSync'
+
+/** Trigger tasks weekly on Sunday at 2am */
+export const weeklyTasks = functions.pubsub
```
chrismclarke, line 11: This code had been mostly just copied from the removed `src/exports/tasks.ts` file, git just doesn't track the changes fully

functions/src/scheduled/tasks.ts
```diff
@@ -0,0 +1,25 @@
...
+  .schedule('0 2 * * SUN')
+  .onRun(async (context) => {
+    console.log('weekly tick', context)
+    const backupStatus = await BackupDatabase()
+    console.log(backupStatus)
```
thisislawatts, line 16: **thought:** I think there is probably value in this being emitted to a Slack channel as a signal of a healthy system. Not something to be done as part of this PR but a useful enhancement going forward.

.eslintignore
```diff
@@ -1,6 +1,5 @@
 node_modules
 functions
-functions-cron
```
thisislawatts, line 3: **praise:** It is really good to clean this directory out of the root. Very happy to see this change!

thisislawatts (approved):

functions/src/scheduled/tasks.ts
chrismclarke: agreed, would be nice if there were some sort of integration that would at least pass any errors flagged from the functions to some channel somewhere (pretty sure these logs have no value in any terms, but I'll leave for now to at least check output for first run)

PR #1715, opened 2022-06-05
feat: drop Avatar component

thisislawatts:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Developer experience (improves developer workflows for contributing to the project)

## Description

The `Avatar` component seemed to duplicate the MemberBadge component, so extended `MemberBadge` to capture the missing behaviour and dropped `Avatar` entirely. 

Closes https://github.com/ONEARMY/community-platform/issues/1714

PR #1736, opened 2022-06-07
Feat/lazy images

chrismclarke:
PR Checklist

PR Type

## Description

On most pages we use image that load eagerly (the moment the page loads images try to load even when not on screen). This is expensive on bandwidth and also can result in a lot of network requests that impact on SEO (e.g. googlebot limits 25 resources and may miss out on key js/css if loading lots of images). We use a virtual scroller to limit this impact on the main howto list page, however individual howto items can also often contain 20+ images so good to have here too

This PR tries to add a `loading='lazy'` strategy for all images that could be found (including ones generated through styled components as `styled(image)`

## Review Notes
@thisislawatts - I searched for `<Image` to match all theme-ui and vanilla image tags, as well as `styled(image)`, but let me know if you think there might be any more image tags I could have missed

@davehakkens - The idea here is that the images only load just a little before they are scrolled into view, and so the user should not notice that they are not loaded. But it would be good if you could take a quick browser on the preview site in case you notice any pages where the images appear to load extra slowly

## Git Issues

Closes #

## Screenshots/Videos

**Before** - all images loaded on page load

https://user-images.githubusercontent.com/10515065/172474119-765d1f78-3129-4bd8-a5a5-ba21185d32df.mp4

**After** - images are loading on demand (while scrolling)

https://user-images.githubusercontent.com/10515065/172473799-95837434-36f5-41cb-8cc2-6232e8a417ac.mp4

---

## What happens next?

### DIFF
--- a/src/components/Avatar/index.tsx
+++ b/src/components/Avatar/index.tsx
@@ -76,6 +76,7 @@ export class Avatar extends Component<IProps, IState> {
     return (
       <>
         <Image
+          loading="lazy"
           className="avatar"
           sx={{ width: width ? width : 40, borderRadius: '25px' }}
           height={width ? width : 40}
--- a/packages/components/src/MemberBadge/MemberBadge.tsx
+++ b/packages/components/src/MemberBadge/MemberBadge.tsx
@@ -14,6 +14,7 @@ export const MemberBadge = (props: Props) => {
 
   return (
     <Image
+      loading="lazy"
       className="avatar"
       sx={{ width: size ? size : 40, borderRadius: '50%' }}
       height={size ? size : 40}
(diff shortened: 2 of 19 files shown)


src/components/Avatar/index.tsx
```diff
@@ -76,6 +76,7 @@ export class Avatar extends Component<IProps, IState> {
     return (
       <>
         <Image
+          loading="lazy"
```
thisislawatts, line 79: **nitpick:** This is not needed, we are looking to drop the Avatar component entirely in #1715

thisislawatts (approved):

davehakkens (approved):
makes sense and looks good!

PR #1740, opened 2022-06-08
Feat/seo clientside tags

chrismclarke:
PR Checklist

PR Type

## Description

The platform is limited in terms of SEO as it is a single page app, so whatever page a searchbot navigates to it will see the same metadata tags stored in the page's index.html `<head` section.

This PR adds support for dynamically updating those tags depending on content, specifically:

- Add methods to populate SEO tags dynamically from client-side code
- Integrate howto SEO tags (using howto title, description and feature image)
- Integrate top-level route SEO (names of main pages, e.g. map, how-to, research)

This is still unlikely to provide the correct metadata to searchbots for cases like social sharing as most bots simply read the tags once and do not wait for dynamic update, however it is possible that using a service like prerender may cache the correct updates.  

## Git Issues

Closes #

## Screenshots/Videos

Navigating the site now updates top URL bar. Individual howtos now also populate custom metadata

https://user-images.githubusercontent.com/10515065/172537721-0a793d16-971e-4dc2-9217-92f9d8bc53a0.mp4

Back button menu (not shown in video)

image

Putting the preview site into https://seositecheckup.com/analysis shows that maybe the google search meta might come through (won't be able to test on real Googlebot until released onto a domain that we have proper access to)

image

Using facebook debugger shows that it will not populate dynamic content when social sharing on whatsapp/facebook
https://developers.facebook.com/tools/debug/

image

---

## What happens next?


src/pages/Howto/Content/Howto/Howto.tsx
```diff
@@ -122,10 +123,18 @@ export class Howto extends React.Component<
   public async componentDidMount() {
     const slug = this.props.match.params.slug
     await this.store.setActiveHowtoBySlug(slug)
+    seoTagsUpdate({
```
chrismclarke, line 126: dynamic content (e.g. howtos) can also be updated using data retrieved from the db. Similar code could be used for other modules

src/pages/Howto/Content/Howto/Howto.tsx
```diff
@@ -122,10 +123,18 @@ export class Howto extends React.Component<
...
+      title: this.store.activeHowto?.title,
+      description: this.store.activeHowto?.description,
+      imageUrl: this.store.activeHowto?.cover_image.downloadUrl,
+    })
     this.setState({
       isLoading: false,
     })
   }
+  public async componentWillUnmount() {
+    seoTagsUpdate({})
```
chrismclarke, line 136: I've added this just to ensure any howto metadata is removed when navigating away (although it should actually be overwritten anyways by the general router implementation), so possibly overkill

src/pages/index.tsx
```diff
@@ -68,6 +69,7 @@ export class Routes extends React.Component<
                     key={page.path}
                     render={() => (
                       <React.Fragment>
+                        <SeoTagsUpdateComponent title={page.title} />
```
chrismclarke, line 72: The router component handles updating seo for top-level routes (e.g. map, how-to, research, academy)

thisislawatts:
@chrismclarke you should be able to test out how this would be interpreted by prerender.io using the following curl command

```sh
curl -H "X-Prerender-Token: <YOUR\_PRERENDER\_TOKEN>" https://service.prerender.io/https://www.example.com/
```

chrismclarke:
More details can also be seen from the pre-render console
https://dashboard.prerender.io/

image

davehakkens (approved):
Looks good to me. Nice work! 🔥🔥🔥

src/pages/Howto/Content/Howto/Howto.tsx
thisislawatts: @chrismclarke can we add a Cypress test to validate the these tags have been populated?

thisislawatts (changes requested):

chrismclarke:
@thisislawatts 
Just added an seo spec for howtos to check metadata populated. I haven't added a general one for the rest of the pages as I don't think the code currently works how we want it to (default hardcoded SEO values are for precious plastic site only, but likely we will want to improve to have different values depending on site/theme). 

We can either decide whether to try include here before merge (alongside any additional seo for things like research, map, etc.), or leave for follow-up. My preference would be to get merged as I think the current improvements help unblock work integrating prerender

thisislawatts (approved):

PR #1650, opened 2022-05-03
fix: responsive comment component

hunjunChoi:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)
- [x] - Latest `master` branch merged

PR Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

- Made responsive comment component of `how-to` page
- Page was tested in the dimension 401x853

## Review Notes

The issue marked two pages, `research` and `how-to`, to solve the bug, but I noticed that the comment component of the `research` page has been corrected already. 

## Git Issues

Closes [one](https://github.com/ONEARMY/community-platform/issues/1610#issuecomment-1100616771) of #1610 

### Appearance before the adjustment
image

### After the change took effect 
image

---

## What happens next?

Thanks for the contribution! We try to make sure all PRs are reviewed ahead of a monthly dev call (first Monday of the month, open to all!).

If the PR is working as intended it'll be merged and included in the next platform release, if not changes will be requested and re-reviewed once updated.

If you need more immediate feedback you can try reaching out on slack in the `platform-dev` channel.

PR #1742, opened 2022-06-08
feat: extract create comment component

### LINKED ISSUE #1719: Help to migrate src/components/Comment/CommentTextArea.tsx
We are looking to migrate `src/components/Comment/CommentTextArea.tsx` to `packages/components/src/CommentTextArea/CommentTextArea.tsx`

**Why?**

-   Improve structure of the components
-   Improve documentation for our existing components
-   Make it easier for our design team to review and refine the components
-   Decouple our components from application state
-   Convert our class based components to functional components
-   Allow reuse of our components outside of the primary single page application (SPA)#

**Definition of Migrated**

-   The component file exists in `packages/components/src`
-   There is a companion `*.stories.tsx` which includes an example for each use case of the component

**How?**

1.  [Create a new component](https://github.com/ONEARMY/community-platform/tree/master/packages/components#creating-a-new-component) in `./packages/components`
2.  Move existing component into new component
3.  Update all references to import the new component from `oa-components`
4.  Delete all files from `./src/components`

🐉 **Things to watch out for:**

-   If the original component file exported _multiple_ components these should be split into multiple separate files
-   If the component you're working on imports another component from `src/components` you will need to migrate that item first. Please make sure to claim ownership on the associated issue.
-   We want to use functional components, [read more around how to refactor a class component](https://nimblewebdeveloper.com/blog/convert-react-class-to-function-component)
-   You may spot an opportunity to create a new component to encapsulate some elements which are being used frequently across the UI. Yay, a new component is born 🐥
- The component may not need to exist.

thisislawatts:
PR Checklist

PR Type

## Description

Refactors existing `CommentTextArea` component into shared
`CreateComment` component which groups the user avatar and
submission button.

This change also improves the mobile rendering which removes
the need for #1650.

Whilst I think there is still further improvement to be made
on this component for a mobile user that can be done in a
future iteration with oversight from the Design team.

Closes #1719 

Screenshot 2022-06-08 at 14 28 13
Screenshot 2022-06-08 at 14 28 40


davehakkens (approved):
nice. Only small detail I would have is:
Would visually be nice if the font-family + font-size from the `CommentTextArea` is the same layout as the body text from the comments.

But also happy to merge like this if it makes things complex

thisislawatts:
@davehakkens should we align all of out inputs to match this larger style? For example the text inputs within the Settings page. Ideally we should aim for consistency across the Platform so users can easily identify input elements from non interactive components.

davehakkens:
Yeh I think its years of different developers and not super strict design guidelines (focus on features first) that make it a bit inconsistent.  How it actually should be is that the body of the comments should have the same font (`Inter`) as the general body text and input fields. Not what there is currently (`Varela-Round`) Varela was not for body text, just titles and headers.

Would be nice to sync everything but it would require some deeper design work to align over the entire platform.
Visually not super needed at this point unless it makes it easier to develop or migrate components then it might be worth it.
Let me know if you think it makes sense then I can have a look at it

Issue #1729, opened 2022-06-05
[bug] Profile image upload doesnt seem to work? (both member and space) 


thisislawatts:
**Describe the bug**

It is not possible to upload a Profile image for either a member or space user type.

**To Reproduce**
Steps to reproduce the behaviour:

**Expected behaviour**

A user should be able to upload a Profile image

AlfonsoGhislieri:
Also happy to pick this one up

PR #1747, opened 2022-06-10
1729 profile image upload doesnt seem to work

AlfonsoGhislieri:
PR Checklist

- [ x ] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [  x ] Bug fix (non-breaking change which fixes an issue)

## Description

_What this PR does_

Fixed the bugs that caused error messages when trying to edit profile settings, added correct badge svgs to each styled theme. Also fixed right badges displaying depending on correct theme. 

## Git Issues

Closes #

https://github.com/ONEARMY/community-platform/issues/1729

## Screenshots/Videos

Screenshot 2022-06-10 at 16 24 05

## What happens next?

### DIFF
--- a/src/modules/profile/SupportedProfileTypesFactory.ts
+++ b/src/modules/profile/SupportedProfileTypesFactory.ts
@@ -26,7 +26,7 @@ export const ProfileType = {
 } as const
 
 const DEFAULT_PROFILE_TYPES =
-  'member,workspace,community-builder,collection-point,machine-builder'
+  'member,space,workspace,community-builder,collection-point,machine-builder'
 
 const PROFILE_TYPES: IProfileType[] = [
   {
--- a/src/pages/Settings/content/formSections/Focus.section.tsx
+++ b/src/pages/Settings/content/formSections/Focus.section.tsx
@@ -9,7 +9,18 @@ import { useTheme } from '@emotion/react'
 
 function ProfileTypes() {
   const theme = useTheme()
-  const profileTypes = getSupportedProfileTypes()
+
+  function filteredProfileTypes() {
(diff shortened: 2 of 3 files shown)


AlfonsoGhislieri:
Might be a bit of a hacky solution to filter for the right profile types inside the actual component...  I just wanted to get it to work without the bugs and make sure everything shows correctly like it should. 

https://github.com/ONEARMY/community-platform/blob/master/src/modules/profile/index.ts#L13 --> This function here is what I originally was messing around, but struggled to make progress there.. 

I think I defer to you @thisislawatts for any ideas for a cleaner solution! :)

thisislawatts:
Thanks for raising this PR @AlfonsoGhislieri, this looks good. I agree though that there is something uncomfortable about the way these member types work. The repeated declarations spread across the module **and** theme suggest that the abstraction is not quite right yet. Some more thought and iteration needed 🤔  

I've taken the changes you've made here and used them as the base for [a new PR](https://github.com/ONEARMY/community-platform/pull/1748) which removes the hardcoded theme references. I have added you as a [co-author of the commit](https://github.com/ONEARMY/community-platform/pull/1748/commits/2fe626218121c058f9d5ae0c1c84a9c1abe9c439) to reflect your contribution here ✨ 
https://github.com/ONEARMY/community-platform/pull/1748

AlfonsoGhislieri:
Much better without the hardcoded values, not sure why I didn't just think of using the theme labels for filtering... Definitely would be nice to refactor the logic down the line!

PR #1753, opened 2022-06-11
fix: add support for ff member types on map

### LINKED ISSUE #1731: [bug] Fixing Fashion Map shows wrong spaces
**Describe the bug**
The map for Fixing Fashion isntance still shows the Precious Plastic spaces.
It should show the ones from Fixing Fashion
- Member
- Space

**Screenshots**
afbeelding

thisislawatts:
Closes https://github.com/ONEARMY/community-platform/issues/1731

PR Checklist

PR Type

## Git Issues

Closes #1731

### DIFF
--- a/packages/components/src/MemberBadge/MemberBadge.stories.tsx
+++ b/packages/components/src/MemberBadge/MemberBadge.stories.tsx
@@ -62,3 +62,10 @@ export const TypeMachineBuilder: ComponentStory<typeof MemberBadge> = () => (
     />
   </>
 )
+
+export const TypeSpace: ComponentStory<typeof MemberBadge> = () => (
+  <>
+    <MemberBadge size={100} profileType={'space'} />
+    <MemberBadge size={100} profileType={'space'} useLowDetailVersion />
+  </>
+)
--- a/packages/components/src/MemberBadge/MemberBadge.tsx
+++ b/packages/components/src/MemberBadge/MemberBadge.tsx
@@ -1,6 +1,7 @@
 import type { ImageProps } from 'theme-ui'
 import { Image } from 'theme-ui'
 import { useTheme } from '@emotion/react'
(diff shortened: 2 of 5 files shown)


davehakkens:
Looks good. But the icons don't seem correct, sorry didnt spot this before. 
We use different badges (avatars) when showing them in small size
[Member Icon](https://github.com/ONEARMY/community-platform/blob/master/src/assets/images/themes/fixing-fashion/avatar_member.svg)
[Space Icon](https://github.com/ONEARMY/community-platform/blob/master/src/assets/images/themes/fixing-fashion/avatar_space.svg)

Also on mobile it says "others", would be better if that was not there...
afbeelding

thisislawatts:
@davehakkens, I've pushed an update to ensure the correct badges are being loaded. 

Regarding the other point around the hanging _others_ title for the mobile filters, I would prefer to iterate on that in a separate PR. The work [I flagged to @kfern here](https://github.com/ONEARMY/community-platform/issues/1197#issuecomment-1153105097) would likely be the best place to address this behaviour so I don't want use to tread on each other toes here.

davehakkens:
nice, thanks!
Just double checking to make sure it doesn't belong in this PR.
Since it does seem somewhat related to theming, because the "others" filter does work for other instances like Precious Plastic when there are multiple pins
afbeelding

Just wanted to give that piece of information. Will trust your judgement though 
Happy to merge and no problem if we iterate on it later!

thisislawatts:
It _could_ be done in this PR but the core issue is around the datastructure used to populate the dropdown. This is something which also needs to be resolved to support the `Verified` option which we want to appear without a group heading. I have a few open threads/PRs open so would rather defer this work for now.

I will open a separate Issue to track this item and assign it to the FF milestone.

Screengrab showing Verified _without_ an _Others_ heading
Screenshot 2022-06-12 at 10 40 59

PR #1758, opened 2022-06-13
feat: add verified option to map dropdown

### LINKED ISSUE #1197: Add verified filter to the map

thisislawatts:
PR Checklist

PR Type

## Description

Closes #1197

### DIFF
--- a/src/pages/Maps/Content/Controls/GroupingFilterMobile.tsx
+++ b/src/pages/Maps/Content/Controls/GroupingFilterMobile.tsx
@@ -1,17 +1,14 @@
-import { inject } from 'mobx-react'
-import { MemberBadge } from 'oa-components'
 import { Component } from 'react'
 import checkmarkIcon from 'src/assets/icons/icon-checkmark.svg'
-import type { IMapGrouping } from 'src/models/maps.models'
+
 import type { MapsStore } from 'src/stores/Maps/maps.store'
-import { Flex, Image, Text } from 'theme-ui'
-import { transformSpecialistWorkspaceTypeToWorkspace } from './transformSpecialistWorkspaceTypeToWorkspace'
+import { Flex, Text, Image } from 'theme-ui'
+import type { FilterGroup } from './transformAvailableFiltersToGroups'
 
 interface IProps {
-  items: Array<IMapGrouping>
-  entityType: string
-  onChange?: (selectedItems: string[]) => void
+  items: FilterGroup[]
   selectedItems: Array<string>
+  onChange?: (selectedItems: string[]) => void
 }
 
 interface IState {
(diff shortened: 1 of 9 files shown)


thisislawatts:
@davehakkens at the moment the filters are cumulative so if you select `workspace` and `member` you will get all map pins which match either of those criteria. Do we want to preserve the same behaviour for the `verifed` filter? 

I was thinking that a user may want to see _only_ `members` who are also `verified`. If we want to take that approach we may want to adjust the visual styling so that it looks different because it behaves differently.

thisislawatts:
Yes I have currently implemented it so that it will match `verified` AND type, but I wasn't sure if this was the intended behaviour.

davehakkens:
yeh its awesome! Very happy to see this one being developed :)

davehakkens (approved):

src/stores/Maps/filter.ts
```diff
@@ -1,16 +1,23 @@
...
-  // filter pins to include matched pin type or subtype
-  // excluding items which have been marked as deleted=true
-  return mapPins.filter(
-    (pin) =>
-      !pin._deleted &&
-      (pin.subType
-        ? filters.includes(pin.subType)
-        : filters.includes(pin.type)),
-  )
+  const filterList = new Set(filters)
```
chrismclarke, line 9: ooh, fancy! will get a nice performance boost if we ever expand our list of filters into the thousands too :D

src/pages/Maps/Content/Controls/GroupingFilterMobile.tsx
```diff
@@ -50,69 +43,57 @@ class GroupingFilterMobile extends Component<IProps, IState> {
 
   handleChange(item: string) {
     if (this.props.onChange) {
+      console.log(`handleChange`, { item })
```
chrismclarke, line 46: Think this debug log may have been left hanging, no big deal but might be good to remove next time working on the code

src/pages/Maps/Content/Controls/transformAvailableFiltersToGroups.tsx
```diff
@@ -0,0 +1,86 @@
...
+const ICON_SIZE = 30
+
+function asOptions(mapStore, items: Array<IMapGrouping>): FilterGroupOption[] {
+  return (items || [])
+    .filter((item) => {
+      return !item.hidden
+    })
+    .map((item) => {
+      const filterType =
+        (item.type as string) === 'verified'
```
chrismclarke, line 18: the use of double ternary makes this a bit harder to read, and the order of checking first for verified type, then subtype then back to type again is a bit confusing (assume a result of confusing code elsewhere and/or just trying to get a quick fix working). If you have time to add a comment to explain briefly the behaviour that would be handy

chrismclarke:
Really nice to have this in, thanks for the addition @thisislawatts!

PR #1781, opened 2022-06-16
Refactor: automated backups system

chrismclarke:
PR Checklist

PR Type

## Breaking Changes
To enable the function on other projects additional service account permissions need to be granted - namely the `Cloud Datastore Import Export Admin` role on the default service agent (usually `PROJECT_ID@appspot.gserviceaccount.com`). More info and links in created docs (screenshot below)

## Description

The current system for automating backups has been broken (due to changes in how service agents are authorized), and was never designed to work across multiple firebase projects (required custom app-engine for cron scheduling, as well as custom service accounts provide through environment variables)

This PR rewrites the code used to automated backups to use the functions scheduler (previously updated), and authenticate against the default firebase service agent.

**Additional Changes**
To try and test this I added support for pubsub within the functions emulator, however as the cloud UI doesn't support sending custom messages on demand I also had to add a REST API endpoint (extending the existing emulator-only API endpoints) that would proxy calls to publish messages in the same way firebase does in scheduler.

This all is working, however it turns out the export operation isn't supported in the emulator anyways so all in vain. 

## Review Notes
Testing this function is a pain as it is only triggered once per week. I wrote a temporary REST API to trigger and published to the onearmy-next account to check working, which it does (e.g. see storage bucket for onearmy-next project). 

Otherwise will have to wait until first Sunday after release to test

## Git Issues

Closes #

## Screenshots/Videos

**Successful export populated to storage bucket from function**
image

**New documentation page**
localhost_3001_Backend%20Development_firestore-backup

---

## What happens next?


functions/src/emulator/index.ts
```diff
@@ -39,5 +40,53 @@ app.get('/seed-clean', (req, res) =>
...
+ * https://github.com/firebase/firebase-tools/issues/2034
+ * https://blog.minimacode.com/publish-message-to-pubsub-emulator/
+ *
+ * This allows pubsub topics to be tested when running in emulators as an api endpoint
+ * It includes additional prefix (/topic or /schedule) depending on trigger
+ * @example
+ * `http://localhost:4002/community-platform-emulated/us-central1/emulator/pubsub/schedule/dailyTasks`
+ * `http://localhost:4002/community-platform-emulated/us-central1/emulator/pubsub/topic/test-topic`
+ */
+app.post('/pubsub/:trigger/:topic', async (req, res) => {
```
chrismclarke, line 55: This is the api endpoint that only exists when running code in emulators that allows publishing a message via pubsub from a HTTP request. It can be used to trigger the pubsub calls for things like daily and weekly task functions as part of testing.

PR #1788, opened 2022-06-20
fix: update hosting-preview action

chrismclarke:
PR Checklist

PR Type

## Description

Our automation action to deploy previews to firebase seems to be failing, e.g.
https://github.com/ONEARMY/community-platform/runs/6974256199?check_suite_focus=true

The error suggests an issue with how the 3rd-party firebase-hosting-deploy action calls the firebase cli (`firebase-tools`) using npx.
This installs the firebase cli globally instead so that it should be available when the action calls also.

Should be working now, although as the default action relies on pull_request_target it will only use the code from this function after merge into master (I manually forced the action to run for the demo below by changing the trigger to pull_request):
https://github.com/ONEARMY/community-platform/runs/6975113104?check_suite_focus=true

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/.github/workflows/pr-preview.yml
+++ b/.github/workflows/pr-preview.yml
@@ -13,9 +13,9 @@ on:
   #     - '.github/workflows/pr-preview.yml'
   #     - 'package.json'
   #     - 'yarn.lock'
-  push:
-    branches:
-      - fix/hosting-preview
+  pull_request:
+    types: [labeled, synchronize]
+
 jobs:
   build_and_preview:
     # NOTE - as we are going to check out and build from forks we also need to add manual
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-preview.yml
```diff
@@ -7,10 +7,12 @@ on:
     types: [labeled, synchronize]
     # Only create a preview if changes have been made to the main src code or backend functions
     paths:
-      - "src/**"
-      - "functions/**"
-      - "package.json"
-      - "yarn.lock"
+      - 'src/**'
+      - 'functions/**'
+      - 'packages/components/**'
```
chrismclarke, line 12: include additional paths that can be used when triggering previews

thisislawatts (approved):

Issue #1543, opened 2022-03-15
@mention feature request


thisismattia:
Hey team,

Not sure how big a deal this is but as I was using the comments it came up the need of being able to mention users. 
Guess behaviour could be as follow:
User A @mention User B, User B gets a notification. 

Not urgent, but maybe nice to have in the future
<img width="1265" alt="image" src="https://user-images.githubusercontent.com/17761877/158327349-bae411c9-b8a4-49dc-9c62-c76f829350dc.png">

knoopx:
alright, taking this one. I suggest to start with a simple implementation and develop further as we go.

the bare minimum implementation is:

1) within the comment form submit handler, simply use a regex to match `@mentions` within the comment body
2) lookup the matched mentions to user ids and validate they actually exist
3) resolve the id of the page the user commented on and the id of the comment itself
4) create a notification for every matched mention id and the referring page/comment id. 

* notification text will be something like `User <username> mentioned you on <page title>`.
* whenever the user clicks over the notification, she will be redirected to the page the comment belongs to, and scroll height adjusted accordingly.
* mentions are highlighted and turned into links to corresponding user profiles
* no double notifications are triggered when the commenting user re-edits the comment?

for this to work we need:

* usernames to not contain spaces, otherwise will be very unreliable to parse mentions.
* comments to have unique html ids, so we can `scrollTo` corresponding html elements.
* commenting system to be somewhat properly abstracted, so that we don't need to re-implement mention handling all over the different sections/pages.
* be able to uncollapse comments if they are on request (when for example page#comment-n is given)

---

as a second stage, an auto-completion popup can be added:

1) the user types `@` and some characters and he gets a suggestion list of matching usernames above/below the cursor.
2) whenever he selects an username, we temporarily store the mentioning user ids alongside the comment body itself
3) when he submits the comment we already have all the necessary data to create the notification

---

I haven't checked (yet) how the data is modeled so this is pretty much a pseudo-implementation, please let me know if this makes any sense beforehand and if I should be aware of any other considerations.

davehakkens:
hi @knoopx Sorry for the slow reply. Missed the notification on this one!
But sounds like a good plan forward. Exciting.
- Usernames are without spaces.
- Comments don't have unique html yet, we need to add that.

Let my double check with our maintainer to see if they have additional suggestions.

thisislawatts:
[...]

PR #1774, opened 2022-06-15
chore: migrate notifications to component

thisislawatts:
PR Checklist

- [X] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [X] Developer experience (improves developer workflows for contributing to the project)

## Screenshots/Videos

Screenshot 2022-06-15 at 22 45 52

Screenshot 2022-06-15 at 22 46 04

Screenshot 2022-06-15 at 22 46 01

Screenshot 2022-06-15 at 22 46 00Screenshot 2022-06-15 at 22 45 58

PR #1789, opened 2022-06-20
Feature/comment mentions

patrycjapraczyk:
PR Checklist

PR Type

## Description

This PR adds a feature for mentioning users in comments under research and howtos.

Users can be mentioned by @user_name. The following is then added to the db: @[id:userId]. This adds a notification to the mentioned users with a link to the relevant comment. Clicking on this link takes the user to the relevant comment via scrolling.

## Git Issues

Closes [#1543](https://github.com/ONEARMY/community-platform/issues/1543)

## Screenshots/Videos

https://user-images.githubusercontent.com/35103888/174682439-b21e2ee3-c363-46c7-a94d-a62842103d23.mp4

---

## What happens next?


davehakkens:
Nice @patrycjapraczyk  Works good here, looking forward to use it!
Two questions:
1. If i'm the author of content (How-to or Research) I get a notification if someone comments + also if someone mentions me in a comment. Double notification , image below.  Would it be possible to remove one. Something like: If mentioned in a comment > Don't trigger content comment 

afbeelding
2. Is it (easily) possible to extend this behaviour to how-to's and research content? In this way authors can mention contributors or ask specific help from certain members. If not no worries we raise a new issue for it.

packages/components/src/CommentItem/CommentItem.tsx
```diff
@@ -56,8 +58,21 @@ export const CommentItem = (props: Props) => {
     setShowMore(!isShowMore)
   }
 
+  const parseMentions = (text: string): ReactNode => {
```
thisislawatts, line 61: **suggestion (blocking):** I would move this out of the component into a higher level component. Consider as an enrichment method which is applied to comments as they are retrieved from the data layer.

The current approach teaches our component about the format `@id[]` which it really doesn't need to know about.

src/stores/Research/research.store.tsx
```diff
@@ -199,6 +202,19 @@ export class ResearchStore extends ModuleStore {
         runInAction(() => {
           this.activeResearchItem = createdItem
         })
+        const mentionedUsers = Array.from(mentions.mentionedUsers)
```
thisislawatts, line 205: **question:** There seems to be some duplication here, have you looked at extracted triggering notifications from text content to a reusable function? This would make it possible to perform some unit tests on the logic.

thisislawatts (commented):
This looks great @patrycjapraczyk, thanks for the work on this so far ✨ 

I have raised a few comments but overall the approach looks solid

packages/components/src/CommentItem/CommentItem.tsx
patrycjapraczyk: I can see this could be improved by adding a child component like TextWithMentions rather than the upper level component. This could be reused for mentioning users in research or howtos usecase Dave mentioned. Is this what you meant @thisislawatts?

thisislawatts: How would a child component prevent knowledge of the database representation leaking into the component library?

The current implementation also hardcodes the user profile route into the UI library. Consider if we decide to have a different routes per deployment, for example:
1. Community.preciousplastic.com/u/:userId
2. Community.projectkamp.com/profile/:userId

An approach that would support full separation here is for the comment text to include anchor tags, for example `<a href="/u/userId">@username</a>`. This would then be passed into the CommentItem

src/stores/Research/research.store.tsx
thisislawatts: That sounds great!

thisislawatts:
Also @patrycjapraczyk, a heads up that I am doing some refactoring work around the Notification components atm:
https://github.com/ONEARMY/community-platform/pull/1774

packages/components/src/CommentItem/CommentItem.tsx
patrycjapraczyk: We could remove this **id:** bit, as in use `@[sampleUser] `rather than `@[id:sampleUser]`. You need some way of indication to front-end components that a link should be created in that place. I think it's simpler to parse this than url. You need to pass userId to front-end, since that's what's getting used for user profile route. Well, it's slug that's used in profile route but it's the same as userId. I could add a back-end function to get slug from userId and then pass that to front-end but it's not going to fix the issue of not exposing userId in front end.

thisislawatts:
@patrycjapraczyk, yes I believe there will be merge conflicts. How would you prefer to handle this?

patrycjapraczyk:
@thisislawatts hmm not sure. On one hand, you know your reactors better and it'd be easier for you to refractor the changes I made in this PR. But I don't want to delay you, since you got approval to your PR already. I'm also going on holidays for over a week on Friday and am busy tomorrow, so not sure if I'll have time to push the improvements to this PR soon.

packages/components/src/CommentItem/CommentItem.tsx
patrycjapraczyk: Hey @thisislawatts I've just pushed a commit in which I separated comment text to `TextWithMentions` component and changed the format of comments to `@[part1:part2]`, where the first part1 is link text and part2 would be the link ur;, for example with `@[ppraczyk:/u/ppraczyk]` **@ppraczyk** would be the link text and **/u/ppraczyk** would be the link url

patrycjapraczyk:
Hey @davehakkens @thisislawatts I've added the changes addressing your comments. 
@davehakkens now only one notification is triggered when the user is mentioned under howto/research they're an author of.
I changed the format of mentioning users, so the old mentions you tested will not work.

davehakkens:
nice. thanks for removing that duplicate notification, works good now!
But it looks like there are still some conflicts before we can merge 😬
Whats the best way to go about these @thisislawatts?

thisislawatts:
Hope you had a good holiday break! Would you be up for rebasing this branch @patrycjapraczyk or would you rather I did it?

thisislawatts:
Many Thursday evenings have passed since then, and there have been significant changes to the component structure since this PR was raised. I would like to use this as a reference and split the changes into two streams:

1. Introduce the capability to `@mention` users with their handle in the following locations, 
     1. Comments
     2. Howto introductions
     3. Howto steps
     4. Research introductions
     5. Research updates
2. Notify a User whenever they have been `@mention` elsewhere on the Platform.

davehakkens:
hehe a few Thursdays indeed! 
Just to double check. You mean starting with the ability to @mention (step 1) and later add the actual notification (step 2)?

thisislawatts:
That's correct @davehakkens, introduce the `@mention` everywhere then in a second PR introduce notifications based off of those mentions.

davehakkens:
sounds good!

davehakkens:
Should we make a new issue on this @thisislawatts to take on part two?
`Notify a User whenever they have been @mention elsewhere on the Platform.`

thisislawatts:
I don't think so @davehakkens, this looks to be pretty well defined within https://github.com/ONEARMY/community-platform/issues/1543, I have assigned both this PR and Issue to me. Will work on this next.

thisislawatts:
100% agree here @davehakkens, @patrycjapraczyk thank you so much for all of the time and thought you put into your original proposal. Really great work. I hope you're doing good and life is treating you well ✨ Hopeful that we will get the opportunity to work together again.

PR #1795, opened 2022-06-21
fix: adds is editable property to comments

### LINKED ISSUE #1794: [bug] We used to have the option to edit comments, with a little edit button. Possible that this is gone now?
**Describe the bug**
We used to have the option to edit comments, with a little edit button.

**To Reproduce**

1. Go to a how to article and write a comment
2. Click on submit
3. Attempt to edit newly submitted comment

**Expected behaviour**
An author can edit a comment they create.

**Screenshots**

image (5)

thisislawatts:
- test: introduce coverage for editable comments
- fix: adds isEditable check to comments

PR Checklist

PR Type

## Description

This change fixes a regression caused by the comoponent migration. We had removed awareness of the logged in user state from the components. This means that instead of a checking if the current logged in user within each CommentItem we perform this check as we fetch comments from the store.

The benefit of this approach is that it makes the component easier to work with in isolation and more portable. 

The "upside" of this bug is that it showcased there were gaps in our browser tests around the commenting functionality.  

In addition to the functional fix
- we now have test coverage for interacting with comments
- there is a new UserComment type with an  _isEditable_ property as part of the contract for the underlying component

## Git Issues

Closes #1794


src/components/CommentList/CommentList.tsx
```diff
@@ -32,12 +32,12 @@ export const CommentList: React.FC<{
...
         comments
           .slice(0, shownComments)
-          .map((comment: IComment) => (
+          .map((comment) => (
             <CommentItem
               key={comment._id}
               {...comment}
               isUserVerified={!!comment.userIsVerfied}
-              isEditable={false}
+              isEditable={!!comment.isEditable}
```
laianbraum, line 40: don't need not not operator here because isEditable is a required prop

src/pages/Research/Content/ResearchUpdate.tsx
```diff
@@ -138,7 +138,7 @@ const ResearchUpdate: React.FC<IProps> = ({
             </Box>
             <ResearchComments
               update={update}
-              comments={comments}
+              comments={comments as any}
```
laianbraum, line 141: `as any`

laianbraum:
Sorry. I didn't see it was a draft PR. Removed my comments.

thisislawatts:
Feel free to add some more comments @laianbraum 😉

src/pages/Research/Content/ResearchArticle.tsx
```diff
@@ -128,4 +130,15 @@ const ResearchArticle = observer((props: IProps) => {
     return isLoading ? <Loader /> : <NotFoundPage />
   }
 })
+
+function transformToUserComment(
```
chrismclarke, line 134: nit(non-blocking) I expect we will usually want to know if comments are by the current user and editable, so it might make more sense to move this higher up. 

For now I think the highest we could go would be the research store level (as comments are nested within individual research items), although longer term I expect it might make more sense to have comments as a standalone collection/subcollection, at which point we could standardise across stores and move the logic to the db level

packages/cypress/src/integration/research/comments.spec.ts
```diff
@@ -0,0 +1,56 @@
+describe('[Research]', () => {
+  beforeEach(() => {
+    cy.visit('/research/qwerty')
+  })
+
+  it('[Open comments]', () => {
```
chrismclarke, line 6: Thanks again for adding all the extra tests

packages/cypress/src/integration/research/comments.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+      cy.visit('/research/qwerty')
+      cy.get('[data-cy="ResearchComments: button open-comments"]').click()
+      cy.get('[data-cy="comment"]').should('have.length.gte', 2)
+      cy.get('[data-cy="comment"]')
+        .last()
+        .get(`[data-cy="CommentItem: edit button"]`)
+        .should('exist')
+    })
+
+    it('allows comment author to delete', () => {
```
chrismclarke, line 45: nit(non-blocking) I'm not sure whether previously it was possible for admins to remove comments or not. If yes it would be worth adding a test to check admins can and non-admins can't. If not we should check with @davehakkens if that's something expected and create an issue for it.

chrismclarke (approved):
Thanks for fixing this up @thisislawatts , I can see how trying to manage the editable state outside of the core component makes a lot of sense, but also unravels some further messiness about the best way to go about it.

I think in an ideal world `_isEditable` would be a computed property generated on all comments, and comments would sit in a somewhat more distinct data layer (not just as properties within howto and research items), but a pr like this definitely isn't the place for such large-scale changes and I think what you've done is perfect for now.

Looking at the code also made me want to flag that editing and deleting are coupled together (if you can do one you can do the other, which makes sense to me), however the only person who ever receives those privileges appears the comment's original author (i.e. not admins). This might be something to consider @davehakkens, and possibly open up a follow-up issue to change the behaviour if desirable

But for this PR all looks good to me :D

Issue #1418, opened 2022-01-29
Improving SEO #1 - Indexable by search engine.


davehakkens:
### Description

The platform is currently very poorly indexed by search engines.
Our content is rendered dynamically and needs server-side rendering to improve.

### Page related
This issue is mainly important for the [how-to's](https://community.preciousplastic.com/how-to) since they contain the most indexable content.

### Additional work
Besides that, there are a few other smaller things to improve. You can find more of that below in the Additional Report
we had from an SEO Analist. Might be helpful..

1. [Sitemap]: you should add it in public folder, it's recommended to make it auto generated ( I can help in this ), don't try to use online website to generate it, because in your case it's SPA with CSR, so it's will detected just the / page ...
2. try to use href as possible because it's help a lot.
3. try to add "alt" to image.
4. maybe use pagination it's will help to have better performance ( you have an issue in this, and it's have very big impact)
5. you should use google search console to have a better view of the page that are indexed, and also to send request to google to index your page ( https://search.google.com/ ) 
6. [meta data] : you should insert meda data for every page with dinamyc content, to make this it's simple => use React Helmet 
7. if you have a server like apache you should make configuration to set 404 status error for errorPage, this is a solution for SPA CSR, in case of nextjs or ssr it's ok

thisislawatts:
I think the highest priority item here is to get key content pages such as the How-to listing and individual How-to articles rendered via SSR. 
I am going to prototype an approach which will render these pages on the server, most likely via Google Cloud Functions so we can stick with our existing stack.

chrismclarke:
Getting a decent dev workflow going for testing these things is always a bit of a pain (most testing tools require site to be live), although if running from localhost you could possibly use something like https://ngrok.com/ to temporarily expose to public web, or if you want access to a spare firebase platform to test deployments too I think we probably have a few lying around! 

I think server-side rendering is definitely the way to go here (with an unironic solution to dynamically create the static pages). Although if you did decide rather you'd want the client-side code to update meta tags and pass back to bots you could use something like [react-helmet](https://github.com/nfl/react-helmet) (or vanilla js) to update meta tags and then a custom API endpoint + [rendertron](https://github.com/GoogleChrome/rendertron) (or just puppeteer/playwright) to intercept bot requests, generate dynamic page and serialize back to static html.

If we really wanted to get fancy google has additional content types that you can try to align to (although feel like their howto might be harder to match, perhaps easier something like article)... but that could always be a future optimisation also.
[...]

PR #1787, opened 2022-06-20
Feat/seo sitemap

chrismclarke:
PR Checklist

- [ ] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Updates 2022-06-20
- Merged 1781 so diffs should be cleaner for review
- Have added tests which should cover the main behaviour of the sitemap generator. This also required extra test setup code to allow passing of runtime config to the emulator (what would be populated via firebase functions:config:get)
- Realised we already have a `deployment.site_url` populated in firebase that can be used for generating sitemap so will use that for now until we sort out environment variables more generally (opened #1790 to work on)
- Plan to hold-off on robots.txt creation until we resolve the point above (as ideally we want the robots.txt generated as part of our deployment pipeline, but that doesn't have access to the variables set in firebase config)

## WiP TODOs

- [x] Builds off #1781 (needs pubsub emulator) so should wait until that has been merged and update here to show cleaner diff

- [x] Requires a `SITE_URL` environment variable to specify prefix used for URLs (e.g. https://community.preciousplastic.url). Need to decide if possible/better to set via firebase config or circleCI, and what the best name for this variable is. Alternatively could generate on request and extract from request URL, although that would make it harder to specify sites that we don't want to generate sitemaps for (e.g. dev site), likely still requiring some env variable set somewhere.

- [x] Possibly generate a robots.txt file to let robots know where sitemap is located. This will also require some means of populating the full URL of the site
https://developers.google.com/search/docs/advanced/sitemaps/build-sitemap#addsitemap

- [ ] Create future issue for sitemap diffing on a more regular basis (e.g. daily), and notifying prerender ahead of google

## Description

This adds functions to automate the generation of a sitemap from database content. Currently, it is limited to just howtos and research articles

This is done in a few steps:
1. A scheduled function that generates a full sitemap weekly (to prevent too many database requests if generating on-demand). This sitemap is stored in the local storage bucket

2. A sitemap proxy function that takes requests for the sitemap, and returns the contents of the generated sitemap populated in the storage bucket

[...]

PR #1805, opened 2022-06-25
feat: improve appearance when shared sociallly and when viewed in search engine result pages

thisislawatts:
PR Checklist

PR Type

## Description

Introduces a Google Cloud function that receives all requests for individual Research and How-to articles. If the useragent for the request appears to belong to a bot account the page is fetched via [prerender.io](https://prerender.io/) which parses/executes the JS and returns a HTML string which can be better understood by the bot.

The intention here is to improve the appearance of Platform article pages on search engines.

A prerender.io API token is required on the target Firebase instance to ensure that it works. This can be done with: 

```bash
firebase functions:config:set prerender.api_key=XXX
```

## Git Issues

Partial resolution for issue #1418

## Screenshots/Videos

Screenshot 2022-06-24 at 19 01 48


functions/src/seo-render/index.ts
```diff
@@ -0,0 +1,92 @@
...
+export const seoRender = functions.https.onRequest(
+  async (request, res: any) => {
+    const requestUserAgent = (request.headers['user-agent'] || '').toLowerCase()
+    const xPrerender = request.headers['x-prerender']
+    const pathName = request.path.replace('/seo', '')
+
+    console.log(`Request:`, { pathName, requestUserAgent })
+
+    let response: any
+    if (!xPrerender && containsOneOfThem(BOT_AGENTS, requestUserAgent)) {
```
chrismclarke, line 64: nit(non-blocking) - this might be tidier if moved to a separate function such as `isBotRequest` or `shouldUsePrerender` or similar

functions/src/seo-render/index.ts
```diff
@@ -0,0 +1,92 @@
...
+            headers: {
+              'X-Prerender-Token': CONFIG.prerender.api_key,
+            },
+          },
+        )
+        response = await response.text()
+
+        return res.send(response)
+      } catch (error) {
+        console.log(`Error fetching response from prerender:`, { error })
```
chrismclarke, line 79: nin(non-blocking) - might be better to use firebase `logger.error` function so that it gets flagged in the firebase console dashboard and could be actionable in the future

functions/src/seo-render/index.ts
```diff
@@ -0,0 +1,92 @@
...
+      }
+    } else {
+      console.log(`Non bot user detected.`)
+      try {
+        response = fs.readFileSync('./index.html').toString()
+      } catch (error) {
+        console.log(`Unable to load index.html`, { error })
```
chrismclarke, line 86: I'm not sure what could lead to this case, but I'm guessing we should just send some sort of error back instead of a success with undefined content. E.g. `return res.status(404).send()`

functions/src/seo-render/index.ts
chrismclarke, line 79: nit(non-blocking) - same as comment below, probably better not to leave undefined response to filter through to the end, I'd suggest either sending an 404 response or possibly trying a redirect to the home page

functions/package.json
```diff
@@ -33,6 +33,7 @@
     "google-auth-library": "^6.1.1",
     "googleapis": "^61.0.0",
     "log-update": "^4.0.0",
+    "node-fetch": "2",
```
chrismclarke, line 36: nit(non-blocking) - We already have axios in the project, is there a good use case for having two libs that do similar things?

functions/package.json
```diff
@@ -9,7 +9,7 @@
     "start": "ts-node scripts/start.ts",
     "serve:live": "yarn copyDevConfig && yarn serve",
     "lint": "tslint --project tsconfig.json",
-    "build": "webpack",
+    "build": "webpack && cp ../build/index.html ./dist/",
```
chrismclarke, line 12: nit(non-blocking) - this makes me feel slightly anxious as it means that deploying functions depends on the latest frontend to be built, but no real checks in place to ensure this has happened. I've used this approach previously myself and come unstuck when manually doing a quick `firebase deploy --only functions` to try and hotfix, or similar scenarios

A better alternative might be to use an additional HTTP request to fetch the live index.html file instead of relying on having the file copied as described in: https://stackoverflow.com/questions/53904044/accessing-firebase-hosting-files-from-firebase-function-triggers. I expect the request to have pretty negligible time as a) the index file is super small and b) likely will be located in the same region/data-centre.

functions/src/seo-render/index.ts
```diff
@@ -0,0 +1,92 @@
+import * as functions from 'firebase-functions'
+import * as fs from 'fs'
+import fetch from 'node-fetch'
+import { CONFIG } from '../config/config'
+
+const BOT_AGENTS = [
```
chrismclarke, line 6: nit(non-blocking) - do we want to maintain our own list? An alternative would be to use an npm package that does it, e.g. https://www.npmjs.com/package/isbot

firebase.json
```diff
@@ -7,6 +7,14 @@
         "source": "/api",
         "function": "api"
       },
+      {
+        "source": "/how-to/**",
```
chrismclarke, line 11: I had originally thought we would proxy all endpoints via the seoRender, however I can see the benefit of having it specifically for those that we are setup to handle

functions/src/seo-render/index.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    let response: any
+    if (!xPrerender && containsOneOfThem(BOT_AGENTS, requestUserAgent)) {
+      console.log(`Bot user detected.`)
+      try {
+        response = await fetch(
+          `https://service.prerender.io/${CONFIG.deployment.site_url}${pathName}`,
+          {
+            headers: {
+              'X-Prerender-Token': CONFIG.prerender.api_key,
```
chrismclarke, line 71: issue(blocking) - If we reach this point but don't have a prerender api key defined I'm guessing this will throw an error trying to access `api_key` on what would be `undefined`. So I think we will need to have a check earlier for `CONFIG.prerender?.api_key`.

chrismclarke (changes requested):
Thanks Luke, I definitely think this is a really nice addition to have. 

I've added a few non-blocking comments, mostly around how much we want to let errors pass by unnoticed (e.g. catch and continue) vs structuring error response codes. There's also a few implementation details I want to try and sense-check with you (as I have used similar code in my own projects before but am starting to think whether there might be better options available)

I've also flagged one issue I expect may throw errors in production, will be keen to have your feedback. 

As for testing, #1787 contains code that will allow populating things like the prerender API key into the runtime config for use in emulator testing, so once that is merged we should be able to add tests for this if useful

functions/package.json
thisislawatts: @chrismclarke, I did take the HTTP approach initially but I think it's preferable to load from the file system. I was referencing the same SO article. It was the final point around performance that persuaded me to avoid the overhead and uncertainties of a network request. 

> If you just want some static files available in Cloud Function without having to make an external request, simply copy those files in to your functions folder and deploy them all along with your code. You will be able to read them locally, you won't need to enable billing to try it, and you won't incur egress and ingress charges. _**It'll also be faster.**_

I share your concerns around the fragility this introduces into the build process. I wonder what options we have to protect ourselves here, whether that's process, automated checks or documentation. WDYT? Is it worth the overhead of a HTTP request affecting each and every user request in normal operations compared to increased risk during non standard scenarios.

chrismclarke: Yeah, it's hard to comment on these things when we don't really know what _faster_ actually means in this context - like I understand why theoretically it will be, but are the amounts of time we're talking about perceptible? I'm guessing answering that would take more effort than is really worth at this point.

I can't really think of any sensible means to prevent updating of functions without updated code, so personally would probably take the potential performance hit over introducing fragility - however I really don't have a strong opinion so happy to leave as is and address if it comes up as an issue in the future. By that time we may even have SSR working properly, in which case it all becomes a non-issue.

functions/package.json
thisislawatts: not at all, will switch to axios

functions/src/seo-render/index.ts
thisislawatts: Good catch :) Will introduce some logic to catch these edge cases.

functions/src/seo-render/index.ts
thisislawatts: Good catch! Will introduce a fix for this

chrismclarke (approved):
Thanks for making the updates, definitely looks cleaner to me and nice to have some tests also. Will be excited to see it in action!

functions/package.json
chrismclarke: One other minor issue I just picked up on is that cp doesn't work on windows... not hugely significant unless trying to build and run docker emulators locally on a windows machine

chrismclarke:
There's a few conflicts from the work I did on SEO sitemaps, I'll go ahead and give a quick tidy to put them both under the same folder structures

Issue #1706, opened 2022-06-01
Enable Digital events


davehakkens:
### Description

Digital events are becoming more popular in the community. (hello 2022) 
This issue would allow members to share them in the events page.
If the checkbox "digital" event is selected it should show the `Digital.png` icon on the frontend.

⭐️Bonus: Ideally we can add this:
If the checkbox digital is selected > the field `In which city is the event taking place?`  is hidden or not a mandatory input field anymore.

### Page related

Will be used in : [Events](https://community.preciousplastic.com/events)

### Notes
Be aware the input fields are rearranged to make it visually more logical to follow.

### Mockups

![even_digital_02](https://user-images.githubusercontent.com/13672737/171496783-0b479224-38a7-45ec-b467-404a17bea33c.jpg)
![event_digital](https://user-images.githubusercontent.com/13672737/171496794-c6586387-64fb-4e62-9d00-663ea1b239ee.jpg)

### Icon attached
![Digital](https://user-images.githubusercontent.com/13672737/171496819-8805d6d3-9356-4228-9107-f91ca7f36890.png)

missalyss:
Hi! Can I take on this issue? I've got code running and more docs to read but noticed all the other 'Good first issue's were taken 😊

thisislawatts:
Absolutely @missalyss ✨ I have assigned the issue to you. Please let me know if you've any further questions. 

Additional details to be aware of:

1. The _﻿﻿In which city is the event taking place?_ field, which is currently marked as required. I would expect this field to be optional or completely hidden. 
2. For visibility, I'm [working on a PR](https://github.com/ONEARMY/community-platform/pull/1701) at the moment which refactors all Input elements, hoping to merge that in the next day or two. But note you'll want to rebase after #1701 is merged.

davehakkens:
hey @missalyss how is this going? Need any help or guidance?

davehakkens:
Good to hear some of it is working @missalyss! Did you manage to follow [these steps ](https://onearmy.github.io/community-platform/)to run locally?

missalyss:
Yep—I believe I just had errors I couldn’t identify so localhost was hanging. but I have it fixed now. The event form is submitting correctly, just some cleanup and extra tasks before the PR will be ready to put up. 😊

davehakkens:
Nice! Looking forward to it 💪

PR #1815, opened 2022-06-28
feat: enable digital events 1706

missalyss:
PR Checklist

- I like to squash my commits into one but wasn't sure if it was preferred to keep them separate

PR Type

## Description

Adds in `isDigital` checkbox to EventsCreate. `location` is an optional form field and renders only when `isDigital` is false. The form will not submit if there if a location has not been selected or `isDigital` is false). `isLocationSelected` is set to false when search input is cleared with browser `x` button. On EventsCards, "Digital" appears in lieu of country and flag.  

Repro:
1. http://localhost:3000/events/create
2. Create an event. Should be able to select either location or digital event. Should not be able to select both or neither. 
3. See digital tag on your newly created event.

## Git Issues

Closes #[1706](https://github.com/ONEARMY/community-platform/issues/1706)

## Screenshots/Videos
image
image
image

---

## What happens next?


src/components/Form/Checkbox.tsx
```diff
@@ -0,0 +1,21 @@
...
+  labelText,
+  ...rest
+}: FieldProps) => {
+  return (
+    <>
+      <input id={id} type="checkbox" {...input} {...rest}></input>
+      <StyledLabel htmlFor={id}>{labelText}</StyledLabel>
+    </>
+  )
+}
```
thisislawatts, line 21: **praise:** Great to see a new generic component being introduced to the system. We are in the middle of migrating all of our reusable components our to an isolated package at `./packages/components`. Read more about it here: https://github.com/orgs/ONEARMY/projects/3

It would be preferable to move that to this new directory, but I do not consider this a blocker and we can action that as part of a follow up PR.

src/pages/Events/Content/EventsCreate/EventsCreate.tsx
```diff
@@ -223,35 +220,75 @@ export class EventsCreate extends React.Component<IProps, IState> {
...
+                              <>
+                                <Label htmlFor="location">
+                                  In which city is the event taking place?
+                                </Label>
+                                <Field
+                                  id="location"
+                                  name="location"
+                                  className="location-search-create"
+                                  validateFields={[]}
+                                  customClear={() => {
```
thisislawatts, line 238: **question (blocking):** Is this a callback that needs to be handled within the child element? It seems like something which is only relevant within the EventsCreate context so shouldn't need to leak any further into the dependency tree. 

Did you consider using the [`componentDidUpdate`](https://reactjs.org/docs/react-component.html#componentdidupdate) method?

thisislawatts (commented):
Thanks for the PR @missalyss, this looks good. I don't think there's too much left to do here. I've added one question around the implementation and then I wondered, have you looked at the e2e tests we run using Cypress? 

There's a specific user journey which I think is worth duplicating as a base for verifying that a user can create a digital event. What do you think, would you be up for working on that?

src/components/Form/Checkbox.tsx
missalyss: Hey, sorry for the late follow up. Last time I worked on this, I wasn't sure about the `theme`. Do I need to import it into the package? Is there some theme context I can use in the package? (I didn't want to hardcode the css, but I can do that as well, obviously)

src/pages/Events/Content/EventsCreate/EventsCreate.tsx
missalyss: I believe my reasoning was so to not call `componentDidUpdate` with each render, and that it didn't seem I'd be able to access the form values until submit. However, I didn't try `componentDidUpdate`, so I'll give that a go and see if I can detect the location value changes from there! 

Let me know if there's a timeline you'd prefer for this to be complete on, I recognize I've been fairly delayed in my updates 😅 🙏🏻

thisislawatts: I am not aware of a specific timeline from us on this @missalyss. Although I am happy to help out on the PR if your availability has changed for the foreseeable future. Let me know what you would prefer!

src/components/Form/Checkbox.tsx
thisislawatts: Apologies for missing this comment @missalyss. Within the `packages/components` directory we can call `useTheme` within a component to make this available. The typing on that object is incomplete at the moment but you can see an example here:
https://github.com/ONEARMY/community-platform/blob/master/packages/components/src/MemberBadge/MemberBadge.tsx#L15

The source theme definition can be found here: https://github.com/ONEARMY/community-platform/blob/master/src/themes/precious-plastic/styles.ts#L228

thisislawatts:
Thanks for all your time and event on this @missalyss. We really appreciate the contribution and really happy to merge this improvement to the platform as part of #1935. I raised the separate PR using a branch on this repository so I could push a few modifications in.

✨

PR #1816, opened 2022-06-28
Feat/emulator runtime config

chrismclarke:
PR Checklist

PR Type

## TODO

- [x] Builds from #1805 so should be updated after merge for clean diff

## Description

A few minor bits of housekeeping related to docker emulators, namely:
- Add support for runtimeconfig within docker emulators (same way that we use to test functions emulator locally)

- Update emulator seed data. This is mostly just because it is currently the only way to force a rebuild of the emulator core which is required as the package.json dependencies have been updated recently. 
I was hoping to update from the fixed automated weekly export (instead of usual manually triggered export), however this needs a few minor tweaks first (see below) so created manual export. Hopefully should open pathway after for automating the full process

- Remove user `revisions` subcollection - these are quite large and just not very well thought out (keep copy of entire user profile when states change, but really should just track diffs - will add issue to address). Given that we create a weekly backup anyway no reason each should contain full history of user profile modifications

- Tidy jest runtime config setup scripts to allow config to also be imported into docker build.

- Minor docker build logging updates

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?


chrismclarke:
As this shouldn't have any real knock-ons elsewhere I think safe to unblock and just merge

PR #1821, opened 2022-07-05
fix: return sitemap as file

chrismclarke:
PR Checklist

PR Type

## Description

Small update to the sitemap generator to run via `/sitemap.xml` instead of `/sitemap` and return as a file instead of a string. Previously it was assumed not possible (static files should take priority over rewrites), however reading around and testing in local emulators suggests this should actually work.

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -16,7 +16,7 @@
         "function": "seoRender"
       },
       {
-        "source": "/sitemap",
+        "source": "/sitemap.xml",
         "function": "seo-sitemapProxy"
       },
       {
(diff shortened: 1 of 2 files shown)


chrismclarke:
As I can only test this properly by deploying to dev and changes are minimal will merge here without review and revert if not working as intended

chrismclarke:
Looks to be working so will leave as is
https://dev.onearmy.world/sitemap.xml

image

PR #1865, opened 2022-08-05
fix(functions): updating user info in howto comments

### LINKED ISSUE #1469: [bug] Country flag doesnt change or shows on older comments
**Describe the bug**
It seems that if a user changes or adds their location in the profile it doesnt update their flag on older comments.

**To Reproduce**

1. Go to how-to
2. Leave a comment
3. Change location
4. Check the comment
5. Leave a new comment to see the difference

**Expected behavior**
Ideally, this is in sync across the entire platform. Change your location (flag) > see the change everywhere.
in howto, research, comments etc.
(This functionaity does seem to work in user profiles. Change location and the flag changes there)

**Screenshots**
Same username, different location
afbeelding

**Additional context:**

The Community Platform is built on top of [a document orientated database](https://firebase.google.com/docs/firestore#how_does_it_work) called FireStore. As a NoSQL data store it is not simple to perform join queries across multiple collections, for example Howto articles and Users. 

Existing behaviour:

1. We have the `User` entity, this is where a user can define their location, which we use to render a flag next to their username. 
1. We display the user name in multiple places across the site. Each query/read of a User document is considered a billable unit by Firebase.  
To reduce the number of billable operations we copy the location data into each of the following:
    1. Howto article
    1. Howto comment
    1. Research article
    1. Research comment
   
The implications of this approach are that each time a User changes their location we need to update the location data in each of the documents generated by that User.

The current implementation for syncing this content, is run as a [Cloud function triggered by changes in Firestore](https://firebase.google.com/docs/functions/firestore-events).
Code: https://github.com/ONEARMY/community-platform/blob/master/functions/src/userUpdates/index.ts#L16-L17

The drawbacks with the current approach are:
1. Each time we add the username location to a new place in our document store we need to extend our sync mechanism to ensure it stays up to date
1. The update method could fail and we have low visibility into this.
1. Duplicates the User model data structure in multiple places across the system. This means if we made a change to the User data structure we would have to remember to update; the sync mechanism, all existing documents and UI logic.

danitrod:
PR Checklist

PR Type

## Description

Fixes the bug where updating a country in user profile wouldn't reflect in the comments the user made. The logic runs in a function which triggers from user updates. The function queries for all the howtos, then filters and updates the comments owned by the user. An optimization could be made here by querying directly for the howtos where the user participated, with something involving `array-contains` and the userName, but would require an additional aggregation data structure, as [discussed on Slack](https://onearmy.slack.com/archives/C01CXSDSMC2/p1657933042198259). This can be looked at as an optimization in the future.

## Git Issues

Closes #1469

### DIFF
--- a/functions/src/userUpdates/index.ts
+++ b/functions/src/userUpdates/index.ts
@@ -119,6 +119,7 @@ async function updateUserCommentsInfo({
           (comment) => comment.creatorName === originalUserName,
         )
       ) {
+        let toUpdateCount = 0
         const updatedComments = howto.comments.map((comment) => {
           if (comment.creatorName !== originalUserName) return comment
 
@@ -132,7 +133,7 @@ async function updateUserCommentsInfo({
             updatedComment.creatorCountry = country
           }
 
-          count += 1
+          toUpdateCount += 1
           return updatedComment
         })
 
@@ -140,6 +141,7 @@ async function updateUserCommentsInfo({
           await doc.ref.update({
             comments: updatedComments,
           })
+          count += toUpdateCount
         } catch (error) {
           console.error('Error updating comment: ', error)
         }
--- a/functions/src/userUpdates/index.ts
+++ b/functions/src/userUpdates/index.ts
@@ -1,6 +1,6 @@
 import * as functions from 'firebase-functions'
 import { db } from '../Firebase/firestoreDB'
-import { DB_ENDPOINTS, IDBDocChange, IUserDB, IHowtoDB } from '../models'
+import { DB_ENDPOINTS, IDBDocChange, IHowtoDB, IUserDB } from '../models'
 import { backupUser } from './backupUser'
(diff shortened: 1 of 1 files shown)


davehakkens:
thanks for the pr @danitrod and trying to find a simpler solution! Did it work for you locally?
After a quick test here. (Leave comment, change location, leave comment again) it seems that the bug is stil there?
afbeelding

danitrod:
It did work locally using the Firebase emulator; as the Firebase functions don't deploy in the preview environment we have to merge first to test, so we should be good.

thisislawatts:
Thanks for raising the PR @danitrod. Are there any automated tests we can introduce here to verify this behaviour? I will not making that a blocking comment for merging this PR but would welcome your perspective on how to implement this.

thisislawatts (approved):

Issue #1506, opened 2022-02-24
Make simplified profile default


davehakkens:
**Is your feature request related to a problem? Please describe.**
We just released a simplified profile in #1267 this is used for `members. `
The extended (original) profile is used by `Spaces`. 
However by default, the extended profile is still used. 
This should be changed to the simplified layout. 

**Describe the solution you'd like**
If a user signs up and does not choose a profile type `member`, `space` etc. the default layout should be the simplified profile.

![default_profile](https://user-images.githubusercontent.com/13672737/155536144-b76a5258-34f8-4b75-87eb-90317f61d298.jpg)

Kiebert:
Hello @davehakkens.
This was an easy fix. No need for the bounty.
I will create a pull request.

Kiebert:
Oops. Not an easy fix. The Cypress tests fails... I will have an attempt on fixing this...

Kiebert:
Added a fix to pass the Cypress test.

Kiebert:
Main outcome now according to this communication flow:
https://github.com/ONEARMY/community-platform/pull/1961

"When users signup and didn't select a profiletype yet (not space or member), they got the layout of the 'SpaceProfile'. But it should show the 'MemberProfile'"

PR #1966, opened 2022-10-02
fix: member type shows member profile

Kiebert:
PR Checklist

PR Type

## Description

A 'Workspace' type should also show the 'SpaceProfile'. 

## Git Issues

https://github.com/ONEARMY/community-platform/issues/1506

### DIFF
--- a/src/pages/User/content/index.tsx
+++ b/src/pages/User/content/index.tsx
@@ -77,11 +77,10 @@ export class UserPage extends React.Component<
     }
     return (
       <>
-        {user.profileType === ProfileType.SPACE ||
-        user.profileType === ProfileType.WORKSPACE ? (
-          <SpaceProfile data-cy="spaceProfile" user={user} />
-        ) : (
+        {user.profileType === ProfileType.MEMBER ? (
           <MemberProfile data-cy="memberProfile" user={user} />
+        ) : (
+          <SpaceProfile data-cy="spaceProfile" user={user} />
         )}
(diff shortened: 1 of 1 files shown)


thisislawatts:
@Kiebert can we introduce a breaking test that recreates the bug before adding a fix? This will protect us against future regressions. At the moment the majority of our tests are in Cypress and this would be the specific suite to extend:
https://github.com/ONEARMY/community-platform/blob/master/packages/cypress/src/integration/profile.spec.ts

Let me know if you've got any issues writing/running the tests.

Kiebert:
> @Kiebert can we introduce a breaking test that recreates the bug before adding a fix? This will protect us against future regressions. 
Hi @thisislawatts. You mean to recreate the scenario that the user hasn't choosen a profile type yet?

thisislawatts:
Yep, that's right @Kiebert that would ensure we have the desired behaviour under test.

Kiebert:
Ok. I will write a test for it.
I will make time for that this week.
If you want the profile to react as in the original state, you can merge the code changes of the PR her above.

davehakkens (approved):
Will merge for now so the profile looks correct.
But would be cool to see that test added!

Kiebert:
@davehakkens @thisislawatts I wrote this simple profileType test, you can view it in this PR: https://github.com/ONEARMY/community-platform/pull/1979
Could you please check if it is sufficient?

PR #1981, opened 2022-10-13
fix: support more username variants

### LINKED ISSUE #1980: [bug] @mentions dont work with all characters
**Describe the bug**
Some usernames dont show well when mentioned in the comments.
Below a screenshot of the user [@paul-r-](https://community.projectkamp.com/u/paul-r-) on Project Kamp Instance. (last one)
probably because of the last character` - `

**Screenshots**
afbeelding

**Problem**
This is the problematic line, https://github.com/ONEARMY/community-platform/blob/master/src/stores/common/mentions/index.ts#L48

This is where a test needs to be added that verifies the fix https://github.com/ONEARMY/community-platform/blob/master/src/stores/common/mentions/index.test.ts#L74

**Expected behaviour**
@mentions should accept the same formatting as usernames. (Use the same characters, no uppercase, no spaces)

thisislawatts:
PR Checklist

PR Type

## Description

Fixes a missing edge case in how we convert userRefence to an `@mention`

## Git Issues

Closes #1980

### DIFF
--- a/packages/components/src/LinkifyText/LinkifyText.stories.tsx
+++ b/packages/components/src/LinkifyText/LinkifyText.stories.tsx
@@ -16,5 +16,7 @@ export const Default: ComponentStory<typeof LinkifyText> = () => (
 export const SupportsMentions: ComponentStory<typeof LinkifyText> = () => (
   <LinkifyText>
     In addition to a URLs, it is also possible to @mention somone.
+    Although there are edge cases where using
+    @&#8288;a-mention should <b>not</b> be a link.
   </LinkifyText>
 )
--- a/packages/components/src/LinkifyText/LinkifyText.stories.tsx
+++ b/packages/components/src/LinkifyText/LinkifyText.stories.tsx
@@ -15,8 +15,8 @@ export const Default: ComponentStory<typeof LinkifyText> = () => (
 
 export const SupportsMentions: ComponentStory<typeof LinkifyText> = () => (
   <LinkifyText>
-    In addition to a URLs, it is also possible to @mention somone.
-    Although there are edge cases where using
-    @&#8288;a-mention should <b>not</b> be a link.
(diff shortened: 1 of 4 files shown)


davehakkens:
Does it by any chance also make it all lowercase (like the format used for usernames)
For instance this one `@OhKaspian` (on top in the original issue) Also didnt work due to the caps used.
Seems better to force-format any @mention to lowercase only.


afbeelding

thisislawatts:
@davehakkens could you drop a link to the affected comment? I don't think mentions should work unless there is an exact match (lowercase) with an existing user.

davehakkens:
I already changed it back manually to lowercase, now it works.
But here is a newly created one: 
https://dev.onearmy.world/research/support-mention-in-the-key-research-textareas-and-comments
afbeelding

davehakkens:
yeh looks like you can mention anything, hehe
afbeelding

thisislawatts:
Ah interesting, this is slightly separate to the `@mention` implementation. Instead we have a plugins which automatically turns `@aNytHING` into [community.projectkamp.com/u/aNytHING](https://community.projectkamp.com/u/aNytHING) a profile link _without_ verifying that the user exists.

thisislawatts:
Okay fun, have found a fix for this by introducing a zero-width character so that we can have some `@mention` which become links and other `@invalid-username` which do not become links.

Screenshot 2022-10-14 at 09 10 27


> U+2060 Word-Joiner &#8288; representative by no visible character, it prohibits a line break at its position.
https://stackoverflow.com/a/28405917/3292677

davehakkens:
Sounds good. 
But tested the preview, still seem to be able to mention anything? (except `@more`. hehe)
Perhaps it only runs a check on the first one?
afbeelding

davehakkens:
that seems to do the trick here! 
Works good now

davehakkens (approved):

PR #1984, opened 2022-10-15
fix: user references after updates

thisislawatts:
PR Checklist

PR Type

## Description

Recently we introduced an `@mention` feature. This enables users to mention others at multiple places within a Research and Howto article; these are are main description, update description and comments.

This behaviour relies on `addUserReference` having been applied to each of those fields to convert a plain `@mention` string to a user reference ID: `@@{userId:username}`.

Every method which modifies the ResearchItem|HowtoArticle first clones and saves the ResearchItem which means the save object does _not_ have a userReferenceId available. 

I think that there is value in refactoring the internal flow of the store so that we can avoid this rework but first let's fix the immediate bug and introduce some tests to make it safer to refactor the implementation. 

---

## What happens next?

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,6 +143,7 @@
     "@commitlint/cli": "^16.2.3",
     "@commitlint/config-conventional": "^16.2.1",
     "@commitlint/cz-commitlint": "^16.2.3",
+    "@faker-js/faker": "^7.6.0",
     "@semantic-release/changelog": "^6.0.1",
     "@semantic-release/git": "^10.0.1",
     "@testing-library/jest-dom": "^5.11.4",
--- a/src/stores/Howto/howto.store.test.ts
+++ b/src/stores/Howto/howto.store.test.ts
@@ -1,0 +1,247 @@
+jest.mock('../common/module.store')
+import type { IHowtoDB } from 'src/models'
+import { FactoryComment } from 'src/test/factories/Comment'
+import { FactoryHowto } from 'src/test/factories/Howto'
+import { FactoryUser } from 'src/test/factories/User'
+import type { RootStore } from '..'
+import { HowtoStore } from './howto.store'
+
+async function factory(howtoOverloads: Partial<IHowtoDB> = {}) {
+  const store = new HowtoStore({} as RootStore)
+  const howToItem = FactoryHowto(howtoOverloads)
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  store.setActiveUser(
+    FactoryUser({
+      _id: 'fake-user',
+    }),
+  )
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  store.db.get.mockResolvedValue(howToItem)
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  store.db.getWhere.mockReturnValue([howToItem])
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  store.userStore = {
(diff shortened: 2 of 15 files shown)


package.json
```diff
@@ -143,6 +143,7 @@
     "@commitlint/cli": "^16.2.3",
     "@commitlint/config-conventional": "^16.2.1",
     "@commitlint/cz-commitlint": "^16.2.3",
+    "@faker-js/faker": "^7.6.0",
```
chrismclarke, line 146: Here's hoping the devs don't go rogue again 😅

src/test/factories/Comment.ts
```diff
@@ -0,0 +1,16 @@
+import { faker } from '@faker-js/faker'
+import type { IComment } from 'src/models'
+
+export const FactoryComment = (
```
chrismclarke, line 4: nit(non-blocking)   
Adding some factories to generate content for easier testing definitely makes a lot of sense to me. I'm just wondering if the naming is a bit verbose? I.e. would it be clearer for other devs if we rename these methods to prefix `createMock`, e.g. `createMockComment`? 
I'd say just a minor preference, so if you feel more strongly about keeping as factories still happy to keep as is

src/stores/Research/research.store.tsx
```diff
@@ -346,6 +355,35 @@ export class ResearchStore extends ModuleStore {
     }
   }
 
+  /**
+   * Invokes Set on passed dbRef after
+   * applying common transformations on
+   * supplied researchItem object
```
chrismclarke, line 361: nit(non-blocking)
Sounds a bit fancy to me - can we simplify the language a bit? Would "update research item db ref, with side-effects (e.g. converting `@mentions` to user refs)" be reasonably accurate?

src/stores/Howto/howto.store.test.ts
```diff
@@ -0,0 +1,247 @@
+jest.mock('../common/module.store')
+import type { IHowtoDB } from 'src/models'
+import { FactoryComment } from 'src/test/factories/Comment'
+import { FactoryHowto } from 'src/test/factories/Howto'
+import { FactoryUser } from 'src/test/factories/User'
+import type { RootStore } from '..'
+import { HowtoStore } from './howto.store'
+
+async function factory(howtoOverloads: Partial<IHowtoDB> = {}) {
```
chrismclarke, line 9: Really nice to start adding some testing stubs for the stores! Especially as the logic is getting a bit on the complex side, so great to be able to build out some simpler unit tests

chrismclarke (approved):
My first reaction was not quite as dystopian as I would have expected for PR 1984, or maybe just the current state of the world has hardened me to what would otherwise be some radical change proposed (unit testing??? really?!)

The test additions look great, nice to be able to stub out more mocks and tests for some of the more complex things going on. 

I don't think I fully understand the flow/logic of how the mentions get handled, but I can see how the code changes work to ensure the mentions are in place when updating so I'd say makes sense to include for now and revisit later. Otherwise I've just added a few minor non-blocking nits but happy to approve overall

chrismclarke:
Hey @thisislawatts! Just to quickly check in, if you get a chance to review my minor comments in the next week or so that would be great, otherwise I'll just make some minor edits and merge in as I think the code is in pretty good shape otherwise

src/test/factories/Comment.ts
thisislawatts: My preference would be to keep the current naming convention but open to reviewing this at a later date if it proves hard for contributors to reason with.

thisislawatts:
@chrismclarke updated based on the feedback. I can merge and see this one once it's approved ✅

chrismclarke:
Thanks @thisislawatts! will get the release shipped asap

PR #1986, opened 2022-10-16
chore: enforce sorting on class attributes

thisislawatts:
- chore(deps): introduces eslint sort rule
- chore(formatting): apply sort order rule to classes

PR Checklist

PR Type

## Description

> ESLint rule for enforcing consistent ES6 class member order.
https://github.com/bryanrsmith/eslint-plugin-sort-class-members

Because we shouldn't expect humans to care too much about this 👼🏻


chrismclarke (approved):
Sounds reasonable to me, VSCode lint looks clear enough to me that I don't think devs should have a major problem with it and yeah nice to have a bit more consistency for things like properties and render method

image

PR #1357, opened 2021-12-28
[FIX] #1238 howto search cleaning after navigation

jparadasb:
PR Checklist

- [x] - Latest `master` branch merged
- [x] - PR title descriptive (can be used in release notes)

PR Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

I added a flush of the search value in the howToStore when the component invokes `componentWillUnmount` and the respective e2e in cypress test.

## Git Issues

Closes #1238 

## Screenshots/Videos

---

## What happens next?

Thanks for the contribution! We try to make sure all PRs are reviewed ahead of a monthly dev call (first Monday of the month, open to all!).

If the PR is working as intended it'll be merged and included in the next platform release, if not changes will be requested and re-reviewed once updated.

If you need more immediate feedback you can try reaching out on slack in the `platform-dev` channel.

PR #1995, opened 2022-10-29
fix: keep filters in sync with url after back/forward navigation on how-to page

### LINKED ISSUE #1358: [bug] howto filtering by tags persists after going back in history browser
**Describe the bug**
When using the how-to filter by tags after selecting at least two and then going back forward in the browser history, the filters still appear selected although not present in the URL.

**To Reproduce**

1. Go to 'How-to'
2. Select one or more filters tags
3. Press the history back button on the browser
4. The content still is filtered and it appears selected in the SelectBox 

**Expected behavior**
When the filter by tags is selected if I go one time back in the browser history the last tag selected shouldn't appear and the content shouldn't be filtered by that tag, should have consistency with what the URL is showing.

**Screenshots**
gif

**Additional context**

Ulian:
PR Checklist

PR Type

## Description

After using the back and forward navigation buttons the URL and the filters were out of sync, with this fix instead of updating `window` directly we use react-router which keeps track of the props for the page.

In `componentDidUpdate` we received the new props and updated the storage to update the filters if the query params get updated.

Also, I apply the same fix on #1357 to the category filter.

## Git Issues

Closes #1358 

## Screenshots/Videos

https://user-images.githubusercontent.com/15958372/198843730-c5ad53de-cf4b-4f88-b15d-a81ce741da8b.mp4

In #1358 there is a gif with the old behavior.


davehakkens:
Thanks for the PR @Ulian!
We have another little issue related to this:
If you click a how-to to open and click the `back button`to go to overview it doesnt show anything. It requires a refresh (example below)
You think this fix could be added to this PR?

bug


And it seems you are not passing our lint test. I think your commit message should be `fix:` not `fix(how-to-list): `
Would you be able to change that?

Ulian:
I change the commit message, now it does pass the lint!

--- 

I saw that issue yesterday while I was testing my changes. After many attempts, I found a solution but I'm not happy with it (that's why I didn't commit it). It also happens with Create how-to page, if you go back nothing shows.

The solution I found was to call in the constructor of `VirtualizedFlex` `this.generateRowData(props)` inside a `setTimeout`, for some reason `setState`  is not working if is not inside a `setTimeout`.

```
  constructor(props: IProps) {
    super(props)
    this.state = { data: [], dataRows: [], totalColumns: -1 }

    setTimeout(() => this.generateRowData(props), 0)
  }
```

The main reason for the bug I think is that react-router is not destroying the `HowtoList` component because they share the start of the url (/how-to).

Will try to find another solution, maybe someone with more experience with class components can give me a hint.

Ulian:
Fixed @davehakkens.

Will wait for feedback to see if is a good solution 😆. I wasn't able to find out why setState does not work in the constructor.

davehakkens:
Works nice here, thanks!
Let me check in with our devs to see how they feel about it :)

packages/cypress/src/integration/howto/search.spec.ts
```diff
@@ -3,17 +3,56 @@ describe('[How To]', () => {
...
-    cy.get('[data-cy=how-to-search-box]')
-      .invoke('val')
-      .then((searchText) => expect(searchText).to.equal(''))
+      cy.get('[data-cy=how-to-search-box]')
+        .invoke('val')
+        .then((searchText) => expect(searchText).to.equal(''))
+      cy.get('[data-cy=category-select]').should('have.value', '')
+    })
+
+    it('should remove category filter after back navigation', () => {
```
chrismclarke, line 25: thanks for adding all these tests, it's a tricky feature to get right so definitely nice to be able to avoid regression errors

src/pages/Howto/Content/HowtoList/HowtoList.tsx
```diff
@@ -60,23 +71,8 @@ export class HowtoList extends React.Component<any, IState> {
...
-      const searchQuery = searchParams.get('search')?.toString()
-      if (searchQuery) {
-        this.injected.howtoStore.updateSearchValue(searchQuery)
-      }
-      const referrerSource = searchParams.get('source')?.toString()
-      if (referrerSource) {
-        this.injected.howtoStore.updateReferrerSource(referrerSource)
-      }
-    }
+    this.syncUrlWithStorage()
```
chrismclarke, line 75: thanks for moving this into a function, much tidier for sure

chrismclarke (approved):
Thanks @Ulian , changes look great to me!

I couldn't quite figure out exactly what was happening with the constructor method and whether we were getting the behaviour desired or not, but I've tried adding a `componentDidMount` binding in 964fbf84fb4f8b13e9d8f0bf40fca442f913a798 which should hopefully ensure things are calculated each time the page navigates forwards/back (now possibly re-rendering an extra time, but could be optimised in the future with better memorization)

@davehakkens - could you give one more check to see if things are behaving how you would expect them to? If so I'd say good to go

davehakkens:
thanks for checking @chrismclarke, looks good here
Also there was a bounty for making the PR @Ulian and reviewing @asheerrizvi 
More info here on how to claim it: https://onearmy.github.io/community-platform/Contributing/bounties
(let me know if this is unclear, we just starting using this new bounty system :))

PR #1998, opened 2022-10-31
Chore: Update emulators to use latest pp data

chrismclarke:
PR Checklist

PR Type

## Description
This started out as a simple update for emulator data but quickly realised the emulator build process wasn't working correctly. It turns out the emulators and backend functions were using different versions of various firebase dependencies, which had different sets of breaking changes to resolve. So as well as updating the emulator data I've added various minor fixes for the backend functions and emulator build processes, namely:

**Emulator Updates**
- Update seed data when running with docker emulators
- Fix bug where docker build was not working.. This hasn't been working since we added SEO functions as the docker build expects an index.html file from the frontend. Added a small workaround to populate a placeholder index file (instead of building the full frontend), as we're the updated index can be passed at runtime anyway so just need to get past the initial build/test.
- Add support for passing buildtime args to the process, used to pin specific versions of firebase-tools installed

**Functions updates**
- Update firebase dependencies
- Refactor firebase imports to avoid creating all services (e.g. firestore, realtime, auth) in single file
- Refactor firebase imports to use `getFirestore` style imports for updated deps
- Fix bug where realtimeDB emulator would not work (not super important as we rarely use, but was breaking build)
- Add tests to check database emulator working

I've also added some additional notes below, more for documentation in case anyone's looking through the code and wants to understand a bit more about the changes in this PR.

## Review Notes
I manually ran the github action with current branch to check runs as expected. Successful result here:
https://github.com/ONEARMY/community-platform/actions/runs/3358211942/jobs/5564858602

Ideally it would be good to test that somebody can also run the emulators locally and see the updated howtos via 
```
yarn start:emulated:docker
```
It may take a few minutes to download the latest image from dockerhub. Eventually you should see a message with the updated data listed.
image

Assuming the emulators are working as intended, the only slight room for knock-ons is with the deployed functions - everything looks fine when running locally however there can sometimes be slight differences running in the production google environment, so it would be good to keep an eye on logs after deployment

## Additional Notes
**Issue**
Emulator not compiling functions, not able to process bootstrap scripts

**Diagnosis**
Firebase made various changes to how emulators worked from v9 ->v10  ->v11 versions of firebase-tools. This was hard for us to detect as our emulators were set to download the latest version of firebase-tools (except also used caching so what was 'latest' would also just be whatever was cached)... so we didn't really know what version would be installed on various test runs

**Fix**
I've fixed the docker build to automatically pick the same version of firebase-tools as used in the backend which fixes the core issue. This enforces using the older v9 firebase-tools which is what is currently in the backend. I also added a fix for some of the 

At the same time I tested upgrading firebase-tools to v11. There appeared to be a few minor breaking changes which I've tried to implement. I had hoped to split this PR into 2 parts, although things are a bit too interconnected to be done easily so 

Looking through the admin docs it appears no breaking changes to note, so should just work:

https://github.com/firebase/firebase-admin-node/releases (migrate v9.9 -> v11.2)
https://github.com/firebase/firebase-functions/releases (migrate v3.2 -> v4.0)

Closes #

## Screenshots/Videos
**Emulators running locally with updated howto data**
image

---

## What happens next?


davehakkens (approved):
smooth

PR #2011, opened 2022-11-25
chore(deps): remove unused use-debounce

thisislawatts:
PR Checklist

PR Type

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,6 @@
     "styled-system": "^5.1.5",
     "theme-ui": "^0.14.1",
     "ts-node": "^10.4.0",
-    "use-debounce": "^7.0.1",
     "web-vitals": "^1.1.2",
     "workbox-window": "^6.1.5",
     "yup": "^0.32.9"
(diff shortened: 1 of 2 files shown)


Ulian:
It seems to be used in [OsmGeocoding.tsx](https://github.com/ONEARMY/community-platform/blob/chore/deps-remove-unused-use-debounce/packages/components/src/OsmGeocoding/OsmGeocoding.tsx#L4)

thisislawatts:
@Ulian good catch! That file is in another yarn workspace, so it's dependencies are declared in [./packages/components/package.json](https://github.com/ONEARMY/community-platform/blob/master/packages/components/package.json). The copy I am removing here is from the root package.json definition.

Ulian:
For some reason on my localhost it only shows me one package.json when I search on all files 🤔 My bad!

chrismclarke (approved):
Thanks for double checking these @thisislawatts and @Ulian , looks good to me

PR #2029, opened 2022-12-12
Fix/chunk load errors

chrismclarke:
PR Checklist

PR Type

## Description
ChunkLoad errors are those which are seen when a user is trying to load a page within the platform which no longer exists on the server. This is usually due to cases where the site is loaded from a cache but not all routes have been stored, and the user tries to visit an uncached page with reference to the old code instead of fetching fresh updates. For the user they will just see a white page as the javascript fails to load.

Unpicking the exact behaviour is quite difficult, however this PR aims to improve on the current handling in the following ways:

- Add a custom error handler specifically for cases when chunkLoadError occurs (white screen of death). 
This will attempt to reload the page one time to see if that fixes, and if not will display a message to the user.

- Add metadata to lazy loaded pages to provide specific names to code chunks produced in main build. This makes it easier to test this issue locally and should provide some extra debugging info in the future

- Tidy error handler folder structures

## Testing Notes
In order to see the user interface shown you will need to recreate the environment where the browser is trying to load code from a file that does not exist. One way to do this is:
- Build the app locally via `yarn build`
- Server the built version of the app via `npx serve build`
- Visit the local version of the app (likely running on HTTP://localhost:5000)
- Visit at least one of the main pages, e.g. academy or howtos
- Rename one of the javascript files from the `build/static/js` folder, e.g `Maps.{someHashCode}.js` ->  `Maps.{someHashCode}.renamed.js`
- Try to visit the maps page

The platform will know what JS it expects to load, however will not be able to find it resulting in a ChunkLoadError. You should likely see the screen blink whilst the error handler automatically attempts a reload to see if it fixes, and if not it will show the error message. 

Note - the reload button will not fix the error in this case as the file has been manually moved, but may fix on the server when it is a case of outdated cache/service worker (TBD)

As with all code involving service workers, it will only take effect after n+1 releases (first release will store the code in cache, then after next release it should apply the custom error handling if chunkLoad error occurs for any users that used the site in the first release).  So it would be good to check back in a few weeks to see if many users are still reporting the white screen issue. I've also updated the automated error reporting that goes to sentry to track whether the error persists reloads for future debugging.

## Git Issues
Related to #1680
Closes #

## Screenshots/Videos
**Custom error page - will show whenever chunkLoadError received after first attempting a page reload**
image

**Build chunks - now code related to specific lazy-loaded pages are named for easier debugging and reference (numbered chunks will relate to 3rd party code imports)**
image

---

## What happens next?


src/common/Error/handlers/Reloader.tsx
```diff
@@ -0,0 +1,38 @@
...
+/** Get a localstorage value with an expiry date, returning value only if not expired */
+function getWithExpiry(key: string): string | null {
+  const itemString = window.localStorage.getItem(key)
+  if (!itemString) return null
+
+  const item = JSON.parse(itemString)
+  const isExpired = new Date().getTime() > item.expiry
+
+  if (isExpired) {
+    localStorage.removeItem(key)
```
evakill, line 33: thoughts about using sessionStorage for this? removes need for expiry

evakill (approved):
tested locally and the error boundary works as expected! LGTM

src/common/Error/handlers/Reloader.tsx
chrismclarke: Oh nice idea. I think we might still need the expiry because session storage only clears when the tab is closed and I know some people (me) who pretty much keep the same tab open forever. 

But using session storage instead of local at least will ensure it gets cleared if we ever rename the key, localhost dev etc.
Just updated

PR #2039, opened 2022-12-16
feat(settings): add email frequency setting

### LINKED ISSUE #2028: Add mail notifications setting to user profile (frontend)
### Description

For the upcoming email notifications we need a dropdown in the user profiles where users can select to receive updates.
We're making a notifications digest with all the recent unseen updates.
The dropdown needs to contain the 4 options:
- Never
- Daily
- Weekly
- Monthly

### Page related

Will be used in : [/settings](https://dev.onearmy.world/settings)

### Mockup

mail notifications

evakill:
PR Checklist

PR Type

## Description

This PR adds a new section, Email notifications, to the user settings page. This section has a dropdown with options `never`, `daily`, `weekly`, `monthly` that sets the `user.notification_settings.emailFrequency` field.

## Git Issues

Closes #2028 

## Screenshots/Videos
demo:
- dropdown selection with four selectable options
- setting persists on save
- dropdown defaults to saved setting on load
- setting does not persist if not saved

https://user-images.githubusercontent.com/37253846/208171087-86239db9-c76e-49ab-aec4-92610b0baef1.mov

---

## What happens next?


src/models/user.models.tsx
```diff
@@ -101,11 +101,18 @@ export const NotificationTypes = [
 
 export type NotificationType = typeof NotificationTypes[number]
 
+export enum EmailNotificationFrequency {
+  NEVER = 'never',
```
chrismclarke, line 105: nit(blocking)
Could we change `Never` to an empty string? That way it can be picked up in the same way as users who have never set a preference when checking in the backend whether we need to compile a list of notifications for the user

src/pages/Settings/content/formSections/EmailNotifications.section.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <Box mt={2} sx={{ width: '40%' }}>
+          <Field name="notification_settings.emailFrequency">
+            {({ input }) => {
+              return (
+                <Select
+                  options={emailFrequencyOptions}
+                  defaultValue={emailFrequencyOptions.find(
+                    ({ value }) =>
+                      value ===
+                        this.props.notificationSettings?.emailFrequency ??
```
chrismclarke, line 48: thanks for adding the check for users without any settings (still don't have an nice way to migrate all user profiles when we add new fields). 

nit(blocking)
I like the idea that this will get users without notifications enabled to automatically opt in next time they save their profile, although I think probably to keep things a bit clearer it would be better to make the default value never and we can organise a migration in the future to set another default value to all users if required.

chrismclarke (changes requested):
Looks pretty good to me! A few minor nit-picks to resolve but then I'd say good to go

davehakkens:
looks nice @evakill!

davehakkens (approved):

evakill:
thanks for the reviews! applied your comments @chrismclarke

chrismclarke:
Thanks @evakill , all looks great to me

One last quick thought @davehakkens  - do we want this to be viewable by everyone or just beta-testers/admins to begin with? The dropdown doesn't do anything on its own at the moment and will probably be some time next month we get the email send integration in

chrismclarke (approved):

davehakkens:
Good point @chrismclarke, might be better to start with only visable for with _beta-testers_ and _admins_ if we think it will take a bit more time. Than we can just merge it in now. You know how to add this auth wrapper @evakill?

evakill:
@davehakkens i can give it a go later this week!

chrismclarke:
Just a one-liner so I've added in c6d77b3 . 
I don't think there's actually a way to limit anything to multiple roles (e.g. both beta-testers and admins), so have just set to beta-testers for now and any admins that want to test can assign themselves the role (users can have multiple).

Nice to get this one in, will be away from tomorrow so likely will have to wait until the new year to handle the backend work for sending the emails

PR #2043, opened 2022-12-18
fix: address button inconsistency

thisislawatts:
PR Checklist

PR Type

## Description

It was the misalignment of text and icons in the folllowing screengrab that started me down this path. 

Screenshot 2022-12-21 at 11 40 00

* Remove unused `medium` variant in favour of a sensible default complimented by a small and large variant.
* Remove snowflake transform effect from `Create new How-to` button.
* Introduced documentation for all variants within Storybook

Questions:
* Large is not currently being used anywhere, should we just remove it?
* Inconsistent cases, some buttons are lowercased `action` and others are capitalized `Action`. Do we have a clear preference here? 

## Git Issues

Closes #

## Screenshots/Videos

Screen Shot 2022-12-21 at 11 41 17

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -2,72 +2,137 @@ import * as React from 'react'
 import { Icon } from '../Icon/Icon'
 import type { IGlyphs } from '../Icon/types'
 import type { ButtonProps as ThemeUiButtonProps } from 'theme-ui'
-import { Button as ThemeUiButton } from 'theme-ui'
-import styled from '@emotion/styled'
+import { Button as ThemeUiButton, Flex, Text } from 'theme-ui'
 
 // extend to allow any default button props (e.g. onClick) to also be passed
-export interface IBtnProps extends React.ButtonHTMLAttributes<HTMLElement> {
+interface IBtnProps extends React.ButtonHTMLAttributes<HTMLElement> {
   icon?: keyof IGlyphs
   disabled?: boolean
-  translateY?: boolean
   small?: boolean
-  medium?: boolean
   large?: boolean
-  hasText?: boolean
+  showIconOnly?: boolean
 }
 
-const small = (props: IBtnProps) =>
-  props.small
-    ? {
-        padding: '4px 10px!important',
-      }
-    : null
+type ToArray<Type> = [Type] extends [any] ? Type[] : never
+type AvailableButtonProps = ToArray<keyof BtnProps>
 
-const medium = (props: IBtnProps) =>
-  props.medium
-    ? {
-        padding: '6px 12px!important',
-      }
-    : null
-
-const large = (props: IBtnProps) =>
-  props.large
-    ? {
-        padding: '8px 14px!important',
-      }
-    : null
-
-const translateY = (props: IBtnProps) =>
-  props.translateY
-    ? {
-        '&:hover': {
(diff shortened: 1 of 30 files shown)


davehakkens:
Looks proper!
I'll dump a few screenshot below of things that changed.

Account setting in profile are on one row
afbeelding

Create how-to buttons are a bit unaligned
afbeelding

In general the `save profile` and `publish` button feels a bit small compared to` save as draft `and the notification messgae below 
afbeelding

Trashcan button on Contacts in userprofiles is stretched
afbeelding


Comments button went to the left side, was middle before
afbeelding

thisislawatts:
Thanks for catching those @davehakkens, much appreciated. I've pushed fixes for all of those aside from the Publish button.

What do you think about the following screen grab, where the `large` button variant is used for the primary button `Publish`. 

A few reasons behind this proposal:
* We are not using the Large button in many places, so lets use it or remove it!
* Using size _and_ colours to indicate priority is helpful for folks who don't see colour, whether that is due to the way their visual processing works or their device is configured. (For example: My mobile device only renders in black & white)

Screenshot 2022-12-22 at 14 44 50

davehakkens:
haha you got me convinced at `We are not using the Large button in many places, so lets use it `sounds like a very good reason. 

But I do think it makes sense, indeed for priority

thisislawatts:
Updated to use large variant on following pages/processes
* Login/Register
* Add/edit How to
* Add/edit Research 
* Add/edit Research Update
* Account Settings

packages/components/src/Button/Button.tsx
```diff
@@ -2,72 +2,115 @@ import * as React from 'react'
...
 // extend to allow any default button props (e.g. onClick) to also be passed
 export interface IBtnProps extends React.ButtonHTMLAttributes<HTMLElement> {
   icon?: keyof IGlyphs
   disabled?: boolean
   translateY?: boolean
   small?: boolean
-  medium?: boolean
   large?: boolean
-  hasText?: boolean
+  hideText?: boolean
```
chrismclarke, line 14: Much nicer to have properties that are falsy by default, thanks for tidying this.

(nit) non-blocking
hideText seems a bit of a weird one to me, if we don't want text then probably better to just remove the text. But I'm guessing this is more like an `iconOnly` style variant which handles minor padding/margin tweaks for display without text. Probably best to leave for now to avoid too many knock-ons

chrismclarke (approved):
Thanks @thisislawatts , nice reduction in code and feels sensible to me (particularly grateful for removal of tertiary styles and duplicate theming files.

I'll leave it to @davehakkens for any final comments relating to the additional style changes, but otherwise I'd say good to go

packages/components/src/Button/Button.tsx
thisislawatts: Oh, I like `iconOnly` I will change this property to align with that name. 

The reasoning behind this being an explicit flag rather than relying on the absence of text is that we should _always_ look to include instructive text for users who are not able to parse an icon. I am thinking of people with visual impairments. 

Although it's fair to say I haven't done the legwork to fully validate the current implementation meets [Web Content Accessibility Guidelines](https://www.w3.org/WAI/standards-guidelines/wcag/) (WCAG). Will circle back on this variant. Thanks for catching this.

davehakkens (approved):

PR #2064, opened 2023-01-07
Feat/how to redirect url when name changes

Kiebert:
PR Checklist

PR Type

## Description

The string array 'previousSlugs' is added to the 'How-to' item. 
Per 'How-to' item 3 slugs are saved. If a new slug is added the oldest slug will be removed from 'previousSlugs'.
The database schema and CACHE number is updated. 

## Git Issues
[1168](https://github.com/ONEARMY/community-platform/issues/1168)


src/stores/Howto/howto.store.tsx
```diff
@@ -300,6 +311,18 @@ export class HowtoStore extends ModuleStore {
       }),
     )
 
+    if (howToItem.previousSlugs === undefined) {
+      howToItem.previousSlugs = []
+    }
+
+    if (!howToItem.previousSlugs.includes(howToItem.slug)) {
+      howToItem.previousSlugs.push(howToItem.slug)
+      // remove oldest slug
```
evakill, line 320: why do we have to remove the oldest?

evakill (commented):
two questions, but otherwise looks good to me

src/stores/Howto/howto.store.tsx
Kiebert: Hi @evakill, 

The following was mentioned in the issue story: 
["We need to consider how much history we should keep for a single article."](https://github.com/ONEARMY/community-platform/issues/1168#issuecomment-1013943630)

I thought about how maybe times someone would change the title of the article, and this seems to me a valid amount.
Let me know if you want to change the amount?

evakill: thanks @Kiebert! i would think that if we want the idea of "permalinks" then we shouldn't clear any history. calling @davehakkens for executive decision-making here too

src/stores/Howto/howto.store.tsx
```diff
@@ -438,10 +461,21 @@ export class HowtoStore extends ModuleStore {
       // redefine howTo based on processing done above (should match stronger typing)
       const userCountry = getUserCountry(user)
 
+      // create previousSlugs based on available slug or title
+      const previousSlugs: string[] = []
+      if (values.slug) {
+        previousSlugs.push(values.slug)
+      } else if (values.title) {
+        const titleToSlug = formatLowerNoSpecial(values.title)
+        previousSlugs.push(titleToSlug)
```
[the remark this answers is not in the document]
chrismclarke: Really good point, although I'd say at this point from what I see the PR is only handling the storing of previous slugs, and not any logic for picking which to display when a user navigates to a howto with a given slug. So perhaps we can kick the can down the road on that for a follow-up PR?

I think for the same reason I would probably side on "let's just keep all previous slugs" in an array instead of limiting to a fixed number. Always easier to apply limits later

src/stores/Howto/howto.store.tsx
chrismclarke: I've with @evakill on this, I think we still haven't fully decided how we want to handle cases where one person creates a howto with a URL that matches a renamed howto from another author. So until we have a plan fprobably makes sense to just keep a full history.

src/stores/Howto/howto.store.tsx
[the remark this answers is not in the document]
evakill: > not any logic for picking which to display when a user navigates to a howto with a given slug

@chrismclarke i believe it does [here](https://github.com/ONEARMY/community-platform/pull/2064/files#diff-a2437a58b8d602deb7c8d0b921e348a6c2c561c7af93bcaeb0a13833b0e9541eR122-R130), no? 

even though this is an edge case it does break expected behavior, so i think would be good to at least have a plan before merging this PR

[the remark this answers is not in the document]
chrismclarke: oh whoops, don't know how I completely missed that! yeah you're right, by default it will just pick the first result found. I'd say not ideal, so agree good to hear what @davehakkens thinks.

Assuming we do keep all the previous howto slugs I'd probably again vote in favour of resolving any better behaviour as a follow-up issue, because I expect this will be quite a niche/edge case for quite some time (howtos aren't created very frequently, not renamed very frequently, and usually have a fairly specific title so less likely to duplicate).

src/stores/Howto/howto.store.tsx
davehakkens: yeh full history sounds good to me for now!

src/stores/Howto/howto.store.tsx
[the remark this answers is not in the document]
davehakkens: Thanks for the comment @evakill!
Agree with @chrismclarke, up for keeping an eye on it and make follow up issue if needed.

src/stores/Howto/howto.store.tsx
Kiebert: Hello @evakill, @chrismclarke and @davehakkens.
The PR is updated. All previousSlugs are preserved: full history is in effect.

chrismclarke (approved):
awesome, thanks for the updates @Kiebert , I think we're good to go!

PR #2073, opened 2023-01-23
chore: remove globals in favour of local declaration

thisislawatts:
PR Checklist

PR Type

## Description

The original file location suggested these objects were mocks, instead they are used as part of the UI presentation to end users. As they are only used in a single place, moving their declaration to the relevant component until having them in a shared location is required.


chrismclarke (approved):
Makes sense to me, thanks @thisislawatts

PR #2076, opened 2023-01-25
chore(deps): introduce linting to catch focused tests

thisislawatts:
PR Checklist

PR Type

## Description

Adds automated lining to catch use of `.only` in test suites. It is reasonable to use only when working on a specific test scenario, however can be easily missed by human reviewers. Let us make the robots work for us.


chrismclarke (approved):
Amazing! I couldn't tell you how many times I've marked something as only while testing and then found myself wondering a month later why half a test suite was no longer running

..._Starts copy/paste to all other repos..._

Issue #1618, opened 2022-04-25
Add view count to How-to and Research


davehakkens:
### Description

We want to have a view count on our How-to and Research pages. 

### Page related
[Research](https://community.projectkamp.com/research/can-we-make-an-elevation-height-map-of-the-land#update_3) and [How-to](https://community.preciousplastic.com/how-to/beads-mould---a-lot-of-them) pages. 

### Build suggestion
- It will be build using the aggregations.
- Ideally there is some sort of logic that only counts one visit per **hour**. (To prevent it being mis-used by refreshing 200 times in a row)
- Should be released for `beta-testers` first. So we can test it 

### Mockup

![Views How-to](https://user-images.githubusercontent.com/13672737/165151808-8dd334de-1cd5-4cab-9dd2-892c67e180c8.jpg)
![Views research](https://user-images.githubusercontent.com/13672737/165151817-da9b567f-a9fd-42c3-b88c-58fb33f8569e.jpg)

### Icon 
[download](https://drive.google.com/file/d/1Ljyg74WFQTAxffaQBpr3Kb7ci443kaht/view?usp=sharing)

davehakkens:
Want to take on another issue @amandaqiu1 or @ryanwbass? 👆

ryanwbass:
Hey @davehakkens  I will take it! Go ahead and assign.

davehakkens:
sweet!

chrismclarke:
@ryanwbass 
Thanks for taking this on. There might be a bit of added challenge with this one with the best place to actually store this information. The natural place to think would be the howto itself, adding a `viewCount` property and just increasing by 1 any time the howto page is loaded except this will mean howto docs are constantly updated and the entire doc will need to be synced across all users many times (which is a costly set of db read operations)

There's also the not-unlikely scenario that a howto could be viewed by multiple users at similar times, meaning that any client-side updates could easily become out of sync (e.g. two users both increment the count from 2->3 before having the chance to sync the others' update).

So I think I'd recommend a few things:

1. Use a new collection for tracking raw stats. E.g. add a `howto_stats` or similar endpoint to `shared\db.ts` that will hold the viewCount for each doc.

[...]

PR #2086, opened 2023-02-02
1618 add view count to how to and research

AlfonsoGhislieri:
PR Checklist

PR Type

## Description

Implements a simple way for views to be tracked, using sessionstorage to limit views being inflated by refreshing page. 

## Git Issues

Closes #
Add view count to How-to and Research #1618

## Screenshots/Videos

---

## What happens next?


AlfonsoGhislieri:
@davehakkens the link for the icon download wasn't working so added the star icon in the meantime, I can add the correct icon whenever you have time and can upload it. 

The PR also fixed some bugs I noticed with the download counter which now should be all sorted, state wasn't being updated properly.

I also ended up sticking with using`set` since I couldn't quite get `update`  to work and I also realized that using update function we would run into issues with old howtos and research posts that didn't have the `total_views` field. Not super DRY with both howto and research stores having essentially the same increment function could extract this logic and share it across both stores. But it should all be working :-)

davehakkens:
thanks @AlfonsoGhislieri  
Looks good :) Here you can download the icon 
[views.svg.zip](https://github.com/ONEARMY/community-platform/files/10573620/views.svg.zip)

Could you wrap it in our AuthWrapper so it's only available for `beta-testers`.
([here](https://github.com/ONEARMY/community-platform/pull/2039/files) an example of how it's implemented)

packages/components/src/ViewsCounter/ViewsCounter.tsx
```diff
@@ -0,0 +1,33 @@
+import { Text } from 'theme-ui'
+import { Button } from '..'
+
+export interface IProps {
+  viewsCount: number
+}
+
+export const ViewsCounter = (props: IProps) => {
+  return (
+    <>
```
asheerrizvi, line 10: We don't need a fragment here. It is already returning a single element which is `Button`

packages/components/src/ViewsCounter/ViewsCounter.tsx
```diff
@@ -0,0 +1,33 @@
...
+        sx={{
+          backgroundColor: 'transparent',
+          fontSize: 2,
+          ml: 2,
+          background: 'softyellow',
+          borderColor: 'softyellow',
+          opacity: 1,
+          pointerEvents: 'none',
+          cursor: 'none',
+        }}
```
asheerrizvi, line 23: Let's not add more styling than necessary, I would suggest using the base `Button` styling. Since the background color is slightly different we can probably add it as a `variant` but things like `font-sze`, `opacity`, `pointer-events` and `cursor` should be left to the `Button` component to decide.

src/pages/Howto/Content/Howto/HowtoDescription/HowtoDescription.tsx
```diff
@@ -106,239 +104,237 @@ export default class HowtoDescription extends PureComponent<IProps, IState> {
     }
   }
 
-  public render() {
-    const { howto, loggedInUser } = this.props
+  useEffect(() => {
+    setFileDownloadCount(howto.total_downloads)
+    setViewCount(howto.total_views)
+    incrementViewCount()
+  }, [howto._id])
```
asheerrizvi, line 111: Let's not use an effect to initialize state. This is considered an undesirable pattern see more:

https://beta.reactjs.org/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes

src/pages/Howto/Content/Howto/HowtoDescription/HowtoDescription.tsx
```diff
@@ -44,59 +48,53 @@ interface IProps {
...
 }
 
-interface IInjected extends IProps {
-  howtoStore: HowtoStore
-}
-
-interface IState {
-  fileDownloadCount: number | undefined
-}
+const HowtoDescription: React.FC<IProps> = ({
```
asheerrizvi, line 51: `React.FC` is generally discouraged, see the `Why is React.FC discouraged` section here:

https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components

src/pages/Howto/Content/Howto/HowtoDescription/HowtoDescription.tsx
```diff
@@ -106,239 +104,237 @@ export default class HowtoDescription extends PureComponent<IProps, IState> {
...
+    setFileDownloadCount(howto.total_downloads)
+    setViewCount(howto.total_views)
+    incrementViewCount()
+  }, [howto._id])
 
-    const iconFlexDirection =
-      emStringToPx(theme.breakpoints[0]) > window.innerWidth ? 'column' : 'row'
-    return (
+  const iconFlexDirection =
+    emStringToPx(theme.breakpoints[0]) > window.innerWidth ? 'column' : 'row'
```
asheerrizvi, line 114: I think this can be moved to outside the component without any issue, move it back if you find any issue though.

src/pages/Howto/Content/Howto/HowtoDescription/HowtoDescription.tsx
```diff
@@ -106,239 +104,237 @@ export default class HowtoDescription extends PureComponent<IProps, IState> {
...
-              <Image
-                loading="lazy"
-                src={TimeNeeded}
-                height="16"
-                width="16"
-                mr="2"
-                mb="2"
+            </Box>
+          )}
+          <Box sx={{ flexGrow: 1 }}>
```
asheerrizvi, line 171: Instead of using `flex-grow: 1` here, let's remove `justify-content: space-between` from the parent of these buttons/link and add a `gap` as well. We can then remove left margin from our buttons as well.

src/pages/Research/Content/ResearchDescription.tsx
```diff
@@ -40,6 +47,24 @@ const ResearchDescription: React.FC<IProps> = ({
...
+      const updatedViewCount = await store.incrementViewCount(research._id)
+      setViewCount(updatedViewCount)
+      addIDToSessionStorageArray('research', research._id)
+    }
+  }
+
+  useEffect(() => {
+    setViewCount(research.total_views)
+    incrementViewCount()
+  }, [research._id])
```
asheerrizvi, line 67: Please check the `HowToDescription` comment for the effect.

src/pages/Research/Content/ResearchDescription.tsx
```diff
@@ -89,6 +114,9 @@ const ResearchDescription: React.FC<IProps> = ({
               />
             </Box>
           )}
+          <Box style={{ flexGrow: 1 }}>
```
asheerrizvi, line 117: Please check the `flex-grow` comment within `HowToDescription` component.

AlfonsoGhislieri:
@asheerrizvi thanks for the feedback! Made all the changes and also converted the `viewsCounter` component away from being a modified button to being it's own component. Let me know if anything else needs changing :-)

src/pages/Research/Content/ResearchDescription.tsx
```diff
@@ -40,6 +44,23 @@ const ResearchDescription: React.FC<IProps> = ({
...
+    if (!sessionStorageArray.includes(research._id)) {
+      const updatedViewCount = await store.incrementViewCount(research._id)
+      setViewCount(updatedViewCount)
+      addIDToSessionStorageArray('research', research._id)
+    }
+  }
+
+  useEffect(() => {
+    incrementViewCount()
+  }, [research._id])
```
asheerrizvi, line 63: I think if you can do a lazy initialization for `useState` and not be in need of this effect:

https://kentcdodds.com/blog/use-state-lazy-initialization-and-function-updates

src/pages/Howto/Content/Howto/HowtoDescription/HowtoDescription.tsx
```diff
@@ -106,239 +103,234 @@ export default class HowtoDescription extends PureComponent<IProps, IState> {
     }
   }
 
-  public render() {
-    const { howto, loggedInUser } = this.props
+  useEffect(() => {
+    incrementViewCount()
+  }, [howto._id])
```
asheerrizvi, line 108: See comment on `ResearchDescription` component please:

https://kentcdodds.com/blog/use-state-lazy-initialization-and-function-updates

src/pages/Research/Content/ResearchDescription.tsx
AlfonsoGhislieri: The issue is as far as I know you can't pass an  asynchronous function to do a lazy initialization for `useState` it'll set state to a promise rather than the return value of the function. I had already tried to do this initially and had to revert to `useEffect` for that reason.

asheerrizvi: Ok, I am approving this. Can we create a new issue for removing this effect and work on it separately?

asheerrizvi (approved):

davehakkens:
Thanks @AlfonsoGhislieri Its on live now. 
Looking forward to keep an eye on those numbers :)

AlfonsoGhislieri:
> Thanks @AlfonsoGhislieri Its on live now.
> Looking forward to keep an eye on those numbers :)

Great! Let me know if there’s any issues or anything you’d like changed :)

davehakkens:
Will do. One question @AlfonsoGhislieri  to make sure I got it correct: Whats the formula you used to calculate a view?

AlfonsoGhislieri:
A view gets added each time a resource or howto page is opened or refreshed. However, to limit bloating of views each time you view a page the id is stored in [Session storage ](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) and this is checked every time you load a page. This all gets reset after the session ends (when the tab or browser gets closed). So if you’ll view the page again after that a new view will be added. 

Hope that makes sense!

davehakkens:
makes sense! 
And are all views counted now. 
Or are they only counted from `beta-testers`?

AlfonsoGhislieri:
Right now it counts all views!

davehakkens:
perfect!

davehakkens:
Are you sure about this @AlfonsoGhislieri 
I've been keeping an eye and views are very low.
In fact slower then download counts which seem impossible to me?
(got to view before you can download?)


views

AlfonsoGhislieri:
Damn well this is quite confusing, it definitely works like intended when I run it locally. I just tested it multiple times to make sure, the views get incremented correctly when you open the page even just as a normal user (either loading the url or by clicking on the howto/research directly on the main page).

But you are right this clearly looks something is not working like it should, I have no clue as to why this is the case though since everything appears to be working normally when I test it.  

Is there any chance you could make my account a beta-tester so I can have a look on the production side if I can figure something out?

davehakkens:
Yeh not sure either. And then there is caching always messing things up as well.  🙈
What is your username on the platform?
Could optionally also give you read access to firebase so you can see how much views are actually stored 
afbeelding

AlfonsoGhislieri:
Cryptiks on the platform

Ah so the views are actually being correctly incremented in the backend but just not rendering correctly? Even more confused hmm...

Happy to have a look at the firebase as well, might help understand what's up

davehakkens:
ok. invited as beta tester.
Also invited your email to firebase.
You can find the data here:

firestore>v3_howtos> All ID's are a howto> views are at the bottom.
(easiest way to filter is search for username and find their howto's)

afbeelding

AlfonsoGhislieri:
I just had a look at the howto on my end and the number of downloads and views was quite different, are yours still the same as in your earlier picture?

image

davehakkens:
6 views now..
afbeelding

Issue #2061, opened 2023-01-06
[bug] It wants me to save when I haven’t changed anything


MIfoodie:
**Describe the bug**
On the page to edit your profile it wants me to save even though I didn’t change anything.
**To Reproduce**
Steps to reproduce the behaviour:

1. Go to your profile settings
2. Click on another link on the page such as the map.
3. It will ask you if you want to save even though you haven’t changed anything.

**Expected behaviour**
i expected there to be no pop up.

davehakkens:
thanks for posting these issues @MIfoodie. A bit weird indeed This would be a good one to pick up for someone!

Kiebert:
Hi @davehakkens you can assign this to me.

davehakkens:
how is this one going @Kiebert, need any help?

Kiebert:
Hi @davehakkens I haven't spent a lot of time on this bug.
Tomorrow afternoon I should have time so let's see.

PR #2093, opened 2023-02-16
fix: settings with unsavedchanges dialog

Kiebert:
PR Checklist

PR Type

## Description

Replaced the standard Prompt with the UnSavedChangedDialog component.

## Git Issues

[#2061](https://github.com/ONEARMY/community-platform/issues/2061)


davehakkens:
Thanks @Kiebert!
You want to give a try to see @MIfoodie to see if this fixes your problem?

src/pages/Settings/SettingsPage.tsx
```diff
@@ -186,8 +186,8 @@ export class UserSettings extends React.Component<IProps, IState> {
           const heading = user.profileType ? 'Edit profile' : 'Create profile'
           return (
             <Flex mx={-2} bg={'inherit'} sx={{ flexWrap: 'wrap' }}>
-              <Prompt
-                when={!this.injected.userStore.updateStatus.Complete}
+              <UnsavedChangesDialog
+                uploadComplete={this.injected.userStore.updateStatus.Complete}
                 message={
                   'You are leaving this page without saving. Do you want to continue ?'
                 }
```
thisislawatts, line 193: **question:** @Kiebert did you look at introducing a failing test that captures this bug, which then passes as a result of the changes you have made here?

src/common/Form/UnsavedChangesDialog.tsx
```diff
@@ -39,7 +41,7 @@ export const UnsavedChangesDialog = memo((props: IProps) => {
         render={() => null}
       />
       {/* Render a blocking screen prompt to prevent user leaving page via react navigation */}
-      <Prompt when={shouldPromptUnsavedChanges} message={CONFIRM_DIALOG_MSG} />
+      <Prompt when={shouldPromptUnsavedChanges} message={message} />
```
thisislawatts, line 44: **praise:** Great to see this component extended for use in another place in the Platform. Thanks @Kiebert ✨

src/pages/Settings/SettingsPage.tsx
Kiebert: Hi @thisislawatts. I think I can create a test like that.

davehakkens:
Did a little test here, but not sure what changed with the current live site. 
Could you explain or share some screenshots of the change @Kiebert?

Kiebert:
@davehakkens The test.app website expired. I will first write the test upcoming weekend. 
After that I will check the newly generated web.app.

src/pages/Settings/SettingsPage.tsx
Kiebert: Hi @thisislawatts. I added two tests and when the confirmation is triggered is produces a log in the loglist of Cypress.
Is there another way to check is the confirmation window is active or not?

Kiebert:
> Did a little test here, but not sure what changed with the current live site. Could you explain or share some screenshots of the change @Kiebert?

I added the tests. 
So after this change the confirmation window should on only appear when you want to navigate to a different page but you have changed/edited data in the form.

When no edit has to be done in the form and you want to navigate to a different page the confirmation window should not appear.

src/pages/Settings/SettingsPage.tsx
thisislawatts: Let me take a look @Kiebert and see what I can find.

thisislawatts: Did you see the tests you added fail @Kiebert? I lifted just the 2 new test cases onto the `master` branch and ran them. They both passed, which suggests they are not capturing the buggy behaviour.

I've experimented with Cypress + browser alert modals before, but it looks like theres a path forward here:
https://stackoverflow.com/a/73086525

davehakkens:
> So after this change the confirmation window should on only appear when you want to navigate to a different page but you have changed/edited data in the form.

Thanks for clarifying. Here I currently get no confirmation window at all. When I fill in half my profile and click another page like `how-to` or `event` it just goes there and removes the data.
Is this different when you test @Kiebert ?

Kiebert:
> > So after this change the confirmation window should on only appear when you want to navigate to a different page but you have changed/edited data in the form.
> 
> Thanks for clarifying. Here I currently get no confirmation window at all. When I fill in half my profile and click another page like `how-to` or `event` it just goes there and removes the data. Is this different when you test @Kiebert ?

I am getting this confirmation window when I am changing for example my 'Display Name':
afbeelding

@davehakkens Which browser do you use? And what happens when you clear your browser cache?

davehakkens:
got it. works now. For some reason the local testing didn't work well but the preview link does.
Nice little fix!

src/pages/Settings/SettingsPage.tsx
thisislawatts: Nothing from my side, it looks good. Thanks for the follow up, we need one of the current maintainers to review. @davehakkens

davehakkens (approved):

PR #2067, opened 2023-01-13
spike: update node 16 to node 18

chrismclarke:
PR Checklist

- [ ] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Developer experience (improves developer workflows for contributing to the project)

## Description

Add support for using Node 18. Note this is quite tricky due to the number of incompatibilities with webpack/node at different versions, and the way the monorepo is setup to handle specific package resolutions (see linked issue). 

This PR aims to update everything to add compatibility for Node 18, and to migrate all of our cloud processes to use node 18 as a default. In addition it adds a prompt for local developers to try and match their environment to our recommendations

## Dev Notes
We have numerous workspaces that all are pretty much all built using webpack, but in different ways (frontend via react-scripts, functions manual implementation,  e2e tests via cypress, components storybook, docs via docusaurus). Some of these are setup to use webpack 4, others webpack 5. Trying to get everything upgraded to the newer version is a pain due to the large number of breaking changes, and the ways different packages either do or do not expose means to override configurations.

I think that I've now managed to update all the workspaces and resolve any identified issues to work with node 18 across the board. In the future we may want to consider restructuring the repo to keep better separation across workspaces (namely moving the platform frontend code away from the root folder), and possibly reconsidering some of our tooling options (particularly CRA which really failed in its support for webpack 5 and incredibly slow updates)

## Review Notes
The automated tests should have caught any critical issues when migrating, however would be good to specifically check:

- Checkout branch and try to run the platform (e.g. `yarn start`). It should prompt you to update version of node if you're not already running node 18. Be good to know how this experience is (are instructions clear? do things work?). Note the documentation page linked is updated in this PR so won't be entirely accurate

- Once merged would also be good to check for any deployment issues and to check functions working as expected on dev site (e.g. adding vote to howto)

## Followup-tasks
- [ ] Release new emulators build with updated node version
- [ ] Update firebase 8 -> 9 to reduce error logs (no longer strictly required, could be follow-up)
- [ ] Consider downgrading circle-ci plan (not sure if we need anymore, should monitor resource usage)
[...]

PR #2099, opened 2023-02-18
fix: docs node 18 support

chrismclarke:
PR Checklist

PR Type

## Description

As part of PR #2067 most workspaces were upgraded to support node 18 and webpack 5. The documentation workspace did not inherit those changes due to the way dependencies for documentation had been excluded from the main installer (included an additional `yarn.lock`). This PR updates the documentation workspace to work with node 18 and webpack 5, and no longer be excluded from  main install.

## Review Notes
Just to check that docs run and deploy. Will automatically be triggered as part of github actions post merge

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?


packages/documentation/package.json
```diff
@@ -2,24 +2,29 @@
...
-  "__NOTE__:": "avoid hoisting in case of react/webpack conflicts",
+  "__NOTE__:": "avoid hoisting for webpack and other conflicts",
+  "installConfig": {
+    "hoistingLimits": "workspaces"
+  },
   "scripts": {
-    "docusaurus": "docusaurus",
-    "start": "docusaurus start --port 3001",
-    "build": "docusaurus build",
+    "start": "env-cmd -r ../../.env-cmdrc.js -e webpack yarn docusaurus start --port 3001",
```
thisislawatts, line 10: **question:** Introducing a hard coded dependency here to the project root seems contrary to the idea of compartmentation that  workspaces provide. Is there another way we could manage this dependency?

chrismclarke: Not really, the problem with env-cmd scripts is they don't persist across commands, so need to be defined inline where they are required (so not `yarn workspace shared setup-env && ....`). The only alternatives I can think of would be:

a) Create a shared folder where both platform root and docs import from. This would still need to be a hardcoded path, so it would actually add an extra statement in the root env-cmd calls

b) Duplicate the contents of the file into the workspace. Likely would lead to issues down the line when updating one and forgetting to update the other

c) Add custom scripts to manage all docusaurus commands to ensure they've populated env variables correctly

d) Symlinks (again more trouble than they're worth)

chrismclarke: Not sure any of these sound like a better option (although open to suggestions). Also a bit of a weird/niche/edge case for now so I don't think too bad having a slightly hacky fix

thisislawatts: @chrisclarke yeah it's weird because looking at the latest version of docusauraus there are a few closed issues by them bumping a more recent version of webpack5 internally. 

:shrug:

thisislawatts (approved):

packages/documentation/package.json
chrismclarke: Yeah, guess there's just always going to be weird/annoying issues and even weirder workarounds, a-la https://github.com/facebook/docusaurus/blob/f8edeb081643e2b5b8096f19d178e31e15e1f4b8/.github/workflows/tests-windows.yml#L40

chrismclarke:
Facing annoying knock-in issue where running workspace commands attempts to use webpack from the root workspace. I'm guessing the docs webpack was being hoisted to root and used for functions build, but now not picking from own workspace. 

Can recreate locally at least with a clean install, so will try to fix here before merging

PR #2101, opened 2023-02-18
Chore/yarn update 3 4 1

chrismclarke:
PR Checklist

PR Type

## Description

Core update to bump yarn version

No significant changes mentioned in [changelog](https://github.com/yarnpkg/berry/blob/master/CHANGELOG.md#341), however appears there might be some issue with the yarn cache anyway on ci (functions build failing because workspace command not detecting webpack correctly), so just applying updates anyways to see if it can fix

Minor additional knock-ons flagged by prettier version change, so updated and fixed lint errors

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?


thisislawatts:
@chrismclarke was the prettier change required as part of the yarn update or could it be moved to a separate PR?

chrismclarke:
> @chrismclarke was the prettier change required as part of the yarn update or could it be moved to a separate PR?

It seems like part of the yarn.lock build put prettier and eslint at versions which would fight it out over syntax for `typeof a[number]` vs `(typeof a)[number]`. So the prettier update was to try pin it to a more recent version to see if that fixed things, but didn't really (will still reformat manually when working on the files)

I can find some loose references to the two behaviours `https://github.com/prettier/prettier/issues/3782` and `https://eslint.org/docs/latest/rules/no-extra-parens`, but not a clear way to get them to play along.

So short answer yes, I can revert prettier although not the file lint/format fixes

chrismclarke:
@thisislawatts I've made a partial revert of the prettier changes, turns out that components and platform workspaces both resolved to different versions (along with storybook throwing a 3rd into the mix). I've pinned to the lower of the two versions that were in use (for now), which did at least resolve the eslint-prettier conflict for those type definitions (although added another minor conflict for documentation spacing). But I feel like good enough for now

package.json
```diff
@@ -183,7 +187,7 @@
     "idb": "^6.0.0",
     "lint-staged": "^8.1.5",
     "mobx-react-devtools": "^6.0.3",
-    "prettier": "^2.6.1",
```
thisislawatts, line 186: @chrismclarke this would have been the reason that Prettier was upgraded to 2.8. The caret states a range of anything greater than 2.6 and less than 3

thisislawatts (approved):

package.json
chrismclarke: oh, yeah you're right. In my head caret would pin minor as well but no it only respect major. Makes a lot of sense, thanks and good to know if we do want to update properly

PR #2124, opened 2023-02-24
fix: added fetching from server for views

### LINKED ISSUE #2115: [bug] view incrementation not working in production
**Describe the bug**
Views are at times not incrementing correctly, it appears that at times when a user views a howto or research page the value on the firebase database is being incremented completely incorrectly. It seems like completely wrong values are being read from the database. I think it may even be the case that every user is getting different values?

I assume the issue is caused by something in the increment backend function [here](https://github.com/ONEARMY/community-platform/blob/6902da645130be8e6705bfa47f063612428dadf2/src/stores/Howto/howto.store.tsx#L199), but I have no idea what could be causing users to be getting different values from the database for the same ID.

**To Reproduce**
Hard to reproduce as this does not appear to be an issue on the dev environment, but a bug which takes place when different users view a page in the production environment.

**Expected behaviour**
Expect the views to increment by 1 from the value stored in the database whenever any user views a page. The value on the database should not be dramatically changing when different users visit the pages.

**Screenshots**
This is something I saw earlier, the views on the howto page were 28 when I viewed it, when someone else viewed it it went down to 7. When I refreshed the howto page again it set the value on the database back to 28.

Screenshot 2023-02-22 at 15 53 06
Screenshot 2023-02-22 at 15 51 51

### LINKED ISSUE #2123: [bug] Howto state leaking across multiple views
Recently we got more users saying they see complete how-to's wrong.
Its starts mixing steps from how-to's, made an image below to illustrate the issue with a how-to.
You can try yourself going from one how-to to the other. it tends to remember certain steps and show them in the next one.

caching issue

_Originally posted by @davehakkens in https://github.com/ONEARMY/community-platform/issues/2115#issuecomment-1442668748_

AlfonsoGhislieri:
PR Checklist

PR Type

## Description

Fetches howto data from server rather than cache on view. Temporary fix until serverside incrementation is implemented.

## Git Issues

Closes #2115

### DIFF
--- a/src/stores/Howto/howto.store.test.ts
+++ b/src/stores/Howto/howto.store.test.ts
@@ -427,7 +427,7 @@ describe('howto.store', () => {
       // Act
       await store.incrementViewCount(howToItem._id)
 
-      expect(getFn).toBeCalled()
+      expect(getFn).toBeCalledTimes(1)
       expect(getFn).toHaveBeenCalledWith('server')
     })
     it('increments views by one', async () => {
--- a/src/stores/Howto/howto.store.test.ts
+++ b/src/stores/Howto/howto.store.test.ts
@@ -414,7 +414,7 @@ describe('howto.store', () => {
       // Act
       await store.incrementViewCount(howToItem._id)
 
-      expect(getFn).toHaveBeenCalledTimes(1)
+      expect(getFn).toBeCalled()
       expect(getFn).toHaveBeenCalledWith('server')
     })
     it('increments views by one', async () => {
@@ -422,10 +422,10 @@ describe('howto.store', () => {
 
       const views = howToItem.total_views!
       // Act
-      await store.incrementViewCount(howToItem._id)
+      const updatedViews = await store.incrementViewCount(howToItem._id)
 
       expect(setFn).toHaveBeenCalledTimes(1)
-      expect(howToItem.total_views).toBe(views + 1)
+      expect(updatedViews).toBe(views + 1)
     })
   })
 })
--- a/src/stores/Howto/howto.store.test.ts
+++ b/src/stores/Howto/howto.store.test.ts
@@ -50,6 +50,9 @@ async function factory(howtoOverloads: Partial<IHowtoDB> = {}) {
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
     setFn: store.db.set,
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    getFn: store.db.get,
   }
 }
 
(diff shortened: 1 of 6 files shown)


thisislawatts:
Thanks @AlfonsoGhislieri, are there any tests we can add around this change to provide automated verification that the code is loading data from the server and bypassing the cache? Perhaps extending the test suites in [howto.store.test.ts](https://github.com/ONEARMY/community-platform/blob/master/src/stores/Howto/howto.store.test.ts) and [research.store.test.ts](https://github.com/ONEARMY/community-platform/blob/master/src/stores/Research/research.store.test.ts)

AlfonsoGhislieri:
@thisislawatts I can certainly try, never have tested that before so not too sure how to add it. Maybe using `store.db.clients.serverDB` and `store.db.clients.cacheDB` and checking if they have been called? 

The main issue is I can't even run the firebase emulators and do any backend testing since I keep getting this error:

```
[error] 
[error] Error: Failed to get Firebase project precious-plastics-v4-dev. 
Please make sure the project exists and your account has permission to access it.
```

thisislawatts:
@AlfonsoGhislieri, I haven't done too much work with the emulators so I defer to @chrismclarke on getting up and running with them. I don't see that error message in [the known issues](https://onearmy.github.io/community-platform/Backend%20Development/firebase-emulators-docker#troubleshooting) documentation.

You shouldn't need to the backend here if you can write a smaller unit test in the `howto|research.store.test.ts` files which verify that the `db.get` method is called with `server` as an argument. This unit test is cheaper to set up but doesn't provide the same level of coverage that an integration test does. It is better than nothing though!

AlfonsoGhislieri:
@thisislawatts ah yes well that is definitely a much easier test to get up and running. I also added tests to make sure that the incrementation actually worked. The only issue is I'm not sure how to actually run the tests to check they all work since I get the error I described earlier when I run `yarn workspace functions test:watch`

Is there another way to run the unit tests without running the emulators?

AlfonsoGhislieri:
@thisislawatts ah I instantly jumped to having to making some super complicated tests for some reason.. I added some simple unit tests to both check that the server is being called correctly and that the incrementation and actually works!

AlfonsoGhislieri:
@thisislawatts everything look alright to you?

thisislawatts:
Thanks for adding those tests @AlfonsoGhislieri, it looks good to me. Let's merge and see how it holds up in production.

thisislawatts (approved):

AlfonsoGhislieri:
@thisislawatts thanks! I noticed that another PR had added some code that was breaking the functionality of the views incrementation. Added the fix in this [PR](*https://github.com/ONEARMY/community-platform/pull/2146)

PR #2128, opened 2023-02-26
fix: handle removing first item from Contact and links

thisislawatts:
PR Checklist

PR Type

## Description

Fixes an issue that occurs when attempting to remove the first list item from Contacts and Links.
Seems to be coming from: https://github.com/ONEARMY/community-platform/blob/master/src/pages/Settings/content/formSections/Fields/Link.field.tsx#L136:L136

Related chatter:
* https://github.com/final-form/react-final-form-arrays/issues/128
* https://github.com/marmelab/react-admin/pull/6763/files

## ~Git~ Issues

Closes issue reported: 
https://discord.com/channels/586676777334865928/1079116300133072897/1079116300133072897

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,6 +21,7 @@
     }
   },
   "rules": {
+    "no-alert": "error",
     "sort-class-members/sort-class-members": [
       "error",
       {
--- a/packages/components/.eslintrc.json
+++ b/packages/components/.eslintrc.json
@@ -14,7 +14,8 @@
     {
       "files": ["**/*.stories.tsx"],
       "rules": {
-        "import/no-default-export": "off"
+        "import/no-default-export": "off",
+        "no-alert": "off"
       }
     }
   ]
--- a/packages/cypress/src/integration/settings.spec.ts
+++ b/packages/cypress/src/integration/settings.spec.ts
@@ -53,7 +53,7 @@ describe('[Settings]', () => {
       .click()
   }
 
-  const addContactLink = (link: ILink) => {
+  const addContactLink = (link: Omit<ILink, 'key'>) => {
     if (link.index > 0) {
       // click the button to add another set of input fields
       cy.get('[data-cy=add-link]').click()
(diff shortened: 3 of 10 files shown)


thisislawatts:
Working on a test to capture this fix. Think it will need to be at the e2e level as we don't have much testing infrastructure at a lower level around these pages.

davehakkens:
I get this popup when removing the link.
Might be due to local testing? 
afbeelding

And then this one
afbeelding


And the link doest get deleted..
But i'll try preview

davehakkens:
Same behaviour on Preview.
But thinking about this.. Filling in a contact detail is mandatory at the moment.
Why would there even be an option to delete it? -It can only be replaced by another link..
🤔

thisislawatts:
Thanks for taking a look @davehakkens and catching that debugging alert. I have added an automated linter to catch those in future. I will play about with this a bit more locally, it makes sense that if you add a few more links you should be able to delete the first item. Although I understand that we also expect there to be one item entered.

davehakkens:
> There are multiple links added so any one can be removed. Although it is not possible to safe a profile using an empty value

Nice solution @thisislawatts. However the first/top url doest get removed here when i click delete button. Delete works for all the other ones but not the first

thisislawatts:
Thanks for taking a look @davehakkens, I imagine that is why the tests are failing. Will circle back around to this tomorrow evening.

thisislawatts:
Please do

AlfonsoGhislieri (approved):
Looks good @thisislawatts, great idea to add uuid as a key for the links!

**Thought**: Seeing as we are using a modal for the profile link deletion, as seen [here](https://github.com/ONEARMY/community-platform/blob/162bebeb3b8eea2169aee81090752dc1a6d3b65e/src/pages/Settings/content/formSections/Fields/ProfileLink.field.tsx#L148), it could be nice to create a reusable modal. 

I've noticed there are a few of these delete modals in the code base which would benefit from this as well (eg in `HowtoStep.form `and `OpeningHoursPicker.field`. This would help make sure that any change to the modal design can be applied easily across all components. Moreover, we could also then use it for the deletion of comments as well maybe?

I don't think this should this should fall into the scope of this bug fix, but could be something to as a potential issue if you agree?

thisislawatts:
> **Thought**: Seeing as we are using a modal for the profile link deletion, as seen [here](https://github.com/ONEARMY/community-platform/blob/162bebeb3b8eea2169aee81090752dc1a6d3b65e/src/pages/Settings/content/formSections/Fields/ProfileLink.field.tsx#L148), it could be nice to create a reusable modal.

This is a reusable Modal element from `oa-components`, is there anything about the current implementation that could be improved to make it easier to reuse?

> I've noticed there are a few of these delete modals in the code base which would benefit from this as well (eg in `HowtoStep.form `and `OpeningHoursPicker.field`. This would help make sure that any change to the modal design can be applied easily across all components. Moreover, we could also then use it for the deletion of comments as well maybe?
> 
> I don't think this should this should fall into the scope of this bug fix, but could be something to as a potential issue if you agree?

Definitely worth raising something like _Consistent use of alert/modal_ as a separate issue, these are not consistent across the site and I think it would be great to make this experience a bit smoother. Specifically on this page I do not think an alert/modal is necessary as the removal action is not immediately destructive. The user needs to click 'Save' before the change will be persisted.

davehakkens (approved):

PR #2133, opened 2023-02-26
feat: admin edit user badges

### LINKED ISSUE #2131: Add ability for admin to assign badges Frontend
### Description

The current way to assign badges for an admin is to do it directly in the Firebase Backend. This is complex and risky. We want to add this ability to the user profiles so admins can do it on the frontend.

Currently we use `badges` for `verified` users. But we are introducing a new one for `supporters` in https://github.com/ONEARMY/community-platform/issues/2130

### Page related
Will be used one user edit page(admin only) : https://dev.onearmy.world/u/username/edit

### Mockup
admin badges

thisislawatts:
PR Checklist

PR Type

## Description

- chore: refactor to introduce UserSettings page
- feat: adds admin panel to edit user badges

Outstanding:

- [x] Tests
- [ ] Refactor
- [x] Update user object
- [x] Limit to only allow admins access

## Git Issues

Closes #2131

## Screenshots/Videos

Following screenshot using the standard `red2` colour token. 

Screenshot 2023-02-26 at 22 10 45

### DIFF
--- a/src/pages/Settings/UserBadgeSettings.tsx
+++ b/src/pages/Settings/UserBadgeSettings.tsx
@@ -3,7 +3,10 @@ import { useEffect, useState } from 'react'
 import { Form, Field } from 'react-final-form'
 import { useCommonStores } from 'src'
 import { CheckboxInput } from 'src/common/Form/Checkbox'
+import { capitalizeFirstLetter } from 'src/utils/helpers'
 import { Card, Heading, Box, Button } from 'theme-ui'
+
+const availableBadges = ['supporter', 'verified']
 
 export const UserBadgeSettings = observer((props: { userId: string }) => {
   const { userStore } = useCommonStores().stores
@@ -49,38 +52,24 @@ export const UserBadgeSettings = observer((props: { userId: string }) => {
               Badge settings
               <form onSubmit={props.handleSubmit}>
                 <Box mt={1}>
-                  <div>
-                    <Field
-                      id="verified"
-                      name="verified"
-                      data-cy="verified"
-                      onClick={(v) => {
-                        if (badges) {
-                          badges.verified = !!v.target.checked
-                        }
-                      }}
-                      type="checkbox"
-                      labelText={'Verified'}
-                      checked={!!badges?.verified}
-                      component={CheckboxInput}
-                    />
-                  </div>
-                  <div>
-                    <Field
-                      id="supporter"
-                      name="supporter"
-                      onClick={(v) => {
-                        if (badges) {
-                          badges.supporter = !!v.target.checked
-                        }
-                      }}
-                      data-cy="supporter"
-                      labelText="Supporter"
-                      type="checkbox"
-                      checked={!!badges?.supporter}
-                      component={CheckboxInput}
-                    />
-                  </div>
+                  {availableBadges.map((badge, key) => (
+                    <div key={key}>
+                      <Field
+                        id={badge}
+                        name={badge}
+                        data-cy={badge}
+                        onClick={(v) => {
+                          if (badges) {
+                            badges[badge] = !!v.target.checked
+                          }
+                        }}
+                        type="checkbox"
+                        labelText={capitalizeFirstLetter(badge)}
+                        checked={!!badges?.[badge]}
+                        component={CheckboxInput}
+                      />
+                    </div>
+                  ))}
                 </Box>
                 <Button
                   sx={{ mt: 4 }}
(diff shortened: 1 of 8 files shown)


thisislawatts:
@davehakkens, take a look. Iterating slightly on the designs, the implementation has a standalone "Update badges" button rather than hooking into the main "Save profile" form. The aim is to keep this admin only logic separate from the normal edit profile flow.

davehakkens:
nice @thisislawatts. Yeh works for me like this!

AlfonsoGhislieri:
I can take a look at this one next week!

src/pages/Settings/UserBadgeSettings.tsx
```diff
@@ -0,0 +1,89 @@
...
+export const UserBadgeSettings = observer((props: { userId: string }) => {
+  const { userStore } = useCommonStores().stores
+  const [badges, setBadges] = useState({
+    verified: false,
+    supporter: false,
+  })
+  const [isLoading, setLoading] = useState(true)
+  const [isSaving, setSaving] = useState(false)
+
+  async function fetchUser() {
```
AlfonsoGhislieri, line 20: Could change this to be an arrow function to be in line with the other functions

src/stores/User/user.store.ts
```diff
@@ -151,6 +151,21 @@ export class UserStore extends ModuleStore {
     return lookup2[0]
   }
 
+  public async updateUserBadge(userId: string, badges: IUserBadges) {
+    console.log(`updateUserBadge`, badges)
+    const dbRef = this.db.collection<IUserPP>(COLLECTION_NAME).doc(userId)
+
+    console.log(`updateUserBadge.before:`, toJS(await dbRef.get('server')))
```
AlfonsoGhislieri, line 158: Are these console.log's necessary?

src/stores/User/user.store.ts
```diff
@@ -162,6 +177,7 @@ export class UserStore extends ModuleStore {
     adminEditableUserId?: string,
   ) {
     this.setUpdateStaus('Start')
+    console.log(`updateUserProfile`, { values })
```
AlfonsoGhislieri, line 180: Is this console log also needed?

src/stores/User/user.store.ts
```diff
@@ -194,6 +210,7 @@ export class UserStore extends ModuleStore {
     }
 
     // update on db and update locally (if targeting self as user)
+    console.log(`updateUserProfile.doc.set`, { updatedUserProfile })
```
AlfonsoGhislieri, line 213: Same as above

AlfonsoGhislieri (approved):
@thisislawatts looks good to me, some rogue console logs which I'm sure if are necessary though

src/stores/User/user.store.ts
thisislawatts: Nope, let me remove them.

src/stores/User/user.store.ts
thisislawatts: Removed

thisislawatts:
Thanks for the review @AlfonsoGhislieri, I think you flagged an interesting point here around `console.log` calls, I have raised https://github.com/ONEARMY/community-platform/pull/2148, which aims to add automated checks to prevent `console` calls being used in the primary application.

src/pages/Settings/UserBadgeSettings.tsx
thisislawatts: @AlfonsoGhislieri it is a unclear whether this is something you are asking or stating. If this is a suggestion or issue that would block you from approving this PR I recommend being explicit. 

I like to use the [Conventional Comment pattern](https://conventionalcomments.org/) although I am a little inconsistent with it.

For example your comment could adopt either of these prefixes which would help the PR author understand you intent. 

> **nitpick (non-blocking):** Could change this to be an arrow function to be in line with the other functions

> **issue (blocking):** Please change this to be an arrow function to be in line with the other functions

AlfonsoGhislieri: Ah this is great, thanks for the resource! Updated the comment, was more of a suggestion, not a blocking issue

AlfonsoGhislieri:
> Thanks for the review @AlfonsoGhislieri, I think you flagged an interesting point here around `console.log` calls, I have raised #2148, which aims to add automated checks to prevent `console` calls being used in the primary application.

That's a great idea, I've definitely noticied a few console.log's that snuck into the production code

src/pages/Settings/UserBadgeSettings.tsx
thisislawatts: It's a good idea though, I have opened https://github.com/ONEARMY/community-platform/pull/2149 to see if we can introduce an eslint plugin to enforce arrow style functions. Up until now the project has not had an opinion about this convention. I don't have a preference here but like the opportunity to make the code base more consistent. 

Please take a look over the changes there and add a comment or 👍🏻 if it looks good to you.

davehakkens (approved):

PR #2135, opened 2023-02-27
fix: add default one army flag if user has not declared a country

### LINKED ISSUE #2134: Flag mandatory
Would it be possible to make the country field aka flag on the user settings mandatory?
Having users without it makes it look buggy and more boring 😎

Screenshot_2023-02-26_at_20_08_01

Screenshot_2023-02-26_at_20_10_25

thisislawatts:
PR Checklist

PR Type

## Git Issues

Closes #2134

## Screenshots/Videos

Screenshot 2023-03-01 at 20 55 14


davehakkens:
hey nice speedy PR!
I would prefer to take the approach as outlined here https://github.com/ONEARMY/community-platform/issues/2134#issuecomment-1447179512 though to solve the issue.

But I guess we could still add a placeholder flag for all the excisting users that posted content without adding a country flag?

thisislawatts:
I think a placeholder would be useful whilst we work on a longer term solution and for supporting existing accounts which are missing this information.

davehakkens:
> I think a placeholder would be useful whilst we work on a longer term solution and for supporting existing accounts which are missing this information.

Agree. How do you feel about replacing it with the SVG attached?
[unkown.svg.zip](https://github.com/ONEARMY/community-platform/files/10855327/unkown.svg.zip)
- Dont want to overuse the OA logo. Might get more confusing with useful.
- The placeholder flag should be the least interesting option so people prefer to pick a real country.

thisislawatts:
Updated to introduce the unknown flag. 
Screenshot 2023-03-01 at 20 55 14

thisismattia:
Nice one @thisislawatts

AlfonsoGhislieri (commented):
Looks good to me but I do think a test would be a nice addition.

**suggestion**: Would be good to have a test that checks that the right flag shows up when a user selected a flag or the `flagUnknownSVG` if that's not the case.

thisislawatts:
That is a great idea @AlfonsoGhislieri, I had been avoiding adding any as there is no test infrastructure in the components package at the moment. 

But you're right we should definitely have some coverage here. Any preference on tooling here? We can either add react-testing-library for consistency with the SPA or lean further into storybook land and explore their testing patterns.

thisislawatts:
Will take it for a spin this evening and see how it goes.

thisislawatts:
@AlfonsoGhislieri, I took a look into this and opted for the following set up:
1. `vitest` as the test runner, opted for speed over the Storybook test-runner which seems to rely on a Storybook instance being up and running 🐌 
2. `testing-library/react` for query selector

AlfonsoGhislieri (approved):
@thisislawatts this looks good, definitely seems better than having to have each Storybook be running.. Great to see that we will be able to use the stories to quickly also test components :)

davehakkens (approved):

chrismclarke:
Really cool! Thanks for adding this, really interested to try out the storybook testing scaffolding also :D

dalibormrska:
> > I think a placeholder would be useful whilst we work on a longer term solution and for supporting existing accounts which are missing this information.
> 
> Agree. How do you feel about replacing it with the SVG attached? [unkown.svg.zip](https://github.com/ONEARMY/community-platform/files/10855327/unkown.svg.zip)
> 
> * Dont want to overuse the OA logo. Might get more confusing with useful.
> * The placeholder flag should be the least interesting option so people prefer to pick a real country.

Heyy, I backtracked to this PR, since I was now wondering about the neutral flag. Thank you for the reasonings @davehakkens on why the current flag is what it is. Do you think the _introduction of this flag and specifically the intention that the flag should be least interesting and should motivate people to fill in their profiles_ worked?

The reason I'm asking is that the association I'm getting with the flag currently implemented is that the user is offline, or blocked, or deleted or something...
image

My question would be that if the default flag could be something like an earth flag? Also that people that don't want to associate themselves with a country can have one? What do you think? @benfurber 

image
image
image

PR #2146, opened 2023-03-08
fix: move initialization boolean inside component

AlfonsoGhislieri:
PR Checklist

PR Type

## Description

I noticed that this [line](https://github.com/ONEARMY/community-platform/blob/e2727c9ea7ec9c3e149ce9885fc66f9caf138c45/src/pages/Howto/Content/Howto/HowtoDescription/HowtoDescription.tsx#L56) was added, the issue is that now it will set `didInit` to true after a single howto or research is viewed and then the `useEffect` won't run again for other pages until page refresh.  This fixes the issue.

### DIFF
--- a/packages/components/src/ViewsCounter/ViewsCounter.tsx
+++ b/packages/components/src/ViewsCounter/ViewsCounter.tsx
@@ -7,6 +7,7 @@ export interface IProps {
 
 export const ViewsCounter = (props: IProps) => (
   <Flex
+    data-cy={'ViewsCounter'}
     px={2}
     py={1}
     mb={1}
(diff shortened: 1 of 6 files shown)


thisislawatts:
@AlfonsoGhislieri is there a test we can add that fails due to the buggy behaviour, and then passes as a result of this fix?

AlfonsoGhislieri:
Will have a look at adding some tests next week!

AlfonsoGhislieri:
@thisislawatts I added some cypress tests to make sure the views all show up properly.

thisislawatts (approved):

PR #2148, opened 2023-03-19
chore: remove usage of console.log in primary SPA

thisislawatts:
PR Checklist

PR Type

## Description

- chore: prevent console log
- chore: code changes to introduce logger instead of using console.log

@AlfonsoGhislieri raised a really useful bit of PR feedback around `console.log`. I propose we adopt automated checks to catch this and free up our reviewers time to concentrate on details not solved with tooling.

---

## What happens next?

### DIFF
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -1,0 +1,5 @@
+{
+  "rules": {
+    "no-console": "error"
+  }
+}
--- a/src/common/Error/handler.ts
+++ b/src/common/Error/handler.ts
@@ -1,4 +1,5 @@
 import * as Sentry from '@sentry/react'
+import { logger } from '../../logger'
 import { SENTRY_CONFIG } from '../../config/config'
 
 export const initErrorHandler = () => {
@@ -7,7 +8,7 @@ export const initErrorHandler = () => {
     location.search.indexOf('noSentry=true') !== -1 ||
     location.hostname === 'localhost'
   ) {
-    console.log('No error handler for this environment')
+    logger.info('No error handler for this environment')
     return
   }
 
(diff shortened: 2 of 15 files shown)


AlfonsoGhislieri (approved):
This looks good to me 👍🏻, seems like a great change to prevent these rogue console.log's making into production!

**Question**: All of these should now be showing up in the logflare logs?

thisislawatts:
@AlfonsoGhislieri, at the moment I do not think Logflare (or another aggregated logging tool) has been configured for any of the production instances.

Removing the console output here is the primary outcome of this change, hooking up the platform instances to Logflare will need to be done separately by someone with edit access to CircleCI as the relevant variables, `REACT_APP_LOGFLARE_KEY`,`REACT_APP_LOGFLARE_SOURCE`; are injected during the deployment process.

chrismclarke:
If I remember correctly last time I tried looking into logflare integration there was an issue with pino-logflare compatibility for webpack 5 which we had just updated to, but I think that was fixed in #2067 so could be worth a revisit. We also have sentry and firebase crashlytics available for logging endpoints, although typically neither are actively monitored. 

I'd probably suggest what would be most useful is a way to override at runtime for cases where trying to quickly debug an issue in production (e.g.`/how-to?logger=console`) as I think we're more likely to do that than sift through combined logs.

But for now moving things over to a logger makes any transition easier in the future as we can just configure transports directly there.

src/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "no-console": "error"
```
chrismclarke, line 3: question(non-blocking)
Do you know if there's any way to add a custom message to these errors, like in https://eslint.org/docs/latest/rules/no-restricted-imports ? I'm not aware of any myself, just think might be nice to advise developers something along the lines of "Should use logger instead of console" to make clear. Not super important, just a thought.

thisislawatts:
Thanks @chrismclarke

PR #2149, opened 2023-03-19
chore: enforce arrow style functions

thisislawatts:
PR Checklist

PR Type

## Description

An exploration at using automation to standardise on arrow style function declaration. Based on the following comment in another PR.

> Suggestion (not-blocking) : Could change this to be an arrow function to be in line with the other functions
https://github.com/ONEARMY/community-platform/pull/2133#discussion_r1141333791

---

## What happens next?


AlfonsoGhislieri (approved):
**Praise** @thisislawatts this was so satisfying to look at! Looks good to me 👍🏻

The only thing I was worried about was the implicit returns like mentioned [here](https://stackoverflow.com/questions/28889450/when-should-i-use-a-return-statement-in-es6-arrow-functions), but it seems that all the non one liners are wrapped in parathenses which should not cause any issues.

chrismclarke (approved):
Looks good to me also, I took a quick google of any use-cases when it might be better to keep function notation and couldn't find any examples in the changes mentioned so say should be good to go once conflicts resolved
https://www.javascripttutorial.net/es6/when-you-should-not-use-arrow-functions

thisislawatts:
Thanks @chrismclarke, I think the quickest way to do this would be drop 52454b94d8c8c8bd6b58ec427fcceed146aa55cb, then rebase against master addressing any merge conflicts on `.eslintrc.json` and then run `yarn format`

Issue #2026, opened 2022-12-10
Create email template for notifications


davehakkens:
### Description

We are working on email notifications and need this attached template made in [Handlebars](https://handlebarsjs.com/playground.html). We will later hook it to backend and probably use [postmarkapp](https://postmarkapp.com/) for sending them.

### Mockup

![notifications_mailtemplate_1 (1)](https://user-images.githubusercontent.com/13672737/206877777-babfc139-b05c-490b-bf77-b4a2e4e1fa1c.jpg)

asheerrizvi:
Here's what the template looks like at the moment:

https://codesandbox.io/s/clever-tdd-6xfvfe?file=/index.html

chrismclarke:
Thanks @asheerrizvi - looks good to me, I've created a follow-up issue #2091 to use the template to generate a notification list which I'm handing over to @evakill , so I think good to close the issue here and follow-up with anything as needed. Thanks again!

PR #2150, opened 2023-03-20
feat(notifications): transform function and template

### LINKED ISSUE #2091: [feature request] Email notification list
We've slowly been trying to build infrastructure to allow sending user notifications for various interactions like howto/research like, comment and mention. We would like to send a daily or weekly summary email of any notifications that a user hasn't seen.

As of #2027 we have an aggregation that automatically collates all unread user notifications in a single `user_notifications` db endpoint. Ideally we want to run a scheduled function (like a cron task) that will process this list on a daily basis, collate a single email for each user with a summary of all notifications, send the notifications and mark as sent on the user profile notifications.

For this issue most likely it will be best to focus on just converting the `user_notifications` list into a format ready for email, and populating to a separate `emails` db collection. The format required should match the [firestore-send-email](https://firebase.google.com/docs/extensions/official/firestore-send-email) extension. It will require an additional call to the firebase-auth system to retrieve the email from the user auth id (example code in `functions\src\admin\getUserEmail.ts`)

Sending notifications at the time a notification is triggered instead of a digest, however could become quickly excessive.

**Additional context**
A base template has already been created in #2026, which can be adapted to include retrieved notifications. 

Getting a good workflow setup might be a bit tricky, would likely recommend trying to use the existing spec testing system in functions to test the code that converts the user_notifications list to emails, and viewing the output HTML in a HTML editor to see if it looks as expected.

Additonally as the user_notifications aggregation code has only just been merged it doesn't yet appear in the most recent docker image. I will try to update later this week, but if starting work before then the _easiest_ way to generate the list is to follow the steps from pr #2027 (log in as demo_admin, comment on howto for demo_beta_tester, or vice-versa - this will seed the aggregation). 

image

Alternatively could likely just create some mock data

evakill:
PR Checklist

PR Type

## Description
This PR is the next step in the process to send email notifications:
- [x] Aggregating pending notifications (https://github.com/ONEARMY/community-platform/pull/2027)
- [x] Create template for emails (https://github.com/ONEARMY/community-platform/issues/2026)
- [ ] Trigger a function at a set cadence to aggregate pending notifications and trigger an email <-- we are here
- [ ] Set up email extension (link sendgrid, add template) in firebase

The trigger uses chron to call a function every day at 5pm. The function reads from `user_notifications.pending_emails`, aggregates/transforms the data, then adds to an `emails` collection. This collection will use the [`firestore-send-email`](https://firebase.google.com/docs/extensions/official/firestore-send-email) extension to send emails to users. 

I needed to modify the template to use handlebars, which we can apply using the [firebase handlebars integration](https://firebase.google.com/docs/extensions/official/firestore-send-email/templates). I included in this PR for visibility and so we can version control changes to the template in the future.

## Git Issues
Closes #2091

## Screenshots/Videos

---

## What happens next?


functions/src/emailNotifications/createEmail.ts
```diff
@@ -0,0 +1,106 @@
+import { logger } from 'firebase-functions/v1'
+import { NotificationType } from '../../../src/models'
+import { db } from '../Firebase/firestoreDB'
+import { DB_ENDPOINTS, IUserDB, IPendingEmails } from '../models'
+
+export const TEMPLATE_NAME = 'email_digest'
+
+// TODO: move following 2 fns to handlebars if possible
+const getResourceLabelFromNotificationType = (type: NotificationType) => {
```
thisislawatts, line 9: **praise:** I don't think this needs to be done, I think it is much easier to read keeping this behaviour outside of handlebars.

thisislawatts: **suggestion(non-blocking):** Maybe we could move these out to oa-shared, as we have some duplicate logic in the components library as well which smells a bit funny.
https://github.com/ONEARMY/community-platform/blob/master/packages/components/src/NotificationItem/NotificationItem.tsx#L20-L26

Not a blocker though. We can refactor this at a later date.

functions/src/emailNotifications/createEmail.ts
```diff
@@ -0,0 +1,106 @@
...
+          } else if (
+            [
+              'new_comment',
+              'howto_mention',
+              'new_comment_research',
+              'research_mention',
+            ].includes(notification.type)
+          ) {
+            comments.push(templateNotification)
+          }
```
thisislawatts, line 86: **suggestion (non-blocking):** Consider removing the if / else if here. Two if statements provides the same behaviour and can be easier to read. 

```suggestion
          if (
            [
              'howto_useful',
              'research_useful'
            ].includes(notification.type)
          ) {
            usefuls.push(templateNotification)
          }
          
          if (
            [
              'new_comment',
              'howto_mention',
              'new_comment_research',
              'research_mention',
            ].includes(notification.type)
          ) {
            comments.push(templateNotification)
          }
```

thisislawatts (approved):

PR #2152, opened 2023-03-21
chore: address filename formatting

thisislawatts:
PR Checklist

PR Type

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?


packages/documentation/docusaurus.config.gh-pages.js
```diff
@@ -1,90 +0,0 @@
-module.exports = {
-  title: 'OneArmy Community Platform',
-  tagline: 'Developer Documentation',
-  url: 'https://onearmy.github.io',
-  baseUrl: '/community-platform/',
```
chrismclarke, line 5: Question(blocking)
I can see you've removed this file and workflow step that populates the different config for deployment to github pages. As far as I can remember we needed a different config because github pages hosts documentation on nested path (hence the `baseUrl` defined above which is `/` when running locally).

Do you think this is no longer required? I'm not sure if there's an easy way to test

chrismclarke (changes requested):
Many thanks @thisislawatts , I've got one small question written inline, but otherwise the rest of the changes all look good to me

packages/documentation/docusaurus.config.gh-pages.js
thisislawatts: Good catch @chrismclarke, somehow I missed that property when diffing the two files 🙈 Updated the PR to ignore that file.

PR #2155, opened 2023-03-22
feat: introduce pkg for themes

thisislawatts:
PR Checklist

PR Type

## Description

Refactor to lift Themes out to a standalone module. The aim here is to further improve the boundary between the stateful single page application and rendering layer. Whilst themes _could_ sit alongside components it is better to think of the themes as a visual treatment that can be applied onto components. 

Next up 🦉 Dark Mode(TM) that no one is asking for.


davehakkens:
> Next up 🦉 Dark Mode(TM) that no one is asking for. 
😅

packages/themes/dist/common/button.d.ts
chrismclarke, line 1: question(blocking)
Do we want to be including the dist folder? 

I assume the answer is probably no, and in which case we need to add to the scripting to ensure it's both recompiled when developing locally and compiled as part of build processes

packages/themes/tsconfig.json
chrismclarke, line 1: nit(non-blocking)
I'm not sure if we really need the full boilerplate - most editors have intellisense for the properties anyway.

chrismclarke (changes requested):
See a couple of comments in-line.

Also as this has potential for accidentally breaking theme/style imports I'll suggest @davehakkens  takes a visual check just to make sure colours, logos etc. appear as expected

packages/themes/dist/common/button.d.ts
thisislawatts: Good catch @chrismclarke, I've updated to ignore `dist` and reworked each of the workflows to build themes.

davehakkens (approved):
Looks good to me! (only tried PP thought, will test again when deployed to dev sites)

thisislawatts:
@chrismclarke do you think this should be marked as a breaking change, aka a v2? Originally I was only thinking about this as refactoring but maybe its more meaningful than that.

package.json
```diff
@@ -14,18 +14,20 @@
   "private": true,
   "main": "lib/index.js",
   "scripts": {
-    "start": "concurrently --kill-others --names components,platform --prefix-colors blue,magenta \"yarn start:components\" \"yarn start:platform\"",
+    "start": "concurrently --kill-others --names themes,components,platform --prefix-colors cyan,blue,magenta \"yarn start:themes\" \"yarn start:components\" \"yarn start:platform\"",
```
chrismclarke, line 17: Thanks for updating all of these, appreciate it's a bit of a pain maintaining the long list (on my deep backlog to revisit nx/other options)

chrismclarke (approved):
Thanks for the updates @thisislawatts , confirm running locally behaves as expected (can make changes to theme and see live reloaded). 

I can see there's a few additional minor tweaks to CI scripts so I'd say good to go once passing

chrismclarke (approved):

PR #2158, opened 2023-03-25
chore: refactor to remove use of prototype built ins

thisislawatts:
PR Checklist

PR Type

## Description

It’s better to always call built in methods from Object.prototype.
For example, foo.hasOwnProperty("bar") should be replaced
with Object.prototype.hasOwnProperty.call(foo, "bar").

Further information:
https://eslint.org/docs/latest/rules/no-irregular-whitespace


chrismclarke (approved):
I think some of the code could possibly be more succinct with `key in value` notation given mostly we're just dealing with json-type objects, 
https://masteringjs.io/tutorials/fundamentals/hasownproperty

Even the eslint docs suggests might not be required to impose for this type of case
https://eslint.org/docs/latest/rules/no-prototype-builtins#when-not-to-use-it

But also see no harm in applying it, so sounds good to me

PR #2161, opened 2023-03-25
chore: refactor no-useless-catch

thisislawatts:
> A catch clause that only rethrows the original error is redundant, and has no effect on the runtime behavior of the program. These redundant clauses can be a source of confusion and code bloat, so it’s better to disallow these unnecessary catch clauses.
https://eslint.org/docs/latest/rules/no-useless-catch

PR Checklist

PR Type

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/src/stores/User/user.store.ts
+++ b/src/stores/User/user.store.ts
@@ -1,8 +1,8 @@
 import { action, computed, makeObservable, observable, toJS } from 'mobx'
-import { logger } from 'src/logger'
-import { auth, EmailAuthProvider } from 'src/utils/firebase'
-import { getLocationData } from 'src/utils/getLocationData'
-import { formatLowerNoSpecial } from 'src/utils/helpers'
+import { logger } from '../../logger'
+import { auth, EmailAuthProvider } from '../../utils/firebase'
+import { getLocationData } from '../../utils/getLocationData'
+import { formatLowerNoSpecial } from '../../utils/helpers'
 
 import { ModuleStore } from '../common/module.store'
 import { Storage } from '../storage'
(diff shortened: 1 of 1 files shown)


src/stores/User/user.store.ts
```diff
@@ -1,8 +1,8 @@
 import { action, computed, makeObservable, observable, toJS } from 'mobx'
-import { logger } from 'src/logger'
-import { auth, EmailAuthProvider } from 'src/utils/firebase'
-import { getLocationData } from 'src/utils/getLocationData'
-import { formatLowerNoSpecial } from 'src/utils/helpers'
+import { logger } from '../../logger'
```
chrismclarke, line 2: question(blocking)
I've seen on most PRs you've been changing some src imports to relative (but not all).
I'm wondering if this is something your editor is doing or an active choice, and if so why make the logger reative but not the models or utils below? Right now this feels a little inconsistent (it was before, but the added changes don't resolve)... might just be a lack of understanding on my part

chrismclarke (changes requested):
One minor question but otherwise looks good to me

src/stores/User/user.store.ts
chrismclarke: Otherwise happy with the change, so actually I think this can be non-blocking but would be good to know for future ref

chrismclarke (approved):

src/stores/User/user.store.ts
thisislawatts: A good question, I am intermittently seeing failures owing to `/src/paths` not being resolveable during the build process. It suspect it happening when the require chain is changed due to order of imports across the project. You're right though that it is really inconsistent across the project.
* 609 instances of `from 'src'`
* 359 instances of `from '..?/` (although that pattern is a little greedy)

It would be great to standardise here, ideally with some tooling to help out.

chrismclarke: Yeah path resolution is a total pain, particularly as you have somewhat competing systems trying to manage - cra/webpack has it's own support where it basically just replaces the references at build time via it's own config, and then typescript which uses tsconfig files to resolve as it compiles.

I think my personal preference would be to get all workspaces working with absolute paths. The thing I don't like about relative paths is when they get too nested and you need to make folder structure changes (I've seen vscode make really janky replacements, turning local `../folder/file` to `../../../root/project/folder/file`. You also can get issues if people try to import files from outside the workspace (which may/may not fail depending on compiler), although I think that one is more easily solved with lint rules (think there's some module boundary one that we may already have(?)). 

Thanks for the stats though, nice to know that we are consistently inconsistent enough that probably not worth dealing with for now 😅

PR #2169, opened 2023-03-25
fix(useful-button): use local update instead of optimistic update

evakill:
PR Checklist

PR Type

## Description
we've identified two issues with the "mark as useful" button. one issue is when you smash the useful button the count gets out of sync. another issue is that users report sometimes articles they've marked as useful mysteriously disappear. 

in testing the update behavior, I could verify that the optimistic update causes the first issue. i changed this optimistic update to a local update in the useful button component. this maintains the instant update on click, but prevents aggregation updates that cause counts to get out of sync. 

i am not sure whether this will impact the second issue. i think a good next step would be to add error catching in the update step so we can track unsuccessful updates and notify users if their useful was not recorded in the DB. 

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/packages/components/src/UsefulStatsButton/UsefulStatsButton.tsx
+++ b/packages/components/src/UsefulStatsButton/UsefulStatsButton.tsx
@@ -33,18 +33,29 @@ export interface IProps {
 export const UsefulStatsButton = (props: IProps) => {
   const theme: any = useTheme()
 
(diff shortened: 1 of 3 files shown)


davehakkens:
behaviour feels much better! 💪

PR #2184, opened 2023-03-29
feat(Howto.form): added current character count indicator

### LINKED ISSUE #2015: [bug] Poorly documented max string length for How-to title
**Describe the bug**

It is not possible to type more than 50 characters into the How to title field. 

**To Reproduce**

1. Go to 'How to -> Create'
2. Click on 'Title of your How-to *'
3. Type or copy a long string into the box, for example:
> Duis nec sollicitudin mauris. Orci varius natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Mauris condimentum erat vitae lorem viverra, lobortis iaculis enim tincidunt. Vestibulum tempus ex eget nunc vulputate tempus. Phasellus pulvinar imperdiet ligula in iaculis. Morbi ac gravida mi.
4. Notice the absence of any visual indicators or explanation to the user about why the input is no longer working. 

**Expected behaviour**

It should be possible to type more than 50 characters for a title, or if this hard limit is really required then there should be an indication to the user and explanation on why they are not able to enter more text. For example see attached screengrab of how we implement a similar function in the comments box. 

**Screenshots**

Screen grab demonstrates hitting the max limit where the cursor stops moving forward, there is no explanation for the user. 
Screenshot 2022-11-27 at 16 13 14

Screenshot 2022-11-27 at 16 21 34

cerkiewny:
In order to easier see what the limit of characters in the howto creation for the title is, show current and maximum number of characters next to the text area widget.

PR Checklist

PR Type

## Description

Introduced the better character length indicator on the title input ui for event

## Git Issues

Closes #2015 

## Screenshots/Videos

https://user-images.githubusercontent.com/4504330/228482191-3b00edd0-66c8-4c90-9dd0-09c8d91cee0a.mp4

---

## What happens next?

### DIFF
--- a/packages/components/src/FieldInput/FieldInput.tsx
+++ b/packages/components/src/FieldInput/FieldInput.tsx
@@ -1,3 +1,4 @@
+import { useState } from 'react'
 import type { FieldRenderProps } from 'react-final-form'
 import { Text, Input } from 'theme-ui'
 
@@ -7,6 +8,7 @@ export interface Props extends FieldProps {
   // additional fields intending to pass down
   disabled?: boolean
   children?: React.ReactNode
+  showCharacterCount?: boolean
   'data-cy'?: string
   customOnBlur?: (event: any) => void
 }
@@ -26,14 +28,38 @@ const processInputModifiers = (value: any, modifiers: InputModifiers = {}) => {
   return value
 }
 
+const TextLimitIndicator = ({
+  curSize,
+  maxSize,
+}: {
+  curSize: number
+  maxSize: number
+}) => {
+  const percMax = curSize / maxSize
+  const colorVec = [
+    { value: 1.0, color: 'red' },
+    { value: 0.75, color: 'red2' },
+    { value: 0.6, color: 'yellow.base' },
+  ]
+  const color = colorVec.find((cur) => cur.value <= percMax)?.color ?? 'black'
+  return (
+    <Text color={color} ml="auto" mr={2} mt={2} sx={{ fontSize: 1 }}>
+      {curSize} / {maxSize}
+    </Text>
+  )
+}
+
 export const FieldInput = ({
   input,
   meta,
   disabled,
   modifiers,
   customOnBlur,
+  showCharacterCount,
+  maxLength,
   ...rest
 }: Props) => {
+  const [curLength, setLength] = useState<number>(input?.value?.length ?? 0)
   return (
     <>
       <Input
@@ -41,6 +67,7 @@ export const FieldInput = ({
         variant={meta?.error && meta?.touched ? 'textareaError' : 'textarea'}
         {...input}
         {...rest}
+        maxLength={maxLength}
         onBlur={(e) => {
           if (modifiers) {
             e.target.value = processInputModifiers(e.target.value, modifiers)
@@ -51,11 +78,18 @@ export const FieldInput = ({
           }
           input.onBlur()
         }}
+        onChange={(ev) => {
+          showCharacterCount && setLength(ev.target.value.length)
+          input.onChange(ev)
+        }}
       />
       {meta.error && meta.touched && (
         <Text sx={{ fontSize: 0, margin: 1, color: 'error' }}>
           {meta.error}
         </Text>
+      )}
+      {showCharacterCount && maxLength && (
+        <TextLimitIndicator maxSize={maxLength} curSize={curLength} />
       )}
     </>
   )
--- a/src/pages/Howto/Content/Common/Howto.form.tsx
+++ b/src/pages/Howto/Content/Common/Howto.form.tsx
@@ -263,6 +263,7 @@ export class HowtoForm extends React.PureComponent<IProps, IState> {
                                   component={FieldInput}
                                   maxLength={HOWTO_TITLE_MAX_LENGTH}
                                   placeholder={`Make a chair from... (max ${HOWTO_TITLE_MAX_LENGTH} characters)`}
+                                  showCharacterCount
                                 />
                               </Flex>
                               <Flex sx={{ flexDirection: 'column' }} mb={3}>


davehakkens:
Looks nice, cool color detail.
Might be good to align/intergrate this with other areas in the platform
We already have a similair function in the comments input.
afbeelding

And could use it also in the research module
afbeelding

thisislawatts:
Thanks for taking a look at this @cerkiewny, I think it would be worth adding this counter into: https://github.com/ONEARMY/community-platform/blob/master/packages/components/src/FieldInput/FieldInput.tsx

Consider:
* Making this counter conditional so it is only included if there is a `maxLength` attribute passed through. 
* Using [standard fontsizes](https://github.com/ONEARMY/community-platform/blob/0b3a0119f5ada4dc55c29679e85ef3319a7d51fb/packages/themes/src/project-kamp/styles.ts#L4) from the theme declaration rather the hardcoding pixel values.

cerkiewny:
On related note do you have the theme or some design system from which I could use the colors (like instead of green orange red success/warning/danger colors)? I will change it later today. And can move it to FieldInput as you suggested.

iSCJT:
> 

There's a bit about the theming [here](https://onearmy.github.io/community-platform/Frontend%20Development/overview)

Although I think it needs an update and the themes directory was moved to [here](https://github.com/ONEARMY/community-platform/tree/b2c823a37210dd741acadd9b6c9c557e7990138a/packages/themes) a few days ago.

src/pages/Howto/Content/Common/Howto.form.tsx
```diff
@@ -95,6 +96,24 @@ const Label = styled.label`
...
+  }
+  if (percMax > 0.99) {
+    color = 'red'
+  }
+  return (
+    <Text color={color} mt={2} sx={{ fontSize: '10px' }}>
+      {curSize} / {HOWTO_TITLE_MAX_LENGTH}
+    </Text>
+  )
+}
```
AlfonsoGhislieri, line 116: Like suggested by @thisislawatts try and use theme for both color and fontsizes. Eg: `theme.fontSizes[2]` or `theme.color.red`

src/pages/Howto/Content/Common/Howto.form.tsx
```diff
@@ -95,6 +96,24 @@ const Label = styled.label`
   display: block;
 `
 
+const TextLimitIndicator = ({ curSize }) => {
+  const percMax = curSize / HOWTO_TITLE_MAX_LENGTH
+  let color = 'black'
```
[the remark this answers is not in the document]
davehakkens: Yes would be great to apply to other areas as well!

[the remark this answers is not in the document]
AlfonsoGhislieri: Then we should definitely look into extracting this functionality from just this component and making it reusable @cerkiewny

packages/components/src/FieldInput/FieldInput.tsx
```diff
@@ -26,21 +29,51 @@ const processInputModifiers = (value: any, modifiers: InputModifiers = {}) => {
...
+}: {
+  curSize: number
+  maxSize: number
+}) => {
+  const theme: any = useTheme()
+  const percMax = curSize / maxSize
+  let color = theme.colors.black
+  if (percMax > 0.6) {
+    color = theme.colors.yellow.base
+  }
```
AlfonsoGhislieri, line 44: **Issue**: Same here, once again my bad if I didn't make it clear and confused you leading you use theme rather than making the component theme aware.

But here we should pass `color` as a prop for `sx` so that it has access to the themeing. So we can stop using theme, eg: `theme.colors.red` --> `'red'`. This could also be refactored so we don't need to use all these if statements, you can define an object that maps percentage thresholds and check the `percMax` against the keys:

```
const colorMap = {
  0.6: 'yellow.base',
  0.75: 'red2',
  0.99: 'red',
}
```

src/pages/Howto/Content/Common/Howto.form.tsx
```diff
@@ -263,6 +263,7 @@ export class HowtoForm extends React.PureComponent<IProps, IState> {
                                   component={FieldInput}
                                   maxLength={HOWTO_TITLE_MAX_LENGTH}
                                   placeholder={`Make a chair from... (max ${HOWTO_TITLE_MAX_LENGTH} characters)`}
+                                  showCharacterCount
                                 />
                               </Flex>
```
AlfonsoGhislieri, line 268: **Potential issue**: Don't we need `showCharacterCount` to pass a boolean here? Is it not just passing a Falsey value right now?

packages/components/src/FieldInput/FieldInput.tsx
```diff
@@ -1,12 +1,15 @@
...
 
 type FieldProps = FieldRenderProps<any, any> & { children?: React.ReactNode }
 
 export interface Props extends FieldProps {
   // additional fields intending to pass down
   disabled?: boolean
   children?: React.ReactNode
+  showCharacterCount?: boolean
   'data-cy'?: string
   customOnBlur?: (event: any) => void
```
AlfonsoGhislieri, line 14: **Question**: Is it really necessary to also create this `showCharacterCount` prop? Is passing `maxLength` not enough to decide when the counter should be conditionally rendered?

AlfonsoGhislieri (changes requested):
Thanks @cerkiewny for making some of the suggested changes, sorry for the confusion around the use of themeing earlier. Would be great if we can make the component theme aware.

Some other things which would be great if you could change as suggested by @davehakkens earlier in the conversation:
- The counter currently is on the bottom left of the field, would be better to have it on the bottom right for consistency.
- Could you also add it to the research form?

Let me know if you have any questions or need any help!

packages/components/src/FieldInput/FieldInput.tsx
```diff
@@ -1,12 +1,15 @@
+import { useState } from 'react'
 import type { FieldRenderProps } from 'react-final-form'
 import { Text, Input } from 'theme-ui'
+import { useTheme } from '@emotion/react'
 
 type FieldProps = FieldRenderProps<any, any> & { children?: React.ReactNode }
```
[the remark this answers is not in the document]
cerkiewny: Ok, will adjust this.

packages/components/src/FieldInput/FieldInput.tsx
```diff
@@ -26,21 +29,51 @@ const processInputModifiers = (value: any, modifiers: InputModifiers = {}) => {
...
+  }
+  if (percMax > 0.75) {
+    color = theme.colors.red2
+  }
+  if (percMax > 0.99) {
+    color = theme.colors.red
+  }
+  return (
+    <Text color={color} mt={2} sx={{ fontSize: '10px' }}>
+      {curSize} / {maxSize}
```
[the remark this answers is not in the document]
cerkiewny: Sure, my bad missed it :)

packages/components/src/FieldInput/FieldInput.tsx
cerkiewny: No problem at all. I will change it.

src/pages/Howto/Content/Common/Howto.form.tsx
cerkiewny: I was thinking about it but decided that adding extra property will make it backwards compatible. If you think it is enough I can also base it on maxLength being provided.

packages/components/src/FieldInput/FieldInput.tsx
cerkiewny: I was thinking about it but decided that adding extra property will make it backwards compatible. If you think it is enough I can also base it on maxLength being provided.

src/pages/Howto/Content/Common/Howto.form.tsx
AlfonsoGhislieri: For some reason I was convinced that with the shorthand it would default to undefined since it's an optional parameter!

packages/components/src/FieldInput/FieldInput.tsx
AlfonsoGhislieri: That makes sense, I'd say keep it just to be safe. Also probably a good extra guard to avoid this rendering where we don't explicitly want it!

cerkiewny:
https://user-images.githubusercontent.com/4504330/232135441-28fee4d0-6407-44e5-8876-6c828de8042a.mp4

AlfonsoGhislieri (changes requested):
Thanks for all the changes @cerkiewny looks like we are nearly there! Could you also add this to the [Research.form](https://github.com/ONEARMY/community-platform/blob/master/src/pages/Research/Content/Common/Research.form.tsx) please?

We also need to figure out how come the unit test is failing 🤔... You can run this locally with `yarn test:unit`

packages/components/src/FieldInput/FieldInput.tsx
```diff
@@ -26,21 +28,46 @@ const processInputModifiers = (value: any, modifiers: InputModifiers = {}) => {
...
   disabled,
   modifiers,
   customOnBlur,
+  showCharacterCount,
+  maxLength,
   ...rest
 }: Props) => {
+  const [curLength, setLength] = useState<number>(input.value.length)
   return (
     <>
```
AlfonsoGhislieri, line 64: **Issue**:  I think this line here is breaking the tests because  if `input.value` is `undefined` it cannot access `length`. We should make sure to include some kind of check or conditional before. Eg: `showCharacterCount && maxLength`, this would also need to be included in or before the `onChange` (before is probably better to avoid uncessary processing).

But besides the tests it's best to make sure we don't perform these actions to avoid unecessary state updates, potential bugs and errors.

AlfonsoGhislieri (changes requested):

packages/components/src/FieldInput/FieldInput.tsx
```diff
@@ -51,12 +78,19 @@ export const FieldInput = ({
           }
           input.onBlur()
         }}
+        onChange={(ev) => {
+          setLength(ev.target.value.length)
+          input.onChange(ev)
+        }}
```
AlfonsoGhislieri, line 84: **Issue**: This action should not be taking place unless we actually want to display the character counter like I mentioned in my earlier review comment.

packages/components/src/FieldInput/FieldInput.tsx
```diff
@@ -26,21 +28,46 @@ const processInputModifiers = (value: any, modifiers: InputModifiers = {}) => {
...
   disabled,
   modifiers,
   customOnBlur,
+  showCharacterCount,
+  maxLength,
   ...rest
 }: Props) => {
+  const [curLength, setLength] = useState<number>(input?.value?.length ?? 0)
   return (
     <>
```
AlfonsoGhislieri, line 64: **Issue**: There are many cases where the fieldInput is used but we don't actually want to set or use this functionality, we don't want to burden the component with unecessary state management. An if statement to conditonally set the state would be preferred.

packages/components/src/CreateComment/CreateComment.tsx
```diff
@@ -64,18 +64,6 @@ export const CreateComment = (props: Props) => {
...
-              <Text
-                sx={{
-                  fontSize: 2,
-                  position: 'absolute',
-                  right: 0,
-                  bottom: -5,
-                  pointerEvents: 'none',
-                  padding: 1,
-                }}
-              >
```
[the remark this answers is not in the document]
AlfonsoGhislieri: @cerkiewny Can you get back to me on this?

AlfonsoGhislieri (changes requested):

packages/components/src/FieldInput/FieldInput.tsx
cerkiewny: Hooks can't be conditional AFAIK...

AlfonsoGhislieri: My bad if this wasn't clear, I was referring to setting the state conditonally, not calling it conditionally. But, now I realize I must have quickly glanced over this and didn't realize you used nullish coalecing to take care of conditionally setting it. I was defaulting to something that was more explicit in my head eg:

```
const [curLength, setLength] = useState<number>(
  showCharacterCount && maxLength ? input.value.length : 0
);
```

AlfonsoGhislieri: Also just came to mind but another interesting option could be to use Yup to create a schema and validate it, like was introduced in #2218, you can see an example in code [here](https://github.com/ONEARMY/community-platform/blob/3129cc8e7050e82162f047dd3735f68269c4f51b/packages/components/src/ResearchEditorOverview/ResearchEditorOverview.tsx#L23)

packages/components/src/FieldInput/FieldInput.tsx
cerkiewny: You mean setLength should be conditional? input.onChange has to run, otherwise it breaks inputField

AlfonsoGhislieri: Yes exactly, I don't think we should be updating the state unless we are using the characterCounter right?

cerkiewny: No opinion here, sure, can change that.

packages/components/src/FieldInput/FieldInput.tsx
AlfonsoGhislieri: Just in general if you were interested in adding some schema validation since it's something we recently incorporated in the code base. Not at all necesssary.

AlfonsoGhislieri (approved):
Looks good @cerkiewny, thanks a bunch for the addition 🤩 and sorry for being a pain in the ass!

PR #2192, opened 2023-03-30
fix: fail back to unknown flag for invalid country code

thisislawatts:
PR Checklist

PR Type

## Description

Fall back to unknown flag for invalid country code.

Before:
Screenshot 2023-03-30 at 20 52 56

After: 
Screenshot 2023-03-30 at 20 52 34

---

## What happens next?


packages/components/src/Username/Username.tsx
```diff
@@ -12,6 +12,263 @@ export interface Props {
...
+  'AF',
+  'AX',
+  'AL',
+  'DZ',
+  'AS',
+  'AD',
+  'AO',
+  'AI',
+  'AQ',
+  'AG',
```
AlfonsoGhislieri, line 25: **Issue**: Could we maybe have a file inside `Username` with these codes, this makes the readability of the whole component not great.

We can even just make this more accessible by placing it in some helper folder for other components if we need to reuse it down the line?

AlfonsoGhislieri (changes requested):

packages/components/src/Username/Username.tsx
thisislawatts: Update made in: https://github.com/ONEARMY/community-platform/pull/2192/commits/3a2901d1f24ee4d2a5ff28f6ae673180a904c5be

AlfonsoGhislieri (approved):

PR #2189, opened 2023-03-30
fix: map page missing flags

thisislawatts:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

Include user country so that the Username component renders a flag. 

- chore: refactor to simplify typing
- chore: remove mock pin generation code
- chore: remove unused code
- fix: pass flag through from User object
- chore: refactor to common method

Closes #2171

PR #2209, opened 2023-04-06
Revert "fix: map page missing flags"

thisislawatts:
Attempting to revert ONEARMY/community-platform#2189 and unblock failing e2e.


AlfonsoGhislieri:
@thisislawatts so do we want to revert back after all then for the time being then?

thisislawatts:
I think so @AlfonsoGhislieri, I would rather reintroduce the minor bug and fix the deployment pipeline. I can then investigate the PR commit by commit.

AlfonsoGhislieri (approved):

PR #2207, opened 2023-04-06
feat: adds option to delete ResearchUpdate

thisislawatts:
PR Checklist

- [X] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

Building on the work in #2206 (please review that one first), this change introduces the ability to "Delete" a Research update.

## Git Issues

Closes #

## Screenshots/Videos

_If useful, provide screenshot or capture to highlight main changes_

---

## What happens next?

Thanks for the contribution! We try to make sure all PRs are reviewed ahead of a monthly dev call (first Monday of the month, open to all!).

If the PR is working as intended it'll be merged and included in the next platform release, if not changes will be requested and re-reviewed once updated.

If you need more immediate feedback you can try reaching out on Discord in the [Community Platform `development` channel](https://discord.com/channels/586676777334865928/938781727017558018).

PR #2206, opened 2023-04-05
feat: adds draft status

thisislawatts:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

Adds "draft" status to Research Updates, matched against design pattern for Howto form. A separate PR will be add the ability to delete Research Updates.

⚠️ Marking an update as Draft makes it unreachable, as Draft items are not listed on the Research Article page. One possible workaround for this is being discussed here: #2218 

## Git Issues

Addresses #2199 

## Screenshots/Videos

Screenshot 2023-04-05 at 21 23 10

Screenshot 2023-04-05 at 22 40 38

PR #2205, opened 2023-04-05
feat(platform): add color toggle to follow button on research page

NoHara42:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

First PR to an open-source project, woop!
Adds the same color as the active notification icon to the thunderbolt inside any research item.

## Git Issues

Closes #2200

## Screenshots/Videos

_If useful, provide screenshot or capture to highlight main changes_
image
image

PR #2215, opened 2023-04-09
Reintroduce "fix: map page missing flags"

thisislawatts:
Reintroduce changes, reverted in ONEARMY/community-platform#2209

Originally these changes were reverted in attempt to resolve test failures in an `events.spec.ts`. However, we ended up skipping the failing test (0e49c907f71f00f3ae60be3c67358d9c32776f71) as it consistently failed _after_ the revert was applied.


thisislawatts:
@AlfonsoGhislieri would you be able to take a look at this revert of a revert? 🙈

AlfonsoGhislieri:
@thisislawatts I can absolutely re-review the revert of the revert! However, just at a glance it does seem that this will reintroduce flakiness for the maps e2e tests though, are we okay with this?

thisislawatts:
My understanding of the issue @AlfonsoGhislieri is that the flakey tests were in events.spec.ts which have now been skipped.

AlfonsoGhislieri:
@thisislawatts I noticed that in these old PR's the map spec was having flaky results and even failing outright:
- #2207 
- #2206 
- #2205

thisislawatts:
Good catch @AlfonsoGhislieri, I am being a too lazy here and made the assumption that the test failures all shared a common route. I had missed the failures in `map.spec.ts`

Closing in favour of lightweight #2221, I will raise the various refactors made here as separate PRs later on to nudge things in the right direction. Hope this doesn't put anyone off refactoring.

PR #2218, opened 2023-04-10
feat: adds quick navigation for Research

thisislawatts:
PR Checklist

PR Type

## Description

Introduces an **Update** Card to the sidebar of the Research and ResearchUpdate edit form under the primary action buttons. The intent is to make it easier to navigate between different parts of a Research document with a single click. Currently the only way to navigate was either using the URL bar directly or going back to the public facing Research page, this required multiple clicks 🐌 

## Screenshots/Videos

**Editing a Research article**

Screenshot 2023-04-10 at 20 38 52

**Editing a Research Update**

Screenshot 2023-04-10 at 20 39 01 1

**Note:** The back to Research button here is a bit random, but I wanted a quick way to hop up one level. Alternative suggestions welcome. Or we can merge as and iterate 🥳

---

## What happens next?


thisislawatts:
Ps. The [failing cloud function test](https://app.circleci.com/pipelines/github/ONEARMY/community-platform/4338/workflows/ee59aa12-861c-4b84-8a6c-09bb5435f576/jobs/30298) seems unrelated. Likely an issue in the Cypress world.
Screenshot 2023-04-10 at 20 51 24

```
The Cypress App could not be downloaded.
```

Although it begs the question, **why** are the serverless functions test installing Cypress 🤯

davehakkens:
Interesting feature @thisislawatts! 
Could eventually see this going to the frontend of the research to have an overview of all the steps in a (long) research.
Anyway for now i have this feedback:

- Images dont load well (they either dont show at all or they don't change when toggeling)
- When you "create update" from the quick menu it's a bit confusing to realise you just created a new step. Ideally a new number is added in the quick menu (perhaps as a draft?)
- How would this work with drafts?
- What exact problem are we trying to fix with this? I like adding features and improving the reseach module but also scared of the maintaince of them. Just trying to understand if the added complexcity is worth it.

thisislawatts:
@davehakkens the intention here is to solve the issue you raised in the draft status PR. Where you are unable to access a draft post because it is no longer visible in the front end.

Forcing a full page load would fix most of the bugs you've raised here.

davehakkens:
ah indeed that fixes the image bug. Is there a way to make this more user friendly?

thisislawatts:
@davehakkens aside from the confusing experience when creating a new update would you be able to help me understand what you find user unfriendly about this component?

I thought the Updates list could also show a badge to indicate when the Update has a `draft` status.

davehakkens:
> Its not implemented yet, but here's a sketch of how this listing could handle Draft items

That looks neat!  

> aside from the confusing experience when creating a new update would you be able to help me understand what you find user unfriendly about this component?

- I was mainly refering to the behaviour of the images. That you need a refresh to see the ones that belong to the step. This is not very user friendly/ intuitive. Not sure if it can have an auto refresh or something?

-Perhaps instead of calling the "updates" I would call it something like "research overview"

+ One (minor) thing that is a bit confusing is that they all use numbers, but they are not necesserly the same numbers on frontend. (since drafts don't have a number, step 3 is not step 3) But not sure how to fix this. Just throwing it there maybe you have brilliant ideas. If not no problem

thisislawatts:
The buggy image behaviour should be fixed by this commit: https://github.com/ONEARMY/community-platform/pull/2218/commits/ac3ec635533e944a5d9f49b1fbc057e40b2893ec, that is a long standing bug we've caught as part of this work :)

thisislawatts:
@davehakkens updated the heading and switched to an unordered list which removes the numbers.

Screenshot 2023-04-13 at 21 17 56

davehakkens:
nice. yeh now it works smooth! Nice to fix that bug along  the way.

One additional thought on this. (sorry I didnt have this clear yet, bit of a freestyle :))
The drafts seems to mess up a bit the logic of the steps.
And realistically I dont really see a case that a users wants to write an update, save as draft, create another new update and then get back to an unpublished draft from 3 steps ago and squeeze it in there again between other steps.

I think a draft is only needed for the latest step. A step that is still under development...

What about:
- Only the latest update can be saved as a draft
- We keep the numbers (not bullet points) to keep it clear and aligned.

thisislawatts:
At the moment these changes are separate from the Draft status PR. How do you feel about reviewing this change in isolation and then iterating on how this interacts with the Draft status work? 

One potential future for this quick nav component would be to support reordering Updates through drag and drop.

src/common/Form/ImageInput/ImageInput.tsx
```diff
@@ -128,6 +129,17 @@ export class ImageInput extends React.Component<IProps, IState> {
...
+
+  componentDidUpdate(previousProps): void {
+    if (
+      JSON.stringify(this.props.value) !== JSON.stringify(previousProps.value)
+    ) {
+      this.setState({
+        uploadedFiles: this._getUploadedFiles(this.props.value),
+      })
+    }
+  }
```
AlfonsoGhislieri, line 141: **Suggestion**: While the risk is low in this case since the change is being caused by a prop, it could be good to have a stricter check and avoid any unecessary re-rendering. Could use `shouldComponentUpdate` before eg:

```
shouldComponentUpdate(nextProps, nextState) {
    return JSON.stringify(this.props.value) !== JSON.stringify(nextProps.value)
  }
```

This `componentDidUpdate` does also highlight how much simpler and flexible it would be to instead use `useEffect` if we were using functional components (I think it's also more performant).

thisislawatts: Thanks for the pointer her @AlfonsoGhislieri, I didn't know about this optimisation. I took a [look at the documentation](https://react.dev/reference/react/Component#shouldcomponentupdate-caveats), and they specifically mention the following.
>  * This method only exists as a performance optimization. If your component breaks without it, fix that first.
> * We do not recommend doing deep equality checks or using JSON.stringify in shouldComponentUpdate. It makes performance unpredictable and dependent on the data structure of every prop and state. In the best case, you risk introducing multi-second stalls to your application, and in the worst case you risk crashing it.

My preference here would be to refactor the ImageInput component to be a functional component, however I don't want to bloat this PR too much and would rather defer that change for a separate PR. What do you think?

src/common/Form/ImageInput/ImageInput.tsx
```diff
@@ -198,18 +210,3 @@ export class ImageInput extends React.Component<IProps, IState> {
...
-
-export const bytesToSize = (bytes: number) => {
-  const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB']
-  if (bytes === 0) {
-    return '0 Byte'
-  }
-  const i = Number(Math.floor(Math.log(bytes) / Math.log(1024)))
-  const size = (bytes / Math.pow(1024, i)).toPrecision(3) + ' ' + sizes[i]
-  return size
-}
```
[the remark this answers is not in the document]
thisislawatts: This is unused at the moment, the FileInformation component had its own local implementation. I think that when FileInformation was extracted to be a in components package this duplication was introduced.

src/pages/Research/Content/Common/Update.form.tsx
```diff
@@ -338,6 +339,63 @@ const UpdateForm = observer((props: IProps) => {
...
+                      <Button small variant={'outline'}>
+                        <InternalLink
+                          to={`/research/${store.activeResearchItem?.slug}/edit`}
+                          sx={{ color: 'black' }}
+                        >
+                          Back to research
+                        </InternalLink>
+                      </Button>
+                    </Card>
+                  ) : null}
```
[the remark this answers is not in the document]
thisislawatts: Great idea! I wasn't sure about making it a component until people seemed interested in the approach. 
Here's the work to extract it to the component: f49c5e47b0813a232d58956f331a4a605324bc9a

davehakkens:
> At the moment these changes are separate from the Draft status PR. How do you feel about reviewing this change in isolation and then iterating on how this interacts with the Draft status work?

Sounds good to me!

src/common/Form/ImageInput/ImageInput.tsx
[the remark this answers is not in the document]
AlfonsoGhislieri: Ah makes sense!

src/common/Form/ImageInput/ImageInput.tsx
AlfonsoGhislieri: Yeah that does seem like `shouldComponentUpdate` definitely isn't the correct solution then. Using `key` to reset the state could be maybe be best way then if that works? https://react.dev/reference/react/useState#resetting-state-with-a-key

But otherwise yes I agree, refactoring should definitely be a separate PR.

src/pages/Research/Content/Common/Update.form.tsx
```diff
@@ -338,6 +339,27 @@ const UpdateForm = observer((props: IProps) => {
                       {props.parentType === 'edit' ? 'Save' : 'Add update'}
                     </span>
                   </Button>
+
+                  {store.activeResearchItem ? (
+                    <ResearchEditorOverview
+                      sx={{ mt: 4 }}
```
AlfonsoGhislieri, line 345: **Nitpick**: Using conditional `&& `could be cleaner and more concise than ternary (I know you love those ternary operators 😉)

src/pages/Research/Content/Common/Research.form.tsx
```diff
@@ -327,6 +327,19 @@ const ResearchForm = observer((props: IProps) => {
                     <span>Publish</span>
                   </Button>
                 </Box>
+
+                {props.formValues.updates ? (
+                  <ResearchEditorOverview
+                    sx={{ mt: 4 }}
```
AlfonsoGhislieri, line 333: Same Nitpick as above

packages/components/src/ResearchEditorOverview/ResearchEditorOverview.tsx
```diff
@@ -0,0 +1,80 @@
...
+  } = props
+  return (
+    <Card sx={{ p: 4, ...sx }}>
+      <Heading as="h3" mb={3} variant="small">
+        Research overview
+      </Heading>
+      <Box as={'ul'} sx={{ margin: 0, mb: 4, p: 0, pl: 3 }}>
+        {updates.map((update, index) => (
+          <Box as={'li'} key={index} sx={{ mb: 1 }}>
+            <Text variant={'quiet'}>
```
AlfonsoGhislieri, line 37: **Issue**: Can we add the optional chaining operator `?` after updates `updates?.map ...`. Just to add a extra protection for any runtime errors.

AlfonsoGhislieri (changes requested):
Thanks @thisislawatts new component looks great, just a few small bits to maybe tweak I think.

src/pages/Research/Content/Common/Update.form.tsx
thisislawatts: I found [this blog post](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx) from a few years ago that persuaded me away from the `&&` pattern. I totally agree that it is a little more verbose but in this it is more explicit about the intended behaviour.

packages/components/src/ResearchEditorOverview/ResearchEditorOverview.tsx
thisislawatts: Thanks for catching that @AlfonsoGhislieri, I've added some tests that verify this behaviour **and** introduced yup for object validation. https://github.com/ONEARMY/community-platform/pull/2218/commits/fcabb6919b964e678344104c224b2f01c41220f4

Let me know if this is overkill, but I think its about time we adopted an schema validation language.

src/pages/Research/Content/Common/Update.form.tsx
AlfonsoGhislieri: Wow, super interesting, didn't realize it would render 0! Thanks for the link :-)

AlfonsoGhislieri (approved):
Looks good @thisislawatts, really cool to see object validation being introduced! Nice to get another layer of validation, will be interesting to see if we get widespread use out of it.

Issue #2231, opened 2023-04-21
[bug] Useful votes not working correctly


AlfonsoGhislieri:
**Describe the bug**
Useful count is not working as intended, it seems like the aggregations are messing up and that clicking useful on a howto or research is affecting another unrelated item. This is most likely related to the bugs that have been reported by @thisismattia 

It seems to me like all of the research aggregations may not be actually working properly and that only the howto aggregations are working but still affecting the research articles like seen in the video below.

**To Reproduce**
Follow steps like in video:

At the start the Howto has a useful count of 1, after a research useful count is decreased it also affects the howto's count. They seem to be interlinked.

https://user-images.githubusercontent.com/652368/233655059-fe1ebcde-aa80-4100-99c7-1e624dc8040b.mov

**Expected behaviour**
Clicking useful on a Research or Howto should not affect entities that are not the specific one in question

davehakkens:
ai.. This would be quite an important one to look into

iSCJT:
Can someone tag me on this please.

PR #2233, opened 2023-04-22
feat: add useful count to profile

### LINKED ISSUE #2136: Show total amount of useful on user profile.
### Description

On the user profiles we want to show the total amount of 'useful' a user gathered.
The useful is gathered through their content (how-to's an research) that other users mark as useful

note: If the user gathered no useful yet, nothing should be shown.

### Page related

Will be used in :

- [User profile](https://dev.onearmy.world/u/username) (spaces and members)

### Mockup
useful

AlfonsoGhislieri:
PR Checklist

PR Type

## Description

Adds useful counter to the space and member profile pagaes

## Git Issues

Closes #2136 

## Screenshots/Videos

image
---

## What happens next?


src/pages/User/User.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { Route, Switch, withRouter } from 'react-router-dom'
 import { inject, observer } from 'mobx-react'
 import type { UserStore } from 'src/stores/User/user.store'
-import { UserPage } from './content'
```
thisislawatts, line 5: **question:** Any reason to switch this to using Default exports? That is inconsistent with the other imports here.

src/stores/User/user.store.ts
```diff
@@ -387,6 +387,26 @@ export class UserStore extends ModuleStore {
     }
   }
 
+  public getUserUsefulCount(): number {
+    const contentTypes = ['Howtos', 'Research']
+
+    const usefulCounts: number[] = contentTypes.map((type) => {
+      const userContent = this.user?.stats?.[`userCreated${type}`]
```
AlfonsoGhislieri, line 394: I think it's probably best to calculate the usefulCount rather than storing it since it is so likely to change and I don't think it's that computationally intensive.

Unsure whether this should in the userStore though since it actually has to be calculated for the id of the page that is being viewed rather than the active user.

TODO: This functionality has to be made for not just active user

src/pages/User/User.tsx
AlfonsoGhislieri: Good catch! must have slipped through when I was converting the UserPage component fixed in 924835b

AlfonsoGhislieri:
This should all be good to go, the issue still remains that the aggregations are quite buggy as I mentioned in #2231. So while this functionality should all be working, the underlying useful counts remain a problem. This is something I will try and have a look into but falls outside of the scope of this PR.

src/pages/User/content/index.tsx
```diff
@@ -1,89 +1,88 @@
...
+        const userData = await userStore.getUserProfile(userid)
+        setUser(
+          userData
+            ? {
+                ...userData,
+                receivedTotalUsefulVotes: await userStore.getUserUsefulCount(
+                  userData._authID,
+                ),
+              }
+            : null,
```
AlfonsoGhislieri, line 38: I am unsure where the best place is to set `receivedTotalUsefulVotes` for a user, I didn't want to add it to the  `getUserProfile` function to avoid unnecessarily bloating it. It has to be called when the UserPage is loaded in order to have the most up to date aggregations. 

Happy for any other suggestions on how to implement this.

src/pages/User/content/index.tsx
```diff
@@ -1,89 +1,88 @@
...
-      isLoading: true,
+    const fetchData = async () => {
+      try {
+        const userData = await userStore.getUserProfile(userid)
+        setUser(
+          userData
+            ? {
+                ...userData,
+                receivedTotalUsefulVotes: await userStore.getUserUsefulCount(
+                  userData._authID,
```
AlfonsoGhislieri, line 35: I am unsure where the best place is to set `receivedTotalUsefulVotes` for a user, I didn't want to add it to the  `getUserProfile` function to avoid unnecessarily bloating it. It has to be called when the UserPage is loaded in order to have the most up to date aggregations. 

Happy for any other suggestions on how to implement this.

davehakkens:
Looks nice! Seems to work well here,
Could we release this for `beta-testers` first?
Then we can observe, test it a bit more. Plus it gives some time to look into the underlying aggregations problem

davehakkens:
> Edit: I've added option 2 with just the usefulCount being hidden, so let me know if you want the other option.

like this is perfect!

thisislawatts (approved):

PR #2240, opened 2023-04-24
fix: allcontributors doesnt skip ci

AlfonsoGhislieri:
PR Checklist

PR Type

## Description

As noticed in #2227 there is an issue as the bot skips ci in its commit messages but in doing so this breaks the ability to merge as CI is required.

## What happens next?

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -6,7 +6,6 @@
   "contributorsPerLine": 7,
   "contributorsSortAlphabetically": false,
   "badgeTemplate": "[![All Contributors](https://img.shields.io/badge/all_contributors-<%= contributors.length %>-orange.svg?style=flat-square)](#contributors)",
-  "skipCi": true,
   "types": {
     "maintainer": {
       "symbol": "💪",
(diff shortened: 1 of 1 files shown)


thisislawatts:
I think the issue here is the GitHub configuration. If CI has been skipped then it should be possible to not have the associated checks as required. 

However that seems like a known product gap in GitHub based on this documentation. A weird workaround here:
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks

My preference would be NOT to spin up our CI infrastructure for PRs which don't need it.

thisislawatts (approved):

AlfonsoGhislieri:
This is definitely a weird one, I also saw this workaround but wasn't sure how it could be implemented with circle ci. I agree that ideally we don't want to spin up the CI when it's not necessary. How was this dealt with before, I noticed the bot has pushed plenty of commits in the past.

thisislawatts:
@AlfonsoGhislieri I guess previously I had access to merge without required checks passing.

AlfonsoGhislieri:
Ah right, so that could also be another solution (easiest probably).

Issue #2238, opened 2023-04-24
[feature request] Add file download to research


davehakkens:
**Is your feature request related to a problem? Please describe.**
We can't add files to research topics to download. (besides an external link)

**Describe the solution you'd like**
Would be nice to be able to upload files to a research, like we made in how-to's. https://github.com/ONEARMY/community-platform/pull/1664
This keeps track of the downloads and makes sure the file stays online. The button can be added per step

**Describe alternatives you've considered**
Continue using Gdrive to store files.

**What it would look like**

![download](https://user-images.githubusercontent.com/13672737/233983411-fa546aef-269c-499b-977e-f3b97e34aef8.jpg)

Features:

- [ ] Add the option to upload files/links in a research update
- [ ] Add the counter for downloads for files in updates 
- [ ] Refactor and implement firebase incrementation functions

AlfonsoGhislieri:
The logic for incrementing howto download count can be seen [here](https://github.com/ONEARMY/community-platform/blob/master/src/stores/Howto/howto.store.tsx#L182), and is incorporated into the `HowtoDescription` through a handleClick method that is passed to the `DownloadFiles` component.

The more I think about this, the more I realize that I first introduced this for HowTo's so maybe it's best I take this on since I've been meaning to refactor the whole process and introduce incrementation using the firebase [increment](https://cloud.google.com/firestore/docs/samples/firestore-data-set-numeric-increment) functionality. This whole process has to be cleaned up and I've been putting it off for a while... 🥲

davehakkens:
Thanks for the extra info @AlfonsoGhislieri 
Feel free to take it on of you want. But only do if you have the capacity for it, since you mentioned things will get more busy on your end 🙃

AlfonsoGhislieri:
@davehakkens there is no way to currently add a download link directly when creating/editing like there is for a Howto. I assume right now this is done just by adding a link in the description? We would need to include the option to upload files/links to the form, is this the approach you wanted?

davehakkens:
Yes indeed the method would be the same as for how-tos. File upload in the form. (It could also copy the same functionality as how-to that adds the possibility to use an external link. To keep things coherent, and track downloads. But not sure if this would be an additional issue for later.) sorry for not making this more clear in the original issue

thisislawatts:
![7jgjhm](https://user-images.githubusercontent.com/472589/234367171-cf3ed30e-5142-4325-9144-985963c1e271.jpg)

[...]

Issue #2244, opened 2023-04-26
[bug] Upload file not working in howto


AlfonsoGhislieri:
**Describe the bug**
Can't upload a file and then publish or edit a howto

**To Reproduce**
Steps to reproduce the behaviour:

1. Go to 'howto' you authored / create a howto if you havent
2. Click on 'edit.'
3. Scroll down to 'Or upload your files here'
4. Upload file
5. Try to revert to draft or publish

**Expected behaviour**

You should be able to revert to draft or publish when uploading a file

**Screenshots**

https://user-images.githubusercontent.com/652368/234537935-cd000a2d-8329-4c8a-9c37-2509252635fc.mov

**Additional context**

davehakkens:
I just tried this (in a preview link)
Seems to work here...
<img width="411" alt="afbeelding" src="https://user-images.githubusercontent.com/13672737/234852066-b4f1a94b-d619-4240-845a-9d4177386362.png">
What happend in your case, couldnt click publish?

AlfonsoGhislieri:
This happens everytime when editing the howto, I didn't check when creating it. I will add some tests to address this bug and work from there!

davehakkens:
tried it a bit more.
✅ It works if you edit/add files to a howto that had no files before.
✅ It works if you edit/replace uploaded files by another uploaded file
🉐 It does not work when you try to edit a how-to that had files uploaded before and want to replace it by a URL to download

AlfonsoGhislieri:
Thanks for this @davehakkens, good to know exactly which behaviour is currently bugged!

AlfonsoGhislieri:
So finally understood what the issue is, this issue exists for both updating and creating. Basically the way it is currently setup you can EITHER upload a file or add the download link. You can't do both which I think is the intended design? The issue here is that the user has no way of knowing that they need to remove one of the two file download options in order for them to upload/edit. 

The issue is that the following code seems to not be working:

`  public checkFilesValid = (formValues: IHowtoFormInput) => {
    if (formValues.fileLink && formValues.files.length > 0) {
      this.setState({ showInvalidFileWarning: true })
      return false
    } else {
      this.setState({ showInvalidFileWarning: false })
      return true
    }
  }`

Going to try and get this back to working!

davehakkens:
thanks for digging in it.
Yeh there used to be a message there for the user. Like here in the original issue https://github.com/ONEARMY/community-platform/issues/1609

[...]

PR #2245, opened 2023-04-26
2238/add file download to research

AlfonsoGhislieri:
PR Checklist

PR Type

## Description

This introduced the ability to upload files when creating and editing research articles as well as having the functionality to track the total number of downloads. 

This is PART 1, a lot of this functionality repeats with what already exists for the howtos. In order to avoid bloating this PR too much PART 2 will address refactoring, finally PART 3 will add icrementing the downloads via a firebase function to avoid unnecessary database calls.

## Git Issues

#2238

## Screenshots/Videos

image
---

## What happens next?

### DIFF
--- a/src/models/research.models.tsx
+++ b/src/models/research.models.tsx
@@ -24,9 +24,11 @@ export namespace IResearch {
     total_views?: number
     collaborators: string[]
     subscribers?: UserIdList
-    total_downloads?: number
-    fileLink?: string
-    files?: Array<IUploadedFileMeta | File | null>
+    attachments?: {
+      total_downloads: number
+      fileLink?: string
+      files?: Array<IUploadedFileMeta | File | null>
+    }
   } & Omit<FormInput, 'collaborators'>
 
   /** A research item update */
--- a/src/models/research.models.tsx
+++ b/src/models/research.models.tsx
@@ -24,6 +24,9 @@ export namespace IResearch {
     total_views?: number
     collaborators: string[]
     subscribers?: UserIdList
+    total_downloads?: number
+    fileLink?: string
+    files?: Array<IUploadedFileMeta | File | null>
   } & Omit<FormInput, 'collaborators'>
 
   /** A research item update */
@@ -33,7 +36,6 @@ export namespace IResearch {
     images: Array<IUploadedFileMeta | IConvertedFileMeta | null>
     videoUrl?: string
     comments?: IComment[]
-    total_views?: number
     collaborators?: string[]
     status: 'draft' | 'published'
   }
@@ -46,6 +48,8 @@ export namespace IResearch {
     tags: ISelectedTags
     creatorCountry?: string
     collaborators: string
+    fileLink?: string
+    files?: Array<IUploadedFileMeta | File | null>
   }
 
   /** Research items synced from the database will contain additional metadata */
--- a/src/pages/Howto/Content/Howto/HowtoDescription/HowtoDescription.tsx
+++ b/src/pages/Howto/Content/Howto/HowtoDescription/HowtoDescription.tsx
@@ -290,9 +290,9 @@ const HowtoDescription = ({ howto, loggedInUser, ...props }: IProps) => {
               <Text
                 data-cy="file-download-counter"
                 sx={{
-                  fontSize: '12px',
-                  color: '#61646B',
-                  paddingLeft: '8px',
+                  fontSize: 1,
+                  color: 'grey',
+                  paddingLeft: 1,
                 }}
               >
                 {fileDownloadCount}
--- a/src/pages/Research/Content/Common/Research.form.tsx
+++ b/src/pages/Research/Content/Common/Research.form.tsx
@@ -273,9 +273,9 @@ const ResearchForm = observer((props: IProps) => {
                                 placeholder="A comma separated list of usernames."
                               />
                             </Flex>
-                            <Label>
+                            <ResearchFormLabel>
                               Do you have supporting file you would like to add?
-                            </Label>
+                            </ResearchFormLabel>
                             <Flex
                               sx={{ flexDirection: 'column' }}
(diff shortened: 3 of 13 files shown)


src/setupTests.ts
```diff
@@ -3,3 +3,5 @@
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
 import '@testing-library/jest-dom/extend-expect'
+import ResizeObserver from 'resize-observer-polyfill'
+global.ResizeObserver = ResizeObserver
```
thisislawatts, line 7: **suggestion:** It would be worth explicitly adding this as a dev dependency, otherwise there is a risk it could be removed accidentally.

`yarn add --dev resize-observer-polyfill`

AlfonsoGhislieri: Good idea! Added in 8b1b0e8

thisislawatts:
> This is PART 1, a lot of this functionality repeats with what already exists for the howtos. In order to avoid bloating this PR too much PART 2 will address refactoring, finally PART 3 will add icrementing the downloads via a firebase function to avoid unnecessary database calls.

Would it be worth adding these steps as a task list to the original issue?

AlfonsoGhislieri:
Good idea @thisislawatts, just added them now! Also this is ready to review, sorry for all the commits and changes I maybe should have already broken up this PR more.. 😬

thisislawatts:
**question:** Did you take a look at what tests we could introduce here? I am thinking about the misbehaviour [[bug] Upload file not working in howto](https://github.com/ONEARMY/community-platform/issues/2244) you have found and whether we should prioritise addressing that before moving forward with this PR.

My reasoning is that:
* We should prioritise addressing functional defects in production before new features
* Any testing we introduce for how-to downloads can be adopted for Research downloads.

src/models/research.models.tsx
```diff
@@ -24,6 +24,9 @@ export namespace IResearch {
     total_views?: number
     collaborators: string[]
     subscribers?: UserIdList
+    total_downloads?: number
+    fileLink?: string
+    files?: Array<IUploadedFileMeta | File | null>
```
thisislawatts, line 29: **suggestion:** Would you be up for pushing these down under a common "namespace". The intention here being to remove some noise from the object root and group related properties together. 

Something like: 

```ts
type Item = {
  ...
  attachments?: {
    total_downloads: number;
    fileLink?: string;
    files?:  Array<blah>
  }
}
```

src/pages/Howto/Content/Howto/HowtoDescription/HowtoDescription.tsx
```diff
@@ -290,9 +290,9 @@ const HowtoDescription = ({ howto, loggedInUser, ...props }: IProps) => {
               <Text
                 data-cy="file-download-counter"
                 sx={{
-                  fontSize: '12px',
-                  color: '#61646B',
-                  paddingLeft: '8px',
```
thisislawatts, line 295: **praise:** Great to see this improvement, and happy to wave it through here but really it is not related to Research module so could be split into a separate PR.

src/pages/Research/Content/Common/Research.form.tsx
```diff
@@ -13,25 +13,29 @@ import {
...
 import { COMPARISONS } from 'src/utils/comparisons'
 import { stripSpecialCharacters } from 'src/utils/helpers'
-import { required } from 'src/utils/validators'
+import { required, validateUrlAcceptEmpty } from 'src/utils/validators'
 import { PostingGuidelines } from './PostingGuidelines'
 import { ResearchSubmitStatus } from './SubmitStatus'
 import { CategoriesSelect } from 'src/pages/Howto/Category/CategoriesSelect'
 import { RESEARCH_TITLE_MAX_LENGTH, RESEARCH_MAX_LENGTH } from '../../constants'
 
+const MAX_LINK_LENGTH = 2000
```
thisislawatts, line 30: **nitpick(non-blocking):** Move this to one of the `constants.ts` files as it is probably a constraint shared with the how-to files. Non-blocking as it may be something you had already identified for refactoring.

davehakkens:
Nice. thanks for adding!
Originally the issue was aiming for each step/update to option to attach files. Just as they can add videos or images.
Mainly because reseach is an ongoing proces, when you start you don't have the files yet. You will attach them as you move forward researching

I can imagine this makes it more complex since it's a bit different from the original usage in how-to. 
Your PR could also work for now, but i'd  be curious to hear what would be preffered from a dev point of view.

src/models/research.models.tsx
AlfonsoGhislieri: Really like this idea, great suggestion! Added in: 0e6dcc9. Will also extend this to the howto's when I get to the refactoring!

src/pages/Research/Content/Common/Research.form.tsx
AlfonsoGhislieri: Yes! I have this on the TODO for the refactoring that comes next.

AlfonsoGhislieri:
Ah I see @davehakkens, I didn't know we were going to extend it for updates but that makes sense. Considering there is still a fair bit of work to go to get this to be less messy I added it to the issue TODO list and this will be addressed in Part 4 once I have refactored and implemented firebase functions.

@thisislawatts I agree I think we should freeze for now and I will look into adding some better test coverage that we can then extend to the research articles.

Will mark this as blocked until this is in place.

AlfonsoGhislieri:
#2280 Has now been added to unblock #2244, this PR will also need to be updated with the suitable logic and tests to give feedback when a user tries to add both a link and uploads files. This again just really points towards how important it would be to find a way to combine both howto and research to avoid all this duplication of work and code :(

AlfonsoGhislieri:
Closing this as its become very stale and I’m working on a different implementation

PR #2252, opened 2023-04-26
chore: update snapshots when testing components

AlfonsoGhislieri:
PR Checklist

PR Type

## Description

Updates snapshot files when they are run, documentation explains it here: https://vitest.dev/config/#update

### DIFF
--- a/packages/components/vite.config.mts
+++ b/packages/components/vite.config.mts
@@ -13,4 +13,5 @@ export default defineConfig({
       reporter: ['text', 'json', 'html'],
     },
   },
+  update: true,
 })


thisislawatts:
@AlfonsoGhislieri what value do the snapshot tests provide if we are continually updating them? 

Maybe we would be better off removing the snapshot tests entirely and replacing them with a more explicit assertion that would be less brittle. 

This post expands on the idea a little more:
https://peterhrynkow.com/testing/2019/01/07/the-perils-of-snapshot-testing.html

Is there a specific problem you are looking to solve with this change?

AlfonsoGhislieri:
Ah yes it was late and I didn't really fully explain what the issue is. I noticed there is an issue where certain components are styled and making changes to the style will change the css class which then causes the snapshot test to fail. This happened to UserStatistics when I was updating the styled values for the Icon component. 

This does also likely point to the tests being too brittle which yes could simply mean that we should replace them as well!

thisislawatts:
I think that the way JSS-in-CSS results in dynamic CSS classes including these in the snapshot is likely to be too brittle. I think the better approach here would be to make the test assertations more explicit and avoid using the snapshot.

AlfonsoGhislieri:
Agreed @thisislawatts, will close this in favour of a later PR in making less brittle tests that don't rely on snapshots

PR #2258, opened 2023-04-27
feat(email-notifs): create emailed state and catch send email errors

evakill:
PR Checklist

PR Type

## Description
after testing with the data on dev, a couple more things popped up that needed to be addressed
1. we need to keep track of emails we've sent to users 
i added a boolean field to the notification object that indicates whether the notif has been sent by email, and set that state to true in the create email fn. the aggregation that populates the pending emails collection will skip notifs that users have already been emailed about.
2. better error handling
testing against messy dev data was a good stress test for the function. i found some edge cases that i should account for - like deleted users threw an error when we tried to get their email - and decided to wrap the loop in a try/catch so that one messed up email will not throw an error and crash the fn. we can see when errors are being logged [here](https://console.cloud.google.com/logs/query;query=resource.type%3D%22cloud_function%22%20resource.labels.function_name%3D%22emailNotifications-sendOnce%22%20resource.labels.region%3D%22us-central1%22;cursorTimestamp=2023-04-27T20:47:21.516122Z?project=precious-plastics-v4-dev&authuser=0&hl=en) which helps for debugging if we find emails seem to be missing when we start testing more comprehensively.
3. email notification settings
since email notifications are a new feature, most users don't have the `notification_settings` field set. the aggregation for pending emails only adds notifications for users that have this setting. for testing, i made an http endpoint `setEmailNotificationSettings` that sets everyone's setting to daily. i commented out this function because it shouldn't be deployed without better auth validation, but we'll need to do something similar when we're ready to release this feature on prod, so i left it in for reference. 
 
BUT it was still cool to start seeing real-ish emails pop up in mailtrap

## Git Issues

Closes #

## Screenshots/Videos
Screen Shot 2023-04-27 at 2 15 54 PM

---

## What happens next?


src/models/user.models.tsx
```diff
@@ -91,6 +91,7 @@ export interface INotification {
   type: NotificationType
   read: boolean
   notified: boolean
+  emailed?: boolean
```
thisislawatts, line 94: **question:** What do you think about this property being a timestamp? I can't think of anything concrete we would do with that information at this point, but it seems like useful data to collect.

functions/src/emailNotifications/index.ts
```diff
@@ -39,3 +43,18 @@ exports.sendOnce = functions.https.onCall(async (_, context) => {
     'Emails can be triggered by admins only.',
   )
 })
+
+// exports.setEmailNotificationSettings = functions.https.onRequest(async () => {
```
thisislawatts, line 47: **issue:** As we are not actively using this code here let's drop it from the PR. 

You've caught a really good issue here and it will require a little more thought before implementation. 

My preference would be that new users get the setting enabled by default but existing users need to manually enable it.

functions/src/emailNotifications/createEmail.ts
```diff
@@ -32,6 +32,15 @@ const getActionTypeFromNotificationType = (type: NotificationType) => {
   }
 }
 
+const getUserEmail = async (uid: string): Promise<string | undefined> => {
+  try {
+    const { email } = await firebaseAuth.getUser(uid)
+    return email
+  } catch (error) {
+    return undefined
```
thisislawatts, line 40: **suggestion:** Can we add a console.error to this `catch` so that we can debug if/when this happens. 

Also consider using `null` here rather than `undefined`. We probably want to return an explicitly blank response rather than a yet to be defined value. 

> * null is an assigned value. It means nothing.
> * undefined typically means a variable has been declared but not defined yet.
> * null and undefined are falsy values.

A useful article looking at the difference between the two in JS/TS
https://codeburst.io/javascript-null-vs-undefined-20f955215a2




```suggestion
    console.error('Unable to fetch user email', {error, userId: uid})
    return null
```

functions/src/emailNotifications/createEmail.ts
```diff
@@ -41,69 +50,81 @@ export async function createNotificationEmails() {
...
         },
-      },
-    })
+      })
+      const updatedNotifications = notifications.map((n) => ({
+        ...n,
+        emailed: true,
+      }))
+      await user.ref.update({ notifications: updatedNotifications })
+    } catch (error) {
```
thisislawatts, line 126: **praise:** Great to wrap this entire block in the error handling. :sparkles: 

Do you think we would also want to capture the affected user in the console.error?

thisislawatts (changes requested):

src/models/user.models.tsx
evakill: how about a reference to the email in email collection? could see this being useful for sure!

thisislawatts (approved):

src/models/user.models.tsx
thisislawatts: That sounds really useful 👍🏻

PR #2265, opened 2023-04-29
feat: adopt tslog

thisislawatts:
PR Checklist

PR Type

## Description

Introduce a functional logger for the client side context so we can aggregate usage information from users. This information can then be used in a variety of ways; including debugging, monitoring, and auditing.

One of the primary benefits of logging is that it can help developers debug software issues. When a problem occurs, logs can provide valuable information about what went wrong and why. This information can help developers identify and fix bugs more quickly and efficiently.

Logs can also be used to monitor the performance and health of a system. By analyzing log data, developers can identify trends and patterns that may indicate issues such as performance bottlenecks or security vulnerabilities. This information can be used to proactively address issues before they become critical.

This PR introduces the machinery to emit logs but it will need to be turned on individually for each environment, this is intentional due to the dependency on the _logging serverless function. 

---

## What happens next?

### DIFF
--- a/functions/src/logging/logging.ts
+++ b/functions/src/logging/logging.ts
@@ -51,7 +51,9 @@ export const handleCloudLoggingRequest =
     })
 
 export const logToCloudLogging = functions
-  .runWith({ memory: '512MB' })
+  .runWith({
+    memory: '512MB',
+  })
   .https.onRequest(
     handleCloudLoggingRequest({
       logger,
--- a/package.json
+++ b/package.json
@@ -99,8 +99,6 @@
     "oa-components": "workspace:*",
     "oa-shared": "workspace:*",
     "oa-themes": "workspace:*",
-    "pino": "^7.2.0",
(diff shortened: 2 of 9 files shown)


AlfonsoGhislieri:
How come we are switching from pino to tslog?

thisislawatts:
> How come we are switching from pino to tslog?

There is nothing being sent to Logflare at the moment. That is part of what I have been working on in the tslog PR. When I took it for a spin, pino doesn't work in the browser. No idea how we merged this originally because I can't see that it ever worked.

AlfonsoGhislieri:
> 
> There is nothing being sent to Logflare at the moment. That is part of what I have been working on in this PR. When I took it for a spin, pino doesn't work in the browser. No idea how we merged this originally because I can't see that it ever worked.

Ah that makes sense, I was always wondering what was going on with the logger... Would also be good to include some tests then to make sure that this doesn't happen again.

PR #2267, opened 2023-04-30
fix(email-notifs): fix links in email template

evakill:
PR Checklist

PR Type

## Description

you had suggested switching links based on environment, which i am still open to as a solution, but was having a hard time figuring out how to access env variables from the cloud function and wasn't sure it was worth passing in through the aggregated notifications...

anyway, here's an alternate solution. the templates are stored in the database, so we will have different templates for each environment + project anyway. we can hardcode the URLs as well at the top image and anything else (colors, etc) we might want to add in the future that may differ between projects. 

since this template is purely for version control, i put the prod URLs in, but i can change them to the dev URLs in the pp database (i haven't added templates for the other projects yet, not sure if that'll be included in initial release)

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/functions/src/emailNotifications/pp-template.handlebars
+++ b/functions/src/emailNotifications/pp-template.handlebars
@@ -40,20 +40,24 @@
             {{#if this.isComment}}
               <p>
                 New comment on your
-                <a href='{{this.relevantUrl}}'>{{this.resourceLabel}}</a>
+                <a
+                  href='https://community.preciousplastic.com{{this.relevantUrl}}'
+                >{{this.resourceLabel}}</a>
                 by
                 <a
-                  href='{{this.triggeredBy.userUrl}}'
+                  href='https://community.preciousplastic.com/u/{{this.triggeredBy.displayName}}'
                 >{{this.triggeredBy.displayName}}</a>
               </p>
             {{/if}}
             {{#if isMention}}
               <p>
                 <a
-                  href='{{this.triggeredBy.userUrl}}'
+                  href='https://community.preciousplastic.com/u/{{this.triggeredBy.displayName}}'
                 >{{this.triggeredBy.displayName}}</a>
                 mentioned you in this
-                <a href='{{this.relevantUrl}}'>{{this.resourceLabel}}</a>
+                <a
+                  href='https://community.preciousplastic.com{{this.relevantUrl}}'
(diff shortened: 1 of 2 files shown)


functions/src/emailNotifications/pp-template.handlebars
```diff
@@ -40,20 +40,24 @@
...
             {{#if isMention}}
               <p>
                 <a
-                  href='{{this.triggeredBy.userUrl}}'
+                  href='https://community.preciousplastic.com/u/{{this.triggeredBy.displayName}}'
                 >{{this.triggeredBy.displayName}}</a>
                 mentioned you in this
-                <a href='{{this.relevantUrl}}'>{{this.resourceLabel}}</a>
+                <a
+                  href='https://community.preciousplastic.com{{this.relevantUrl}}'
```
evakill, line 59: `relevantUrl` is a bit of a misnomer, what's stored in the notification is just the slug

thisislawatts: We can't use a hardcoded URL like this, we will instead need to have this defined at the environment level. I believe there should already be a variable floating around we can use:
https://github.com/ONEARMY/community-platform/blob/master/functions/src/seo/index.ts#L14:L14

Although it would require a bit of digging to figure out if this is available in _all_ functions or is instead something we have to add manually.

evakill: tried to explain my reasoning for this in the description - the templates that are used to send the emails are stored in the database, so we will have different templates for each project anyway. i don't know if there's that much benefit to having one template and adding conditionals based on the env variable. let me know if you disagree.

thisislawatts: Oh wow, I completely missed that message in the description. My apologies for bouncing about too fast. I did not understand that the templates are stored in the database 🤯

I wonder how we can make that more obvious for dumb me and Future Contributors

thisislawatts (approved):

functions/src/emailNotifications/pp-template.handlebars
evakill: good point! i think i can add comments to the handlebars template. this email flow is complex enough that it may even warrant a readme in this directory. i'll add better docs to my todo list :)

PR #2282, opened 2023-05-02
fix: remove use of theme from admin subheader

thisislawatts:
PR Checklist

PR Type

## Description

Removes direct usage of the theme object, instead accessing it within the `sx` property.

### DIFF
--- a/packages/themes/src/types/index.ts
+++ b/packages/themes/src/types/index.ts
@@ -117,4 +117,8 @@ export interface ThemeWithName {
   maxContainerWidth: number
   regular: number
   bold: number
+
+  sizes: {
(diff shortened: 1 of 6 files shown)


packages/themes/src/types/index.ts
```diff
@@ -117,4 +117,8 @@ export interface ThemeWithName {
   maxContainerWidth: number
   regular: number
   bold: number
+
+  sizes: {
+    container: number
+  }
```
thisislawatts, line 123: Echo echo. Repetition. I think we could probably use a base theme that is extended for each instance

evakill (approved):

PR #2366, opened 2023-05-20
fix: ensure plain object before save

thisislawatts:
PR Checklist

PR Type

## Description

There was a problem saving issues because the `collaborator` property being sent through to Dexie was still an observable array. This caused an error to be thrown by IndexDB/Dexie. 

This fix forces a plain object to be passed through.

Not addressed by this PR:
* Although an error was thrown by Dexie, the record on Firestore was updated. So when reloading the page the comment would be added as expected. What should the UI do if there is a crash at IndexDB, should it force a page reload to ensure alignment with origin?

---

## What happens next?

### DIFF
--- a/src/stores/Research/research.store.tsx
+++ b/src/stores/Research/research.store.tsx
@@ -6,6 +6,7 @@ import {
   runInAction,
   toJS,
 } from 'mobx'
+import { cloneDeep } from 'lodash'
 import { createContext, useContext } from 'react'
 import { MAX_COMMENT_LENGTH } from 'src/constants'
 import { logger } from 'src/logger'
@@ -162,7 +163,7 @@ export class ResearchStore extends ModuleStore {
 
     const researchData = await toJS(dbRef.get('server'))
     if (researchData && !(researchData?.subscribers || []).includes(userId)) {
-      await this.updateResearchItem(dbRef, {
+      await this._updateResearchItem(dbRef, {
         ...researchData,
         subscribers: [userId].concat(researchData?.subscribers || []),
       })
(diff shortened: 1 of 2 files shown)


src/stores/Research/research.store.tsx
```diff
@@ -162,7 +163,7 @@ export class ResearchStore extends ModuleStore {
 
     const researchData = await toJS(dbRef.get('server'))
     if (researchData && !(researchData?.subscribers || []).includes(userId)) {
-      await this.updateResearchItem(dbRef, {
+      await this._updateResearchItem(dbRef, {
```
thisislawatts, line 166: Refactoring for legibility, adopting a `_` to signal a private method

src/stores/Research/research.store.tsx
```diff
@@ -294,12 +295,12 @@ export class ResearchStore extends ModuleStore {
     update: IResearch.Update | IResearch.UpdateDB,
   ) {
     const user = this.activeUser
-    const item = this.activeResearchItem
+    const researchItem = this.activeResearchItem
```
thisislawatts, line 298: Improve description of value through variable naming

src/stores/Research/research.store.tsx
```diff
@@ -822,6 +689,141 @@ export class ResearchStore extends ModuleStore {
       ) ?? false
     )
   }
+
```
thisislawatts, line 692: One of our rules requires private functions to be listed after public ones.

iSCJT (approved):
Thanks for the fix and renaming @thisislawatts

PR #2376, opened 2023-05-23
feat: remove service workers

### LINKED ISSUE #2324: [bug] White screen of Death 💀
**Describe the bug**

Regular visitors to the site are intermittently presented with a blank screen. This is because the primary JS/CSS bundles have failed to load. This is a result of a new release being rolled out. 

**To Reproduce**
Steps to reproduce the behaviour working locally. 

1. `npm run build`
2. `npx serve -s build`
3. Delete/rename one of the JS files in dist
4. Witness the white screen of Death

It is worth noting that this demonstrates the failure state of the site when part of the JS bundle is missing, rather than strictly recreates the behaviour in our deployments. 

**Desired Behaviour**

* No error state, partial errors should deployments do not cause a white screen of death.
* OR a meaningful error about broken build

**Screenshots**

169112328-1aabdc2d-33f7-4668-8b26-88460f2dcfd5

**Additional context**

* Discord feedback - https://discord.com/channels/586676777334865928/1103192067083087882/1103192067083087882

thisislawatts:
PR Checklist

PR Type

## Description

Completely removes the service-workers and the offline capabilities they add to the website. Whilst this is a really powerful feature we do not have the capacity to maintain it in its current state. It may be something we reintroduce at a later stage but at the moment it is introducing [more problematic behaviour](https://github.com/ONEARMY/community-platform/pull/2376) than benefits from offline experience.

There is minimal data to back up this decision, the basis for removing the service workers is that we have more reports of bugs caused by it than we do feedback on the offline functionality. 

## Git Issues

Closes #2324


evakill (approved):
don't have much context on service workers, but agree better to provide a great experience than lots of extra features 👍

PR #2386, opened 2023-05-29
fix: compare against username property

thisislawatts:
PR Checklist

PR Type

## Description

The bug behaviour here is that the Follow button would not show the correct Follow/Following state for some users. 

Reverts to using the `userName` property of User entity. There is some drift across user entities where the `_id` property could contain either the Firestore Auth ID **or** the username. This is something that will need to be addressed separately but is out of scope for this change.

Original implementation:
https://github.com/ONEARMY/community-platform/commit/bc6985ecff43ed87ae1ea9afb0e431aacdb8a357#diff-404a1450ddfbbd25ee32f597e3e8675f7142774a2efa5a730044cdf6ffed0847R234

### DIFF
--- a/src/stores/Research/research.store.test.ts
+++ b/src/stores/Research/research.store.test.ts
@@ -15,13 +15,21 @@ jest.mock('../../utils/helpers', () => ({
   },
 }))
 
-const factoryResearchItem = async (researchItemOverloads: any = {}) =>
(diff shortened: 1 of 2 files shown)


AlfonsoGhislieri (approved):

AlfonsoGhislieri:
Looks good @thisislawatts! Will definitely need to look into the issue with the `_id `to make it all consistent

PR #2404, opened 2023-06-04
feat: adds logging backend method

thisislawatts:
PR Checklist

PR Type

## Description

Introduces a serverless function which writes data to Google's Cloud Logging platform: 

```bash
curl -X POST https://REGION-PROJECT_ID.cloudfunctions.net/logToCloudLogging-logToCloudLogging -H "Content-Type:application/json" -d '{"message":"Hello, World!"}'
```

This could then be used with by the client side logger introduced as part of #2265 

Outstanding work:
* [x] Introduce tests that verify functional correctness
* [x] Handle malformed requests


functions/src/logging/logging.ts
```diff
@@ -0,0 +1,58 @@
+import * as functions from 'firebase-functions'
+import { Logging } from '@google-cloud/logging'
+import { json } from 'body-parser'
+const jsonParser = json({
+  strict: true,
+})
+
+const logger = new Logging()
+
+export const handlelogToCloudLogging =
```
evakill, line 10: nit: camel case `handleLogToCloudLogging`

functions/src/logging/logging.ts
```diff
@@ -0,0 +1,58 @@
...
+      // If not, return a 400 error status code.
+      if (typeof req.body !== 'object' || req.body === null) {
+        return res.status(400).send('Request body must be a JSON object')
+      }
+
+      if (!Object.keys(req.body).length) {
+        return res.status(400).send('Request body must not be an empty object')
+      }
+
+      // Perform some validation on the request body
```
evakill, line 41: comment on 41 seems like it refers to the checks on 31-39 - remove or move up?

functions/src/logging/logging.ts
thisislawatts: Updated

functions/src/logging/logging.ts
thisislawatts: Ah this is a note about putting some kind verification in that endpoint only logs payloads which match an expected schema. Will remove for the time being :)

evakill (approved):
LGTM

PR #2415, opened 2023-06-09
feat: adds empty service-worker ahead of completely removing

thisislawatts:
This reverts commit 545f80481b3ead992603a031bed39bba79310495.

PR Checklist

PR Type

## Description

Introduces almost completely empty Service Worker script to remove all functionality. This is necessary because simply deleting the service worker file may not lead to the expected behaviour for all clients. Some browsers may not know that the file is meant to be gone and will continue serving the old service worker from their HTTP cache.

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,6 +136,7 @@
     "theme-ui": "^0.15.7",
     "ts-node": "^10.4.0",
     "web-vitals": "^1.1.2",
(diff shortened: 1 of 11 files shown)


davehakkens:
Would be up for merging but some tests seem to be failing?

PR #2435, opened 2023-06-19
fix: remove use of theme from logo

### LINKED ISSUE #2307: [refactor] Remove direct usage of Theme from "Logo.tsx"
Remove the following

```ts
```

Read more about it here: 
https://theme-ui.com/sx-prop

BaltacMihai:
PR Checklist

PR Type

## Description

## Git Issues

Closes #2307

### DIFF
--- a/src/pages/common/Header/Menu/Logo/Logo.tsx
+++ b/src/pages/common/Header/Menu/Logo/Logo.tsx
@@ -1,29 +1,13 @@
 import { Component } from 'react'
-// TODO: Remove direct usage of Theme
-import { preciousPlasticTheme } from 'oa-themes'
-const theme = preciousPlasticTheme.styles
-import { Flex, Image, Text } from 'theme-ui'
-import styled from '@emotion/styled'
+import { Flex, Image, Text, Box } from 'theme-ui'
 import { inject, observer } from 'mobx-react'
 import type { ThemeStore } from 'src/stores/Theme/theme.store'
-
 import { VERSION } from 'src/config/config'
 import { Link } from 'react-router-dom'
 
 interface IProps {
   isMobile?: boolean
 }
-
-const LogoContainer = styled(Flex)`
-  align-items: center;
-  position: relative;
-  padding: 10px 0;
-
-  @media only screen and (min-width: ${theme.breakpoints[1]}) {
-    margin-bottom: -50px;
-    padding: 0;
-  }
-`
 
 interface InjectedProps {
   themeStore: ThemeStore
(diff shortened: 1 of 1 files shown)


src/pages/common/Header/Menu/Logo/Logo.tsx
```diff
@@ -44,44 +28,49 @@ export class Logo extends Component<IProps> {
...
 
     return (
-      <>
-        <LogoContainer>
-          <Link to="/">
-            <Flex
-              ml={[0, 4]}
+      <Box
+        sx={{
+          py: ['10px', '10px', 0], // padding on y axes ( top & bottom )
```
thisislawatts, line 37: **suggestion:** You can reference [the `space` array available in the theme](https://github.com/ONEARMY/community-platform/blob/f5e1444ef627840a91c4dca14007ffbfa30d3855/packages/themes/src/project-kamp/styles.ts#L32-L35).

```suggestion
          py: [2, 2, 0], // padding on y axes ( top & bottom )
```

src/pages/common/Header/Menu/Logo/Logo.tsx
```diff
@@ -44,44 +28,49 @@ export class Logo extends Component<IProps> {
...
+      >
+        <Link to="/">
+          <Flex
+            ml={[0, 4]}
+            sx={{
+              zIndex: 1000,
+              display: 'flex',
+              alignItems: 'center',
+              width: ['50px', '50px', '100px'],
+              height: ['50px', '50px', '100px'],
```
thisislawatts, line 50: **question:** You seem to be including the `px` value here but not when declaring the width on the `<Image/>` element. Is that intentional?

src/pages/common/Header/Menu/Logo/Logo.tsx
```diff
@@ -44,44 +28,49 @@ export class Logo extends Component<IProps> {
...
-                title={nameAndVersion}
-              />
-            </Flex>
-            <Text
-              className="sr-only"
-              ml={2}
-              sx={{ display: ['none', 'none', 'block'] }}
-              color="black"
-            >
-              Precious Plastic
```
thisislawatts, line 80: **praise:** Great catch on this @BaltacMihai ✨ Thanks for fixing this!

src/pages/common/Header/Menu/Logo/Logo.tsx
BaltacMihai: To be honest, the pixel value makes it simpler to read, but I can adjust

thisislawatts: Understood, I agree that the theme abstraction takes some getting used to, but it means that if we wanted to change our 10px value to 9px we would only need to do so in the Theme configuration and not across all of our components.

BaltacMihai: I get it now.

src/pages/common/Header/Menu/Logo/Logo.tsx
BaltacMihai: It wasn't done on purpose. In order to implement the DRY principle (don't repeat yourself), I create one variable everytime the array is repeated.

thisislawatts (approved):

thisislawatts:
@all-contributors add @BaltacMihai for code

PR #2448, opened 2023-06-20
refactor: remove use of theme from ProfileButtons

### LINKED ISSUE #2302: [refactor] Remove direct usage of Theme from "ProfileButtons.tsx"
Remove the following

```ts
```

Read more about it here: 
https://theme-ui.com/sx-prop

BaltacMihai:
PR Checklist

PR Type

## Description

## Git Issues

Closes #2302


src/pages/common/Header/Menu/Profile/ProfileButtons.tsx
```diff
@@ -27,20 +23,22 @@ export class ProfileButtons extends Component<IProps> {
...
+              }}
+            >
               <ProfileButtonItem
                 link={'/sign-in'}
                 text="Login"
                 variant="secondary"
                 sx={{
-                  justifyContent: 'center',
+                  ..._commonMobileBtnStyle, // Merge with common style
                   fontWeight: 'bold',
```
AlfonsoGhislieri, line 39: **Question**: is the intent here to make a shared style to consolidate the mobile button style across the platform?

AlfonsoGhislieri (approved):

src/pages/common/Header/Menu/Profile/ProfileButtons.tsx
BaltacMihai: To be completely honest, I didn't look to see if there were any other parts; instead, I just created a local variable there (note that I added a dash before the variable name to indicate that it is local). I am now aware to follow the DRY principle (don't repeat yourself) and create a variable, function, or component whenever I feel the need to copy and paste something.

AlfonsoGhislieri: That makes sense, I was just confused by what was meant by '/ Merge with common style'

BaltacMihai: The comment serves as a descriptive indication that the _commonMobileBtnStyle and the extra styles listed in the sx prop are being combined. Yes, I get why it can be confusing. Should I get rid of it?

AlfonsoGhislieri: It’s been merged now but yes I would suggest getting rid of it, it’s pretty clear already and the comment won’t be obvious for future contributors/might make us think this is a TODO. Probably best to avoid future confusion!

PR #2453, opened 2023-06-21
fix: how-to most useful sort

iSCJT:
PR Checklist

PR Type

## Description

- Fixes how-to most useful sort
- Fixes display of useful count on how-to list for cards with long title

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/src/pages/Howto/Content/HowtoList/SortSelect.tsx
+++ b/src/pages/Howto/Content/HowtoList/SortSelect.tsx
@@ -9,9 +9,7 @@ interface ISortOption {
   label: string
 }
 
-interface IProps {
-  usefulCounts: { [key: string]: number }
-}
+interface IProps {}
 
 interface InjectedProps extends IProps {
   howtoStore: HowtoStore
@@ -47,7 +45,7 @@ class SortSelect extends Component<IProps, IState> {
   onSortValueChange(value: ISortOption) {
     this.setState({ value })
     if (value.value === 'most useful') {
-      this.injected.howtoStore.sortHowtosByUsefulCount(this.props.usefulCounts)
+      this.injected.howtoStore.sortHowtosByUsefulCount()
     } else {
       this.injected.howtoStore.sortHowtosByLatest()
     }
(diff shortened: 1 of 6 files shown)


src/pages/Howto/Content/HowtoList/SortSelect.tsx
```diff
@@ -9,9 +9,7 @@ interface ISortOption {
   label: string
 }
 
-interface IProps {
-  usefulCounts: { [key: string]: number }
-}
+interface IProps {}
```
AlfonsoGhislieri, line 13: **Question**: do we need to keep the `IProps` interface if it does not have any properties? Or is this being kept here in case we did want to use it in the future?

AlfonsoGhislieri (approved):

thisislawatts (changes requested):
**issue (blocking):** As this is a fix can we introduce a failing test which documents the bug and passes due to the change being made here?

src/pages/Howto/Content/HowtoList/SortSelect.tsx
iSCJT: to be honest I originally removed it completely but then couldn't get the injected store to work correctly so left the empty interface.  Any ideas?

AlfonsoGhislieri: Looking into other examples in the codebase I see that the empty Props iterface seems to be used frequently, so clearly this is an issue that has appeared. The only way I can think of which could work is:

```
@inject('howtoStore')
@observer
class SortSelect extends Component<**any**, IState> {
  constructor(props: any) {
    super(props)
    this.state = { value: null }
  }

  get injected() {
    return this.props as InjectedProps
  }
```

But then we would just be bypassing any type checking, which is something I just saw done in the [header](https://github.com/ONEARMY/community-platform/blob/master/src/pages/common/Header/Header.tsx) and [eventListings](https://github.com/ONEARMY/community-platform/blob/master/src/pages/Events/Content/EventsListings/EventsListings.tsx). But, I think it's best to leave it like this with the empty interface as is seen in other files.

thisislawatts (approved):

PR #2454, opened 2023-06-21
feature: research page card redesign

### LINKED ISSUE #2292: [refactor] Remove direct usage of Theme from "ResearchListItem.tsx"
Remove the following

```ts
```

Read more about it here: 
https://theme-ui.com/sx-prop

### LINKED ISSUE #2450: [feature request] Research page card new design

You may find it difficult to understand the cards on the Research page since there are many elements as you scroll, namely the right portion with the three numbers (Useful, comments, and updates).

In order to check whether there was something similar or how the issue was solved, I also looked at the other pages.
One solution would be to relocate the top icons from the "table header" to each card and add a tooltip that would explain each icon's meaning when the mouse is hovered over it. This is how we address the issue of visibility.

Why pagination isn't utilized for "tables" so we may restrict the number of entries per page is one thing I don't understand.

Therefore, pagination is used to reduce the amount of entries

**Additional context**
I quickly mocked up the version I was envisioning.
Group 98

Additionally, I'm not sure why the user needs to know how often it was updated when he only needs to know when the most recent update occurred.

BaltacMihai:
PR Checklist

PR Type

## Description

Place the top icons on each individual card instead of the "table header". I used a placeholder until we get an icon for the "Updates" sign.

--
Additionally, I changed the empty state symbol from "-" to "0" because the user will now understand it better. In the old form, the user might have wondered if the data had loaded or not.

## Git Issues

Closes #2450 
Closes #2292

## Screenshots/Videos

image


src/pages/Research/Content/ResearchListItem.tsx
```diff
@@ -1,46 +1,27 @@
...
   const collaborators = item['collaborators'] || []
   const usefulDisplayCount =
-    item.votedUsefulCount > 0 ? item.votedUsefulCount : '-'
+    item.votedUsefulCount > 0 ? item.votedUsefulCount : '0'
+
+  const _commonStatisticStyle = {
+    display: 'flex',
+    alignItems: 'center',
+    fontSize: [1, '16px', '16px'],
+  }
```
AlfonsoGhislieri, line 24: **Thought** interesting to see that we dont have 16 in the fontsize array, I'd suggest adding it, but I'm not sure it gets used elsewhere in the codebase and that would require updating the other fontSizes..

BaltacMihai: Only three files, including this one, use it, as far as I can tell. I suggest using a pre-existing value, such as 14 or 18 pixels, for linearity.

AlfonsoGhislieri: That sounds good to me, much better for future maintenance as well if we can standardise fontsizes across the platform!

BaltacMihai: I designed 2 types of icons that I think maybe fit.

Group 104

A @davehakkens I posted some comments on Figma about the overall icons

davehakkens:
Looks good!
Here you can find the icon @BaltacMihai 
research-icon

davehakkens:
Just noticed the font-size of the numbers is quite large, 18px.
Could this be reduced to 14px @BaltacMihai  (like on the how-to overview page)

PR #2455, opened 2023-06-21
fix: add optimistic locking to research updates

### LINKED ISSUE #2393: [bug] published research updates can be overwritten
**Describe the bug**
If 2 users have the same research article open at the same time to add an update then whoever publishes first can have their update overwritten when the second user publishes.

**To Reproduce**

1. Open the same research article in 2 different browsers
2. Click on add update in both browsers
3. Add update and save using browser 1, refresh the article to see that the update has been added
4. Add update and save using browser 2, refresh the article to see that the update has been added and browser 1 update is no longer present.
5. Refresh browser 1 and see that the original update is no longer there.

**Expected behaviour**
Updates should be atomic and any updates added to the research article should not be overwritten.

**Screenshots**

**Additional context**
Same problem would occur if any kind of modification is being made to the research article at the same time, possibly even a comment.  Need to reduce / eliminate use of cached docs for any crud operations for docs that have multiple potential users modifying.  Should also be more specific about which fields are being updated in the doc rather than sending a full new doc each time.  In progress pr for useful buttons incudes an update function that can handle partial docs.

CDeighton:
PR Checklist

PR Type

## Description

Currently if there are 2 people editing the same research article, the person to save second will have their changes silently overwrite the first persons. This raises an error to notify that someone else has saved first

## Git Issues

Closes #2393

## Screenshots/Videos

First save runs as expected

Screenshot 2023-06-21 at 11 45 38 PM

This still needs a bit of work to surface it to users, but proof that it blocks the secondary update:

Screenshot 2023-06-21 at 11 45 47 PM

### DIFF
--- a/src/models/common.models.tsx
+++ b/src/models/common.models.tsx
@@ -17,6 +17,10 @@ export type IModerationStatus =
   | 'rejected'
   | 'accepted'
 
+export interface ILockable {
+  lockVersion?: number
+}
+
 export interface IModerable {
   moderation: IModerationStatus
   _createdBy?: string
--- a/src/stores/Research/research.store.tsx
+++ b/src/stores/Research/research.store.tsx
@@ -521,13 +521,22 @@ export class ResearchStore extends ModuleStore {
       .collection<IResearch.Item>(COLLECTION_NAME)
       .doc(values._id)
     const user = this.activeUser as IUser
-    const updates = (await dbRef.get())?.updates || [] // save old updates when editing
+    const record = await dbRef.get()
+    const updates = record?.updates || [] // save old updates when editing
     const collaborators = Array.isArray(values?.collaborators)
       ? values.collaborators
       : (values.collaborators || '')
           .split(',')
           .map((s) => s.trim())
           .filter(Boolean)
+
+    // TODO: if lockVersion is an acceptable idea, then the next step would be to add a nice message to the UI.
+    // Potentially just an 'Another user just updated this article, you'll need to refresh and try again' message in the
+    // Upload progress modal
+    if (values.lockVersion !== record?.lockVersion)
+      throw new Error(
+        `lock version ${values.lockVersion} | ${record?.lockVersion}`,
+      )
 
     try {
       // populate DB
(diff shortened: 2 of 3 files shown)


src/models/common.models.tsx
```diff
@@ -17,6 +17,10 @@ export type IModerationStatus =
   | 'rejected'
   | 'accepted'
 
+export interface ILockable {
+  lockVersion?: number
```
CDeighton, line 21: It'd be nice if there was a way of defaulting this to zero? Couldn't see a clean way with a quick google, but `ts` is not my forte 🙂

src/stores/Research/research.store.tsx
```diff
@@ -521,14 +521,20 @@ export class ResearchStore extends ModuleStore {
...
       ? values.collaborators
       : (values.collaborators || '')
           .split(',')
           .map((s) => s.trim())
           .filter(Boolean)
 
+    // TODO: if lockVersion is an acceptable idea, then the next step would be to add a nice message to the UI.
+    // Potentially just an 'Another user just updated this article, you'll need to refresh and try again' message in the
+    // Upload progress modal
+    if (values.lockVersion !== record?.lockVersion) throw new Error(`lock version ${values.lockVersion} | ${record?.lockVersion}`)
```
CDeighton, line 536: Standard optimistic locking things

If the `lockVersion` has not yet been set, both will be undefined, so it will save
If the `lockVersion` of the form matches the one in the db, it will save
If the `lockVersion` of the form does not match the db, someone else must have saved at some point, so block saving

src/models/common.models.tsx
```diff
@@ -17,6 +17,10 @@ export type IModerationStatus =
   | 'rejected'
   | 'accepted'
 
+export interface ILockable {
```
CDeighton, line 20: Potentially a bit YAGNI to have an interface for this when we could just add the field to the one model. But this approach can be used with any form that allows concurrent edits, so might be useful down the track 🤷

iSCJT: You could potentially look at adding the property to IModerable and defaulting the value in the _setDocMeta function in [src/stores/databaseV2/DocReference.tsx](https://github.com/ONEARMY/community-platform/blob/master/src/stores/databaseV2/DocReference.tsx)

benfurber:
This is quite old now(!) Is it still relevant and needed?

davehakkens:
@benfurber It's for an edge case that realistically doesnt happen much (never as far as i'm aware.)
So up for parking it

PR #2458, opened 2023-06-23
[refactor] Remove direct usage of Theme from "ImageInput.tsx"

### LINKED ISSUE #2285: [refactor] Remove direct usage of Theme from "ImageInput.tsx"
Remove the following

```ts
```

Read more about it here: 
https://theme-ui.com/sx-prop

jonboiser:
PR Checklist

PR Type

## Description

Implements the refactoring in #2285.

However, in converting the styled-components code to theme UI, I am not sure whether there is a better way of reusing styles, aside from defining a the base style object, then spreading it into the `sx` props.

## Git Issues

Closes #2285

## Screenshots/Videos

TODO, but the before/after looks the same to me. I could not get the overlay to activate because dropzone wouldn't work in dev mode.

Before

After
---

## What happens next?

### DIFF
--- a/src/common/Form/ImageInput/ImageInput.tsx
+++ b/src/common/Form/ImageInput/ImageInput.tsx
@@ -1,12 +1,9 @@
 import * as React from 'react'
-import { Box, Image } from 'theme-ui'
-import styled from '@emotion/styled'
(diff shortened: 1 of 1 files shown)


AlfonsoGhislieri (approved):
Great work here @jonboiser, really appreciate the refactoring work! :)

PR #2465, opened 2023-06-24
feat: add categories admin to adminv2

CDeighton:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

Add category viewing/adding/updating to the /admin_v2 screen

## Git Issues

https://github.com/ONEARMY/community-platform/issues/2321

PR #2463, opened 2023-06-24
feature: add tags admin to admin_v2

CDeighton:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

Add tag viewing/adding/updating to the `/admin_v2` screen

## Git Issues

Handles half of #2321 - categories still need to be moved

PR #2466, opened 2023-06-24
Feature/merge to v2 drop legacy

### LINKED ISSUE #2321: [feature request] Merge admin and adminv2

There are two admin areas:

* Legacy – http://localhost:3000/admin
* New – http://localhost:3000/admin_v2

These should be consolidated into a single admin area.
* Ideally, ensure all of the functionality from legacy admin (`/admin`) is available in the new admin (`/admin_v2`).
* New admin routes are moved to exist on the `/admin` URL

CDeighton:
PR Checklist

PR Type

## Description

This depends on #2463 and #2465

Drops the old `/admin` code now that all of the features are in `admin_v2`

## Git Issues

Closes #2321


src/modules/index.ts
```diff
@@ -8,14 +8,14 @@ export enum MODULE {
...
-  ADMIN = 'admin_v2',
+  ADMIN = 'admin',
 }
 
 export const getSupportedModules = (): MODULE[] => {
   const envModules: string[] =
     getConfigurationOption(
       'REACT_APP_SUPPORTED_MODULES',
-      'howto,map,events,research,academy,user,admin_v2',
+      'howto,map,events,research,academy,user,admin',
```
CDeighton, line 18: Drops legacy admin and shifts `admin_v2` routes to `admin`

iSCJT (changes requested):
@CDeighton one additional consideration with this is that the beta testers admin functionality isn't yet replicated in the new users page so that will also need to be replicated somewhere before fully deprecating v1 admin.

AlfonsoGhislieri:
Going to assign this over to you @iSCJT!

CDeighton:
> @CDeighton one additional consideration with this is that the beta testers admin functionality isn't yet replicated in the new users page so that will also need to be replicated somewhere before fully deprecating v1 admin.

@iSCJT oh yikes sorry 🤦 good catch! I thought the new `/admin_v2/users` route covered it but it looks like there is no way to add roles to users on that page.

If it is just the roles, how do you feel about having them on the `u/${username}/edit` screen (mockup below) - 

Screenshot 2023-06-28 at 11 29 37 PM

I'd hide the `super-admin` checkbox if a non-super-admin was looking at the screen of course. This way feels a little better to me because it consolidates the user management into a single place. Though it is very possible that I've missed another detail somewhere 😅 

I'm happy to just lift the existing screens as they are if you prefer though, I think it's roughly the same amount of work either way

iSCJT:
@CDeighton No problem at all, it's not the most obvious and intuitively you'd expect it to be in the new users screen.  At the moment we're trying to centralise the admin functionality into one place rather than have functionality spread across different pages so just lift the existing page into the v2 admin please.  Thank you 🌟

CDeighton:
@iSCJT should be good now 🤞 I'd recommend reviewing it commit by commit. Theres one for each of the admin and beta tester screens, then one to drop the old admin screen

PR #2469, opened 2023-06-26
refactor: rename "yellow" variable to "accent"

BaltacMihai:
PR Checklist

PR Type

## Description

When I was looking through the application, I noticed that we use a color variable called yellow, which has 2 statuses: base and hover. Looking at the styles for each project, I came to the conclusion that this name is for the facade (it's not always yellow). I also looked in the documentation to see if this yellow appears as something "standard".

I changed the name from yellow to accents, and I changed it everywhere I found it used. Thus, it will be easy for the developer to understand the role of color.


AlfonsoGhislieri (approved):
**Praise**: great catch here @BaltacMihai, this definitely seems to be confusing with some colors not actually even being yellow..

Good work!

PR #2472, opened 2023-06-26
refactor: imageConverter general theme removed

### LINKED ISSUE #2288: [refactor] Remove direct usage of Theme from "ImageConverter.tsx"
Remove the following

```ts
```

Read more about it here: 
https://theme-ui.com/sx-prop

BaltacMihai:
PR Checklist

PR Type

## Description

## Git Issues

Closes #2288

### DIFF
--- a/src/common/Form/ImageInput/ImageConverter.tsx
+++ b/src/common/Form/ImageInput/ImageConverter.tsx
@@ -2,10 +2,6 @@ import * as React from 'react'
 import { Flex } from 'theme-ui'
 import imageCompression from 'browser-image-compression'
 import type { IConvertedFileMeta } from 'src/types'
-import styled from '@emotion/styled'
-// TODO: Remove direct usage of Theme
-import { preciousPlasticTheme } from 'oa-themes'
-const theme = preciousPlasticTheme.styles
 
 interface IProps {
   file: File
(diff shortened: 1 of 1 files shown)


src/common/Form/ImageInput/ImageConverter.tsx
```diff
@@ -80,9 +65,19 @@ export class ImageConverter extends React.Component<IProps, IState> {
...
+          backgroundPosition: 'center',
+          backgroundRepeat: 'no-repeat',
+          width: '100%',
+          height: '100%',
+        }}
+        sx={{
+          border: '1px solid ',
+          borderColor: 'offwhite',
+          borderRadius: 1,
         }}
```
AlfonsoGhislieri, line 81: **Nitpick**: Not a big deal but the correct conversion for the `borderRadius` should be 1.25 rather than 1 if we want it to equal 5px (4px is the default).

BaltacMihai: Mmm, it appears to be 5px in the console. I chose 1 because I use sx, which allows you to design elements directly within a webpage using values taken from your theme. We assumed that the border radius in our situation could be space and entered 1, which is the second element of the space vector, or 5px. To confirm that it actually reads that vector and operates, I also tested it with other values.

AlfonsoGhislieri: Ah good catch, I didn't realize someone had added value scales for border radii. Just saw that it was added to overrite the default scales..

This makes me think we should probably add some better documentation for contributors to follow where all this is nicely outlined 🤔

Also the fact that these values are the same for the most part also does point to some more refactoring.. Not very clean with all this duplication

PR #2475, opened 2023-06-26
refactor: remove usage of theme from spaceprofile

### LINKED ISSUE #2297: [refactor] Remove direct usage of Theme from "SpaceProfile.tsx"
Remove the following

```ts
```

Read more about it here: 
https://theme-ui.com/sx-prop

benfurber:
PR Checklist

PR Type

## Description

Remove direct usage of Theme from "SpaceProfile.tsx".

## Git Issues

Closes #2297


src/pages/User/content/SpaceProfile.tsx
```diff
@@ -207,14 +143,47 @@ const renderMachineBuilderXp = (machineBuilderXp: Array<IMAchineBuilderXp>) => (
...
 )
 
+const MobileBadge = ({ children }) => (
+  <Flex
+    sx={{
+      flexDirection: 'column',
+      marginBottom: 0,
+      position: 'relative',
+      width: '100%',
+      '@media only screen and (min-width: 52em)': {
```
benfurber, line 173: I really don't like that I'm defining this breakpoint. Any ideas?

src/pages/User/content/SpaceProfile.tsx
```diff
@@ -207,14 +143,47 @@ const renderMachineBuilderXp = (machineBuilderXp: Array<IMAchineBuilderXp>) => (
...
+        alignItems: 'center',
+      },
+      '@media only screen and (min-width: 70em)': {
```
benfurber, line 176: Ditto above comment.

src/pages/User/content/SpaceProfile.tsx
thisislawatts: Thanks for calling this out @benfurber, I think that we should be able to write this in a shorthand manner using:

```suggestion
      height: ['300px', '300px', '500px'],
```
This would mean that:
* The default height is 300px
* At the 0th breakpoint & larger, height is 300px
* At the 1st breakpoint & larger, height is 500px

It took me some time to get used to this convention of the theming systems but I believe this slightly opaque syntax is the only way to work with breakpoints. 
https://theme-ui.com/sx-prop#responsive-values

benfurber: Thanks @thisislawatts, I think I get it now(!) Does it look right to you now?

benfurber:
As you can see I've made a mess with the rebasing! I'll close this PR later and apply my changes fresh.

PR #2477, opened 2023-06-26
[refactor] consolidate common style variables

### LINKED ISSUE #2474: [refactor] consolidate common sytle variables

We have multiple different files for the different platforms and there is a lot of overlap and duplication of the different variables, these are the files in question below:

[Project kamp](https://github.com/ONEARMY/community-platform/blob/master/packages/themes/src/project-kamp/styles.ts)
[Fixing fashion](https://github.com/ONEARMY/community-platform/blob/master/packages/themes/src/fixing-fashion/styles.ts)
[Precious plastics](https://github.com/ONEARMY/community-platform/blob/master/packages/themes/src/precious-plastic/styles.ts)

Consolidation of variables that are the same across all the different platforms into common files. This makes it easier to make changes, maintain and keep uniformity across the platform.

Similar to what was done for:
[colors](https://github.com/ONEARMY/community-platform/blob/master/packages/themes/src/common/colors.ts)
[text variant](https://github.com/ONEARMY/community-platform/blob/master/packages/themes/src/common/textVariants.ts)

BaltacMihai:
PR Checklist

PR Type

## Description

Consolidation of variables that are the same across all the different platforms into common files. This makes it easier to make changes, maintain and keep uniformity across the platform.

## Git Issues

Closes #2474


AlfonsoGhislieri:
Thanks for having already raised this so quickly @BaltacMihai, wasn’t expecting this so fast! Unfortunately the amount of overlap here really does highlight like @thisislawatts mentioned in #2474 that the better approach should be to use a base theme instead. 

Sorry about this and really do appreciate the work you’ve put into this already!

BaltacMihai:
Yes, I will try to make a common theme, I think it's a good step that I managed to remove all the repeating elements, so it's easier for me to create this common theme. Instead, I have to think about how I could modify it, so that it would be easy to overwrite certain elements, add something, etc.

packages/themes/src/common/index.ts
```diff
@@ -0,0 +1,159 @@
+class AuxiliaryStyle {
+  static _commonInputStyle = {
```
AlfonsoGhislieri, line 2: **Question**: why are we using a class here? If we are just using it to store static variables and no methods we can just use an object instead right?

packages/themes/src/common/index.ts
```diff
@@ -0,0 +1,159 @@
...
+    body: `'Inter', Arial, sans-serif`,
+  }
+  static _commonSpace = [
+    0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55, 60, 65, 70, 75, 80, 85, 90,
+    95, 100, 105, 110, 115, 120, 125, 130, 135, 140,
+  ]
+}
+
+export const baseTheme = {
+  colors: AuxiliaryStyle._colors,
```
AlfonsoGhislieri, line 52: **Thought**: Should we maybe separate the `baseTheme` and `AuxiliaryStyle` into two separate files? I know these two are closely related but it might make things more readable and easier to maintain down the line.

packages/themes/src/common/index.ts
```diff
@@ -0,0 +1,159 @@
...
+    borderRadius: 1,
+    '&:focus': {
+      borderColor: '#83ceeb',
+      outline: 'none',
+      boxShadow: 'none',
+    },
+  }
+
+  static _colors = {
+    white: 'white',
```
AlfonsoGhislieri, line 12: **Issue**: the use of underscores here also goes against convention for private variables. Underscores are used as to indicate private variables or function that are not meant to be accessed directly from outside the class/object. Since these are intended for use outside the class, it would be better to remove the underscores.

packages/themes/src/common/index.ts
AlfonsoGhislieri, line 2: **Nitpick(not blocking)**: these all seem to be common style variables, we should change the name of the class/objec to reflect this more accurately (eg: commonStyles)?

AlfonsoGhislieri (changes requested):
Thanks @BaltacMihai good work! I also am wondering

packages/themes/src/common/index.ts
BaltacMihai: Yes, my first thought was that I want a class that does not have the possibility to change the variables, that's why I put static, so we encourage the subsequent modification only on these components to have linearity. But yes, it is possible to return to an object, since currently there is no direct modification on common styles.

packages/themes/src/common/index.ts
BaltacMihai: Yes, that's a good idea.

packages/themes/src/common/index.ts
BaltacMihai: Here is my mistake, I put "_" in the initial version where I had separate variables and I didn't take it into account when I refactored.

packages/themes/src/common/index.ts
BaltacMihai: Yes, that seems like a more accurate name for what we have.

AlfonsoGhislieri (approved):
Thanks @BaltacMihai looks good to go now! Great work

PR #2479, opened 2023-06-26
fix: switch notification store to use username instead of userid

iSCJT:
PR Checklist

PR Type

## Description

It has been noticed duplicate users are getting reintroduced into the system and the 'wrong' users are in some cases the only ones getting updated. This was first noticed in the notifications records saved for the user records being different. As a result all uses of the user _id are being switched for userName so that if duplicates are created that it's definitely the user record that has an _id that matches the username that receives updates.  This pr:

- Switches use of user._id to user._userName in the notifications store.
- Fixes 'notifications' typos

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/src/stores/User/notifications.store.ts
+++ b/src/stores/User/notifications.store.ts
@@ -92,7 +92,7 @@ export class UserNotificationsStore extends ModuleStore {
           ? [...toJS(user.notifications), newNotification]
           : [newNotification]
 
-        await this._updateUserNofications(user, notifications)
+        await this._updateUserNotifications(user, notifications)
       }
     } catch (err) {
       logger.error(err)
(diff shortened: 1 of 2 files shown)


thisislawatts:
@iSCJT could you explain why this change is being made?
> Switches use of user._id to user._userName

iSCJT:
> @iSCJT could you explain why this change is being made?
> 
> > Switches use of user._id to user._userName

I've updated the pr description @thisislawatts

src/stores/User/notifications.store.ts
```diff
@@ -72,7 +72,7 @@ export class UserNotificationsStore extends ModuleStore {
           _created: new Date().toISOString(),
           triggeredBy: {
             displayName: triggeredBy.displayName,
-            userId: triggeredBy._id,
+            userId: triggeredBy.userName,
```
evakill, line 75: would it be a big diff to change this to be called userName for clarity?

iSCJT: had a think and would definitely be clearer although we'd need to run some kind of data migration for existing records which I'm not sure is worth the time to code right now.  I'd prefer to hold off and then get a proper db with a relational notifications table 🤓

evakill: fair! in that case could you add a comment / todo to this code explaining the reasoning?

evakill (approved):

PR #2483, opened 2023-06-26
fix: switch user _id to username

iSCJT:
PR Checklist

PR Type

## Description

Further updates to switch the use of user _id's to usernames in the code base with the aim of preventing duplicate users being created.

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/src/stores/User/user.store.ts
+++ b/src/stores/User/user.store.ts
@@ -384,7 +384,7 @@ export class UserStore extends ModuleStore {
 
         const dbRef = this.db
           .collection<IUser>(COLLECTION_NAME)
-          .doc(updatedUser._id)
+          .doc(updatedUser.userName)
 
         await dbRef.set(updatedUser)
         //TODO: ensure current user is updated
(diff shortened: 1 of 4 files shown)


src/stores/User/user.store.ts
```diff
@@ -384,7 +384,7 @@ export class UserStore extends ModuleStore {
 
         const dbRef = this.db
           .collection<IUser>(COLLECTION_NAME)
-          .doc(updatedUser._id)
```
thisislawatts, line 387: **question (blocking):** This seems to be the only part of the code in this PR which may affect whether a new User is being created. 

Could you explain a bit more about the change you are looking to make here? My understanding was that updatedUser._id should be equal to updatedUser.userName.

thisislawatts (changes requested):

src/stores/User/user.store.ts
iSCJT: Duplicate users are getting reintroduced into the system again and the 'wrong' users are in some cases the only ones getting updated.  Trying to clear all uses of the _id so that if duplicates are created that it's definitely the user record that has an _id that matches the username that gets updated.

thisislawatts:
Closing to keep the concept of `_id` and `userName` as separate entities.

PR #2405, opened 2023-06-06
fix(emails): use userName in links

evakill:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

i was mistakenly using the display name in the url slug instead of the user name. this fixes the issue and updates templates and tests. 

steps to deploy:
- [ ] PR approval
- [ ] @evakill updates templates in dev databases (backwards compatible)
- [ ] deploy function to dev
- [ ] test in dev
- [ ] @evakill updates templates in prod databases
- [ ] merge and deploy

## Git Issues

Closes #2363 

## Screenshots/Videos

_If useful, provide screenshot or capture to highlight main changes_

---

## What happens next?

Thanks for the contribution! We try to make sure all PRs are reviewed ahead of a monthly dev call (first Monday of the month, open to all!).

If the PR is working as intended it'll be merged and included in the next platform release, if not changes will be requested and re-reviewed once updated.

If you need more immediate feedback you can try reaching out on Discord in the [Community Platform `development` channel](https://discord.com/channels/586676777334865928/938781727017558018).

Issue #2364, opened 2023-05-18
[feature request] add 'follow' notifications to emails


davehakkens:
### Description

We recently introduced the 'follow' button in the research module here https://github.com/ONEARMY/community-platform/pull/2145
The notifications for this are already intergrate into website, but we want to add them to email notifications.

### Page related
[Research topics](https://dev.onearmy.world/research)

### More information
[Here in the readme](https://github.com/ONEARMY/community-platform/blob/63aeddff1b07ce0cac790dadf260e09d121c9da0/functions/src/emailNotifications/README.md)

### Currently on website 
<img width="190" alt="afbeelding" src="https://github.com/ONEARMY/community-platform/assets/13672737/34c2e466-027a-498c-af0c-4eaef1de40b8">

<img width="297" alt="afbeelding" src="https://github.com/ONEARMY/community-platform/assets/13672737/8acd9857-120f-4e90-ab7e-594eac4721af">

PR #2507, opened 2023-06-30
refactor(emails): dynamically generate email html

evakill:
PR Checklist

PR Type

## Description

After implementing email templates with handlebars, we recognized the friction this causes when the templates need to be changed. In an effort to reduce firebase dependency and make the email development process more accessible, we decided to start dynamically generating HTML email templates in code. 

this is done by [getNotificationEmailTemplate](functions/src/emailNotifications/getNotificationEmailTemplate.ts) called from [createNotificationEmails](functions/src/emailNotifications/createNotificationEmails.ts) for each set of unread notifications aggregated by user.

this will fix https://github.com/ONEARMY/community-platform/pull/2405 and allow future bugfixes like this one to be much more straightforward.

it will also set us up well to knock out https://github.com/ONEARMY/community-platform/issues/2364 quickly using this pattern.

@davehakkens also requested a design change to the templates: rather than separating the notifications into "comments" and "useful" sections, we should use a flat notification list. the changes are included in this PR

## Git Issues

Closes #

## Screenshots/Videos

email screenshots from test snapshots
Screen Shot 2023-07-04 at 11 22 10 AM
Screen Shot 2023-07-04 at 11 22 30 AM

---

## What happens next?


functions/src/emailNotifications/consts.ts
```diff
@@ -0,0 +1,11 @@
+export const COMMENT_SECTION_IMAGE = `http://cdn.mcauto-images-production.sendgrid.net/cb5a685b085b7e7c/1e6f3be3-e6fd-46c5-b30f-bcde950b0dbc/165x131.png`
```
thisislawatts, line 1: **nitpick:** Elsewhere in the codebase we use constants for these files.

functions/src/emailNotifications/pp-template.handlebars
thisislawatts, line 1: **praise:** 🔥 Farewell templates

functions/src/emailNotifications/createEmailNotifications.ts
```diff
@@ -0,0 +1,103 @@
...
+      )
+    } catch (error) {
+      console.error('Error sending an email', { error, userId: _userId })
+      await updateEmailedNotifications(
+        toUserDoc,
+        pendingNotifications,
+        'failed',
+      )
+    }
+  }
```
thisislawatts, line 102: **issue:** As we are doing some refactoring, it would be useful to extract the complex business logic here into a separate functions/methods and adopt the "return early" pattern to improve code readability and maintainability.

```ts
// Original code
for (let i = 0; i < arr.length; i++) {
    if (arr[i] === null) {
        continue;
    }

    // Complex business logic here...
}

// Refactored code
function handleItem(item: any) {
    if (item === null) {
        return;
    }

    // Complex business logic here...
}

for (let i = 0; i < arr.length; i++) {
    handleItem(arr[i]);
}
```

In the refactored code, the handleItem function encapsulates the complex business logic and uses return early to skip null items, making the loop in the main code simpler and easier to understand.

functions/src/emailNotifications/consts.ts
thisislawatts: Apologies for being unclear, my comment is referencing the filename `consts.ts` rather than `constants.ts`

thisislawatts (approved):

PR #2555, opened 2023-07-04
feat: add validation skip to research description

### LINKED ISSUE #2541: [feature request] allow draft research save without form validation
A human that wants to save an incomplete draft version of a research item will be prevented from doing so unless they have completed all the required fields.  This defeats the point of saving a draft.

When saving a draft research the form validation should be bypassed and they should be allowed to save an incomplete version.  Form validation should only apply when publishing or performing a final save of a research item.

N/A

**Additional context**

benfurber:
PR Checklist

PR Type

## Description

This adds a skip to the descripton field on the research form when saving a draft.

A couple of decisions I made as part of this - very happy to work further if others disagree.

1. I've only applied the validation skip to the description even though I believe the issue considered applying it to the title too. I'm guessing that if a title is left blank it'll cause issues deciding on a slug, etc.? And will be easier for users to navigate drafts if they don't have a load of 'untitled' research or similar?
2. The validation check is essentially off by default and then turned on to publish. That way users won't be annoyed by the error message until it's important. Totally lots of valid arguments to do it the other way around though.

Like I said, very happy to keep working on this if others disagree with my thinking. :)
 

## Git Issues

Closes #2541


evakill (commented):
hey @benfurber! thanks for picking this one up

i pulled down your branch to test and the save as draft is working well without requiring the description, but it appears that i can also click publish with an empty description, which we don't want. i wonder if final form and react state aren't interacting well together? an idea i had was to make the moderation state a field value on the form - then the validation function can access the moderation state through the second param in the validate function (see docs https://final-form.org/docs/react-final-form/types/FieldProps#validate) let me know if that makes sense to you.

another issue i spotted was that once a draft is published it can be moved to "accepted" using this green check box. 
Screen Shot 2023-07-04 at 7 03 43 PM
this doesn't have any safeguards for validity, so incomplete research can be published from the draft state. we'd need to add additional checks here. ideally, they'd be shared through some sort of config between the form and this button. this also makes working with final-form a bit tricky... let me know if you want to chat more on potential solutions!

benfurber:
@evakill Oh wow, apologies for not testing this properly!

I'd love to talk about this more. Are you free/available now-ish through discord?

benfurber:
As we need to workout what to do with the accepted button element too, I'm thinking that the validation needs to be moved to the form level so that the rules can be easily shared. First attempt using [this as guide](https://codesandbox.io/s/github/final-form/react-final-form/tree/master/examples/record-level-validation?file=/index.js) didn't go very well as the whole form went into a loading state every time the title did a validation check(!)

I'm thinking of giving [this pattern](https://codesandbox.io/s/kl9n295n5?file=/index.js:1606-1633) a go tomorrow.

iSCJT:
@benfurber @evakill I've added a separate issue for the fact that moderators can approve a draft item: https://github.com/ONEARMY/community-platform/issues/2571

Will leave Eva to circle back on approaches as she had her head into the problem yesterday.

packages/cypress/src/integration/research/write.spec.ts
```diff
@@ -118,32 +152,16 @@ describe('[Research]', () => {
...
-      ).should('exist')
-
       cy.step('Update the intro')
       cy.get('[data-cy=intro-title]').clear().type(expected.title)
-
       cy.get('[data-cy=submit]').click()
 
       cy.step('Open the updated research article')
-
+      cy.wait(2000)
```
benfurber, line 164: I think this deals with a flaky test...

packages/cypress/src/integration/research/write.spec.ts
```diff
@@ -118,32 +152,16 @@ describe('[Research]', () => {
...
-      ).should('not.exist')
-
-      cy.get('[data-cy=intro-title]')
-        .clear()
-        .type('qwerty')
-        .blur({ force: true })
-      cy.contains(
-        'Titles must be unique, please try being more specific',
-      ).should('exist')
-
```
benfurber, line 139: Flow covered above, I don't think it's needed a second time and thus worth speeding the tests up.

benfurber:
What do you think about this approach @evakill?

packages/cypress/src/integration/research/write.spec.ts
```diff
@@ -12,9 +15,40 @@ describe('[Research]', () => {
     previousSlugs: ['create-research-article-test'],
   }
 
+  describe('[Create draft article]', () => {
```
evakill, line 18: these are great, thanks for taking time to improve tests 🙌

src/pages/Research/Content/Common/Research.form.tsx
```diff
@@ -267,6 +279,25 @@ const ResearchForm = observer((props: IProps) => {
                                 placeholder="A comma separated list of usernames."
                               />
                             </Flex>
+                            <Flex
```
evakill, line 282: i think that we can remove this hidden element. my understanding is that the final form `formValues` state does not _need_ to correspond to fields. `isDraft` will be undefined until we set it with the mutators on the button click. if we want to set `isDraft` more explicitly, we could add it to `src/pages/Research/Content/CreateResearch/Template.tsx` as an initial prop to the form. I'll leave that decision up to you! 

tested without the hidden checkbox to confirm: behavior works as expected, also did a little console.log test of `formValues`  - first log is on load, second on click of "Save as draft" and third is on click of "Publish"
Screen Shot 2023-07-06 at 11 41 04 AM

src/pages/Research/Content/Common/Research.form.tsx
```diff
@@ -99,6 +99,12 @@ const ResearchForm = observer((props: IProps) => {
     return store.validateTitleForSlug(value, 'research', originalId)
   }
 
+  const validateDescription = (value, allValues) => {
+    const isDraft =
```
evakill, line 103: `allValues.isDraft` will be `boolean | undefined` so this can be simplified to 
```
const validateDescription = (value, allValues) => allValues.isDraft ? undefined : required(value)
```

evakill (changes requested):
this looks great @benfurber! also good call @iSCJT to pull out the moderation into a separate issue. left some suggestions on the new approach.

also sorry i missed your invitation to chat, I'll be online most of today if you want to shoot me a message on discord :)

packages/cypress/src/integration/research/write.spec.ts
benfurber: Update! ...It didn't...

src/pages/Research/Content/Common/Research.form.tsx
benfurber: Huh, thanks @evakill. I thought I tested (with exactly the same steps) removing the hidden field and it didn't work, but you're definitely right and it's fine.

src/pages/Research/Content/Common/Research.form.tsx
benfurber: Ugh, thanks for noticing - this is from when I was playing with the visable field and missed this in my tidy up.

benfurber:
@evakill I think that's everything now. I'm also online for the next few hours, so I'd also certainly be up for working on something together.

evakill (approved):
LGTM!

Issue #2571, opened 2023-07-05
[change] remove moderation options from draft research items


iSCJT:
**Is your feature request related to a problem? Please describe.**
Moderators are given the option to approve / reject a draft research item.  This could lead to them approving an item that is not finished.

**Describe the solution you'd like**
Remove the moderation section from the research article view.

**Describe alternatives you've considered**
N/A

**Additional context**

![](https://user-images.githubusercontent.com/37253846/250968784-77f5ee0c-22a0-42dd-b889-f4fa0ccbaf2b.png)

maxbethke:
From my (very limited) understanding of the system, this intermediate state of a research item which you called 'finished draft' would be modeled by the 'awaiting-moderation' moderation status (See IModerationStatus).
In my mind, after an item gets published by the creator, it should still remain hidden from the public but now get shown to moderators in the form of the screenshot above. Once they approve it, it would get assigned the status 'accepted' or 'rejected' respectively.

Let me map out the visibility of an item here:

|  Actor  |  Moderation Status  |  Visible  | 
|  ---  |  ----------------  |  -----------  |
|  Creator  |  Any  |  Yes | 
|  Moderators  |  draft, rejected  |  No  |
|  Moderators  |  awaiting-moderation, accepted  |  Yes  |
|  Everyone  | draft, awaiting-moderation, rejected  |  No  |
|  Everyone  |  accepted  |  Yes  |

Do you @iSCJT, confirm with this view of how its suppost to work?

iSCJT:
Thanks @bagiyal for thinking through a solution and @maxbethke for providing your input on how the moderation is working.

The first thing to point out is that for research we currently automatically set the moderation status to 'accepted' as it's only Project Kamp who can add new research, this will change in future though. The normal moderation flow which we use for how-to's and will be the same for research once used more widely would be: draft > awaiting-moderation > accepted / rejected.

In terms of closing this issue @bagiyal my suggestion would be to make the moderation buttons section conditional on the moderation field being equal to 'awaiting-moderation'.

@maxbethke almost:

|  Actor  |  Moderation Status  |  Visible  | 
|  ---  |  ----------------  |  -----------  |
|  Creator  |  Any  |  Yes | 
|  Moderators  |  Any  |  Yes  |
|  Everyone  | draft, awaiting-moderation, rejected  |  No  |
|  Everyone  |  accepted  |  Yes  |

Moderators do/will want to be able to view any research type if they need to:

[...]

PR #2573, opened 2023-07-05
2550 sort research by moderation status

### LINKED ISSUE #2550: [feature request] pin draft / awaiting moderation / rejected research created by the logged in human to top of listing
If a human using the platform adds a draft research item then they can lose visibility of it.

If the logged in human has any draft, awaiting-moderation or rejected status research created by them then they should be pinned to the top of the research list so they see can always find them.

N/A - Agreed at June meetup

maxbethke:
PR Checklist

PR Type

## Description

Research that is a draft, awaiting moderation or rejected should be displayed on the top of the list of researched regardless of the chosen sorting option, because it can otherwise easily get lost further down the list.

## Git Issues

Closes #2550 

## What happens next?


iSCJT:
@maxbethke thanks for this.  The requirement is a bit more complicated than implemented.  I should have made the issue more explicit, sorry.  The sort logic required is that the logged in user should only see their own items awaiting moderation at the top of the list and then all of the others next, so...

show first: _createdBy === logged in user AND moderation !== 'accepted'.  Then for anything not meeting that requirement use whichever sort option has been chosen.

Hope that makes sense and sorry the issue ticket wasn't more descriptive.

maxbethke:
Thank you for making that clear @iSCJT. My understanding was (Like you can see here https://github.com/ONEARMY/community-platform/issues/2571#issuecomment-1622954402), that the visibility is already set up according to authentication, so only the creator (and moderators) see non-accepted items in the first place.

If thats not the case, I agree with the need to filter by auth like you described above.

I do want to make the case though that this would be non-intuitive to me as a user, as I would expect my research in draft state to not be publicly shown (hence the 'publish') button to proceed in the workflow. To me, this state would rather be called 'Work in Progress', as it is already public but indicated to not be finished. 
Secondly, why have a moderation (in later stages of the platform) if the drafts are already published?

I will rework the feature to take ownership into account but would advise to rethink that system.

iSCJT:
@maxbethke you're right that the visibility is set up according to the authentication and it should stay that way, we don't want to change anything to do with the visibility just the order that things are displayed.

The original request came from the moderation team in relation to the how-to's, they felt as a normal user of the system that there is nowhere that surfaces the non-accepted items easily.  The research and how-to lists are the most visible/frequented related pages so for the user that created the item (and only that user) they wanted the non-accepted items to appear first in the list.  As a logged in moderator they don't want anything to change and the non-accepted items should appear in the list as they do currently.

Another option would be to have a separate panel at the top which showed any of your own items that weren't accepted but we wanted to keep it as simple as possible / least work to implement.

evakill:
hey @maxbethke - are you interested in continuing work on this one? otherwise will open it back up to contributors

maxbethke:
> hey @maxbethke - are you interested in continuing work on this one? otherwise will open it back up to contributors

Hi, even though I'd really like to, I am not able to find the time in the last weeks and it does not seem to change soon.
Maybe someone else can take it from here. From the conversations in on #2550 any contributor should be able to pick up the task. If required, I can write a handover to make clear what still has to be done. If there should be any questions, just mention me and I will see what I can do.

AlfonsoGhislieri:
Closing this as I've taken over the issue and raised a PR: #2739

Issue #2329, opened 2023-05-06
[feature request] allow use of multiple role options in AuthWrapper


iSCJT:
**Is your feature request related to a problem? Please describe.**
[AuthWrapper](https://github.com/ONEARMY/community-platform/blob/master/src/common/AuthWrapper.tsx) currently allows a single role to be provided for the roleRequired property.

**Describe the solution you'd like**
Ability to provide multiple role options as an array, for example ['beta-tester','admin'].

**Describe alternatives you've considered**
Currently if multiple roles required access to a certain feature you would need to duplicate the element, not a good alternative.

**Additional context**
N/A

thisislawatts:
Hey @VincentNikkelen, thanks for taking the time to write up your thoughts. I [totally agree](https://github.com/ONEARMY/community-platform/pull/2595#issuecomment-1627240663) and whilst permissions are not something we have on near term roadmap to revisit this work does need doing. 

I think we probably need a separate Issue/Feature Request as a starting point for that discussion.

PR #2595, opened 2023-07-07
feat: make research available to all users

thisislawatts:
PR Checklist

PR Type

## Description

Currently the Research module is only available for `beta-testers` this is a misleading state for the module however as it is now being regularly used in production by folks on site at Project Kamp.

Switches the system to instead rely on the `research_creator, research_editor` roles. 

- chore: remove beta-tester role from accessing Research
- test: stub out testing around AuthWrapper
- chore: consolidate test:unit into single cmd


davehakkens:
Will this PR make the research module available for all users?

thisislawatts:
> Will this PR make the research module available for all users?

Nope @davehakkens, this will only be available for `research_editor` and `research_creator` roles.

davehakkens:
> Nope @davehakkens, this will only be available for `research_editor` and `research_creator` roles. 

Ah ok. Whats the thinking behind adding those new roles instead of keeping it for beta-tester?

thisislawatts:
`beta-tester` should be reserved for evaluating features until they reach maturity and are ready for general use. The fact that we need to gate Research to a subset of PK accounts is not a result of the feature being `beta` it is instead a specific requirement for the Research Module. 

Weirdly the `research_creator` and `research_editor` roles seem to already exist in the system so I was adopting them for what I imagine is their intended purposed.

davehakkens:
Thanks for explaining. I see the point.

But not sure that is really needed at this point though. The current research module on PK had the unique combination that both the project and module were not mature. So they could develop together in beta. I foresee other instances in the short term would choose to either enable or disable the entire research module. 


Mainly worried about adding more complexity to our roles management (thinking of the role overview chart)

thisislawatts:
@davehakkens would you be able to include a link or screenshot of the "role overview chart" you've mentioned here?

davehakkens:
> Enabling who can write Research articles is currently being limited by the beta-tester role, this is misuse of the role and hard to reason with. 

Agree the only reasons I would have are:
- Currently the only people writing research are in Project Kamp. So we talk with them, explain its experimental/beta and have a close feedback loop.
- It is a temporary state. +/-Next year PK will open it publically and everyone can use it and those specific user roles will likely become obsolete.


Below the roles overview:
afbeelding

davehakkens:
Forgot to mention: Not here to block the PR. Mainly want to highlight the current temporary setup and double check if its worth integrating those extra roles for it in the long term. If you think it makes sense for the codebase feel free to go for it, we will adapt when assigning roles.

thisislawatts:
@davehakkens I think that these research roles are only temporary and ideally over the course of the year we migrate to a more flexible system for managing permissions.

src/common/AuthWrapper.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+        }
+        roleRequired="super-admin"
+        fallback={<div>Fallback Content</div>}
+      >
+        <div>Test Content</div>
+      </AuthWrapper>,
+    )
+    expect(getByText('Fallback Content')).toBeTruthy()
+  })
+})
```
AlfonsoGhislieri, line 79: **Praise**: great work adding these tests for the `AuthWrapper`

AlfonsoGhislieri (approved):
This looks good to me @thisislawatts, great work 💪🏼.

The one thing I did notice when having a read through the code is that I see that we are using this `additionalAdmins` prop for the Authentication but it is never actually passed along anywhere. Is this maybe outdated? Should we consider removing it? Let me know your thoughts!

AlfonsoGhislieri:
Will leave it at @davehakkens to merge this in if he's also happy with it!

iSCJT:
@thisislawatts just scanning through this on my phone, does it also close off https://github.com/ONEARMY/community-platform/issues/2329 ?

thisislawatts:
Well caught @iSCJT, I will close out #2329

PR #2600, opened 2023-07-10
fix: update current imgIndex when clicking through images

iSCJT:
PR Checklist

PR Type

## Description

It was noticed that when viewing the photos from a gallery and then clicking the main image to open it full screen it would always open the first image rather than the selected one.

This pr updates the imgIndex when clicking through the thumbnails in the ImageGallery component so that the correct image is opened fullscreen.

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/packages/components/src/ImageGallery/ImageGallery.tsx
+++ b/packages/components/src/ImageGallery/ImageGallery.tsx
@@ -56,10 +56,11 @@ export const ImageGallery = (props: IProps) => {
     })
   }, [])
 
-  const setActive = (image: IUploadedFileMeta) => {
+  const setActive = (image: IUploadedFileMeta, imgIndex: number) => {
     setState({
       ...state,
       activeImage: image,
+      imgIndex: imgIndex,
     })
   }
 
@@ -101,7 +102,7 @@ export const ImageGallery = (props: IProps) => {
                 mb={3}
                 mt={4}
                 opacity={image === state.activeImage ? 1.0 : 0.5}
-                onClick={() => setActive(image)}
+                onClick={() => setActive(image, index)}
                 key={index}
               >
                 <Image
--- a/packages/components/src/ImageGallery/ImageGallery.tsx
+++ b/packages/components/src/ImageGallery/ImageGallery.tsx
@@ -22,10 +22,9 @@ export interface IProps {
 }
 
 interface IState {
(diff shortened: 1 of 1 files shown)


packages/components/src/ImageGallery/ImageGallery.tsx
```diff
@@ -101,7 +102,7 @@ export const ImageGallery = (props: IProps) => {
                 mb={3}
                 mt={4}
                 opacity={image === state.activeImage ? 1.0 : 0.5}
-                onClick={() => setActive(image)}
+                onClick={() => setActive(image, index)}
```
thisislawatts, line 105: **issue:** Can we introduce a test to `ImageGallery.test.tsx` which captures the desired behaviour and to guard against regression?

thisislawatts (dismissed):

packages/components/src/ImageGallery/ImageGallery.tsx
```diff
@@ -56,10 +56,11 @@ export const ImageGallery = (props: IProps) => {
     })
   }, [])
 
-  const setActive = (image: IUploadedFileMeta) => {
+  const setActive = (image: IUploadedFileMeta, imgIndex: number) => {
     setState({
       ...state,
       activeImage: image,
+      imgIndex: imgIndex,
```
evakill, line 63: with this change i think it makes sense for `activeImage` to become a derived state from `images` and `imgIndex` (maybe rename to `activeImageIndex`) rather than tracking both in the component state.

evakill (changes requested):

packages/components/src/ImageGallery/ImageGallery.tsx
iSCJT: @evakill Yep agree that makes sense, should have looked at the code properly rather than just rushing a fix for Ale so I could get on with general help stuff.  I've refactored.

packages/components/src/ImageGallery/ImageGallery.tsx
iSCJT: @thisislawatts I'll cycle back to the tests as a separate issue as I've not mocked image data before so will need to look into how it's done and would prefer to get this pr cleared.  I'll add another issue for it in a minute.

evakill (approved):
lgtm!

PR #2604, opened 2023-07-11
Speed up CircleCI pipelines

### LINKED ISSUE #2249: [devops] 🐌 Slow CircleCI build times
**Describe the bug**

The dependency installation process runs for 1-2mins at multiple stages within the deployment pipeline. This unnecessarily slows down the feedback cycle and wastes a bunch of compute time. This is bad for the planet and developers. 

**To Reproduce**

1. Take a look at any of [the build pipelines](https://app.circleci.com/pipelines/github/ONEARMY/community-platform).
2. Check one of the jobs, ie. Unit tests
3. See that _install dependencies_ has run for a few minutes or more.

**Expected behaviour**
Dependency installation should run only run twice (once for all dev dependencies and again for production dependencies)

inad9300:
PR Checklist

PR Type

## Description

Disable third-party post-install scripts present in the `package.json`s of dependencies.

## Git Issues

Closes #2249

## Screenshots/Videos

Before:
image

After:
image

There is _a lot_ of variation between runs, but overall my feeling is that it got faster (the "Lint code" step is probably a good way to see this, as it is run without any other parallel step).


thisislawatts:
Some odd permissions here maybe from CircleCI, unable to trigger a run, I've created a new branch on the repo itself [chore/speed-up-circleci-test-runs](https://github.com/ONEARMY/community-platform/tree/chore/speed-up-circleci-test-runs)

inad9300:
Unfortunately, the project seems to have a direct dependency, [sharp](https://www.npmjs.com/package/sharp), which depends on post-install scripts being run (see [sharp's common installation problems](https://sharp.pixelplumbing.com/install#common-problems)). Given that, I'm not sure I see any satisfactory way to speed installation up by way of avoiding post-install scripts: the only ideas I have are either to replace this dependency with some other that doesn't have the same requirement, or to somehow run post-install scripts only for this dependency.

inad9300:
[Researching about this topic some more](https://yarnpkg.com/advanced/lifecycle-scripts#a-note-about-postinstall), yet another alternative would be to switch to Yarn's "Zero-Install" (which in short appears to mean committing Yarn's cache to Git), or move to either a WebAssembly (possibly [sharp](https://github.com/lovell/sharp/issues/3336) itself in the future) or –accepting the [performance penalty](https://sharp.pixelplumbing.com/performance)– a pure JavaScript alternative (like [Jimp](https://www.npmjs.com/package/jimp)).

However! Looking further into the problem, I have found that all sharp-related functions (and, indeed, some others) were actually unused. So, I removed them, but I'm not sure if they were intended to be used at some point or something. If you find it okay to remove them, this is probably the easiest solution.

inad9300:
A well-timed [tweet](https://twitter.com/alvarlagerlof/status/1679285897710247936) gave me the clue on how to be selective about which post-install scripts are run with Yarn, plus an extra potentially beneficial flag, which I gave a try to in an [alternative branch](https://github.com/inad9300/community-platform/commits/speed-up-circleci-test-runs-alt). In fact, I encountered a second dependency used by the project which requires post-install scripts, Cypress, so running some of these scripts is necessary even if sharp is removed (BTW [caching Cypress](https://glebbahmutov.com/blog/cypress-circleci-subfolders/) is another thing I would like to try).

inad9300:
Thanks for the feedback! That is great to hear. I've been making some more experiments in a separate branch and I'm pretty sure I can bring down pipeline times even more, mostly by re-enabling Yarn's cache in a couple of places where it was disabled (commented out). I should be able to show the progress sometime this week.

thisislawatts:
@inad9300 I would be inclined to merge this as is and for any further changes to be raised as a separate PR. There are some good hygiene changes in here which I think need to be merged asap.

I have a bias towards multiple small PRs focused on a single changeset :)

thisislawatts (approved):

PR #2501, opened 2023-06-29
feat: drop Events module from the platform

thisislawatts:
PR Checklist

- [X] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [X] Developer experience (improves developer workflows for contributing to the project)

## Description

There are no instances currently using Events, nor are there plans to enable this. We will remove it for the time being.

PR #2614, opened 2023-07-13
feat: remove moderation options from research items

bagiyal:
PR Checklist

PR Type

## Description
if someone now published research but now it became awaiting-moderation
_What this PR does_

## Git Issues

Closes #

#2571

## Screenshots/Videos

---

## What happens next?


scripts/compose-dev.yaml
```diff
@@ -0,0 +1,12 @@
...
+    entrypoint:
+    - sleep
+    - infinity
+    image: docker/dev-environments-default:stable-1
+    init: true
+    volumes:
+    - type: bind
+      source: /var/run/docker.sock
+      target: /var/run/docker.sock
+
```
AlfonsoGhislieri, line 12: **Issue (blocking)**: why are we adding a new script here? What is the reason behind this?

thisislawatts, line 12: **question:** This is interesting, I do not see how this is related to the change you have proposed here but I would be interested to understand how you were using this?

src/pages/Events/Content/EventsCreate/EventsCreate.tsx
```diff
@@ -0,0 +1,329 @@
+import * as React from 'react'
+import type { RouteComponentProps } from 'react-router'
```
AlfonsoGhislieri, line 2: **Issue**: I thought the issue was to do with research items why are we modifying or adding to events files?

thisislawatts: Could be a merge problem, Events were all removed in: https://github.com/ONEARMY/community-platform/pull/2501

AlfonsoGhislieri: Ah yes that makes sense, could you please remove this file @bagiyal

AlfonsoGhislieri:
Some also some linting issues here @bagiyal that need to be address.

- Code is not passing lint test, you need to make sure you are using prettier: https://prettier.io/
- You need to also make sure the commits are correctly formatted, you can follow conventional commits as a guide: https://www.conventionalcommits.org/en/v1.0.0/

src/pages/Research/Content/Common/Research.form.tsx
```diff
@@ -89,7 +89,7 @@ const ResearchForm = observer((props: IProps) => {
   }, [submissionHandler])
 
   const onSubmit = async (formValues: IResearch.FormInput) => {
-    formValues.moderation = submissionHandler.draft ? 'draft' : 'accepted' // No moderation for researches for now
+    formValues.moderation = submissionHandler.draft ? 'draft' : 'awaiting-moderation' // No moderation for researches for now --> Now Change to awaiting-moderation
     await store.uploadResearch(formValues)
```
AlfonsoGhislieri, line 93: **Issue (blocking)**:  I don't think we should be modfying the moderation values, these should stay in their current state. What needs to be added is conditional rendering of the moderation buttons only when the status of the moderation status being set to 'awaiting-moderation' --> the logic for this has not yet been added.

Just to double check @iSCJT am I correct in these statements?

AlfonsoGhislieri: I am on holiday so can't a zoom call I'm afraid, but here is how @iSCJT nicely summarized what you should do in the issue:

_In terms of closing this issue @bagiyal my suggestion would be to make the moderation buttons section conditional on the moderation field being equal to 'awaiting-moderation'._

So we are not looking to modify the form itself but rather just simply make the moderation buttons section render only when the field is 'awaiting-moderation'.

PR #2626, opened 2023-07-19
fix: increase memory limit

evakill:
PR Checklist

PR Type

## Description

@davehakkens noticed emails were not being sent daily. looked into the logs and saw:
Screen Shot 2023-07-18 at 9 14 29 PM
which, sadly, isn't caught by our error handling (💔) because it occurs within firebase i presume. might be possible to set up an alert to the webhook from functions logs?

i can't find any super obvious memory drain here.. let me know if you see anything I'm missing. if not, seems reasonable to increase the memory usage. testing on dev, 512 seems to do the trick. 
this won't cost us anything extra - as of now we stay well within the 2mil free function invocations per month (https://cloud.google.com/functions/pricing)

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/functions/src/emailNotifications/index.ts
+++ b/functions/src/emailNotifications/index.ts
@@ -5,8 +5,11 @@ import { DB_ENDPOINTS, IUserDB } from '../models'
 import { EmailNotificationFrequency } from 'oa-shared'
 import { withErrorAlerting } from '../alerting/errorAlerting'
 
-exports.sendDaily = functions.pubsub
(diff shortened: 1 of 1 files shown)


AlfonsoGhislieri:
I've been pretty removed from all the email functionality, so probably better that @thisislawatts gives his thoughts. But I agree that there should be at least an alert to prevent this being a recurring issue which we don't have visibile of.

PR #2642, opened 2023-07-21
fix: remove image compression from uploads

### LINKED ISSUE #2548: [feature request] add image resize for uploaded photos: thumbnail, cover, original
It is expected that this issue will be completed by a maintainer due to the various backend permissions required to implement the required functionality but if you have a strong interest in completing this ticket then please reach out to use in the comments.

All uploaded images are currently cropped to a set resolution and uploaded to the database.  This can result in pixelated or blurry images when the original was a large file or contains lots of text etc.  Reported by Project Kamp in relation to research module.

We'd like to introduce / save 3 image sizes:

- Thumbnail - for use when displayed images are very small.
- Cover / cropped - for use in areas such as the how to listing.
- Original - To be used when clicking into images added to a how-to or research article.

Maintainer taking it on to determine exact sizing and process requirements.

thisislawatts:
PR Checklist

PR Type

## Description

Removes in browser compression in favour of browser side compression.

## Git Issues

Closes #2548

Outstanding:

* [ ] Do not display original images in browser, only resized assets
   * This should be backwards compatible, ie. For images that do not have generated resize assets.
* [ ] Install Resize Images plugin on all target environments

Questions:

The Firebase extension doesn't interact with the models so there is no way to **know** that the images exist. Instead the FE components would assume that an image exists at the target path. This seems sub optimal.


thisislawatts:
Update on this one, we had to remove the Firebase extension where we had it installed for PP as it caused non obvious error states for folks modifying map pins. 😢 

Needs further investigation to recreate the error state outside of production environment and verify whether the issue is tied to our application logic, user data or the Firebase extension.

PR #2609, opened 2023-07-13
feat: allow draft how-to save without form validation

benfurber:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

This allows users to save a draft how-to with (at a minimum) a title provided. The smallest UI change has been done to enable this. But a lot of store changes were required for this to happen smoothly.

## Git Issues

Closes #2540 

## Screenshots/Videos

Screenshot 2023-07-13 at 09 11 42

PR #2654, opened 2023-07-22
Fix: implement validation and error messages for form field lengths

### LINKED ISSUE #2655: [bug] Field character limits are not being enforced or giving warnings
**Describe the bug**

A lot of the fields thata are using the character counter or have limits on the total number of characters do not actually have these enforced. The only validation is for the most part just checking that the field is not left blank rather than checking against mininum and maximum character lengths.

This is an issue in the following files:
- Howto.form
- HowtoStep.form
- Research.form
- ResearchUpdate.form

**To Reproduce**

1. Go to 'Create How-To'
2. Scroll down to 'Short description of your How-to'
3. Add more than 400 characters to the text field
4. See that there is nothing enforcing the limit on the amount of characters, and no warning when the limit is reached

**Expected behaviour**

There should be some feedback given to the user and these limits should be enforced.

**Potential solutions**
The first solution could be to stop further characters from being able to be typed by the user once the max limit is reached, this would remove the need for any validation for upper limits. But this doesn't solve the issue for minimum limits.

The best approach I can think of is to create some reusable validation functions that can be reusued across the codebase, here is an example of one of the validation functions already in the [codebase](https://github.com/ONEARMY/community-platform/blob/a88a3c3d6368bc28b20a8772ac52d26dab838eb6/src/pages/Howto/Content/Common/HowtoStep.form.tsx#L83)

AlfonsoGhislieri:
PR Checklist

PR Type

## Description

Adds the character counter for all the fields that should have it and implements custom validators to enforce them and give correct errors to the user.

## Git Issues

Closes #2655

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/packages/components/src/FieldTextarea/FieldTextarea.tsx
+++ b/packages/components/src/FieldTextarea/FieldTextarea.tsx
@@ -41,6 +41,8 @@ export const FieldTextarea = ({
     <>
       <Textarea
         disabled={disabled}
+        minLength={minLength}
+        maxLength={maxLength}
         variant={meta?.error && meta?.touched ? 'textareaError' : 'textarea'}
         {...input}
         {...rest}
--- a/src/pages/Howto/Content/Common/Howto.form.test.tsx
+++ b/src/pages/Howto/Content/Common/Howto.form.test.tsx
@@ -138,36 +138,6 @@ describe('Howto form', () => {
       ).not.toBeInTheDocument()
     })
   })
-
-  describe('HowtoStep', () => {
-    describe('Description field', () => {
-      it('validation fails when description is over the limit', async () => {
-        // Arrange
-        const maxLength = HOWTO_STEP_DESCRIPTION_MAX_LENGTH
-        const longText = faker.random.alpha(maxLength + 1)
-        const formValues = FactoryHowto({
-          steps: [FactoryHowtoStep({ text: '' })],
-        })
-
-        // Act
-        const wrapper = getWrapper(formValues, 'edit', {})
-        const descriptionTextAreaElement =
-          wrapper.getByTestId('step-description')
-        descriptionTextAreaElement.focus()
-        fireEvent.change(descriptionTextAreaElement, {
-          target: { value: `${longText}` },
-        })
-        descriptionTextAreaElement.blur()
-
-        // Assert
-        expect(
-          wrapper.getByText(
-            `Descriptions must be less than ${HOWTO_STEP_DESCRIPTION_MAX_LENGTH} characters`,
-          ),
-        ).toBeInTheDocument()
-      })
-    })
-  })
 })
 
 const getWrapper = (formValues, parentType, navProps) => {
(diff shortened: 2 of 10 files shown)


AlfonsoGhislieri:
I think the next step here would be to extract the validation of titles/descriptions/textFields into some helper functions that can be used across the different forms and just require the max and min lengths to be passed across. This should make the whole implementation a lot cleaner and easy to maintain!

src/pages/Research/Content/Common/Research.form.tsx
```diff
@@ -93,16 +103,6 @@ const ResearchForm = observer((props: IProps) => {
...
 
-  const validateTitle = async (value: any) => {
-    const originalId =
-      props.parentType === 'edit' ? props.formValues._id : undefined
-    return store.validateTitleForSlug(value, 'research', originalId)
-  }
-
-  const validateDescription = (value, allValues) => {
-    return allValues.isDraft ? undefined : required(value)
-  }
```
AlfonsoGhislieri, line 104: I removed this because it only seems to be in use for this one specific form description, if this is a functionality we want to have it would be good to implement for other forms as well (directly into the validator functions)

src/pages/Howto/Content/Common/Howto.form.test.tsx
```diff
@@ -138,36 +138,6 @@ describe('Howto form', () => {
       ).not.toBeInTheDocument()
     })
   })
-
-  describe('HowtoStep', () => {
-    describe('Description field', () => {
-      it('validation fails when description is over the limit', async () => {
-        // Arrange
-        const maxLength = HOWTO_STEP_DESCRIPTION_MAX_LENGTH
```
AlfonsoGhislieri, line 146: Removed this test because it is no longer possible to type more characters than the character limit

packages/components/src/FieldTextarea/FieldTextarea.tsx
```diff
@@ -41,6 +41,8 @@ export const FieldTextarea = ({
     <>
       <Textarea
         disabled={disabled}
+        minLength={minLength}
+        maxLength={maxLength}
         variant={meta?.error && meta?.touched ? 'textareaError' : 'textarea'}
```
AlfonsoGhislieri, line 46: This adds the limits directly just like is done for the title fields (removes need for any validation for maxLength of the field)

AlfonsoGhislieri:
I do have some questions here @davehakkens:

1) Currently research can be saved as a draft even if it has no description, this kind of execption is not allowed for howtos. Do we want this to stay? Should it maybe also similarly be added to Howtos?

2) Should there be a minimum length for Research descriptions? Currently there is just an upper limit.

davehakkens:
>     1. Currently research can be saved as a draft even if it has no description, this kind of execption is not allowed for howtos. Do we want this to stay? Should it maybe also similarly be added to Howtos?

Yep same should be for how-to.

>     2. Should there be a minimum length for Research descriptions? Currently there is just an upper limit.
This isn't really a problem we have at the moment (since we upload ourselves) but we could align it with how-to's for possible future issues and to keep it clear and aligned?

AlfonsoGhislieri:
> 
> Yep same should be for how-to.
> 


Okay, I'm going to leave it for now since I know @benfurber  is working on this I think in #2609

> This isn't really a problem we have at the moment (since we upload ourselves) but we could align it with how-to's for possible future issues and to keep it clear and aligned?

Up to you! I can quickly add it if you want, we can always add it later when it becomes public if you prefer. I'd say it would be good to maybe just have a small limit just to already have the logic and testing in place though (if you have a nmber let me know)

thisislawatts (approved):

PR #2656, opened 2023-07-23
chore: fixes tests so there are no more errors

### LINKED ISSUE #2646: Fix "act" errors when running unit tests
When running Jest component tests, there are some errors that make it hard to read the test output. Additionally, the errors are a clue that we're doing something wrong.

Repro steps:
- run `yarn test:unit`
- type "." to select tests by pattern. I think this selects all tests
- The tests will run, and you should see some errors that look like:

```sh
console.error
      Warning: An update to ReactFinalForm inside a test was not wrapped in act(...).

      When testing, code that causes React state updates should be wrapped into act(...):

      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */

      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
```

Additional context:
- The URL provided in the error output redirects to a `legacy` documentation page. https://legacy.reactjs.org/docs/test-utils.html#act , this is a clue that something we're doing is out-dated. A quick search finds the following [stack overflow thread](https://stackoverflow.com/questions/60113292/when-to-use-act-in-jest-unit-tests-with-react-dom) that mentions react testing library may wrap all actions in `act` already.

- We appear to be using test-library/react 11 (`@testing-library/react@npm:^11.1.0`), and the [github page](https://github.com/testing-library/react-testing-library) appears to be on version 14. Could we fix this issue by upgrading our testing library?

- the documentation for running unit tests and component tests is out of date. We should file a separate issue to update the docs, OR, update the documentation as part of this issue. I believe it's simply `yarn test:unit` but that might only test .tsx files?

Suggested Next Steps:
- investigate whether a test library update fixes the issue
  - if so, we might want to update the library, remove all instances of act, and ensure all existing tests pass
  - if not, maybe just wrap all tests in act that need it. And consider filing a separate issue to update the testing library anyways (as it's 3 major versions behind)
- Consider updating the documentation for running unit tests and component tests (I'm calling unit tests anything with `*.test.ts` and component tests anything with `*.test.tsx`) OR create a separate issue to update the documentation
  - Consider updating the following: 
    - https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--getting-started
    - https://docs.platform.onearmy.earth/Testing/overview
    - ?

asdFletcher:
PR Checklist

PR Type

## Description

- Fixes all the errors related to tests.
  - Act errors
    - In tests, it appears that any async code or code that causes side-effects in the wrapper factories (i.e. anything inside the render function) will cause an act error, as all these changes must be settled before we're testing what the user experiences. This includes the dom, state, etc.
    - My initial suspicion that act could be removed was not correct
  - form decorator errors
    - memoizing decorator function definitions prevents each render from re defining those functions, and prevents a test error that was complaining that: `form decorators should not change from one render to the next as new values will be ignored`
  - duplicate key errors
    - in one place in the app we had duplicate keys assigned to 2 child elements

## Git Issues

Closes #2646

## Screenshots/Videos

No errors/warnings!
image

---

Note:
- the act syntax is super confusing:
```ts
      let wrapper
      await act(async () => {
        wrapper = await getWrapper(formValues, 'edit', {})
      })
```

It can be fixed if we upgrade testing-library/react (requires updating react as well)

### DIFF
--- a/src/pages/Howto/Content/Common/Howto.form.test.tsx
+++ b/src/pages/Howto/Content/Common/Howto.form.test.tsx
@@ -1,4 +1,4 @@
-import { render, fireEvent } from '@testing-library/react'
+import { render, fireEvent, act } from '@testing-library/react'
 import { Provider } from 'mobx-react'
 import { HowtoForm } from './Howto.form'
 import { MemoryRouter } from 'react-router'
@@ -27,6 +27,8 @@ jest.mock('src/index', () => {
             Database: false,
             Complete: false,
           },
+          validateTitleForSlug: jest.fn(),
+          uploadHowTo: jest.fn(),
         },
         tagsStore: {
           categoryTags: [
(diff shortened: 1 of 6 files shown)


thisislawatts (approved):
✨ I am so happy to see this change @asdFletcher. I will think fondly about this PR each time I see a green test run. 💚

PR #2669, opened 2023-07-25
chore: upgrade cypress@12.17.2

### LINKED ISSUE #1522: [bug] - e2e tests fail with firefox
**Describe the bug**
As noted in #1521, firefox has been given quite a lot of flaky and failed tests. Ideally should be investigated further in the future (probably after a few more core updates are in to things like cypress and firebase)

benfurber:
PR Checklist

PR Type

## Description

Locally I'm finding no problem with the firefox tests, running on ci to see if there's issues.

## Git Issues

Closes #1522

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -499,7 +499,7 @@ workflows:
           matrix:
             parameters:
               CI_NODE: [1, 2, 3, 4]
-              CI_BROWSER: ['chrome']
+              CI_BROWSER: ['chrome', 'firefox']
           filters:
             branches:
               ignore:
--- a/commitlint.config.ts
+++ b/commitlint.config.ts
@@ -1,10 +1,10 @@
 import type { UserConfig } from '@commitlint/types'
-import path from 'path'
 
 export const Configuration: UserConfig = {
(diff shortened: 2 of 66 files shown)


package.json
```diff
@@ -172,6 +172,7 @@
     "concurrently": "^6.2.0",
     "cross-env": "^7.0.3",
     "cspell": "^6.1.0",
+    "cypress": "12.17.2",
```
thisislawatts, line 175: **issue (blocking):** It seems like there may have been a small oversight here. I've noticed that the `cypress` package has been added to the root application. We should avoid introducing this dependency in the root package.json file. Would you mind looking into this?

thisislawatts (changes requested):
**praise:** Thanks for picking this item up @benfurber. One small fix and then it looks ready to merge.

package.json
benfurber: Yes! Sloppy of me, apologies.

benfurber:
@iSCJT - I forgot to ask earlier. Did you mean to add to this branch?

benfurber:
@evakill Rebased your change in.

evakill (approved):

PR #2693, opened 2023-07-30
refactor: improve updates around user profile

iSCJT:
PR Checklist

PR Type

## Description

This pr:

- Aims to reduce the number of operations that set the full document when updating the user profile.  The reason for this is to remove the number of possible causes of 'duplicate' user docs being created.
- It also introduces two additional properties to the user profile to track when and why a user profile doc was created.  These will be used to help determine the cause of the duplicate records being created and then be removed again at a later date.
- Prevents user docs being returned by the getUserProfile function if the authId and Id are the same

I would suggest reviewing each commit separately as should make more sense that way.

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/src/stores/User/user.store.ts
+++ b/src/stores/User/user.store.ts
@@ -80,7 +80,7 @@ export class UserStore extends ModuleStore {
         photoURL: authReq.user.photoURL,
       })
       // populate db user profile and resume auth listener
-      await this._createUserProfile()
+      await this._createUserProfile('registration')
       // when checking auth state change also send confirmation email
       this._listenToAuthStateChanges(true)
     }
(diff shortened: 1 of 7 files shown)


src/stores/User/user.store.ts
```diff
@@ -161,12 +161,7 @@ export class UserStore extends ModuleStore {
 
     if (lookup.length > 1) {
       logger.warn('Multiple user records fetched', lookup)
-      return lookup
-        .sort(
-          (a, b) =>
-            new Date(a._created).getTime() - new Date(b._created).getTime(),
-        )
-        .filter((user) => user._lastActive)[0]
+      return lookup.filter((user) => user._id !== user._authId)[0]
```
thisislawatts: **question:** Given this change, I am guessing we are at a point where we can remove the `lookup2` block from this method?

thisislawatts (approved):

src/stores/User/user.store.ts
iSCJT: I initially thought the same but not just yet as there's still some legacy records that don't meet the initial lookup. I'm going to write a function to migrate the legacy records and then we can drop it.

PR #2694, opened 2023-07-31
fix: research comments not showing

thisislawatts:
PR Checklist

PR Type

## Description

The problem is that the comments were being looked up via a pointer
reference on an array of Research Updates. This look up would be inaccurate
when there are updates in the array which are not visible
due to their `status=draft`. The result is
that items would be loaded off of the wrong ResearchUpdate.

As the comment list is available on the Research Update document
there is no need to look it up. Instead it can be passed through
and formatted.


iSCJT (approved):
Top 🐛 squashing @thisislawatts, thanks!

Issue #2684, opened 2023-07-27
[feature request] list form validation errors under publish button if how-to / research form validation fails


iSCJT:
Replicate any validation errors to a list under the publish button.

**Is your feature request related to a problem? Please describe.**
It is not always immediately obvious which fields have not been completed when creating/updating how-to and research and pressing the publish button.

**Describe the solution you'd like**
We would like to replicate any individual validation errors that appear beneath each field into a combined list beneath the Publish button if the form validation fails.

To be clear we would like the existing validation messages to stay where they are and for these to be duplicated into the list.

**Describe alternatives you've considered**
Modal - not considered a good user experience as requires clicking out of it and then you lose the list.

benfurber:
Working on this now. 

**Here is as suggested:**
<img width="548" alt="Screenshot 2023-08-11 at 18 16 50" src="https://github.com/ONEARMY/community-platform/assets/16688508/ab636e4a-45d2-4273-980b-a68acc0fb4bf">

The problem with this is that on smaller screens you won't be able see all the errors on first look if there are many, you'll have to scroll down to the bottom of the page.

**Another approach:**
<img width="1009" alt="Screenshot 2023-08-11 at 18 17 46" src="https://github.com/ONEARMY/community-platform/assets/16688508/e6ffa15e-3878-439a-9dd5-f3359430bf0a">

At the top of the form.

Something for you @davehakkens or @thisismattia?

davehakkens:
nice to see an overview of errors @benfurber!
I would say it could go underneath the button. Dont think its often many errors + its closer to the publish button.

This would also be nice to align for user profiles. That currently says this: 
<img width="500" alt="afbeelding" src="https://github.com/ONEARMY/community-platform/assets/13672737/85bf2e2c-08e4-4752-9012-314b5b40ff61">

benfurber:
@davehakkens How's this? I'd say the second sentence of the user profile error button isn't needed in this case as the reset of the text explains what the user needs to change. Happy to change it though. As it's definitely going in the sidebar too I've tidied up the spacings.

<img width="473" alt="Screenshot 2023-08-15 at 14 09 36" src="https://github.com/ONEARMY/community-platform/assets/16688508/138265f0-611f-4364-b750-04343ac1a473">


[...]

PR #2717, opened 2023-08-19
feat: add generic error rendering component

benfurber:
PR Checklist

PR Type

## Description

This is a generic errors component doing all the presentational lifting for the error components that are about to be created for the how-to, research and user settings forms.

## Git Issues

Part of #2684

### DIFF
--- a/src/common/Form/ErrorsContainer.tsx
+++ b/src/common/Form/ErrorsContainer.tsx
@@ -1,0 +1,78 @@
+import { Box, Card, Text } from 'theme-ui'
+
+import { headings } from './labels'
+
+interface IErrors {
+  [key: string]: string
+}
+
+interface ILabels {
+  [key: string]: { title: string }
+}
+
+interface IErrorsListSet {
+  errors: IErrors[]
+  keys: string[]
+  labels: ILabels[]
+  title?: string
+}
+
+interface IProps {
+  errorsListSet: IErrorsListSet[]
+  isVisible: boolean
+}
+
+export const ErrorsContainer = ({ errorsListSet, isVisible }: IProps) => {
+  if (!isVisible || errorsListSet.length === 0) return null
+
+  return (
+    <Box paddingTop={2} data-cy="errors-container">
+      <Card
+        sx={{
+          padding: 3,
+          flexDirection: 'column',
+          fontSize: 1,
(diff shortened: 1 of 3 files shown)


src/common/Form/ErrorsContainer.tsx
```diff
@@ -0,0 +1,91 @@
...
+    </Box>
+  )
+}
+
+const ErrorsListSet = ({ errorsListSet }) => {
+  return errorsListSet.map((errorsList, index) => {
+    if (errorsList === undefined) return
+    return <ErrorsList key={index} errorsList={errorsList} />
+  })
+}
```
evakill, line 56: this file feels like a bit too much componentization to me. i think this could all be written in 1 or maybe 2 components. i tend to agree with kent c dodds on most react things, and here's his take on this: https://kentcdodds.com/blog/when-to-break-up-a-component-into-multiple-components

that being said, this is a stylistic preference - moreso just offering my opinion here than requesting a change. your code works and is still readable, so do what feels best for you :)

evakill (commented):
going to review the next 3 PRs to understand how this will be used, but looks good so far

src/common/Form/ErrorsContainer.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+interface IErrorsListSet {
+  errors: IErrors[]
+  keys: string[]
+  labels: ILabels[]
+  title?: string
+}
+
+interface IProps {
+  errorsListSet: (IErrorsListSet | undefined)[]
```
evakill, line 21: why the undefined here? and since this is an array of `IErrorsListSet` can we call this `errorsListSets`

evakill: would recommend not allowing the undefined prop in the generic component, unless there is a special handling for it - but currently we're just returning null

src/common/Form/ErrorsContainer.tsx
benfurber: @evakill I've reduced a couple of them. How's about now?

benfurber:
@evakill Changes made.

src/common/Form/ErrorsContainer.tsx
evakill: lgtm!

PR #2707, opened 2023-08-04
refactor: add abstractions for howto forms

benfurber:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Developer experience (improves developer workflows for contributing to the project)

## Description

With the upcoming work in #2683 there will be...
- Even more text flying around the how-to form so I thought I'd bring forward a suggested change for feedback/consideration. I propose abstracting out all the labels/text needed for the forms.
- More repetitive layout changes needed for the guidance so adding the wrapper should help keep it DRY.

PR #2720, opened 2023-08-19
feat: add errors to how-to

benfurber:
PR Checklist

PR Type

## Description

Building off #2717 and #2707(!), this adds an error component to the how-to form.

## Git Issues

Part of #2684.

### DIFF
--- a/src/pages/Howto/Content/Common/HowtoErrors.test.tsx
+++ b/src/pages/Howto/Content/Common/HowtoErrors.test.tsx
@@ -1,0 +1,50 @@
+import { render, screen } from '@testing-library/react'
+
+import { HowtoProvider } from 'src/test/components'
+import { HowtoErrors } from '.'
+
+describe('HowtoErrors', () => {
+  it('renders component when visible and has errors', async () => {
+    const errors = {
+      title: 'Make sure this field is filled correctly',
+    }
+
+    render(
+      <HowtoProvider>
+        <HowtoErrors isVisible={true} errors={errors} />
+      </HowtoProvider>,
+    )
+
+    await screen.findByText(errors.title, { exact: false })
+  })
+
+  it('renders errors for steps', async () => {
+    const text = 'Make sure this field is filled correctly'
+    const errors = {
+      steps: [{}, { text }],
+    }
+
+    render(
+      <HowtoProvider>
+        <HowtoErrors isVisible={true} errors={errors} />
+      </HowtoProvider>,
+    )
+
+    await screen.findByText(text, { exact: false })
+    await screen.findByText('Step 2')
+  })
+
+  it('renders nothing when not visible', async () => {
+    const errors = {
+      title: 'Make sure this field is filled correctly',
+    }
+
+    const { container } = render(
+      <HowtoProvider>
+        <HowtoErrors isVisible={false} errors={errors} />
+      </HowtoProvider>,
+    )
+
+    expect(container.innerHTML).toBe('')
+  })
+})
--- a/src/pages/Howto/Content/Common/HowtoErrors.test.tsx
+++ b/src/pages/Howto/Content/Common/HowtoErrors.test.tsx
@@ -4,9 +4,11 @@ import { HowtoProvider } from 'src/test/components'
 import { HowtoErrors } from '.'
 
 describe('HowtoErrors', () => {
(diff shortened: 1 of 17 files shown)


src/pages/Howto/Content/Common/HowtoErrors.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const errors = stepsWithErrors.map((keys, index) => {
+    const errors = stepErrors[index]
+    const labels = steps
+    const title = `${steps.heading.title} ${index + 1}`
+
+    if (keys.length > 0) return { errors, keys, labels, title }
+  })
+
+  return errors.filter((error) => error !== undefined)
+}
```
benfurber, line 20: This is a lot of code(!) Would love a better approach. To explain. the steps errors info might come in looking like:
```
steps: {
  { },
  { title: 'error message' },
} 
```
And we want the output to ultimately add 'Step 2' to the second child object and remove the first child object.

evakill: we can condense this to one map fn:
```
  const errors = stepErrors.map((errors, index) => {
    const labels = steps
    const title = `${steps.heading.title} ${index + 1}`
    const keys = Object.keys(errors)

    if (keys.length > 0) return { errors, keys, labels, title }
  })
```

src/pages/Howto/Content/Common/HowtoErrors.test.tsx
```diff
@@ -0,0 +1,53 @@
+import { render, screen } from '@testing-library/react'
+
+import { HowtoProvider } from 'src/test/components'
+import { HowtoErrors } from '.'
+
+describe('HowtoErrors', () => {
```
evakill, line 6: since there is some transformational logic going on, i'd recommend unit tests for the `stepErrors` and `introErrors`  functions

evakill (commented):
looking good! i think your approach to the generic error container and per-form transformations is super smart. 

would encourage you to read through and add a few comments - i'd say especially for people who aren't familiar with final-form, some of the error object shapes can be confusing.

benfurber:
Incorporating your comments resulted in a lot of changes(!) - Especially tidying up all the typing. What do you think now?

benfurber:
^ @evakill

src/pages/Howto/Content/Common/HowtoErrors.test.tsx
evakill: looks good!!

evakill (commented):
the code looks good, but i keep running into an error testing on preview that's a bit weird. as i'm filling out the form, sometimes my changes reset when i press publish or "enter" - see here

https://github.com/ONEARMY/community-platform/assets/37253846/69c9ec92-bf1e-42a4-b4d9-bd2a86926e05

could you check this out and see if you can repro on your end?

benfurber:
Weird. I'll start looking into this today. I'll see if it's happening on the other forms too.

benfurber:
@evakill My cypress tests are failing locally (a good sign considering the bug you found). But that lead to see why circle didn't pick them up. Running the feature tests isn't in the CI flow, is that deliberate or has something gone wrong somewhere?

benfurber:
@evakill I replicated the bug and tested the fix.

evakill (commented):
thanks @benfurber, i can't repro anymore. what was the fix?
in the future, especially for big PRs, it's helpful for the reviewer to keep commit history clean by not force pushing - then i can filter by commits added since my last review! 

@davehakkens can you give final sign off on this? preview should be up to date

evakill:
> Running the feature tests isn't in the CI flow, is that deliberate or has something gone wrong somewhere?

yeah, i've noticed on your branches that the CI flow is marked as "Unauthorized" and not run. i'm going to look into it - doesn't seem to be an issue for others. i will keep you updated

evakill:
@benfurber saw a PR come through https://github.com/ONEARMY/community-platform/pull/2750 which affects this work - looks like a rebase is required and it may affect some of the error mapping

benfurber:
@evakill Rebase done. I've keep the string in the label file for the error rendering.

evakill (approved):
woooo! this looks great, thanks for all the hard work @benfurber! talked to dave and he approved 🚢

PR #2723, opened 2023-08-24
fix(stores): fixed issue #2709

### LINKED ISSUE #2709: [bug] Sort Research by latest update (not useful,comment or following)
**Describe the bug**
Currenty the research overview show all the research topics, with the `latest updated` on top.
However at the moment an update means:
- Posting a new update to the research
- Editting an existing update
- Leaving a new comment
- Marking the research as useful
- Following the research

**To Reproduce**

1. Go to [research overview](https://dev.onearmy.world/research)
2. Interact on a research (useful comment etc)
3. See how the research goes to the top

**Expected behaviour**
What we would prefer to include as `latest update` is: 
- Posting a new update to the research

The other actions should not change the `update` status

**Screenshots**
afbeelding

**Additional context**

goratt12:
PR Checklist

PR Type

## Description

Fix issue #2709 
Now only the following action will update '_modified' meta-data field:
- (V) Posting a new update to the research
- (X) Editing an existing update
- (X) Leaving a new comment
- (X) Marking the research as useful
- (X) Following the research

## Git Issues

Closes #2709

## Screenshots/Videos

---

## What happens next?


balhelhas:
Hello @goratt12, thank you for your input. I have checked in the preview environment and the issue is still present. Could you check again your implementation?

davehakkens:
Seems to work here now! nice

src/stores/Research/research.store.tsx
```diff
@@ -212,6 +212,7 @@ export class ResearchStore extends ModuleStore {
   public async toggleUsefulByUser(
     docId: string,
     userName: string,
+    keep_modified_timestamp = true,
```
evakill, line 215: i know it isn't 100% consistent, but we try to use camelCase for variable names in code rather than snake_case

evakill (commented):
nice fix! please fix the casing of the variable and run lint (`yarn lint`) then this is good to go

iSCJT:
Sorry I just randomly saw an email on my phone, not near a laptop to check and might not have full context but we shouldn't be changing when the modified time stamp is updated.

There's a separate field we should be using called content_last_modified, something like that. If the sorts not working then it should be updated to use that field instead.

goratt12:
> Sorry I just randomly saw an email on my phone, not near a laptop to check and might not have full context but we shouldn't be changing when the modified time stamp is updated.
> 
> There's a separate field we should be using called content_last_modified, something like that. If the sorts not working then it should be updated to use that field instead.

goratt12:
> Sorry I just randomly saw an email on my phone, not near a laptop to check and might not have full context but we shouldn't be changing when the modified time stamp is updated.
> 
> There's a separate field we should be using called content_last_modified, something like that. If the sorts not working then it should be updated to use that field instead.

I think you are correct, the field you are referring named '_contentModifiedTimestamp' and its only set if the option 'set_last_edit_timestamp' is set to true when calling DocReference.set / DocReference.update.
Currently these functions are called with this option set to true only when:
- Uploading an update to a research item
- Deleting an update from a research item
- Uploading a HowTo item

this behavior is happening because sorting by Modified is to sort by the '_modified' field here:
https://github.com/ONEARMY/community-platform/blob/74c6a16aafc5afa655ecc4ac2fd4485b5cf10341/src/stores/common/FilterSorterDecorator/FilterSorterDecorator.ts#L74-L77

Should we open a new issue or address it in here? I can fix it and commit the changes

evakill:
thanks @iSCJT !! i didn't realize we had implemented that. 

ok so.. good find @goratt12 - in that [FilterSorterDecorator](https://github.com/ONEARMY/community-platform/blob/74c6a16aafc5afa655ecc4ac2fd4485b5cf10341/src/stores/common/FilterSorterDecorator/FilterSorterDecorator.ts#L74-L77), we should change the sort from looking at `_modified` to `_contentModifiedTimestamp`. this should give the desired outcome without having to conditionally update `_modified`. let me know if you have any questions about the approach or implementation!

goratt12:
Changed the sort by Modified to use the '_contentModifiedTimestamp' field.
And set the default sort to be Modified

I don't have permissions to create/update research articles, can you please check the changes does not break the functionality?

evakill:
@goratt12 thanks, looks good! we only need the changes in the file `src/stores/common/FilterSorterDecorator/FilterSorterDecorator.ts`, can you remove the changes in the other files please?

evakill:
also sorry, i think i gave you the wrong lint command - `yarn format:style` will auto fix the errors!

goratt12:
I got an error in CircleCI automated test that indicate the array `sortingOptions` in undefined and I cannot use the `find` function on it. 
but when I'm running the application on my computer this error doesn't occur,
Even when I debug this function, the array is always populated.
Can someone help me understand why CircleCI got this error?

evakill (approved):
looks like tests are passing and preview is working now @goratt12! i think the change with the default sorting option fixed it. thanks for your contribution!

evakill:
sorry @goratt12 - i spoke to soon. we should update the test [here](https://github.com/ONEARMY/community-platform/blob/e1f07b156cc05a4ed813b768c0f2502b8d2a8620/src/stores/common/FilterSorterDecorator/FitlerSorterDecorator.test.ts#L114) before we merge. could you update the `mockItems` to include the `_contentModifiedTimestamp` field, and update the `modified` test accordingly? 

thanks for helping us maintain tests!

evakill (approved):

PR #2724, opened 2023-08-25
Fix #2315 - Possible to reuse previous slugs causing old links to break

### LINKED ISSUE #2315: [bug] previously used how to & research slugs not prevented from being used
**Describe the bug**
If the title of a how to or research item is updated then the slug is also updated, previously used slugs are saved for that item so that old links can be maintained however nothing prevents a new item from being created with a previously used slug.

**To Reproduce**

1. Edit how-to / research item and change title
2. Open item using original and new slugs to confirm that accessible using both
3. Create how-to / research item and use the old title of the previous item
4. Complete details of item and submit
5. Observe that new item uses original slug and loads correctly but that it is not possible to access the edited item using the original slug.

**Expected behaviour**
That current and previous slugs are globally unique across item type.

**Screenshots**

**Additional context**
My suggestion is a global lookup table of slugs and corresponding document id's which is checked before assigning a slug.  If a slug is in use then either return the existing error about the title not being unique or auto append an incrementing number.

benkelaar:
PR Checklist

PR Type

## Description

Adds a check on previousSlugs on title validation.

## Git Issues

Closes #2315 

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/packages/cypress/src/integration/howto/write.spec.ts
+++ b/packages/cypress/src/integration/howto/write.spec.ts
@@ -176,6 +176,15 @@ describe('[How To]', () => {
         'Titles must be unique, please try being more specific',
       ).should('exist')
 
+      cy.step('Warn if title is identical with a previously existing one')
+      cy.get('[data-cy=intro-title]')
+        .clear()
+        .type('Make glassy beams')
+        .blur({ force: true })
+      cy.contains(
+        'Titles must be unique, please try being more specific',
+      ).should('exist')
+
       cy.step('Warn if title has less than minimum required characters')
       cy.get('[data-cy=intro-title]').clear().type('qwer').blur({ force: true })
       cy.contains(
--- a/packages/cypress/src/integration/research/read.spec.ts
+++ b/packages/cypress/src/integration/research/read.spec.ts
@@ -2,7 +2,7 @@ describe('[Research]', () => {
   const SKIP_TIMEOUT = { timeout: 300 }
   const totalResearchCount = 2
   const researchArticleUrl = '/research/qwerty'
-  const authoredResearchArticleUrl = '/research/A%20test%20research'
+  const authoredResearchArticleUrl = '/research/a-test-research'
 
   beforeEach(() => {
     cy.visit('/research')
@@ -84,7 +84,7 @@ describe('[Research]', () => {
         cy.get('[data-cy="go-back"]:eq(0)').as('topBackButton').click()
 
         cy.step('Views show on second research article')
-        cy.visit('/research/A%20test%20research')
+        cy.visit('/research/a-test-research')
         cy.get('[data-cy="ViewsCounter"]').should('exist')
       })
     })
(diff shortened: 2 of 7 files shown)


evakill (approved):
good stuff, thanks @benkelaar! really appreciate the comments and testing 💪

benkelaar:
@evakill Thanks!  That was quick!  You don't think the concern [I describe here](https://github.com/ONEARMY/community-platform/issues/2315#issuecomment-1694488376) is an issue?

evakill:
> So I do have a concern, coming out of the tests as well. The test data for the research items did not contain previousSlugs arrays at all, does this mean that there might be how tos or research items in production databases that don't have this property either?

oops, wasn't following the thread on the issue! let me look into this first thing, i believe the research and howto both track previous slugs in the prod db

> Do you guys do database migrations? Because I couldn't find any so far, but that might be a good mechanism to fix this.

i wish 😅 I'm not aware of migrations frameworks that work w firebase. if you have experience here would love to hear it. absolutely agree we could improve our database hygiene generally - it is hard to keep track of schemas, especially with the multiple database instances

evakill:
@benkelaar yeah, it seems it's just that the research test data was missing the previous slugs - confirmed that the howtos and research both use that field in the DB. 

thanks for modifying the research mocks, and for better test coverage we should add a similar test to research as you did [here](https://github.com/ONEARMY/community-platform/pull/2724/files#diff-b04f255b1accab30e685dcb0eb02e7cbb3edb8cf6f40571d9deacb914a93df4aR179-R187). let me know if you'd like to do that, or i can make an issue for someone else to pick up.

sorry for not reviewing the tests thoroughly enough and thanks for bringing up this concern!

benkelaar:
Since it was just a variant of the already existing "Title already exists" use case I thought it might not be necessary to add another e2e test in both variants (Especially since they use the same code).

I did actually get a unit test to work just now, so would it be acceptable if I cover all variants with unit tests instead?

benkelaar:
> i wish 😅 I'm not aware of migrations frameworks that work w firebase. if you have experience here would love to hear it. absolutely agree we could improve our database hygiene generally - it is hard to keep track of schemas, especially with the multiple database instances

On this, I haven't seen it on a firebase-backed application either.  Theoretically shouldn't be too hard to roll your own though, you just need to store a version in the database somewhere and then keep an index of migrations against version number.  But the devil's in the details of course and I'm not sure how high up the priority list that would be for you.

Quick google gets me via https://stackoverflow.com/questions/35147780/firebase-database-migration to https://github.com/kevlened/fireway , but that seems to have support issues, but API design seems solid (Love flyway).  There is a more recently updated fork here https://github.com/dev-aces/fireway but that's once again only supported by one person so you'd basically have to adopt the library if you want to be dependent on it.

evakill:
> I did actually get a unit test to work just now, so would it be acceptable if I cover all variants with unit tests instead? 

yep definitely, i tend to prefer the unit tests, the cypress tests can be super slow

benkelaar:
@evakill Here's the promised unit tests: https://github.com/ONEARMY/community-platform/pull/2730

PR #2749, opened 2023-09-05
feat: add thumbnails to research lists

### LINKED ISSUE #2682: [feature request] add thumbnail images to the research listing page
To include picture and better indication of what the 3 icons / numbers mean for each research item.

goratt12:
PR Checklist

PR Type

## Description

Added thumbnails to research list

## Git Issues

Closes #2682

## Screenshots/Videos

image

---

## What happens next?


AlfonsoGhislieri:
Thanks for this @goratt12, this makes me think we need to have a think about how to actually move forward in regards to generating and serving thumbnails. The current option that is proposed in this PR I think poses some issues for the resources that are needed and the sacalability moving forwards. To avoid having to pay off this debt down the line, I think we  should re-think the implementation.

Discussion on the topic in: 
https://discord.com/channels/586676777334865928/1148720148509900862

The two options that stand out to me:

- Storing the thumbnail in the DB (adding it to the items on upload and edit to avoid them going stale) --> this does require us to migrate all old howtos
- Using a cdn to service these images --> I wonder if we do have enough traffic/images to really need this approach, might be overengineering at this point?

I think this is definitely an area that could use some discussion before going ahead with implementing thumbnails, @ONEARMY/maintainers.

davehakkens:
Putting this here for visibility. Its an issue that had been opened up before https://github.com/ONEARMY/community-platform/issues/2548

No strong opinion on CDN vs resizing. But agree good moment to tackle the image-size issue side wide!

davehakkens:
hi @goratt12 meanwhile we figure out the discussion on the images. Would you be up for making it mobile ready? A bit bumpy atm
afbeelding

goratt12:
Yeah sure, I’ll have a look at it @davehakkens

goratt12:
Now the thumbnail will be hidden on mobile.

src/assets/images/default-research-thumbnail.jpg
thisislawatts, line 1: @goratt12 As this image is only being used for a block that is ~100x100px I've downscaled the image to be 200x200px. This reduces the image size from 1.31mb to 11.7 kb

src/pages/Research/Content/ResearchListItem.tsx
```diff
@@ -180,6 +196,18 @@ const ResearchListItem = ({ item }: IProps) => {
   )
 }
 
+const getItemThumbnail = (researchItem: IResearch.ItemDB): string => {
+  if (researchItem.updates?.length) {
+    const latestImage = researchItem?.updates
```
thisislawatts, line 201: **issue(blocking):** Please note that we should only load images from Updates which are publicly available. It is possible for an author to mark an update as either draft or deleted. In this case the author would not expect the image to be visible to the public.

Here's the logic [available elsewhere](https://github.com/ONEARMY/community-platform/blob/cd6ef960b50f64e59d2c3b50935c0a7e43b45e4d/src/pages/Research/Content/ResearchArticle.tsx#L254C1-L256C2):
```ts
const isUpdateVisible = (update: IResearch.UpdateDB) => {
  return update.status !== 'draft' && !update._deleted
}
```

It is being used in the same file to [filter updates for display](https://github.com/ONEARMY/community-platform/blob/cd6ef960b50f64e59d2c3b50935c0a7e43b45e4d/src/pages/Research/Content/ResearchArticle.tsx#L193C1-L195C40). It is probably worth introducing a shared helper for fetching these updates, perhaps something like `getPublicUpdates(researchItem.updates)`

thisislawatts, line 201: **issue(blocking):** Please note that we should only load images from Updates which are publicly available. It is possible for an author to mark an update as either draft or deleted. In this case the author would not expect the image to be visible to the public.

Here's the logic [available elsewhere](https://github.com/ONEARMY/community-platform/blob/cd6ef960b50f64e59d2c3b50935c0a7e43b45e4d/src/pages/Research/Content/ResearchArticle.tsx#L254C1-L256C2):
```ts
const isUpdateVisible = (update: IResearch.UpdateDB) => {
  return update.status !== 'draft' && !update._deleted
}
```

It is being used in the same file to [filter updates for display](https://github.com/ONEARMY/community-platform/blob/cd6ef960b50f64e59d2c3b50935c0a7e43b45e4d/src/pages/Research/Content/ResearchArticle.tsx#L193C1-L195C40). It is probably worth introducing a shared helper for fetching these updates, perhaps something like `getPublicUpdates(researchItem.updates)`

thisislawatts (changes requested):

src/pages/Research/Content/ResearchListItem.tsx
goratt12: I created a shared function in the `src\utils\helpers.ts` file named `getPublicUpdates`.
linked both the thumbnail code and the `ResearchArticle.tsx` code

thisislawatts (approved):

PR #2752, opened 2023-09-06
feat: pin user howtos that require moderation/draft

AlfonsoGhislieri:
PR Checklist

PR Type

## Description

Pins a users howtos to the top of the howto list if they are a draft/rejected/awaiting-moderation

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/src/stores/Howto/howto.store.tsx
+++ b/src/stores/Howto/howto.store.tsx
@@ -219,7 +219,10 @@ export class HowtoStore extends ModuleStore {
       this.searchValue,
     )
 
-    return validHowtos
+    return this.filterSorterDecorator.getSortedItems(
+      this.activeUser,
+      validHowtos,
+    )
   }
 
   public async incrementDownloadCount(howToID: string) {
--- a/src/stores/Howto/howto.store.tsx
+++ b/src/stores/Howto/howto.store.tsx
@@ -77,8 +77,12 @@ export class HowtoStore extends ModuleStore {
     // the given endpoint and emits changes as data is retrieved from cache and live collection
     super(rootStore, COLLECTION_NAME)
     makeObservable(this)
-
-    this.activeSorter = ItemSortingOption.None
+    this.allDocs$.subscribe((docs: IHowtoDB[]) => {
+      runInAction(() => {
+        this.filterSorterDecorator = new FilterSorterDecorator<any>(docs)
+        this.updateActiveSorter('Created')
+      })
+    })
     this.selectedCategory = ''
     this.searchValue = ''
     this.referrerSource = ''
@@ -219,10 +223,9 @@ export class HowtoStore extends ModuleStore {
       this.searchValue,
     )
 
-    return this.filterSorterDecorator.getSortedItems(
-      this.activeUser,
-      validHowtos,
-    )
+    this.filterSorterDecorator.allItems = validHowtos
+
+    return this.filterSorterDecorator.sort(this.activeSorter, this.activeUser)
   }
 
   public async incrementDownloadCount(howToID: string) {
--- a/src/stores/Howto/howto.store.tsx
+++ b/src/stores/Howto/howto.store.tsx
@@ -84,7 +84,7 @@ export class HowtoStore extends ModuleStore {
 
       runInAction(() => {
         this.allHowtos = activeItems
-        this.filterSorterDecorator = new FilterSorterDecorator<any>(
+        this.filterSorterDecorator = new FilterSorterDecorator<IHowtoDB>(
           this.allHowtos,
         )
         // sets default starting sort filter for howto list items
@@ -101,8 +101,13 @@ export class HowtoStore extends ModuleStore {
   }
 
   public updateActiveSorter(sorter: ItemSortingOption) {
-    this.allHowtos = this.filterSorterDecorator?.sort(sorter, this.activeUser)
     this.activeSorter = sorter
+
+    this.allHowtos = this.filterSorterDecorator?.sort(
+      sorter,
+      this.allHowtos,
+      this.activeUser,
+    )
   }
 
   public getActiveHowToComments(): IComment[] {
@@ -224,9 +229,11 @@ export class HowtoStore extends ModuleStore {
       this.searchValue,
     )
 
-    this.filterSorterDecorator.allItems = validHowtos
-
-    return this.filterSorterDecorator.sort(this.activeSorter, this.activeUser)
+    return this.filterSorterDecorator.sort(
+      this.activeSorter,
+      validHowtos,
+      this.activeUser,
+    )
   }
 
   public async incrementDownloadCount(howToID: string) {
--- a/src/stores/Howto/howto.store.tsx
+++ b/src/stores/Howto/howto.store.tsx
@@ -77,18 +77,21 @@ export class HowtoStore extends ModuleStore {
     // the given endpoint and emits changes as data is retrieved from cache and live collection
     super(rootStore, COLLECTION_NAME)
     makeObservable(this)
+    super.init()
+
     this.allDocs$.subscribe((docs: IHowtoDB[]) => {
+      logger.debug('docs', docs)
       const activeItems = [...docs].filter((doc) => {
         return !doc._deleted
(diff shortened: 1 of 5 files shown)


src/stores/Howto/howto.store.tsx
```diff
@@ -206,7 +206,10 @@ export class HowtoStore extends ModuleStore {
       this.searchValue,
     )
 
-    return validHowtos
+    return this.filterSorterDecorator.getSortedItems(
+      this.activeUser,
+      validHowtos,
+    )
```
AlfonsoGhislieri, line 212: I was forced to pass the `validHowtos` to the `filterSorterDecorator.getSortedItems` function rather than modifying the `filterSorterDecorator` allItems directly like for [research items](https://github.com/ONEARMY/community-platform/blob/8022687ab9b4609d60e09d34e31a5c3e0b355586/src/stores/Research/research.store.tsx#L120)

I'm not sure why but when I try the same implementation for the howtoStore it just clears the `filterSorterDecorator.Items `array ...

src/stores/Howto/howto.store.tsx
```diff
@@ -206,7 +208,9 @@ export class HowtoStore extends ModuleStore {
       this.searchValue,
     )
 
-    return validHowtos
+    this.filterSorterDecorator.allItems = validHowtos
```
evakill, line 211: why are we resetting `allItems` to `validHowtos`? if we want only the valid items sorted by `getSortedItems` i would suggest updating `getSortedItems` optionally accept a `listItems` argument as we do in the other methods

evakill (changes requested):
updating the active sorter "unpins" the moderation/draft PRs from the top because it uses the `sort` function rather than the `getSortedItems` function. not sure if this is expected behavior? but the research does not do that - items stay pinned despite the sort

src/stores/Howto/howto.store.tsx
AlfonsoGhislieri: So this was my original thinking as well... I replicated the implementation that was originally also present in the research store.

The reason we are resetting it is because the `getSortedItems` function uses [allItems](https://github.com/ONEARMY/community-platform/blob/1e8155d4573417986fc6c63681fcf9c8c989ba8a/src/stores/common/FilterSorterDecorator/FilterSorterDecorator.ts#L136) from it's own class, I had originally implemented it like you suggested but changed it back to avoid making more changes.

Happy to go back and do this though!

AlfonsoGhislieri:
So the latest commit makes sure that everything works, I do think that there could be some re-thinking about how it's all implemented. I'll have a look at refactoring and improving this in a future PR. Let me know what you think @evakill and if you think we should pass the valid items to `filterSorterDecorator` or leave it like it is now (just like with the research store, but I'll have a look at improving this as well in a next PR)

evakill (changes requested):
the more i play with this the weirder things get 😅 
some observations:
- changing the sorter doesn't always trigger a resort? not sure why
- sometimes things get unpinned still
- maybe we should have a "backup" unique sort key that's the id or something. that means that every time you change a sorter, you'll get the same result. without this, the sorter will not move things with the same value, so orders are not consistent (lmk if this makes sense it's hard to explain lol)


https://github.com/ONEARMY/community-platform/assets/37253846/2efa9ea4-60bd-429d-b7c9-74973525f319

AlfonsoGhislieri:
Hmmm well this is quite interesting...

>changing the sorter doesn't always trigger a resort? not sure why

This doesn't seem to be the case at all locally... So I wonder why on the preview it's happening.. Will have a look into this, maybe something with caching hmm

>sometimes things get unpinned still

This is by design, it's because when you are sorting by other filters eg: category or the direct search that don't match then it will disregard any post that doesn't match those parameters. Regardless of it being your own howto that is in awaiting-moderation, rejected  or draft. I thought this was the desired design as it may then just be very confusing if they are always showing even when doing a detailed search (it's the same for researches). Or do we really always want all of them pinned regardless of whether or not they are even in the right category for example? 

>maybe we should have a "backup" unique sort key that's the id or something. that means that every time you change a sorter, you'll get the same result. without this, the sorter will not move things with the same value, so orders are not consistent (lmk if this makes sense it's hard to explain lol)

Here you lost me a little @evakill! Not exactly sure what you mean here

src/stores/Howto/howto.store.tsx
AlfonsoGhislieri: Sure, I think this is probably the better approach as well, will be more in line with the other functions as well. We can pass the `validItems` to the `getSortedItems` directly.

src/stores/common/FilterSorterDecorator/FilterSorterDecorator.ts
```diff
@@ -168,12 +168,12 @@ export class FilterSorterDecorator<T extends IItem> {
   }
 
   @action
-  public sort(query: string): any[] {
+  public sort(sorter: ItemSortingOption, activeUser?: IUser): T[] {
     const sortingOption: ItemSortingOption =
-      ItemSortingOption[query as keyof typeof ItemSortingOption]
+      ItemSortingOption[sorter as keyof typeof ItemSortingOption]
```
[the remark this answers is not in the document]
AlfonsoGhislieri: Good catch, too much time away fro TS.. Removed in 6c0172e

src/stores/Howto/howto.store.tsx
AlfonsoGhislieri: Passing validItems to the `sort` function which then passes them along to `getSortedItems` 6c0172e. 

I think a case could be made to merge `sort` and `getSortedItems` into one function.. `getSortedItems` is never used outside of `sort`, not sure we really need these to be separate functions? But this should probably be one for a refactor PR..

AlfonsoGhislieri:
Some of the weird behaviour that you observed @evakill was due to the shuffling of the items in the lists when items had equal values, the fix in #2761 should address that. And I think this is in line with the reasoning you were suggesting here?

> maybe we should have a "backup" unique sort key that's the id or something. that means that every time you change a sorter, you'll get the same result. without this, the sorter will not move things with the same value, so orders are not consistent (lmk if this makes sense it's hard to explain lol)

src/stores/Research/research.store.tsx
```diff
@@ -91,17 +92,17 @@ export class ResearchStore extends ModuleStore {
...
+          this.activeSorter,
+          activeItems,
+        )
       })
     })
   }
 
   public updateActiveSorter(sorter: ItemSortingOption) {
-    this.allResearchItems = this.filterSorterDecorator?.sort(sorter)
     this.activeSorter = sorter
```
AlfonsoGhislieri, line 106: this is no longer needed since `this.activeSorter` being observable triggers the computed get `filteredResearches` value

src/stores/Howto/howto.store.tsx
```diff
@@ -77,26 +77,33 @@ export class HowtoStore extends ModuleStore {
...
-      logger.debug('docs', docs)
-      this.filterSorterDecorator = new FilterSorterDecorator<any>(docs)
-      // Sets default starting sort filter for howto list items
-      this.updateActiveSorter(ItemSortingOption.Created)
-    })
   }
 
   public updateActiveSorter(sorter: ItemSortingOption) {
-    this.allHowtos = this.filterSorterDecorator?.sort(sorter)
     this.activeSorter = sorter
```
AlfonsoGhislieri, line 107: this is no longer needed since `this.activeSorter `being observable triggers the computed get `filteredHowtosvalue` to be re-run

evakill:
cool! this looks like an overall improvement, thanks for the updates and comments @AlfonsoGhislieri. i'm ready to stamp the code, but still getting something weird on the preview where only 8 howtos show up and i'm not able to add a new one. is that happening for you? might just need a rebase, not sure

AlfonsoGhislieri:
@evakill yeah I think it must have been an issue with the preview, now that I rebased it looks to be working like it should (all still good locally as well)

evakill (approved):
woooo same! 🚢 🚢 🚢

PR #2769, opened 2023-09-12
fix: research updates and comments count

### LINKED ISSUE #2760: [bug] old research updates not deleted properly
**Describe the bug**

Correct number of updates are not actually being stored, this probably has to do with updates in the past not having been correctly deleted from the item in the database? Not sure how big of an issue this is, and considering this is no longer an issue as deletion is correctly handled, if we even want to go back and fix it?

**To Reproduce**

1. Go to 'research'
2. Look at a 'make a house from recycled plastic' researches total updates
3. Click on 'make a house from recycled plastic'
4. See that there are not actually the same number of updates

**Expected behaviour**

**Screenshots**
image

image
**Additional context**

laviodias:
PR Checklist

PR Type

## Description

This PR fixes the amount of updates and comments on researches. The comments from deleted updates were being considered and now they're ignored. Also fixed a small type problem on a function that was returning `boolean | ""` when it was supposed to return only `boolean`.

## Git Issues

Closes #2760

## Screenshots/Videos

Since the updates were not actually marked as `_deleted` (instead, this field was null), they should appear on the research page.
image

As you can see, now all the 7 updates are being shown.

---

## What happens next?

Obs: My first contribution to the project, so sorry if I did something wrong. I read every docs I found, so I hope it's all according to the standard.

### DIFF
--- a/src/pages/Research/Content/ResearchListItem.tsx
+++ b/src/pages/Research/Content/ResearchListItem.tsx
@@ -198,7 +198,7 @@ const getItemDate = (item: IResearch.ItemDB, variant: string): string => {
 const calculateTotalComments = (item: IResearch.ItemDB) => {
   if (item.updates) {
     const commentOnUpdates = item.updates.reduce((totalComments, update) => {
-      const updateCommentsLength = update.comments ? update.comments.length : 0
+      const updateCommentsLength = !update._deleted && update.comments ? update.comments.length : 0
       return totalComments + updateCommentsLength
     }, 0)
 
@@ -212,7 +212,7 @@ const getUpdateText = (item: IResearch.ItemDB) => {
   return item.updates?.length
     ? String(
         item.updates.filter(
-          (update) => update.status !== 'draft' && update._deleted !== true,
+          (update) => update.status !== 'draft' && !update._deleted,
         ).length,
       )
     : '0'
--- a/src/pages/Research/Content/ResearchArticle.tsx
+++ b/src/pages/Research/Content/ResearchArticle.tsx
@@ -252,7 +252,7 @@ const ResearchArticle = observer((props: IProps) => {
 })
 
 const isUpdateVisible = (update: IResearch.UpdateDB) => {
-  return update.status !== 'draft' && update._deleted === false
+  return update.status !== 'draft' && !update._deleted
 }
 
 const transformToUserComment = (
--- a/src/utils/helpers.ts
+++ b/src/utils/helpers.ts
@@ -143,7 +143,7 @@ export const isAllowedToDeleteContent = (doc: IEditableDoc, user?: IUser) => {
   return (
     roles.includes('admin') ||
     roles.includes('super-admin') ||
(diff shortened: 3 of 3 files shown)


src/pages/Research/Content/ResearchListItem.tsx
```diff
@@ -198,7 +198,8 @@ const getItemDate = (item: IResearch.ItemDB, variant: string): string => {
 const calculateTotalComments = (item: IResearch.ItemDB) => {
   if (item.updates) {
     const commentOnUpdates = item.updates.reduce((totalComments, update) => {
-      const updateCommentsLength = update.comments ? update.comments.length : 0
+      const updateCommentsLength =
+        !update._deleted && update.comments ? update.comments.length : 0
```
AlfonsoGhislieri, line 202:
```suggestion
        !update._deleted && update.status !== 'draft' && update.comments ? update.comments.length : 0
```

**Issue**: we are also not checking if the update status is NOT a draft, can we please add this check or else we will be showing the wrong number of comments.

AlfonsoGhislieri (changes requested):
Thanks for raising this @laviodias, one issue is that there is some shared functionality here between the `ResearchListItem` component and the [`FilterSorterDecorator` class](https://github.com/ONEARMY/community-platform/blob/ec6da2514c220ad8ac9924f5bfd50c19c0c4c22a/src/stores/common/FilterSorterDecorator/FilterSorterDecorator.ts#L39), they are both using a calculateTotalComments function. 

Let's make sure they are both using the same function in order to retain the desired functionality and keep it maintable. Can we pull this function out and add it to the [helpers](https://github.com/ONEARMY/community-platform/blob/master/src/utils/helpers.ts) file please?

laviodias:
Done!
I added some tests with general mocks using the array `fill` method only with the arguments that I needed for that function. If you believe that it's better to manually write more complete items, let me know.

src/utils/helpers.test.ts
```diff
@@ -200,4 +201,69 @@ describe('src/utils/helpers', () => {
...
+            {
+              status: 'draft',
+              _deleted: false,
+              comments: Array.from({ length: 6 }),
+            },
+          ]),
+      } as any
+      expect(calculateTotalComments(item)).toBe('4')
+    })
+  })
```
AlfonsoGhislieri, line 268: Thanks for adding these tests, could we use [research factory items](https://github.com/ONEARMY/community-platform/blob/master/src/test/factories/ResearchItem.ts) for the tests just to check against proper items, will help to make sure everything keeps working in future if we make changes to research items/updates/comments

laviodias: Updated!

AlfonsoGhislieri:
Lovely, thanks very much for all the changes and the fix @laviodias 🌟

AlfonsoGhislieri (approved):

AlfonsoGhislieri:
@all-contributors add @laviodias for code

PR #2780, opened 2023-09-13
fix: prevent signup if username contains invalid characters

### LINKED ISSUE #2408: [feature request] prevent signup if username contains invalid characters
Username that is saved is different to the one entered by the user when joining as all special characters are stripped.  e.g. if you used sean@example.com as the username in the database it would save seanexamplecom

I think it's preferable to tell users they can only use certain characters for the username and check this as part of the form validation prior to creating the user.

request a display name on the user signup form and then use an auto generated username based on the email address in the background.  Where the username currently appears in the ui then show the display name.

**Additional context**
N/A

jgable:
PR Checklist

PR Type

## Description

- Add a new `noSpecialCharacters` validator along with a `getSpecialCharacters` helper.
- Refactored the SignUp form to use the new validator via `composeValidators`.

## Git Issues

Closes #2408 

## Screenshots/Videos

Screenshot 2023-09-13 at 11 51 01 AM

I also tested the signup flow worked when using a display name without special characters and ran the SignUp Cypress tests locally.

### DIFF
--- a/src/utils/validators.test.ts
+++ b/src/utils/validators.test.ts
@@ -79,6 +79,6 @@ describe('noSpecialCharacters', () => {
   it('returns proper message for email values', () => {
     const result = noSpecialCharacters('someones@email.com')
 
-    expect(result).toContain("Only letters and numbers are allowed")
+    expect(result).toContain('Only letters and numbers are allowed')
   })
 })
--- a/src/utils/validators.test.ts
+++ b/src/utils/validators.test.ts
@@ -1,4 +1,8 @@
-import { draftValidationWrapper, validateTitle } from './validators'
+import {
+  draftValidationWrapper,
+  noSpecialCharacters,
+  validateTitle,
+} from './validators'
 
 // Mock out module store to limit impact of circular dependency
 jest.mock('src/stores/common/module.store')
@@ -62,5 +66,19 @@ describe('validateTitle', () => {
       'Titles must be unique, please try being more specific',
     )
     expect(isReusedMock).toHaveBeenCalledWith(duplicatedTitle, id)
+  })
+})
+
+describe('noSpecialCharacters', () => {
+  it('returns undefined when no special characters are present', () => {
+    const result = noSpecialCharacters('validUsername')
+
+    expect(result).toBeUndefined()
+  })
+
+  it('returns proper message for email values', () => {
+    const result = noSpecialCharacters('someones@email.com')
+
+    expect(result).toContain("Only letters and numbers are allowed")
   })
 })
--- a/src/utils/validators.ts
+++ b/src/utils/validators.ts
@@ -1,5 +1,5 @@
 import isUrl from 'is-url'
-import { stripSpecialCharacters } from './helpers'
+import { getSpecialCharacters, stripSpecialCharacters } from './helpers'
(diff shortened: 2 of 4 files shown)


src/utils/validators.test.ts
```diff
@@ -64,3 +68,18 @@ describe('validateTitle', () => {
...
+  it('returns undefined when no special characters are present', () => {
+    const result = noSpecialCharacters('validUsername')
+
+    expect(result).toBeUndefined()
+  })
+
+  it('returns proper message for email values', () => {
+    const result = noSpecialCharacters('someones@email.com')
+
+    expect(result).toBeDefined()
```
AlfonsoGhislieri, line 82: **Nitpick**: we don't need this assertion, we are already checking that the result will contain a string  after

src/utils/validators.ts
```diff
@@ -13,6 +13,15 @@ import type { Mutator } from 'final-form'
 const required = (value: any) =>
   value ? undefined : 'Make sure this field is filled correctly'
 
+const noSpecialCharacters = (value: string) => {
+  const specialCharacters = getSpecialCharacters(value)
+  return specialCharacters.length > 0
+    ? `Cannot contain special characters; ${specialCharacters
+        .map((c) => `'${c}'`)
+        .join(', ')}`
+    : undefined
```
AlfonsoGhislieri, line 22: **Issue**: not so sure about the error message here, it's a little confusing with the semi colon followed by a special characters in question..

Maybe we can use something simpler and easier for the user to understand at just a glance, eg: "Only letters and numbers are allowed." Taking this from [Microsoft 365](https://learn.microsoft.com/en-us/office/troubleshoot/office-suite-issues/username-contains-special-character)

AlfonsoGhislieri (changes requested):
Thanks for this @jgable just a couple things I would change!

src/utils/validators.ts
```diff
@@ -13,6 +13,15 @@ import type { Mutator } from 'final-form'
 const required = (value: any) =>
   value ? undefined : 'Make sure this field is filled correctly'
 
+const noSpecialCharacters = (value: string) => {
+  const specialCharacters = getSpecialCharacters(value)
+  return specialCharacters.length > 0
+    ? `Cannot contain special characters; ${specialCharacters
+        .map((c) => `'${c}'`)
+        .join(', ')}`
```
AlfonsoGhislieri, line 21:
```suggestion
    ? "Only letters and numbers are allowed."
```

Issue: not so sure about the error message here, it's a little confusing with the semi colon followed by a special characters in question..

Maybe we can use something simpler and easier for the user to understand at just a glance, eg: "Only letters and numbers are allowed." Taking this from Microsoft 365

AlfonsoGhislieri:
Thanks for the changes here @jgable, just linting test failing since prettier wasn’t run on one of the files (validator test file)

AlfonsoGhislieri:
Thanks for the changes @jgable, we are failing the lint test in the `src/utils/validators.test.ts` file because prettier wasn't run on the file. Could we fix that please? :-)

https://prettier.io/

thisislawatts (approved):
Thanks @jgable for the time and effort on this PR ✨ Appreciate you for helping to build the community platform.

thisislawatts:
@all-contributors add @jgable for code

Issue #2379, opened 2023-05-24
[feature request] send email notifications when how-to/research/map pin approved or 'rejected'


iSCJT:
**Is your feature request related to a problem? Please describe.**
Users do not currently know if their post has been approved or needs amendments before being approved.

**Describe the solution you'd like**
Email notification to be sent when how to or research article is approved or rejected.  Email should include link to post and if 'rejected' include [comments left by admins](https://github.com/ONEARMY/community-platform/issues/2377)

Need to include new map pin status needs improvement as well (clarify exact naming).

**Describe alternatives you've considered**
User continues to log in to check if post approved.

**Additional context**

thisislawatts:
@iSCJT is this useful without the _comments left by admins_ work or should we consider this blocked by #2377?

iSCJT:
@thisislawatts I think it would still be useful as a standalone but agree nicer to implement following 2377.

thisislawatts:
@iSCJT given the dependency on #2377, should we remove it from the June sprint?

iSCJT:
@evakill I've added a pr for displaying the comments now, do you want me to look at this or do you think you'll still have time to?

evakill:
@iSCJT if you have time to pick it up, go for it! and if you want we can pair sometime this week

iSCJT:
@sigolenej if you could add required wording one this ticket that would be appreciated.

sigolenej:
Thanks @iSCJT ! 

Can you confirm these will be sent by the community@preciousplastic.com email? 

In regards to the design, if it will look like a normal email for the user (which is fine), can we include our automatic signature?
![Screenshot 2023-07-27 at 16 21 51](https://github.com/ONEARMY/community-platform/assets/138245356/15e24614-a029-47b6-8985-dd0ba17a92be)


Here are the emails we currently send to humans related to their HT. 

We have been using dynamic content, that will be shown between " {} ".

                1. How-To submission confirmation
**_---- Title ---_**
Your How-to -{HT title}- has been submitted

**_---- Content ---_**
Hey {username},

Your How-To -{HT title}- has been received! 

Our team is currently reviewing it and will get back to you within a few business days. 

Thank you for taking the time, the community will benefit from this knowledge and more plastic will be recycled thanks to YOU. 

Cheers,
Charlie your Precious Plastic Community Manager

                2. How-To approval
**_---- Title ---_**
Your How-to -{HT title}- has been published!

**_---- Content ---_**
[...]

PR #2783, opened 2023-09-13
feat: add unsubscribe link in emails

### LINKED ISSUE #2863: [feature request] Easy unsubscribe from notifications emails
We want to make it easy to unsubscribe from notifications (without logging in)

Add unsubscribe to bottom of email 

**Layout** 
afbeelding

**Notes** 
Pr is already submitted. 
https://github.com/ONEARMY/community-platform/pull/2783

evakill:
PR Checklist

PR Type

## Description
we want an unsubscribe link in the emails we send out to users to avoid spamming them unwantedly. this should ideally not require users to log in, in case they've lost their login info and don't want to reenter the platform.

the "quick and easy" solution i came up with is creating a new unsubscribe page that pulls the username from the URL and updates the user's profile to set their emailFrequency to "never". this will stop them from getting any more emails from the platform. 

this is not a high-security solution, but the risk of malicious users going around and unsubscribing people is... small. hopefully.

added a cypress test for the new page.

cc @davehakkens 

## Git Issues

Closes #2863

## Screenshots/Videos

unsubscribe page:
Screen Shot 2023-09-13 at 5 29 18 PM

new email template:
Screen Shot 2023-09-13 at 5 44 17 PM

---

## What happens next?


davehakkens:
Works here!

thisislawatts:
Given the security concerns you mentioned here I am reluctant to introduce this technical debt. I am wondering instead if there is a relatively small tweak we can make here that would greatly improve the security posture.

For example looking into `handlePendingEmailEntry`, we could make the following additions:

```ts
const uniqToken = uuid(); // or nano id 

// Adding emails to this collection triggers an email notification to be sent to the user
const sentEmailRef = await db.collection(DB_ENDPOINTS.emails).add({
  to: toUserEmail,
  message: getNotificationEmailTemplate(toUser, pendingNotifications, /* New argument */ uniqToken ),
  // New attributes
  userId: toUser.userName,
  uniqToken, 
})
```

The unsubscribe link could then be formatted as `https://community.preciousplastic.com/unsubscribe/{uniqToken}`

Application logic would then be:
1. Load unsubscribe view providing uniqToken in route param
1. Query `emails` collection filtering on uniqToken fetched from route
1. If matching email document found update associated user record based on userId.  

Open questions:
1. How long do messages persist in `emails` collection for?
1. Can we add additional attributes to documents in the emails collection or will it cause the extension to throw an error 🤔

evakill:
ok, fair enough @thisislawatts. you can count on the security guy to not take shortcuts ;) 

> Can we add additional attributes to documents in the emails collection or will it cause the extension to throw an error?

not sure - docs don't say either way. i don't think we should deviate from the suggested schema, however, as the collection is tied to this extension. even if it works now, an update to the extension could possibly break it. i would rather update the user object with the unique token. the lookups would also be simpler, as we can still use the user store on the FE to get the token. 

i can make this update. i'm almost done with #2379 so will finish and put up a PR for that first then come back to this.

evakill:
@thisislawatts ready for re-review

thisislawatts (approved):

PR #2768, opened 2023-09-11
feat: add moderation notifications to emails

evakill:
PR Checklist

- [ ] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos

_If useful, provide screenshot or capture to highlight main changes_

---

## What happens next?

Thanks for the contribution! We try to make sure all PRs are reviewed ahead of a monthly dev call (first Monday of the month, open to all!).

If the PR is working as intended it'll be merged and included in the next platform release, if not changes will be requested and re-reviewed once updated.

If you need more immediate feedback you can try reaching out on Discord in the [Community Platform `development` channel](https://discord.com/channels/586676777334865928/938781727017558018).

PR #2792, opened 2023-09-16
feat(emails): send howto approval email on moderation update

evakill:
PR Checklist

PR Type

## Description

see #2791 for the first PR in this stack
depends on #2790 

**This PR (2)**
- trigger this function on moderation status change
- add template and styles for the how to approval email
- add testing for the flow

**PR 3**
- add templates and functions for how to rejected email, map pin approval email, and map pin rejected email

**future work**
- add PP custom footer (see issue)
- add emails for submitted resources
- add emails for research

## Git Issues

#2379 

## Screenshots/Videos
here is a rendering of the how-to approval email:
Screen Shot 2023-09-16 at 10 22 55 AM

---

## What happens next?

### DIFF
--- a/functions/src/emailNotifications/createModerationEmails.spec.ts
+++ b/functions/src/emailNotifications/createModerationEmails.spec.ts
@@ -1,0 +1,132 @@
+import { FirebaseEmulatedTest } from '../test/Firebase/emulator'
+import { DB_ENDPOINTS, IUserDB } from '../models'
+import { EmailNotificationFrequency } from 'oa-shared'
+import { HOW_TO_APPROVAL_SUBJECT } from './templates'
+import { setMockHowto } from '../emulator/seed/content-generate'
+import { handleHowToModerationUpdate } from './createModerationEmails'
+
+jest.mock('../Firebase/auth', () => ({
+  firebaseAuth: {
+    getUser: () => ({
+      email: 'test@test.com',
+    }),
+  },
+}))
+
+jest.mock('../config/config', () => ({
+  CONFIG: {
+    deployment: {
+      site_url: 'https://community.preciousplastic.com',
+    },
+  },
+}))
+
+const userFactory = (_id: string, user: Partial<IUserDB> = {}): IUserDB =>
+  ({
+    _id,
+    _authID: _id,
+    ...user,
+  } as IUserDB)
+
+describe('create email test', () => {
+  const db = FirebaseEmulatedTest.admin.firestore()
+
+  beforeAll(async () => {
+    await FirebaseEmulatedTest.clearFirestoreDB()
+    await FirebaseEmulatedTest.seedFirestoreDB('emails')
+
+    await FirebaseEmulatedTest.seedFirestoreDB('users', [
+      userFactory('user_1', {
+        displayName: 'User 1',
+        userName: 'user_1',
+        notification_settings: {
+          emailFrequency: EmailNotificationFrequency.DAILY,
+        },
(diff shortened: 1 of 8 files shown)


functions/src/emailNotifications/createModerationEmails.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    const countSnapshot = await db.collection(DB_ENDPOINTS.emails).count().get()
+    expect(countSnapshot.data().count).toEqual(1)
+
+    const querySnapshot = await db.collection(DB_ENDPOINTS.emails).get()
+    querySnapshot.forEach((doc) => {
+      const {
+        message: { html, subject },
+        to,
+      } = doc.data()
+      expect(html).toMatchSnapshot()
```
thisislawatts, line 87: **question:** Can we replace these snapshot tests for a set of specific assertions which verify user focused content. Such as _Huzzah! Your How-To ${resource.title} has been approved._

This move these tests to better align with how humans will use the emails.

evakill: i'm happy to have both! i think the snapshots are important to check for HTML changes, but agree that they aren't the most readable and therefore not as effective for catching content-related bugs (case in point - #2768)

evakill: added more testing! and outsourced [updating the emailNotification tests ](https://github.com/ONEARMY/community-platform/issues/2795) :)

functions/src/emailNotifications/createModerationEmails.ts
```diff
@@ -5,7 +5,16 @@ import { DB_ENDPOINTS, IUserDB } from '../models'
...
 import { getUserEmail } from './utils'
 
-// TODO: trigger this function on howto moderation change
+export const handleHowToModerationUpdate = async (change) => {
+  const howto = change.after.exists ? change.after.data() : null
+  const prevHowTo = change.before.exists ? change.before.data() : null
+  const prevModeration = prevHowTo?.moderation
+  const currModeration = howto?.moderation
+  if (currModeration && prevModeration !== currModeration) {
+    await createHowtoModerationEmail(howto)
```
thisislawatts, line 14: **question:** do we want to release this to all users or do a staggered roll out to beta-users first?

evakill: maybe a good one for @davehakkens to weigh in on - but i would feel comfortable testing the flow on dev and then releasing to all users. the email system has been stable for notifications and the testing we've set up feels comprehensive. 

also, these will only be sent to people who have set their email notification settings in their profile, which is already marked as a beta feature. 
Screen Shot 2023-09-18 at 8 16 52 AM

davehakkens: Not sure if i remember this correct from the meetup. But I thought we would end up with two different mail streams:
1. Notifications emails. These are nice to have notifications for engagement, can be turned off in user profile (see the image above)
2. Transactional emails. These are triggered after certain actions. (map pin approved, how-to etc) You cannot opt out from these. 

Dont really mind if the approuch is different then this, just trying to understand how it will go

evakill: ooh ok - i forgot that we decided there's no opt-out for transactional emails. i can make that change.

do you think the transactional emails should start as a beta feature then?

thisislawatts (approved):

PR #2803, opened 2023-09-23
fix(deps): remove react-icons/all-files

thisislawatts:
PR Checklist

PR Type

## Description

The @react-icons/all-files and react-icons are both related to the react-icons library, which provides icons for React projects. The icons in this library come from various sources like Font Awesome, Material Design, and more.

However, there are reasons to consider not using @react-icons/all-files and to use react-icons directly instead:

Bundle Size:
Importing icons from @react-icons/all-files can potentially lead to larger bundle sizes because it contains every icon from the various icon libraries it supports. This can significantly bloat the size of your JavaScript bundle if not used judiciously.
On the other hand, with react-icons, you can import only the icons you need, ensuring your bundle remains as lightweight as possible.

Easier to Use with Tree Shaking:
Modern bundlers like Webpack and tools like Create React App support tree shaking, which removes unused code from the final bundle. When you use react-icons and import only the icons you need, tree shaking is more efficient.

Better Developer Experience:
It's more intuitive to import only what you need. When using react-icons, you can directly import from specific libraries like react-icons/fa (Font Awesome), react-icons/md (Material Design), and so on. This makes it clear which icon set you are using and can be more maintainable in the long run.
Performance:

Loading unnecessary icons can affect the performance of your app. By only loading the icons we need, you ensure faster load times for your users.


thisislawatts:
Hm… this not be behaving as expected.

thisislawatts:
Closing due to https://github.com/react-icons/react-icons/issues/154, this will require more time and effort than is currently available.

PR #2820, opened 2023-09-27
feat: adds email notification when verified badge added

thisislawatts:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

Screenshot 2023-09-27 at 19 29 43

## Git Issues

Closes #2688

PR #2842, opened 2023-09-30
feat(emails): needs improvements emails

### LINKED ISSUE #2668: [feature request] add moderation feedback display for map pins that 'need improvement' or are 'rejected'
Map pins that have been 'rejected' / need updates before they can be approved need to be able to show any moderation feedback left by admins so that the user can make changes and re-submit.

The feedback comments should be shown in 2 places:
- On the map pin card displayed on the map along with a link to edit their map pin
- On the profile settings page where the map pin is edited it should display the status and if 'rejected' the feedback comments.

Describe alternatives you've considered
N/A

evakill:
PR Checklist

PR Type

## Description

adds needs improvements transactional emails with moderator comments

will wait to merge until #2820 

## Git Issues

Closes #2668

## Screenshots/Videos

Screen Shot 2023-09-30 at 9 19 42 AM
Screen Shot 2023-09-30 at 9 14 13 AM

---

## What happens next?

### DIFF
--- a/functions/src/emailNotifications/templates/index.ts
+++ b/functions/src/emailNotifications/templates/index.ts
@@ -8,6 +8,8 @@ type SupportedEmailTemplates =
   | 'how-to-rejected'
   | 'map-pin-rejected'
   | 'verified-badge-added'
+  | 'how-to-needs-improvements'
+  | 'map-pin-needs-improvements'
 
 export function getEmailHtml(emailType: SupportedEmailTemplates, ctx: {}) {
   const layoutTmpl = Handlebars.compile(
--- a/functions/src/emailNotifications/createModerationEmails.spec.ts
+++ b/functions/src/emailNotifications/createModerationEmails.spec.ts
@@ -2,9 +2,11 @@ import { FirebaseEmulatedTest } from '../test/Firebase/emulator'
 import { DB_ENDPOINTS, IUserDB } from '../models'
 import {
   HOW_TO_APPROVAL_SUBJECT,
+  HOW_TO_NEEDS_IMPROVEMENTS_SUBJECT,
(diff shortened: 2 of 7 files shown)


functions/src/emailNotifications/templates/index.ts
```diff
@@ -8,6 +8,8 @@ type SupportedEmailTemplates =
   | 'how-to-rejected'
   | 'map-pin-rejected'
   | 'verified-badge-added'
+  | 'how-to-needs-improvements'
+  | 'map-pin-needs-improvements'
```
thisislawatts, line 12: **thought:** I wonder if we can auto generate this type based on the files within this directory 🤔

evakill:
```
export function getEmailHtml(emailType: string, ctx: {}) {
  fs.readdir(__dirname, (err, files) => {
    if (err) console.log(err)
    else {
      if (!files.includes(`${emailType}.html`)) {
        throw new Error(`Email template ${emailType}.html does not exist.`)
      }
    }
  })
 ...
```
is this better?

evakill: ooh they have a synchronous method. might be able to generate the type after all

thisislawatts: I like the suggested template name that the type gives us at the moment. I might be overcomplicating things. Although the error handler you've added there should definitely be added! Much more robust implementation.

evakill: > I like the suggested template name that the type gives us at the moment

yeah, i think we'll only get that w the static definition. will keep as is and add the error handling

thisislawatts (approved):

PR #2854, opened 2023-10-01
feat: track firebase extension configuration in code

thisislawatts:
PR Checklist

PR Type

## Description

Captures the configuration in version control so that we can use the CLI to automatically deploy and configure extensions in a consistent way across environments. 

This will require the addition of a `CP_SMTP_CONNECTION_URI` environment variable for each of the [deployment contexts in Circle CI. ](https://circleci.com/docs/contexts/)


.firebaserc
```diff
@@ -8,5 +8,18 @@
...
+  "etags": {
+    "onearmyworld": {
+      "extensionInstances": {
+        "firestore-send-email": "9de00655522653ba435a4fe752115494349ebef10caf2406c211e7d0f64ba790"
+      }
+    },
+    "precious-plastics-v4-dev": {
+      "extensionInstances": {
+        "firestore-send-email": "c04effef87d680a2da28d3db5fd947cc169d2a0bd5bafb8b9a1447eda472f815"
+      }
```
evakill, line 22: does this enable the extension on dev? if so, should we also integrate some protections to avoid actually sending emails to people from dev? (maybe only send to a test email address - could be enforced in a firestore security rule on insert to the `emails` collection)

thisislawatts: @evakill good catch! I was thinking we could use mailtrap.io here to capture all emails sent from the development environments. What do you think?

extensions/.gitignore
```diff
@@ -0,0 +1 @@
+*.env
```
thisislawatts, line 1: The `.env` files contain secrets which we do **not** want to capture in version control.

.firebaserc
evakill: got it - so to confirm - now we'll define the URI for each environment in the CI context, the prod URI will be our brevo one, and the others will be mailtrap. and the extension does not work locally.

extensions/README.md
```diff
@@ -0,0 +1,30 @@
...
+Populate a `.env` based on the `.config` for each extension. This can be done with the following command:
+
+```bash
+envsubst < extensions/firestore-send-email.config > extensions/firestore-send-email.env
+```
+
+Run the following command from the project root directory:
+
+```bash
+firebase deploy --only extensions [--project project_id]
```
evakill, line 29: when might someone use this script? just for testing? and only maintainers with DB access will be able to do this, correct? (might be worth clarifying these points in the README)

evakill (commented):
does anything in CI need to be updated other than the variables? what's the plan to test the deploy process?

.firebaserc
thisislawatts: That sounds right to me :)

thisislawatts:
> does anything in CI need to be updated other than the variables? what's the plan to test the deploy process?

I expect something _may_ need changing in the deploy step but not sure. Hoping it gets picked up by the current command:

```yaml
# Use version of firebase-tools already installed in functions workspace to deploy
name: Deploy to Firebase
command: ./functions/node_modules/.bin/firebase deploy --debug -P << parameters.alias >> --non-interactive
```

https://github.com/ONEARMY/community-platform/blob/master/.circleci/config.yml#L189

extensions/README.md
thisislawatts: Added in: https://github.com/ONEARMY/community-platform/pull/2854/commits/ca8b92d55ac74049da12371f0f4c6b737c407851

evakill (approved):
cool let’s try!

Issue #2892, opened 2023-10-10
HT Publishing bug


thisismattia:
**Describe the bug**
I was trying to clean up this HT https://community.preciousplastic.com/how-to/build-a-flipflopi-boat

**Expected behaviour**
I made the changes and clicked "publish" under the assumption that I was saving a new version of the HT.

**Bug**
However it looks like the HT has been published again. New publish date + HT count gone to zero from 15 😞 

<img width="1274" alt="image" src="https://github.com/ONEARMY/community-platform/assets/17761877/cfeac463-3a82-46d6-bb5f-ea4805a97c22">

Any insights on what happened?

AlfonsoGhislieri:
Could you just clarify what you mean with:

> HT count gone to zero from 15

The total count of HowTos created onn their profile has gone down to 0?

thisismattia:
Before I edited the HT the useful count was 15. When I saved it went to zero.

thisismattia:
This is happening again, @davehakkens seems pretty serious bug 😬
<img width="1302" alt="image" src="https://github.com/ONEARMY/community-platform/assets/17761877/89a20806-ffbd-4fab-9ef7-95109420121a">

thisislawatts:
@thisismattia were you logged in with the `the-flipflopi` user when making the edits or as an Admin/Moderator user?

thisislawatts:
Confirming that I have been able to recreate this behaviour locally. I am seeing that _updates_ made to a How-to result in: 
1. the `_created` property holds the timestamp the document was modified. 
2. Any `voted_useful` data is dropped from the document during the update. 

Next steps:
* Identify source of the bug and implement a fix
* Data recovery, identify when bug was introduced and work backwards through automated backups to identify missing data.

evakill:
@thisismattia using a local migration script, i was able to scrape useful votes for how-tos that had been overwritten. we could not get 100% of the data back, but we restored **121 useful votes for 13 how-tos**, including the one in your original post. the data suggests this has been a problem since the beginning of September. in case it is useful, here is the record of all the data we recovered: 
```
{
  "0jahBYgJtpDadwhU58lf": {
    "id": "0jahBYgJtpDadwhU58lf",
    "name": "El Tornillo Motor Injection Machine",
    "lastModified": "2023-10-21T10:11:43.354Z",
    "votedUsefulBy": ["preciousplasticukraine", "mattia"],
    "updatedUsefulBy": [
      "preciousplasticukraine",
[...]

PR #2939, opened 2023-10-21
feat: restore useful votes with useful migration

evakill:
PR Checklist

PR Type

## Description

the goal of this function is to restore dropped useful vote data from this bug: #2892 
it creates a https get endpoint that can be called to do this one-time update. the update restores data by:
1. iterating through how to documents (currently 330 in PP prod).
2. querying the user collection to form a list of all users that have marked this how-to useful, per the `votedUserfulHowtos` field.
a. note this field has not been updated since June 2023 ([PR](https://github.com/ONEARMY/community-platform/pull/1607/files)). we won't be able to retrieve all dropped data, but hopefully a good chunk of it. 
3. merge this list with the existing votes, checking for a unique list.
4. update the `votedUsefulBy` list on the howto.

i tested this locally using the emulators and spot-checked documents i expected to be updated.

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?


functions/src/database/usefulMigration.ts
```diff
@@ -0,0 +1,62 @@
...
+import { DB_ENDPOINTS } from 'oa-shared'
+import { db } from '../Firebase/firestoreDB'
+
+export const updateHowToUsefulField = functions
+  .runWith({ memory: '512MB' })
+  .https.onRequest(async (_, res) => {
+    console.log('Starting useful field update')
+
+    const howtosRef = db.collection(DB_ENDPOINTS.howtos)
+    const usersRef = db.collection(DB_ENDPOINTS.users)
```
thisislawatts, line 11: **question:** Did you consider fetching all of the user records up front and process the data then update any how-tos rather than querying user records on each  how-to?

functions/src/database/usefulMigration.ts
```diff
@@ -0,0 +1,62 @@
...
+              .catch((error) => {
+                console.error('Useful update: Error updating howto:', error)
+              })
+          }
+        })
+      })
+      .catch((error) => {
+        console.error('Useful update: Error getting howtos:', error)
+      })
+    res.send('Finished useful field update')
```
thisislawatts, line 61: **praise:** This looks great! I am curious what you think of setting up a script that takes connection details rather than deploying a function? This would reduce the overhead as it _could_ be run on a local developer machine or as part of the deployment process.

functions/src/database/usefulMigration.ts
evakill: yes. thought this would be the cheaper approach. if we are paying for # of records returned and query the entire user DB, that's 18k records. i'm almost certain that is much more than the total # of useful votes (would be an average of 55 votes per how to)

evakill: so maybe a little computationally slower, but more cost- and memory-efficient.

functions/src/database/usefulMigration.ts
evakill: yeah, agree that would be preferable and i was originally attempting to do this. my initial attempts failed and i gave up haha. i was trying to run `node functions/src/database/usefulMigration.ts` and got some module import errors etc. do you know if we have an example of a local script anywhere in the code? 

if we don't feel like pushing this, i can also deploy to prod, run, then take down. not the cleanest, but it gets the job done.

functions/src/database/usefulMigration.ts
thisislawatts: Gotcha, works for me :)

functions/src/database/usefulMigration.ts
evakill: well, was trying to deploy a version that logged out the updates so i could test in prod - but found i don't have the permissions:
```
Error: Missing required permission on project onearmyworld to deploy new HTTPS functions. The permission cloudfunctions.functions.setIamPolicy is required to deploy the following functions:

- database-usefulMigration-updateHowToUsefulField

To address this error, please ask a project Owner to assign your account the "Cloud Functions Admin" role at the following URL:

https://console.cloud.google.com/iam-admin/iam?project=onearmyworld
```
not sure if you can do this or if it has to be @davehakkens 

or we switch to the script approach

thisislawatts: I think you want to use `ts-node` to execute the files so that the code gets compiled from Typescript to JS before execution.

You could use a file like this, where the only dependencies are firebase-admin and a service account json.

```ts
/* eslint-disable no-undef */
import * as admin from "firebase-admin";
// The path to the credential file could also be passed through as an argument for the script
import * as serviceAccount from './path-to-your-firebase-adminsdk.json';  // Make sure to download your Firebase admin SDK JSON file and provide its path here.

admin.initializeApp({
  credential: admin.credential.cert(serviceAccount as admin.ServiceAccount)
});

const db = admin.firestore();


(async () => {
  console.log("Starting useful field update");

  const howtosRef = db.collection('v3_howtos');
  const usersRef = db.collection('v3_users');

  await howtosRef
    .get()
    .then((howtoSnapshot) => {
      howtoSnapshot.forEach(async (howto) => {
        const votedUsefulBy = [];
        await usersRef
          .where(`votedUsefulHowtos.${howto.id}`, "==", true)
          .get()
          .then((querySnapshot) => {
            querySnapshot.forEach((userDoc) => {
              votedUsefulBy.push(userDoc.id);
            });
          })
          .catch((error) => {
            console.error("Useful update: Error getting users:", error);
          });

        if (votedUsefulBy.length) {
          const currentVotedUsefulBy = howto.data().votedUsefulBy ?? [];

          console.log("updating how to", {
            id: howto.id,
            "new-voted-useful-count": votedUsefulBy.length,
            "previous-voted-useful-count":
              howto.data().votedUsefulBy?.length ?? 0,
            "new-votedUsefulBy": votedUsefulBy,
            "previous-votedUsefulBy": currentVotedUsefulBy,
          });

          // Use the set to remove duplicates
          const updatedUsefulBy = [
            ...new Set(currentVotedUsefulBy.concat(votedUsefulBy)),
          ];

          await howtosRef
            .doc(howto.id)
            .update({
              votedUsefulBy: updatedUsefulBy,
            })
            .catch((error) => {
              console.error("Useful update: Error updating howto:", error);
            });
        }
      });
    })
    .catch((error) => {
      console.error("Useful update: Error getting howtos:", error);
    });
})();
```

evakill: oooh amazing, thanks. lemme try this out

evakill:
@thisislawatts i used your recommended script and it worked great! pushed up the final code i used for the migration. not sure it needs to be merged to master, but at least we have a record of it in this PR. you can see the updates made in `results.json`. **we saved 121 useful counts for 13 how to's 🥳**

PR #2947, opened 2023-10-24
feat: add profile contact form

### LINKED ISSUE #2607: [feature request] add ability to directly contact a profile
### Description

We want to add a contact form to user profiles. This is to make it easier for the community to get in touch and collaborate with eachother. (this is our lightweight alternative for not having a direct messaging system)

### Page related

Will be used in : 
- User [profiles](https://community.preciousplastic.com/u/precious-plastic)
- Profile [settings](https://community.preciousplastic.com/u/precious-plastic/edit)

### Mockup

1. Add contact button to profiles
1_button

2. Overlay with input fields to send a message
2_pop-up

3. User can disable the direct contact feature in their profiles
3_usersettings

### How does the flow works.
1. Visitor clicks contact 
2. Visitors writes their name, email and a message. 
3. Send
4. A confirmation email is send to visitors/sender (using the email from the form)
5. An email with all the information is send to the Profile (using their auth email)
6. Optional:  Set daily limits. There is a daily limit of one message per profile per visitor, or max 5 messages per day (to avoid spam)

### Examples
The behaviour is based on the on from [Moneybird](https://www.moneybird.nl/experts/zoeken/)
You can search for profiles, press contact, and send them a message. (they receive a mail)

### Possible email templates

**Message to visitor (sender)**
From: noreply@instance.com
Subject: We send your message!
Message: We send the following message to [insert profile username]: [insert message from sender]

**Message to the Profile (receiver)**
From: noreply@instance.com
Subject: Hey you got a new message from [insert name from sender]
Message: Hey you got a new message from [insert name from sender]: [insert message from sender] Make sure to reply to their email adres, [insert email from sender]

benfurber:
PR Checklist

PR Type

## Description

This adds:
### 1 - Messaging opt-in for profiles
An option in the setting pages to opt-in to receiving messages from users.

### 2 - Contact form
If a profile has opted-in. A contact tab and contact form will the accessable on their profile page to logged in users.

### 3 - Message document collection
When a contact form submission is made. A new document is created in the newly added messages collection. An email address is able to send a maximum of 100 emails.

### 4 - Sending of messages
On creation of a message document, an email (in the existing style) with the message is sent to the profile and the user.

## Git Issues

Closes #2607

## Screenshots/Videos

Screenshot 2023-10-24 at 12 31 06
Screenshot 2023-10-24 at 12 28 54
Screenshot 2023-10-25 at 20 19 40
Screenshot 2023-10-25 at 20 20 15


src/modules/index.ts
```diff
@@ -4,6 +4,7 @@ export enum MODULE {
   CORE = 'core', // This is enabled on all installations
   HOWTO = 'howto',
   MAP = 'map',
+  MESSAGE = 'message',
```
thisislawatts, line 7: **Question:** Is this a standalone module or is it fundamental to interactions between Users so instead we should consider it a standard capability of Users rather than an optional extra?

WDYT @davehakkens

davehakkens: Would indeed see this as a standard feature of the platform for all group profiles.

davehakkens:
Logo/image doesnt seem to load in the email..
afbeelding

benfurber:
> Logo/image doesnt seem to load in the email.. afbeelding

Sorry @davehakkens I forgot to say, this is just because it's taken from the dev environment - it calls the same notification template as every other notification, so if that's working fine on prod for all the others, it'll be working fine for this.

functions/src/emailNotifications/createSubmissionEmails.ts
```diff
@@ -1,11 +1,36 @@
...
+  getReceiverMessageEmail,
+  getSenderMessageEmail,
+} from './templates'
 import { getUserAndEmail } from './utils'
 import { withErrorAlerting } from '../alerting/errorAlerting'
 
+export async function createMessageEmails(message: IMessageDB) {
+  const { _id, email, toUserName } = message
+  const { toUserEmail } = await getUserAndEmail(toUserName)
+
```
thisislawatts, line 17: **issue(blocking):** We need to include some additional checks here before writing to the emails collection.
1. The `toUserEmail` is associated with an actual User within the platform. Otherwise anyone with direct write access to the `messages` collection will be able to send emails.
2. The User record that is being referenced by the `toUserEmail` has the expected value in `isContactableByPublic`
3. The User who created the Message has not exceeded the upper limit on number of messages sent. This control could be bypassed if a User directly manipulated the Messages store.

src/modules/index.ts
thisislawatts: Sounds good to me, so lets remove the module wiring @benfurber

src/pages/User/contact/UserContactForm.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { observer } from 'mobx-react'
+
+import { useCommonStores } from 'src/index'
+import { contact } from 'src/pages/User/labels'
+import {
+  UserContactError,
+  UserContactFieldEmail,
+  UserContactFieldMessage,
+  UserContactFieldName,
+} from './'
```
thisislawatts, line 14: **question:** My preference here would be using direct imports from the file than loading rather than via an index file. What value does that provide?

thisislawatts (changes requested):

functions/src/emailNotifications/createSubmissionEmails.ts
benfurber: [How's this](https://github.com/ONEARMY/community-platform/pull/2947/files#diff-c44dbba562fdb839e6e8a3bbf2d810e55e362a5a4d3d456cbac8bf4328926ddcR43) @thisislawatts?

functions/src/emailNotifications/utils.ts
```diff
@@ -40,6 +41,39 @@ export const getUserAndEmail = async (userName: string) => {
   return { toUser, toUserEmail }
 }
 
+export const isValidEmailCreationRequest = async ({
+  email,
+  toUserName,
+}: IMessageDB) => {
+  const user = await getUserAndEmail(email)
+  if (user.toUserEmail !== email) {
+    const error = new Error(`'${email}' attempted to send a message but isn't a user`)
```
thisislawatts, line 50: **issue:** Can we avoid including dynamic attributes in the Error name? Static error messages are easier to search for in codebases and logs. If we are looking through logs trying to diagnose an issue, having a consistent error message helps. 

A consistent error message also allows us to add some alerting if we wanted to. For example if there are X instances of this error in Y minutes we want to raise an alert as it suggests a sustained attack.

functions/src/emailNotifications/utils.ts
```diff
@@ -40,6 +41,39 @@ export const getUserAndEmail = async (userName: string) => {
...
+  }
+
+  const { docs } = await db
+    .collection(DB_ENDPOINTS.messages)
+    .where('email', '==', email)
+    .get()
+  if (docs.length >= EMAIL_ADDRESS_SEND_LIMIT) {
+    const error = new Error(`User email '${email}' has exceeded email message count`)
+    sendErrorAlert('validEmailCreationRequest', error)
+    return false
```
thisislawatts, line 62: **suggestion:** As we have this method wrapped in `withErrorAlerting`, I believe all we need to do is throw an error and let the error handler log it. 

```suggestion
    throw new Error(`User has exceeded email message count`)
```

src/stores/Message/message.store.tsx
```diff
@@ -20,8 +20,11 @@ export class MessageStore extends ModuleStore {
   @action
   public async upload(values: IMessageInput) {
     const messagesByEmailer = await this.messagesByEmailer(values.email)
-    if (messagesByEmailer.length > EMAIL_ADDRESS_SEND_LIMIT) {
-      return Promise.reject(new Error('Too many messages sent'))
+    if (messagesByEmailer.length >= EMAIL_ADDRESS_SEND_LIMIT) {
+      const error = new Error('Too many messages sent')
+
+      logger.error(`${error} by ${values.email}`)
```
thisislawatts, line 26: **suggestion(blocking)** Can we avoid using dynamic variables in the logger and instead pass them through as a separate argument?

Although I am wondering about whether we should log this email address as it represents PII which will mean if we get a GDPR request we would need to purge our logs 😬

thisislawatts (approved):

thisislawatts (approved):

thisislawatts:
@benfurber before merging can you rebase/squash the history so that these commits are merged into a single releasable change. This will ensure a single meaningful message when generating the release notes. 
https://github.com/ONEARMY/community-platform/releases

PR #3000, opened 2023-11-13
fix: hide member mappins on precious plastic

### LINKED ISSUE #2927: [feature request] Hide "member" mappins by default on PP
The map is getting crowded (cluttered) which makes it harder for workspaces to pop out + slow loading times.

afbeelding

We would like to hide the "want to get started" pins by default on [Precious Plastic map](https://community.preciousplastic.com/map)

**Notes**
This is a specific setting for Precious Plastic with a mature community.
Other projects (instances) probably dont want to hide this in their early stages. 
So ideally it's a setting that can be configured per instance.

benfurber:
PR Checklist

PR Type

## Description

A more robust attempt at trying to hide member pins by default on PP. This fixes two problems:
1. While it worked locally, how the platform version is selected for production builds is different.
2. The pins were hidden completely and couldn't be filtered if a user wanted to.

I tried first to apply the logic necessary to the display components but I didn't find an approach that worked so I applied it to the store logic instead.

## Git Issues

Closes #2927


benfurber:
I'm getting a different behaviour on the preview branch than locally. Reverting to a draft.

davehakkens:
Thanks for the try @benfurber. Feel free to leave it for someone else if it gets to complex. Can be challenging with all the instances and configurations

benfurber:
@davehakkens Thanks for the offer Dave. The issue yesterday was just some incomplete logic I'd written. The behaviour is what is required now, just the issue of checking my new approach to determining PP for production...

src/utils/isPreciousPlastic.ts
```diff
@@ -0,0 +1,6 @@
+export const isPreciousPlastic = (): boolean => {
+  return (
+    (process.env.REACT_APP_PLATFORM_THEME ||
```
thisislawatts, line 3: **question:** Can we load this from src/config rather than accessing the process.env variables directly. I am keen to standardise on src/config for sourcing and determining configuration.

Is there a configuration option we could introduce for the map module that would introduce this feature using a configuration parameter rather than making inferences based on the theme.

For example:
* MAP_PROFILE_TYPES_HIDDEN_BY_DEFAULT

benfurber: I've moved to config. Only change from your suggestion is that as only a single value is required now, I've kept it a single value.

davehakkens:
Should be released but member pins are still here.. @benfurber 
Or am I missing something?
afbeelding

benfurber:
@davehakkens Next time a build is release can you please check this again? It turns out that `REACT_APP_PLATFORM_THEME` wasn't being set as an environment variable for the production build of PP.

PR #3015, opened 2023-11-22
fix: handle empty images

thisislawatts:
PR Checklist

PR Type

## Description

Introduce tests for rendering User profile pages, introducing edge case where User has no Cover Images uploaded. Adds graceful fallback for this scenario.

Note the screengrabs below are for two different users, this is due to the bug being visible in prod (before) then fixed locally in (after).

Before:
Screenshot 2023-11-22 at 20 46 29

After:
Screenshot 2023-11-22 at 20 46 19

### DIFF
--- a/src/pages/User/content/SpaceProfile.tsx
+++ b/src/pages/User/content/SpaceProfile.tsx
@@ -22,7 +22,15 @@ import { useCommonStores } from 'src/index'
 import { ProfileType } from 'src/modules/profile/types'
 import { UserContactForm } from 'src/pages/User/contact'
 import { formatImagesForGallery } from 'src/utils/formatImageListForGallery'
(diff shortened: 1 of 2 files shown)


davehakkens:
Are you aware we did have a fallback image in case user didnt upload?

thisislawatts:
@davehakkens I think that is true for the member page, but this is for a workspace User.

Use desktop browser to see example of the bug here:
https://community.preciousplastic.com/u/metamorfica-lab

davehakkens:
hmm... I see.

There just to be a placeholder image there for group profile headers saying something like: "sorry no image uploaded yet. booooring" But I was just searching the reposity and cant seem to find the file. So happy to use this one!

benfurber (approved):

PR #3016, opened 2023-11-26
feat(patreon): set up auth with patreon

evakill:
PR Checklist

PR Type

## Description

this PR sets up a basic "Log in with Patreon" button on the platform

### Implementation
Follows the steps in the docs here: https://docs.patreon.com/#oauth
1. Create keys, set redirect uris to `http(s)://<host>/patreon`

2. Add keys to the firebase config (functions) and env variables (React app) to securely access in app

3. Create button in the user settings page (is this the right place? @davehakkens)
Screen Shot 2023-11-25 at 9 46 10 PM

4. Button redirects to patreon's login auth flow
Screen Shot 2023-11-25 at 9 49 47 PM

5. Patreon's auth flow redirects back to the platform with the path `/patreon` and URL param with a single use token. React app renders a loading screen while the page sends a HTTP request to our serverless functions with the token sent back from Patreon.
Screen Shot 2023-11-25 at 9 50 05 PM
6. The function hits the patreon API auth endpoint with the single-use token. The Patreon API sends back an `access_code` that can be used to hit other parts of the Patreon API.
7. Hit the patreon API with the `access_code` to fetch details about a user, then update our user in the database to include the auth tokens and user info. Now our CP user is "integrated" with the details from the linked patreon account.
Screen Shot 2023-11-25 at 10 39 27 PM

#### UI
1. Is the login button in the right place? 

#### Technical
1. Do we need to store the patreon client ID in a React env variable? 
  i. Is there a better place for it? Is it even that sensitive that we need an env variable?
3. Do we expect this to be testable in the local environment? 
  i. If so, we need a method for granting access to secrets. Maybe [google secret manager](https://cloud.google.com/secret-manager)? Not sure if anyone else has preferences for this
4. How often do we need to refresh the patreon's user data? On each login? 

### Next steps

- [x] Sort out environmental variables (see [comment](https://github.com/ONEARMY/community-platform/pull/3016#issuecomment-1826940206))
- [ ] Figure out refresh strategy
- [ ] Make frontend updates 

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -559,6 +559,7 @@ workflows:
           NOTIFY_SLACK: true
           context:
             - circle-ci-slack-context
+            - circle-ci-patreon-context
             - community-platform-dev
       - deploy:
           name: 'Deploy: dev.community.projectkamp.com'
@@ -569,6 +570,7 @@ workflows:
           NOTIFY_SLACK: true
           context:
             - circle-ci-slack-context
+            - circle-ci-patreon-context
             - project-kamp-dev
       - deploy:
           name: 'Deploy: dev.community.fixing.fashion'
@@ -579,6 +581,7 @@ workflows:
           NOTIFY_SLACK: true
           context:
             - circle-ci-slack-context
+            - circle-ci-patreon-context
             - fixing-fashion-dev
       #---------------------- Development Instances Build and Deploy ----------------------
       - release:
@@ -601,6 +604,7 @@ workflows:
           NOTIFY_SLACK: true
           context:
             - circle-ci-slack-context
+            - circle-ci-patreon-context
             - fixing-fashion-prod
       - deploy:
           name: 'Deploy: community.preciousplastic.com'
@@ -611,6 +615,7 @@ workflows:
           NOTIFY_SLACK: true
           context:
             - circle-ci-slack-context
+            - circle-ci-patreon-context
             - community-platform-production
       - deploy:
           name: 'Deploy: community.projectkamp.com'
@@ -621,4 +626,5 @@ workflows:
           NOTIFY_SLACK: true
           context:
             - circle-ci-slack-context
+            - circle-ci-patreon-context
             - project-kamp-production
--- a/functions/src/Integrations/patreon.ts
+++ b/functions/src/Integrations/patreon.ts
@@ -2,6 +2,7 @@ import * as functions from 'firebase-functions'
 import { DB_ENDPOINTS } from 'oa-shared'
 import { db } from '../Firebase/firestoreDB'
 import { CONFIG } from '../config/config'
+import { IUserDB } from '../models'
 
 const PATREON_CLIENT_ID = CONFIG.integrations.patreon_client_id
 const PATREON_CLIENT_SECRET = CONFIG.integrations.patreon_client_secret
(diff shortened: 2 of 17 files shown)


davehakkens:
Nice Eva. Good to see this moving forward!  🔥
When testing the preview link i get this error. Should it be tested in preview, locally or emulator?
afbeelding

Made this sketch a while back to explain where the location should be to connect. (profile settings)
The content in there is not final. Probably better to wait until we know the exact steps to connect Patreon
afbeelding

(Probably be an interesting PR for you to check @AlfonsoGhislieri )

evakill:
> When testing the preview link i get this error. Should it be tested in preview, locally or emulator?

yeah that is expected, this relates to the question about config variables. i can give you the steps to test locally with the emulator, but we don't have a process to expose secret keys (like the client id) right now

cool, design looks good. i think best to make an issue for that once we've settled the technical details in this PR! will wait for input from @AlfonsoGhislieri  and maybe also @thisislawatts

evakill:
update on variable strategy: 

frontend environmental variable `REACT_APP_PATREON_CLIENT_ID` has been stored in the `circle-ci-patreon-context`. i added the context to the deploy steps of all environments. this enables the frontend redirect when you hit the "Login with Patreon" button. 

the backend variables `patreon_client_id` and `patreon_client_id` i have added to firebase configs and deployed to all environments. this will enable the backend patreon API requests.

for the preview branch, i want to add the client ID as a github secret and load it into the [deploy action](https://github.com/ONEARMY/community-platform/blob/master/.github/workflows/pr-preview.yml) as we do with several other FE env variables. the preview branch will have access to the config variables on the backend through firebase.

this means that the integration will work on all deployed environments!

locally, the integration will not work because the local env doesn't have access to these config variables. i think this is OK. our discord and slack integrations also don't work locally. if someone needs to, those with firebase access can pull the client id and client secret from the firebase config and bootstrap their variables.

i think this should work, let me know if anyone has concerns!

.circleci/config.yml
```diff
@@ -559,6 +559,7 @@ workflows:
           NOTIFY_SLACK: true
           context:
             - circle-ci-slack-context
+            - circle-ci-patreon-context
```
thisislawatts, line 562: **question:** Why do we want to use the same CircleCI context for **all** environments? Would it be better to isolate and create a distinct Patreon app/connection for each instance:

I am thinking this would be useful to:
* Reduce the impact if one of the connection credentials are compromised.
* Improve the user experience with meaningful descriptions/icongraphy that relates to the instance they are attempting to sign into:

functions/src/Integrations/patreon.ts
```diff
@@ -0,0 +1,99 @@
...
+  )
+    .then((res) => res.json())
+    .then(async ({ access_token, refresh_token, expires_in }) => {
+      const patreonUser = await getCurrentPatreonUser(access_token)
+      try {
+        await db
+          .collection(DB_ENDPOINTS.users)
+          .doc(uid)
+          .update({
+            patreon: {
```
thisislawatts, line 76: **issue(blocking):** It is worth noting that any properties stored on a User record are currently available to _anyone_ accessing the site. With an `access_token`, the application can access the user's data and perform actions according to the granted scopes. If this token is compromised, an attacker could gain the same level of access to the user's data as the legitimate application.

We should look into an alternative place we can persist this data and ensure the appropriate firestore rules are in place to protect the data so only the author can read/write to them, rather than the **very** liberal access definition we have at the moment: https://github.com/ONEARMY/community-platform/blob/master/firestore.rules

This may be a new collection entitled, `user_integrations` with something like the folllowing schema:
```ts
{
  user_id: string (link to User record)
  user_auth_id: string (link to Authentication user)
  patreon: {
    accessToken: access_token,
    refreshToken: refresh_token,
    expiresAt: expires_in,
  }
}
```

I do not think that this level of protection is needed for the patreonUser property.

.circleci/config.yml
evakill: all instances are under the one army patreon, so they use the same credentials! https://www.patreon.com/one_army

functions/src/Integrations/patreon.ts
evakill: good call! making this adjustment

evakill:
@thisislawatts sorry for how bloated this PR has become 😵

updates since first review (last 3 commits):
- add `user_integrations` collection and write access codes to that
- define security rules for the `user_integrations` collection and update test
- add info and development tips to readme

making another PR for model and FE changes!

thisislawatts (approved):

PR #3043, opened 2023-12-02
feat(question): select tags

### LINKED ISSUE #2964: [QuestionCreate] Select tags
It should be possible to define the tags which are appropriate for the content based on a predefined set of tags. Look at Research/How-to for an example.

Stretch. Refactor our current implementation to remove duplicate collections.

onim-at:
PR Checklist

PR Type

## Description

- Refactor question form to have a single form component for both create and edit question
- Fix bug in the questionEdit component. The edit page redirected to the view question page even if the user is not admin
  - removed _isLoading_ dependency from useEffect to avoid double rendering
  - redirects to _questions_ page if the user navigates to a non existing _question/{slug}/edit_
- Select tags

## Git Issues

Closes #2964 

## Screenshots/Videos

https://github.com/ONEARMY/community-platform/assets/45094836/8293c570-461e-4a1b-8206-08a1300b6f0f

---

## What happens next?


evakill (commented):
looking good! i noticed while testing some unexpected tags are showing up in the dropdown. only tags that include `question` in their `categories` field should appear (which is probably 0 in the database right now). 

this looks like an existing bug in the tags store - when a category is set, [the filtered tags aren't updated properly. ](https://github.com/ONEARMY/community-platform/blob/98eb009c7c7c098987e75991957cc20bde99df5e/src/stores/Tags/tags.store.tsx#L26)

should be an easy fix we can slip into this PR! and if you want to create tags in the database for testing, would recommend running locally with the emulators and adding tag objects manually through the emulator UI (http://localhost:4001)

evakill (approved):
thanks for the bug fix :)

PR #3044, opened 2023-12-02
fix: select component highlights focused element on keyboard navigation

onim-at:
PR Checklist

PR Type

## Description

The select component didn't highlight correctly the focused item on keyboard navigation

## Git Issues

Closes #

## Screenshots/Videos

https://github.com/ONEARMY/community-platform/assets/45094836/31ed5e81-69ba-4e12-8f0c-458eb6858ca0

https://github.com/ONEARMY/community-platform/assets/45094836/1d05eb47-7001-4e47-9785-a37a6dafb045

---

## What happens next?


evakill:
nice catch @onim-at! could you take a look at this @davehakkens? want to make sure you like the highlight color :)

packages/components/src/Select/Select.tsx
```diff
@@ -47,17 +47,14 @@ export const Select = (props: Props) => {
       },
     }),
 
-    option: (provided) => ({
+    option: (provided, state) => ({
       ...provided,
-      backgroundColor: theme.colors.background,
+      backgroundColor: state.isFocused
```
evakill, line 52: i think we should change color on hover rather than focus - if the mouse exits the dropdown, the highlight should be removed

onim-at: I tried to make it work but didn't manage, the `state: OptionProps` doesn't contain a field that cover this case and the `:hover` is completely overridden by the `state.isFocused` condition. 
I see that both standard html `select` and `react-select` have this behavior that keeps the field highlighted when the mouse exits the dropdown, what do you suggest to do?

evakill: gotcha - if it isn't straightforward then i say let's go with what you wrote originally! thanks for checking it out

evakill (approved):

Issue #2962, opened 2023-10-29
[QuestionPage] Comments


thisislawatts:
Use `Discussion` module once it has been merged https://github.com/ONEARMY/community-platform/issues/2564

evakill:
reminder to also add updates for comments made on questions when a user's username or country changes; see discussion [here](https://github.com/ONEARMY/community-platform/pull/3053#pullrequestreview-1785261150)

thisislawatts:
Marking as closed here as Comments are now available in Questions
https://dev.onearmy.world/questions

PR #3053, opened 2023-12-06
feat(question): add username and Ask-On date

### LINKED ISSUE #2958: [QuestionPage] Show Username

goratt12:
PR Checklist

PR Type

## Description

Add username and Ask-On date to question card
The code relies on the `creatorCountry` property of question, this property wasn't populated before this PR so only new questions will show county flags (older questions will show grey flag)

## Git Issues

Closes #2958

## Screenshots/Videos

image
image

---

## What happens next?


evakill (changes requested):
hiya @goratt12!

there are 2 things that i would love to see happen in this PR - let me know what you think
 
1. we store `creatorCountry` on the resources because it saves us a DB lookup for each load. however, we need to make sure we keep it up to date if the user changes location. we do this with a firebase function triggered on user update. when the user's country changes, we call a function to update all of their posts with their updated country. you can find the code for this [here](https://github.com/ONEARMY/community-platform/blob/master/functions/src/userUpdates/index.ts#L66), and we'll need to add some logic to do the same update on questions as we do on research and howto.
2. this little flag + username + published on would be a great candidate for a global component, as it is used across all our modules. we don't have a ton of these, but a good example i can think of is `SortFilterHeader`. would you be game for converting this into a reusable component and adding it to How to, Research, and Questions? i think it would be a step in the right direction in terms of reusability!

goratt12:
> hiya @goratt12!
> 
> there are 2 things that i would love to see happen in this PR - let me know what you think
> 
> 1. we store `creatorCountry` on the resources because it saves us a DB lookup for each load. however, we need to make sure we keep it up to date if the user changes location. we do this with a firebase function triggered on user update. when the user's country changes, we call a function to update all of their posts with their updated country. you can find the code for this [here](https://github.com/ONEARMY/community-platform/blob/master/functions/src/userUpdates/index.ts#L66), and we'll need to add some logic to do the same update on questions as we do on research and howto.
> 2. this little flag + username + published on would be a great candidate for a global component, as it is used across all our modules. we don't have a ton of these, but a good example i can think of is `SortFilterHeader`. would you be game for converting this into a reusable component and adding it to How to, Research, and Questions? i think it would be a step in the right direction in terms of reusability!

Hey! both sound good, I'll get on it :)

goratt12:
I'm not sure I understand how [this function](https://github.com/ONEARMY/community-platform/blob/a65cce28a29c45a5692020372c2c253638b2f394/functions/src/userUpdates/index.ts#L66) is called.
I see that the function is called from this code: https://github.com/ONEARMY/community-platform/blob/a65cce28a29c45a5692020372c2c253638b2f394/functions/src/userUpdates/index.ts#L20-L26

Could you explain what how this works? 
How does the `id` in the ``${DB_ENDPOINTS.users}/{id}`` is changed and create a listener for every user?

goratt12:
I think we should also update all the user comments, currently only on how-to but in the future also for question comments

evakill:
> I'm not sure I understand how [this function](https://github.com/ONEARMY/community-platform/blob/a65cce28a29c45a5692020372c2c253638b2f394/functions/src/userUpdates/index.ts#L66) is called. I see that the function is called from this code:
> 
> https://github.com/ONEARMY/community-platform/blob/a65cce28a29c45a5692020372c2c253638b2f394/functions/src/userUpdates/index.ts#L20-L26
> 
> Could you explain what how this works? How does the `id` in the `${DB_ENDPOINTS.users}/{id}` is changed and create a listener for every user?

yep - firebase will run this function every time a user is updated. it sounds like a lot, but somehow under the hood firebase optimizes for this 🤷‍♀️

goratt12:
What defines Firebase to set a listener for every user?
how does new users are handled?

evakill:
> I think we should also update all the user comments, currently only on how-to but in the future also for question comments

yes good call! the function that does that is `updateUserCommentsInfo` in the same fire - you might have seen it - and would be great to add the updates for question comments in there as well

evakill:
> I think we should also update all the user comments, currently only on how-to but in the future also for question comments

yes good call! the function that does that is `updateUserCommentsInfo` in the same fire - you might have seen it - and would be great to add the updates for question comments in there as well

> What defines Firebase to set a listener for every user? how does new users are handled?

here are the docs on the triggers - https://firebase.google.com/docs/functions/firestore-events 

we use `onUpdate` and `onCreate` quite a bit in the functions directory. every export from `functions/src/index.ts` is deployed as a firebase function. there are different types: triggers, scheduled functions, and HTTP requests, that can all be deployed with firebase functions

goratt12:
> > What defines Firebase to set a listener for every user? how does new users are handled?
> 
> here are the docs on the triggers - https://firebase.google.com/docs/functions/firestore-events
> 
> we use `onUpdate` and `onCreate` quite a bit in the functions directory. every export from `functions/src/index.ts` is deployed as a firebase function. there are different types: triggers, scheduled functions, and HTTP requests, that can all be deployed with firebase functions

Oh I see, thanks for the explanation!

goratt12:
> > I think we should also update all the user comments, currently only on how-to but in the future also for question comments
> 
> yes good call! the function that does that is `updateUserCommentsInfo` in the same fire - you might have seen it - and would be great to add the updates for question comments in there as well

I see that there is a Question Comments issue currently blocked - #2962.
The issue is blocked because there is a Discussion module coming up 🥳

I think we should wait with the comments updating for this PR and handle it in the Discussion module
What do you think?

evakill:
> #2962

ah yes - sounds reasonable. i will add a comment to that issue as a reminder

evakill (approved):
everything looks great!! thanks @goratt12 🥳

PR #3074, opened 2023-12-14
feat: change to save only minimal impact fields

benfurber:
PR Checklist

PR Type

## Description

Previously, fields for impact data was being stored that was about presentation only and thus wrongly tied presentation to stored data (like a label) that might easily need changing.


benfurber:
@evakill I think this balances not saving presentational data while also being realistic that this is only for PP right now.

evakill (approved):

PR #3085, opened 2023-12-21
feat: update comments ui for how-tos

benfurber:
PR Checklist

PR Type

## Description

This is the first step in updating the design of all comments across the platform leading to nested comments. This updates the design of the current how-to comments data structure.

This is the end state we're aiming for: [Figma Designs](https://www.figma.com/file/In2EriHbHulWjq6Cb1Mdfm/create-how-to---upgrades?type=design&node-id=102-357&mode=design)

## Screenshot
Screenshot 2023-12-21 at 15 32 45


packages/components/src/Button/Button.tsx
```diff
@@ -97,7 +98,9 @@ export const Button = (props: BtnProps) => {
       sx={{
         ...props.sx,
         ...getSizeProps(size, !!props.icon),
-        ...(props.showIconOnly ? { pr: 0 } : {}),
+        ...(props.showIconOnly
+          ? { pr: 0, borderRadius: props.fullRoundedCorners ? 999 : 'auto' }
+          : {}),
       }}
     >
       {props.icon && (
```
thisislawatts, line 106: **issue(blocking):** As there is a new prop, can we please extend the Button.stories.tsx file so that this variant is documented.

thisislawatts:
@benfurber could you provide a link here to the issue or Figma with the target designs? Not sure how to evaluate these changes otherwise.

benfurber:
@thisislawatts Totally, will do. Rushing atm as I was going to include more in this PR then we had that chat on discord.

packages/components/src/CreateComment/CreateComment.tsx
```diff
@@ -1,29 +1,42 @@
...
   onChange: (value: string) => void
   comment: string
+  buttonLabel?: string
   userProfileType?: string
   sx?: ThemeUIStyleObject
 }
 
 export const CreateComment = (props: Props) => {
   const { comment, maxLength, isLoggedIn } = props
+  const buttonLabel = props.buttonLabel || 'Leave a comment'
```
thisislawatts, line 22: **suggestion:** Can we introduce a test that verifies passed `buttonLabel` properties will be rendered as expected?

packages/components/src/UserEngagementWrapper/UserEngagementWrapper.tsx
```diff
@@ -0,0 +1,26 @@
...
+        flexDirection: 'column',
+        marginTop: 4,
+        paddingBottom: 5,
+        paddingTop: 5,
+      }}
+    >
+      {children}
+    </Flex>
+  )
+}
```
thisislawatts, line 26: Can we introduce a `UserEngagementWrapper.stories.tsx` to document this components behaviour? When adding new components, if you use the included `yarn new-component …` script it will automatically generate these documentation files for you.

src/pages/common/Discussion/Discussion.tsx
```diff
@@ -0,0 +1,47 @@
...
+  comments: UserComment[]
+  parent: CommentableModel
+  store: CommentableStore
+}
+
+export const Discussion = ({ comments, parent, store }: IProps) => {
+  const heading = setCommentsHeading(comments.length)
+
+  return (
+    <Flex
```
thisislawatts, line 21: **Question:** Why not use the `Card` component here? It provides the visual styling and also makes any future changes quicker to implement. We can then use a nested `Flex` element to provide the layout.

src/pages/common/Discussion/Discussion.tsx
```diff
@@ -0,0 +1,47 @@
...
+        borderStyle: 'solid',
+        borderWidth: 2,
+        flexDirection: 'column',
+        gap: 2,
+        padding: 3,
+        width: '100%',
+      }}
+    >
+      <Heading as="h3">{heading}</Heading>
+      {comments.map((comment) => (
```
thisislawatts, line 35: **question:** Why are we dropping `CommentList` component here to make visual changes? This removes pagination behaviour
https://github.com/ONEARMY/community-platform/blob/master/packages/components/src/CommentList/CommentList.tsx#L61

src/pages/common/UserAvatar/UserAvatar.tsx
```diff
@@ -0,0 +1,44 @@
...
+  userName: string
+}
+
+export const UserAvatar = ({ userName }: IProps) => {
+  const [src, setSrc] = useState<string>(DefaultMemberImage)
+  const { userStore } = useCommonStores().stores
+
+  useEffect(() => {
+    const getUrl = async () => {
+      const user = await userStore.getUserByUsername(userName)
```
thisislawatts, line 19: **issue:** This will result in the client doing a fetch/read operation for each User record. Read operations are the billable unit in Firestore so we want to be sure these queries are being cached effectively.

Is data fetching something we want to handle within the a nested component? An alternative approach would be to lift it higher up the chain. This would mean when fetching the comment list from the DB there is a follow up enrichment process which handles loading additional user profile data and annotating the comment list data structure for consumption and display by the CommentItem component.

This would allow the UI changes to be made in isolation with `CommentItem` updated to support an avatar parameter and fail gracefully if it is not available or loading.

src/pages/common/Discussion/Discussion.tsx
benfurber: You're right for this PR, though an approach including nested meant I was anticipating having to change this a lot and I missed this today.

src/pages/common/UserAvatar/UserAvatar.tsx
benfurber: I'll have a look at lifting it up the chain but I'm pessimistic as the data isn't currently part of the data structure. 

So I agree with you it probably needs to be covered in the nested model changes.

thisislawatts: Makes sense, we can and should iterate on the UI components in isolation from the data available in the app as long as the component fails gracefully if the expected data is not available :)

benfurber:
Closing in favour of a bunch of small component change PRs.

PR #3092, opened 2023-12-22
feat: introduce Discussion model and store for persisting data

thisislawatts:
PR Checklist

PR Type

## Description

Introduces new Discussion model and store with basic CRUD operations. This new data structure will be used to host all comment lists going forward. It extends the existing Comment model slightly to support a new property `parentCommentId`. Setting the `parentCommentId` for a comment item creates a Reply.

The comments will be stored as a flat list which allows:

* Simplified Data Management: Handling comments in a flat list simplifies database operations, making it easier to manage and query data without complex hierarchical queries.
* Improved Performance: Flat lists can offer better performance, especially in large-scale applications, as operations like sorting and filtering are generally more efficient compared to nested structures.
* Ease of Pagination: Implementing pagination is simpler with a flat list. Hierarchical structures can complicate pagination across multiple levels of nested comments, which is avoided in a flat approach.
* Consistency in Data Representation: A flat list provides a consistent method for representing comments, regardless of their level of nesting. This can simplify the design of APIs and data interfaces.

Drawbacks
* Loss of Hierarchical Context: Flat lists don't inherently show the parent-child relationships, making it difficult for users to follow threaded discussions and understand the context of replies.
* Challenging User Interface Design: Representing nested comments in a user-friendly way can be more challenging with a flat list, as it requires additional UI considerations to clearly convey the relationships between comments.
* Complexity in Maintaining Relationships: In a flat list, additional fields or indicators are often needed to denote relationships between comments, which can add complexity to the data structure and the logic for interpreting these relationships.

Split from the work in #3008, with some rework to expand test coverage. 

Outline of follow up tasks:
* [ ] Integrate with Questions using existing UI components.
* [ ] Support nesting in UI components
* [ ] Integrate with **new** Research
* [ ] Migrate comment content for existing Research items
* [ ] Integrate with **new** How-to
* [ ] Migrate comment content for existing How-to items

This work could happen in parallel, however the interfaces are unlikely to be entire stable so there will be refactoring and adjustments which could introduce complexity when merging long lived feature branches.

### DIFF
--- a/src/models/discussion.models.tsx
+++ b/src/models/discussion.models.tsx
@@ -1,7 +1,7 @@
 import type { IComment } from './comment.model'
 
 export type IDiscussionComment = IComment & {
-  parentCommentId?: string
+  parentCommentId: string | null
 }
 
 export type IDiscussion = {
--- a/src/models/discussion.models.tsx
+++ b/src/models/discussion.models.tsx
@@ -7,6 +7,6 @@ export type IDiscussionComment = IComment & {
 export type IDiscussion = {
   _id: string
   sourceId: string
-  sourceType: string
+  sourceType: 'question'
   comments: IDiscussionComment[]
 }
--- a/src/models/discussion.models.tsx
+++ b/src/models/discussion.models.tsx
@@ -1,0 +1,12 @@
+import type { IComment } from './comment.model'
+
+export type IDiscussionComment = IComment & {
+  parentCommentId?: string
+}
+
+export type IDiscussion = {
+  _id: string
+  sourceId: string
+  sourceType: string
+  comments: IDiscussionComment[]
+}
--- a/src/models/index.ts
+++ b/src/models/index.ts
@@ -17,5 +17,4 @@ export * from './question.models'
 export * from './discussion.models'
 export interface UserComment extends IComment {
   isEditable: boolean
-  showReplies: boolean
 }
--- a/src/models/index.ts
+++ b/src/models/index.ts
@@ -14,6 +14,8 @@ export * from './user.models'
 export * from './userPreciousPlastic.models'
 export * from './notifications.models'
 export * from './question.models'
+export * from './discussion.models'
 export interface UserComment extends IComment {
   isEditable: boolean
+  showReplies: boolean
 }
--- a/src/stores/Discussions/discussion.store.test.ts
+++ b/src/stores/Discussions/discussion.store.test.ts
@@ -1,0 +1,156 @@
+jest.mock('../common/module.store')
+import type { IDiscussion } from 'src/models'
+import { DiscussionStore } from './discussions.store'
+import type { RootStore } from '..'
+import { FactoryUser } from 'src/test/factories/User'
+import { FactoryDiscussion } from 'src/test/factories/Discussion'
+import { FactoryComment } from 'src/test/factories/Comment'
+
+const factory = async (
+  discussions: IDiscussion[] = [FactoryDiscussion({})],
+) => {
+  const store = new DiscussionStore({} as RootStore)
+
+  const discussionItem = discussions[0]
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  store.db.get.mockReturnValue(discussionItem)
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  store.setActiveUser(FactoryUser({ _id: 'fake-user' }))
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  store.aggregationsStore = {
+    aggregations: {
+      users_verified: ['fake-user'],
+    },
+  }
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  store.userStore = {
+    activeUser: {
+      _id: 'fake-user',
+    },
+  }
+
+  return {
+    store,
+    discussionItem,
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    setFn: store.db.set,
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    getFn: store.db.get,
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    getWhereFn: store.db.getWhere,
+  }
+}
+
+describe('discussion.store', () => {
+  describe('uploadDiscussion', () => {
+    it('creates a new discussion with sourceId and sourceType provided', async () => {
+      const { store, discussionItem, setFn } = await factory()
+
+      await store.uploadDiscussion(
+        discussionItem.sourceId,
+        discussionItem.sourceType,
+      )
+
+      const [newDiscussion] = setFn.mock.calls[0]
+      expect(setFn).toHaveBeenCalledTimes(1)
+      expect(newDiscussion.sourceId).toBe(discussionItem.sourceId)
+      expect(newDiscussion.sourceType).toBe(discussionItem.sourceType)
+    })
+  })
+
+  describe('addComment', () => {
+    it('adds a new comment', async () => {
+      const { store, discussionItem, setFn } = await factory()
+
+      //Act
+      await store.addComment(discussionItem, 'New comment')
+
+      // Assert
+      const [newDiscussion] = setFn.mock.calls[0]
+
+      expect(setFn).toHaveBeenCalledTimes(1)
+      expect(newDiscussion.comments[0]).toEqual(
+        expect.objectContaining({ text: 'New comment' }),
+      )
+      expect(newDiscussion.comments[1]).toBeUndefined()
+    })
+
+    it('adds a reply to a comment', async () => {
+      const { store, discussionItem, setFn } = await factory([
+        FactoryDiscussion({
+          comments: [FactoryComment({ text: 'New comment' })],
+        }),
+      ])
+
+      //Act
+      await store.addComment(
+        discussionItem,
+        'New reply',
+        discussionItem.comments[0]._id,
+      )
+
+      const [newDiscussion] = setFn.mock.calls[0]
+
+      // Assert
+      expect(setFn).toHaveBeenCalledTimes(1)
+      expect(newDiscussion.comments).toEqual(
+        expect.arrayContaining([
+          expect.objectContaining({ text: 'New reply' }),
+        ]),
+      )
+    })
+  })
+
+  describe('deleteComment', () => {
+    it('removes a comment', async () => {
+      const { store, setFn } = await factory()
+
+      //Act
+      const comment = FactoryComment({ text: 'New comment' })
+
+      const discussion = FactoryDiscussion({ comments: [comment] })
+
+      await store.uploadDiscussion(discussion.sourceId, discussion.sourceType)
+
+      const [newDiscussion] = setFn.mock.calls[0]
+
+      await store.deleteComment(newDiscussion, comment._id)
+
+      const [newDiscussionWithoutComment] = setFn.mock.calls[1]
+
+      // Asert
+      expect(setFn).toHaveBeenCalledTimes(2)
+      expect(newDiscussionWithoutComment.comments).toHaveLength(0)
+    })
+  })
+
+  describe('formatComments', () => {
+    it('get formated comments from IComment[] type to UserCommentType[]', async () => {
+      const { store, discussionItem } = await factory([
+        FactoryDiscussion({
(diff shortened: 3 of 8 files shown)


src/stores/Discussions/discussions.store.tsx
```diff
@@ -0,0 +1,251 @@
...
+    return (
+      toJS(
+        await this.db
+          .collection<IDiscussion>(COLLECTION_NAME)
+          .getWhere('sourceId', '==', sourceId),
+      )[0] || null
+    )
+  }
+
+  public formatComments(
```
benfurber, line 40: Should a formating function be in the store? Feels like a helper function?

src/models/index.ts
```diff
@@ -14,6 +14,8 @@ export * from './user.models'
 export * from './userPreciousPlastic.models'
 export * from './notifications.models'
 export * from './question.models'
+export * from './discussion.models'
 export interface UserComment extends IComment {
   isEditable: boolean
+  showReplies: boolean
```
benfurber, line 20: Is this a presentational type? If it is, considering your point on discord yet now, should it be in a model type?

But maybe I'm wrong, I'm not sure from this PR I can see what this field is for.

src/stores/Discussions/discussions.store.tsx
```diff
@@ -0,0 +1,251 @@
...
+        ...comment,
+        text: changeUserReferenceToPlainText(comment.text),
+        isUserVerified:
+          !!this.aggregationsStore.aggregations.users_verified?.[
+            comment.creatorName
+          ],
+        isEditable:
+          [
+            this.userStore.activeUser?._id,
+            this.userStore.activeUser?.userName,
```
benfurber, line 57: Do we have user ids? I thought userName was the primary key anyway?

src/stores/Discussions/discussions.store.tsx
```diff
@@ -0,0 +1,251 @@
...
+          ].includes(comment._creatorId) ||
+          isAllowedToEditContent(item, this.userStore.activeUser || undefined),
+        showReplies: false,
+      }
```
benfurber, line 61: There's a lot going on in this map return, could some of it be pulled into variables above to aid readability?

src/stores/Discussions/discussions.store.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  @action
+  public async addComment(
+    discussion: IDiscussion,
+    text: string,
+    commentId?: string,
+  ): Promise<IDiscussion | undefined> {
+    try {
+      const user = this.activeUser
+      const comment = text.slice(0, MAX_COMMENT_LENGTH).trim()
```
benfurber, line 93: I don't think this is right behaviour. If a comment is longer that what should be stored, it should error and give the user the chance to correct rather than persisting something different to what the user intended.

src/stores/Discussions/discussions.store.tsx
```diff
@@ -0,0 +1,251 @@
...
+      if (user && comment) {
+        const dbRef = this.db
+          .collection<IDiscussion>(COLLECTION_NAME)
+          .doc(discussion._id)
+
+        const currentDiscussion = toJS(await dbRef.get())
+
+        const newComment: IDiscussionComment = {
+          _id: randomID(),
+          _created: new Date().toISOString(),
```
benfurber, line 104: Minor thought. OK, it's only on one line, but could the date creation be a helper? Then we can use that helper across all the stores ensuring consistency

src/stores/Discussions/discussions.store.tsx
```diff
@@ -0,0 +1,251 @@
...
+          _creatorId: user._id,
+          creatorName: user.userName,
+          creatorCountry: getUserCountry(user),
+          text: comment,
+          parentCommentId: commentId,
+        }
+
+        if (!currentDiscussion) {
+          throw new Error('Discussion not found')
+        }
```
benfurber, line 114: Should this before newComment?

src/stores/Discussions/discussions.store.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  @action
+  public async editComment(
+    discussion: IDiscussion,
+    commentId: string,
+    text: string,
+  ): Promise<IDiscussion | undefined> {
+    try {
+      const user = this.activeUser
+      const comment = text.slice(0, MAX_COMMENT_LENGTH).trim()
```
benfurber, line 134: Same comment as above.

src/stores/Discussions/discussions.store.tsx
```diff
@@ -0,0 +1,251 @@
...
+import { getUserCountry } from 'src/utils/getUserCountry'
+import { MAX_COMMENT_LENGTH } from 'src/constants'
+import type { DocReference } from '../databaseV2/DocReference'
+import type {
+  IDiscussion,
+  IDiscussionComment,
+} from 'src/models/discussion.models'
+import type { IUserPPDB, UserComment } from 'src/models'
+import type { RootStore } from '..'
+import { changeUserReferenceToPlainText } from '../common/mentions'
```
benfurber, line 20: Probably for the test file too but certainly for this one, maybe you please forgive my request for import ordering pedantry - it helps be understand the shape of a files connections. Ideally the package imports, gap, local imports, gap, types. A compromise I think is at least the type imports seperated from the rest.

benfurber (changes requested):
The store is very easy to read, thank you. And I'm totally with use adding proper return values to everything. 

Is adding Discussion to the `IStores` list on `src/stores/index` missing?

src/stores/Discussions/discussions.store.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+const COLLECTION_NAME = 'discussions'
+
+export class DiscussionStore extends ModuleStore {
+  constructor(rootStore: RootStore) {
+    super(rootStore)
+    makeObservable(this)
+  }
+
+  public async fetchDiscussion(sourceId: string): Promise<IDiscussion | null> {
```
benfurber, line 30: sourceId or _id?

src/models/discussion.models.tsx
```diff
@@ -0,0 +1,12 @@
+import type { IComment } from './comment.model'
+
+export type IDiscussionComment = IComment & {
+  parentCommentId?: string
+}
+
+export type IDiscussion = {
+  _id: string
+  sourceId: string
+  sourceType: string
```
benfurber, line 10: Shouldn't sourceType be stricter? e.g.:
````type SourceType = 'howto' | 'research' | 'comment'```

Comment (thisislawatts) [reply src/models/discussion.models.tsx]: Good idea, fixed in https://github.com/ONEARMY/community-platform/pull/3092/commits/b2a8693f96c1efb7b3f6df81e87cd8dfc9076bf4

File: src/stores/Discussions/discussions.store.tsx (hunk 9 above)
Comment (thisislawatts) [reply src/stores/Discussions/discussions.store.tsx]: Happy to adopt a standard here but please lets make that something that can be enforced (and ideally fixed) using automated tools. Otherwise the burden sits on human reviewers to keep this consistent. I think dropping the format method a3275d59277062df24a5ab6b13d15fa5bab920aa, fixed this for the store file though.

File: src/stores/Discussions/discussions.store.tsx (hunk 6 above)
Comment (thisislawatts) [reply src/stores/Discussions/discussions.store.tsx:72]: That is valid feedback but I would prefer to see this abstraction deferred until there are more call sites which make it a necessary. Until then I believe it is more readable to keep this inlined.

File: src/stores/Discussions/discussions.store.tsx (hunk 5 above)
Comment (thisislawatts) [reply src/stores/Discussions/discussions.store.tsx:61]: I believe that is something which would already be caught when interacting with the form components. This is a defensive mechanism in case someone has slipped (or hacked) past the UI protection.

Comment (thisislawatts) [reply src/stores/Discussions/discussions.store.tsx:61]: At the moment this behaviour also exists in the howto and research stores. If we want to change it we should aim for the modules to be consistent.

File: src/stores/Discussions/discussions.store.tsx (hunk 1 above)
Comment (thisislawatts) [reply src/stores/Discussions/discussions.store.tsx]: A very valid point, I thought about it but it had a test so I gave it a cheeky pass. Removed in a3275d59277062df24a5ab6b13d15fa5bab920aa

File: src/stores/Discussions/discussions.store.tsx (hunk 4 above)
Comment (thisislawatts) [reply src/stores/Discussions/discussions.store.tsx]: Fixed in a3275d59277062df24a5ab6b13d15fa5bab920aa

File: src/stores/Discussions/discussions.store.tsx (hunk 7 above)
Comment (thisislawatts) [reply src/stores/Discussions/discussions.store.tsx]: Fixed in: 3c5a6e8179f0f4a0f8446cd435ecee039ee86315

File: src/models/index.ts (hunk 2 above)
Comment (thisislawatts) [reply src/models/index.ts:20]: Good catch 🦅 👀, fixed in 4cdffbbfe3bd4a2e05d461ccbf7a7764aed02465

File: src/stores/Discussions/discussions.store.tsx (hunk 10 above)
Comment (thisislawatts) [reply src/stores/Discussions/discussions.store.tsx:25]: My preference would to use camelcase for variable names rather than have to many leading underscores floating about.

File: src/stores/Discussions/discussions.store.tsx (hunk 3 above)
Comment (thisislawatts) [reply src/stores/Discussions/discussions.store.tsx]: No longer relevant as the enclosing method has been removed.

**Comment by thisislawatts** [comment]
> Is adding Discussion to the IStores list on src/stores/index missing?
Good spot, I missed this as I have not integrated this store anywhere.

File: src/stores/Discussions/discussions.store.tsx (hunk 10 above)
Comment (benfurber) [reply src/stores/Discussions/discussions.store.tsx:25]: Sorry my question wasn't clear. With the name of `fetchDiscussion` I assume this method is for fetching an individual dicussion by the unique identified, isn't `_id` the unique identifier?

File: src/models/discussion.models.tsx [hunk 12]
```diff
@@ -0,0 +1,12 @@
+import type { IComment } from './comment.model'
+
+export type IDiscussionComment = IComment & {
+  parentCommentId?: string
+}
```
benfurber, line 5: (Thought I could have had on the last review) What's the thinking of extended IComment rather than adding `parentCommentId` to IComment?

src/stores/Discussions/discussion.store.test.ts
```diff
@@ -0,0 +1,249 @@
...
+import { faker } from '@faker-js/faker'
+
+const factory = async (
+  discussions: IDiscussion[] = [FactoryDiscussion({})],
+) => {
+  const store = new DiscussionStore({} as RootStore)
+
+  const discussionItem = discussions[0]
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
benfurber, line 17: Having all these exclusions inline makes the readability more difficult. Can you not exclude once at the top of the file?

benfurber (changes requested):
Still requesting changes as I think what fetchDiscussion should be doing is wrong.

src/stores/Discussions/discussions.store.tsx
thisislawatts: Ah sorry, I misunderstood. The intention is for a Discussion document to contain a reference to the source content item (question, how-to or research). I have renamed the method to better reflect this relationship: fd41225c087be066ab930ac75a6ea249ccf5247b

src/models/discussion.models.tsx
thisislawatts: I thought that adding this to the parent IComment type would run the risk of "leaking" the concept of parentCommentId to code interacting with the existing Comment data objects and it is really only something we want to exist within a Discussion data object. Eventually when all of the content is migrated across to use the new comment format we should be able to merge these interfaces and drop one of them.

thisislawatts:
@benfurber, I have made some changes to the fetchDiscussion method. Could you take another look. 

I appreciate the readability remark on the code comments around the test setup. We have this in a few different places so maybe we can look at this in a separate PR. Do you think it should block the other additions being made here?

benfurber (approved):
Lovely. I think we're all good now.

src/stores/Discussions/discussions.store.tsx
```diff
@@ -0,0 +1,221 @@
...
+      }
+    } catch (err) {
+      logger.error(err)
+      throw new Error(err?.message)
+    }
+  }
+
+  @action
+  public async deleteComment(
+    discussion: IDiscussion,
```
mariojsnunes, line 140: should be enough to pass the discussionId

src/stores/Discussions/discussions.store.tsx
```diff
@@ -0,0 +1,221 @@
...
+      .doc(newDiscussion._id)
+
+    return this._updateDiscussion(dbRef, newDiscussion)
+  }
+
+  @action
+  public async addComment(
+    discussion: IDiscussion,
+    text: string,
+    commentId?: string,
```
mariojsnunes, line 59: this should be called parentCommentId to avoid confusion

PR #3100, opened 2023-12-24
feat: integrate discussions with Question content

thisislawatts:
PR Checklist

PR Type

## Description

Loads comments from Discussions for Question items or creates them if there is no Discussion associated with the collection yet.


src/pages/Question/QuestionPage.tsx
```diff
@@ -1,36 +1,50 @@
...
+          }
+        }
       }
 
       setIsLoading(false)
     }
 
     fetchQuestions()
-  }, [isLoading, question])
+  }, [isLoading])
```
onim-at, line 47: by having this dependency the useEffect will be triggered a second time after the `setIsLoading` at line 43, with a new non required call to `store.fetchQuestionBySlug(slug)`

thisislawatts: Thanks for taking a look so quickly @onim-at ✨ That was entirely my understanding. However I observed that when `question` was included as a dependency of the useEffect it was being invoked in a loop. Welcome any ideas on why that is happening.

onim-at: I think it would be better to remove both `question `and `isLoading `from the dependency and let the useEffect be called only once by leaving the empty array

thisislawatts: I've pushed a new commit which switches to the `slug` dependency, which when thinking about it further is the only thing that should trigger a fresh call to the server. 
https://github.com/ONEARMY/community-platform/pull/3100/commits/facd3301b62ecaaaeb3c504757dd8936fb72a402

onim-at: yes, that's even better 🙂

src/pages/Question/QuestionPage.tsx
```diff
@@ -1,24 +1,29 @@
...
 import { isAllowedToEditContent } from 'src/utils/helpers'
 import { Box, Button, Card, Heading, Text, Flex } from 'theme-ui'
 import { ContentAuthorTimestamp } from '../common/ContentAuthorTimestamp/ContentAuthorTimestamp'
+import { QuestionComments } from './QuestionComments'
+import { useDiscussionStore } from 'src/stores/Discussions/discussions.store'
 
 export const QuestionPage = () => {
   const { slug } = useParams()
   const store = useQuestionStore()
+  const discussionStore = useDiscussionStore()
```
benfurber, line 15: I thought we were moving to accessing the stores via `useCommonStores()` did I just make that up?

src/stores/Discussions/discussions.store.tsx
```diff
@@ -18,20 +18,29 @@ const COLLECTION_NAME = 'discussions'
...
   constructor(rootStore: RootStore) {
-    super(rootStore)
+    super(rootStore, COLLECTION_NAME)
     makeObservable(this)
+    super.init()
   }
 
-  public async fetchDiscussionBySourceId(
+  @action
+  public async fetchOrCreateDiscussionBySource(
```
benfurber, line 27: I don't think this makes sense as a public interface. Aren't I always getting a discussion returned? So why do I care about the creation element and if I did, shouldn't I be doing that as a seperate step?

src/stores/Discussions/discussions.store.tsx
```diff
@@ -18,20 +18,29 @@ const COLLECTION_NAME = 'discussions'
...
           .getWhere('sourceId', '==', sourceId),
       )[0] || null
-    )
+
+    if (foundDiscussion) {
+      return foundDiscussion
+    }
+
+    // Create a new discussion
+    return (await this.uploadDiscussion(sourceId, sourceType)) || null
```
benfurber, line 43: Why would I get null back? Shouldn't it error if a discussion can't be created?

src/pages/Question/QuestionPage.tsx
```diff
@@ -27,15 +32,24 @@ export const QuestionPage = () => {
...
+        if (question && discussionStore) {
+          const discussion =
+            await discussionStore.fetchOrCreateDiscussionBySource(
+              question._id,
+              'question',
+            )
+          if (discussion) {
+            setComments(discussion.comments)
+          }
+        }
```
benfurber, line 44: Should the application be responsible for creating a discussion seperately from the question? Shouldn't the question store as standard create a discussion for the question?

src/pages/Question/index.tsx
```diff
@@ -6,12 +6,18 @@ import {
...
+} from 'src/stores/Discussions/discussions.store'
 
 export const QuestionModuleContainer = () => {
   const rootStore = useCommonStores()
   return (
     <QuestionStoreContext.Provider value={new QuestionStore(rootStore)}>
-      <QuestionRoutes />
+      <DiscussionStoreContext.Provider value={new DiscussionStore(rootStore)}>
+        <QuestionRoutes />
+      </DiscussionStoreContext.Provider>
```
benfurber, line 20: Discussion store is a child of the question store? Won't there be one discussion store for the whole app?

src/stores/Question/question.store.tsx
```diff
@@ -137,5 +137,5 @@ export class QuestionStore extends ModuleStore {
  * The context will be populated with the QuestionStore in the module index
  * (avoids cyclic deps and ensure shared module ready)
  */
-export const QuestionStoreContext = createContext<QuestionStore>(null as any)
+export const QuestionStoreContext = createContext<QuestionStore>('boo' as any)
```
benfurber, line 140: Leftover from testing?

benfurber (changes requested):

src/pages/Question/QuestionPage.tsx
thisislawatts: I might have missed that decision or worse forgotten it 🤔 Do you know if this was documented anywhere? I think generally we are moving towards loading stores via React hooks over class based injection but I am unsure of a strong preference for `useCommonStores` over loading individual stores. Are there any clear benefits/drawbacks to the two approaches?

src/stores/Discussions/discussions.store.tsx
thisislawatts: Yes, I agree the call site doesn't really care about the implementation, simply that they get back a Discussion. I renamed it at the last moment as I thought it was better to be explicit about the implementation in case any consumers at a later date invoked the method and inadvertently ended up creating discussions. 

I thought also about splitting the creation process into a separate step for the call site but thought this would end up in multiple places across the app fetching and then creating a discussion if it didn't exist so thought it would be worth moving into the store. This is based on the assumption that the consumer wants a discussion and doesn't care whether it was pre-existing, ie. created 2 years or 1ms ago.

src/stores/Discussions/discussions.store.tsx
thisislawatts: Yeah good point, I think the error generation/handling could do with some thought here. Let me think on it. Do you think that is a blocker on merging these changes or something which needs to be resolved prior to releasing this functionality?

src/pages/Question/QuestionPage.tsx
thisislawatts: I mentioned this briefly in [the comment here](https://github.com/ONEARMY/community-platform/pull/3100#discussion_r1438442060), if we have a similar flow of logic for Research and How-to documents it seems reasonable to shift responsibility down to the Discussion store rather than duplicating it across the owning documents.

src/stores/Question/question.store.tsx
thisislawatts: Fixed in https://github.com/ONEARMY/community-platform/pull/3100/commits/d1524c020d105c4a1bee1e34a812bc0736415f86, force pushed into a previous changeset

src/pages/Question/index.tsx
thisislawatts: Possibly, but at the moment we are rolling it out only to Questions.

thisislawatts:
@benfurber can I get a re-review on this item please?

src/pages/Question/QuestionPage.tsx
benfurber: Can't really see a real benefit either way. Maybe a minor import reduction if you need more than one store in a file to use `useCommonStores`.

src/stores/Discussions/discussions.store.tsx
thisislawatts: I think there is a slight difference here, we agree that the store should handle fetching and/or creating a discussion.  However we do not seem to agree on the naming convention. To further explain my reasoning, I believe a method should clearly communicate what it is doing, as it that aids maintainability and comprehension for new contributors. It seems unexpected if a method called `fetchDiscussionBySourceId `, actually resulted in additional documents being persisted to the datastore. Do you believe otherwise?

src/pages/Question/index.tsx
thisislawatts: That might be the right choice, I am intentionally not thinking too much about how best this should be implemented for Research/How-to but instead focusing solely on integrating it **only** with the Questions content type.

src/pages/Question/QuestionPage.tsx
```diff
@@ -81,47 +94,52 @@ export const QuestionPage = () => {
...
-              hasUserVotedUseful={store.userVotedActiveQuestionUseful}
-              isLoggedIn={store.activeUser ? true : false}
-              onUsefulClick={onUsefulClick}
-            />
-            <FollowButton
-              hasUserSubscribed={hasUserSubscribed}
-              isLoggedIn={isLoggedIn ? true : false}
-              onFollowClick={onFollowClick}
+        <>
+          <Card sx={{ mt: 4, p: 4, position: 'relative' }}>
```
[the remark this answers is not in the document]
thisislawatts: Understood, but I will not address this feedback in this PR. I agree with your sentiment but stylistic changes like this are not the focus of this PR. These lines have only been marked in the diff due to the additional nesting introduced.

src/stores/Discussions/discussions.store.tsx
thisislawatts: What do you think are the downhill effects of allowing null here?

src/pages/Question/index.tsx
mariojsnunes: I'd tend to agree with ben, but this shouldn't matter if we move to a service approach instead of stores soon.

thisislawatts: We agree that this is reasonable for the time being but may need some refactoring as Discussions are rolled out across different content types.

src/pages/Question/QuestionComments.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const QuestionComments = ({ comments }: IProps) => {
+  return (
+    <>
+      {
+        <Box mt={5}>
+          <CommentList comments={comments} />
+        </Box>
+      }
+    </>
+  )
```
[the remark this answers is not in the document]
thisislawatts: Fixed in 57e00d35e4b88d89412ad758ac505541358c777a

src/pages/Question/QuestionPage.tsx
thisislawatts: Added a try/catch

benfurber (approved):

PR #3117, opened 2023-12-27
feat: add research status to research

iSCJT:
PR Checklist

PR Type

## Description

Introduces the research status field and corresponding research list filter for research status.

Also some minor bug fixes to solve type errors on the research store.

## Git Issues

Closes #

## Screenshots/Videos

image

image

---

## What happens next?

### DIFF
--- a/src/models/research.models.tsx
+++ b/src/models/research.models.tsx
@@ -31,7 +31,7 @@ type ResearchDocumentLock = ResearchDocumentLockInformation | null
 
 export const researchStatusOptions = [
   'In progress',
-  'Complete',
+  'Completed',
   'Archived',
 ] as const
 
--- a/src/models/research.models.tsx
+++ b/src/models/research.models.tsx
@@ -22,6 +22,8 @@ type ResearchDocumentLockInformation = {
 
 type ResearchDocumentLock = ResearchDocumentLockInformation | null
 
+export type ResearchStatus = 'In progress' | 'Complete' | 'Archived'
+
 type UserIdList = UserId[]
 
 export namespace IResearch {
@@ -49,6 +51,7 @@ export namespace IResearch {
     comments?: IComment[]
     collaborators?: string[]
     status: 'draft' | 'published'
+    researchStatus?: ResearchStatus
     locked?: ResearchDocumentLock
   }
 
@@ -61,6 +64,7 @@ export namespace IResearch {
     creatorCountry?: string
     collaborators: string
     previousSlugs?: string[]
+    researchStatus?: ResearchStatus
   }
 
   /** Research items synced from the database will contain additional metadata */
--- a/src/models/research.models.tsx
+++ b/src/models/research.models.tsx
@@ -1,6 +1,13 @@
+import type {
+  DBDoc,
+  IComment,
+  IModerable,
+  ISelectedTags,
+  ISharedFeatures,
+  UserMention,
+} from '.'
 import type { IUploadedFileMeta } from '../stores/storage'
 import type { IConvertedFileMeta } from '../types'
-import type { DBDoc, IComment, IModerable, ISelectedTags, UserMention } from '.'
 import type { IResearchCategory } from './researchCategories.model'
 
 /**
@@ -36,7 +43,6 @@ export namespace IResearch {
     collaborators: string[]
     subscribers?: UserIdList
     locked?: ResearchDocumentLock
-    votesUseful?: UserIdList
   } & Omit<FormInput, 'collaborators'>
 
   /** A research item update */
@@ -55,7 +61,7 @@ export namespace IResearch {
     locked?: ResearchDocumentLock
   }
 
-  export interface FormInput extends IModerable {
+  export interface FormInput extends IModerable, ISharedFeatures {
     title: string
     description: string
     researchCategory?: IResearchCategory
--- a/src/models/research.models.tsx
+++ b/src/models/research.models.tsx
@@ -1,3 +1,5 @@
+import type { IUploadedFileMeta } from '../stores/storage'
+import type { IConvertedFileMeta } from '../types'
 import type {
   DBDoc,
   IComment,
@@ -6,8 +8,6 @@ import type {
   ISharedFeatures,
   UserMention,
 } from '.'
-import type { IUploadedFileMeta } from '../stores/storage'
-import type { IConvertedFileMeta } from '../types'
 import type { IResearchCategory } from './researchCategories.model'
 
 /**
--- a/src/models/research.models.tsx
+++ b/src/models/research.models.tsx
@@ -29,7 +29,13 @@ type ResearchDocumentLockInformation = {
 
 type ResearchDocumentLock = ResearchDocumentLockInformation | null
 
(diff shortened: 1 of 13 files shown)


davehakkens:
hey @iSCJT what design are you building this on, or freestyling it?

iSCJT:
> hey @iSCJT what design are you building this on, or freestyling it?

Freestyling with the view that it will need to be iterated on. I'll send some screenshots when I'm back at my laptop.

davehakkens:
cool!
Could we name it `completed` instead of `complete`. Sound to me like that would make more sense
And Mobile could use a look 
afbeelding

davehakkens:
thanks looks better! But could we remove the colored border on mobile?
Like the idea but needs a bit more thinking with the rest of the design and will look a bit intense for other instances like Fixing Fashion

src/pages/Research/Content/ResearchDescription.tsx
```diff
@@ -126,73 +126,103 @@ const ResearchDescription = ({
...
-                />
-              </Fragment>
-            )}
+            </Flex>
+            <Flex
+              sx={{
+                marginBottom: 'auto',
+                minWidth: '100px',
+                borderRadius: 1,
+                borderBottomRightRadius: 1,
```
thisislawatts, line 202: **suggestion:** It seems like duplication of the borderRadius declaration

```suggestion
                borderRadius: 1,
```

src/pages/Research/Content/ResearchDescription.tsx
```diff
@@ -126,73 +126,103 @@ const ResearchDescription = ({
...
+                  color: 'black',
+                  fontSize: [2, 2, 3],
+                  padding: 2,
+                  margin: 'auto',
+                }}
+                data-cy="research-status"
+              >
+                {research.researchStatus || 'In progress'}
+              </Text>
+            </Flex>
```
thisislawatts, line 225: **question:** It looks like this component could be considered as a variant of the existing `ModerationStatus` component. Did you look at refactoring that to support this usecase? Perhaps a more generic `ContentStatus`. Although I could also argue that it should be a generic `Badge` style which is then used in a few different ways across the Platform.
ThemeUI has a named component which would be a good fit here: https://theme-ui.com/components/badge, although we have naming collision as we already have a named Badge construct.

src/pages/Research/Content/ResearchListItem.tsx
```diff
@@ -29,6 +29,16 @@ const ResearchListItem = ({ item }: IProps) => {
...
   }
+
+  const status = item.researchStatus || 'In progress'
+
+  const statusColour =
+    status === 'Archived'
+      ? 'lightgrey'
+      : status === 'Completed'
+      ? 'betaGreen'
+      : 'accent.base'
```
thisislawatts, line 40: **issue:** Duplicate logic as https://github.com/ONEARMY/community-platform/pull/3117/files#diff-556cc4a5d1fc17ee8506eb98f9d89fed9c46dc1e15549fae69988ab591b5919dR204-R210

src/pages/Research/Content/ResearchDescription.tsx
iSCJT: My bad, thanks.

src/pages/Research/Content/ResearchDescription.tsx
iSCJT: I wouldn't say it's a moderation thing as it's the research creators assigned 'status'.  A `ContentStatus` component could work but I've already started on a separate 'badge' component branch.  I'll add that as a separate pr when done and then refactor to use it.

src/pages/Research/Content/ResearchListItem.tsx
iSCJT: Good point, I've moved the colour logic into a helper for now.

src/pages/common/SortFilterHeader/SortFilterHeader.tsx
```diff
@@ -162,6 +179,34 @@ export const SortFilterHeader = ({
...
         </FieldContainer>
       </Flex>
+      {isResearchStore(currentStore) && (
+        <Flex sx={_inputStyle}>
+          <FieldContainer>
+            <Select
+              options={statusOptions}
+              placeholder="Filter by status"
+              value={
+                (currentStore as ResearchStore).selectedStatus
```
evakill, line 189:
```suggestion
                currentStore.selectedStatus
```
nit(non-blocking): since `isResearchStore` is a type guard, we don't need this cast

src/stores/Research/research.store.tsx
```diff
@@ -297,15 +321,15 @@ export class ResearchStore extends ModuleStore {
   }
 
   public async moderateResearch(research: IResearch.ItemDB) {
-    if (!hasAdminRights(toJS(this.activeUser))) {
+    if (!hasAdminRights(toJS(this.activeUser as IUser))) {
```
evakill, line 324: i try to avoid casting whenever possible; since the type of `this.activeUser` is `IUserPPDB | null | undefined`, i think a preferable approach would be 
```suggestion
    if (!this.activeUser || !hasAdminRights(toJS(this.activeUser))) {
```
& would say the same about the other user casts you added to this file

evakill (approved):
suggestions, no blockers! would be happy to see any/all be addressed in a follow up. tested on preview and LGTM

src/models/research.models.tsx
```diff
@@ -22,6 +29,14 @@ type ResearchDocumentLockInformation = {
 
 type ResearchDocumentLock = ResearchDocumentLockInformation | null
 
+export const researchStatusOptions = [
+  'In progress',
+  'Completed',
+  'Archived',
+] as const
+
+export type ResearchStatus = typeof researchStatusOptions[number]
```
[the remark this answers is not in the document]
iSCJT: Agree that enums are nicer, I'll go through and refactor some bits 😃

src/models/research.models.tsx
```diff
@@ -49,10 +63,11 @@ export namespace IResearch {
     comments?: IComment[]
     collaborators?: string[]
     status: 'draft' | 'published'
+    researchStatus?: ResearchStatus
     locked?: ResearchDocumentLock
   }
 
-  export interface FormInput extends IModerable {
+  export interface FormInput extends IModerable, ISharedFeatures {
```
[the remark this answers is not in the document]
iSCJT: Hmmm, I'll look again.  I added it back in as I was getting type errors in the store related to the useful voting.

src/pages/common/SortFilterHeader/SortFilterHeader.tsx
iSCJT: this was left over from when I first started getting round the typescript errors related to the howto store not having the researchStatus property.  Will clear it up.

src/models/research.models.tsx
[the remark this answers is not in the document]
iSCJT: @evakill for this one I've had a another look and without the ISharedFeatures the store gives type errors for useful etc.  I could make it part of the IModerable declaration or leave it here?

PR #3121, opened 2023-12-29
feat: index commentCount and latestCommentDate

mariojsnunes:
PR Checklist

PR Type

## Description

This function will create `commentCount` and `latestCommentDate` properties for `research`, `howto` and `question` documents.

⚠️Needs testing!
- confirm getCollectionName has the correct cases
- probably need to create an index for these properties

I'm unsure what's the best way to test this.


thisislawatts:
Thanks for raising this @mariojsnunes, do you think this behaviour is something we want to handle in a serverless function that happens as a result of a change being made to a document or within the application level? For example within: https://github.com/ONEARMY/community-platform/blob/master/src/stores/Discussions/discussions.store.tsx#L195

I am aware that we have adopted this pattern in a few places historically. I worry that splitting the business logic between the SPA and serverless functions can be hard to reason with for new contributors and running serverless functions locally with the firebase emulators is an awkward workflow.

mariojsnunes:
Thanks @thisislawatts those are valid points.

Well, in terms of code, I believe the function would be the best way. Events are good to update metadata values in the background. Discussions shouldn't have to know about howto, research, etc

But testing and local development are a pain for sure.
So it's probably best to have the logic on the SPA. I've made the code changes and it looks fine.
(still need to test)

thisislawatts:
I hear you on the event orientated approach @mariojsnunes. Given the additional overhead for devs I appreciate you being open minded to including it in the SPA for the time being. 

This looks like a solid approach, aside from the tests I think this is all in the right direction.

src/stores/Discussions/discussionEvents.ts
```diff
@@ -0,0 +1,47 @@
...
+>
+
+export const updateDiscussionMetadata = (
+  db: DatabaseV2,
+  discussion: IDiscussion,
+) => {
+  const collectionName = getCollectionName(discussion.sourceType)
+
+  if (!collectionName) {
+    return
```
thisislawatts, line 17: **suggestion:** Could we introduce a logger here that logs a warning/error and identifiers for the discussion object, if we have been unable to find a collection name?

src/stores/Discussions/discussionEvents.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const commentCount = discussion.comments.length
+  const latestCommentDate = new Date(
+    Math.max.apply(
+      null,
+      discussion.comments.map((comment) =>
+        new Date(comment._created).valueOf(),
+      ),
+    ),
+  )
```
thisislawatts, line 28: **suggestion(non-blocking):** A small readability tweak, which makes this a bit easier to read. Although I acknowledge this is due with my own familiarity with `...` vs. `.apply` 

```suggestion
  const latestCommentDate = new Date(
    Math.max(
      ...discussion.comments.map((comment) =>
        new Date(comment._created).valueOf(),
      ),
    ),
  )
```

mariojsnunes: yep it's better

mariojsnunes:
I think this could be merged soon. Maybe after https://github.com/ONEARMY/community-platform/pull/3102?

mariojsnunes:
I've tested with the questions module and the respective content item is being correctly populated on every comment create/delete.
Was also able to sort by latestCommentDate and commentCount.

used this code:
```
db.collection(endpoint)
        .orderBy('latestCommentDate', 'desc')
        .get()
        // eslint-disable-next-line no-console
        .then((x) => console.log({ x: x.docs.map((x) => x.data()), s: x.size }))
 ```
 
I think this it's ready to be merged.

thisislawatts (approved):

PR #3145, opened 2024-01-03
chore: remove empty deps array

thisislawatts:
PR Checklist

PR Type

## Description

As part of #3105

### DIFF
--- a/src/common/Form/FileInput/FileInput.tsx
+++ b/src/common/Form/FileInput/FileInput.tsx
@@ -29,7 +29,7 @@ export const FileInput = (props: IProps) => {
     return () => {
       uppy.close()
     }
-  }, [])
+  })
 
   const files = () => {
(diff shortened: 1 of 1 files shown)


mariojsnunes:
Careful with this, an empty dependency array means it will run only on component mount.
If you remove it, it will run on every render.

src/common/Form/FileInput/FileInput.tsx
```diff
@@ -29,7 +29,7 @@ export const FileInput = (props: IProps) => {
     return () => {
       uppy.close()
     }
-  }, [])
+  })
```
mariojsnunes, line 32: I don't see any reason to do this. It's a good practice to keep the empty dependency array.
That return will still run when the component is unmounted.

a useEffect without the dependency array should only be used in a rare scenario when we want to run something after every render, **without blocking the UI**.

Issue #2968, opened 2023-10-29
[QuestionListing] Sorting filter


thisislawatts:
It should be possible for someone to order the Question listings for items which based on various attributes.

Look at the implementation for the how to listing for reference.

davehakkens:
Image as reference 
![sorting](https://github.com/ONEARMY/community-platform/assets/13672737/1878207c-bb69-46b6-bfa7-b22a0e1694da)

mariojsnunes:
Hi, I've been looking into this.
Sorting needs a firestore index. Is that possible? Which fields should allow sorting?

I have already done some draft code:
https://github.com/ONEARMY/community-platform/commit/526a2dfa33cf12bf79f5b072bb9edcfcb0344034

davehakkens:
cool!
It would have these 5 options:

- Newest (default)
- Most useful
- Most comments
- Least comments (To bring unanswered to the top)
- Latest comment (Bonus: Only if possible since it will be linked with this new PR https://github.com/ONEARMY/community-platform/pull/3007)

mariojsnunes:
But on the research module you are getting all results and then filtering client side, correct?

mariojsnunes:
I'd like to propose a different approach from the research module:
- Create the fields we want to sort on the question document (usefulCount, commentCount)
- Populate them with the onWrite trigger in a function https://stackoverflow.com/a/59839989
- Create the index so we can sort/paginate on the firebase query

Pros:
- This will ensure the scalability as we could paginate the results. 
- Would be ready for the fullstack framework migration
- Easier to sort even if we get all the items (if we really wanted to query all and filter/sort on the frontend for caching)

Cons:
- No frontend caching, but we still could do it

Issue #2967, opened 2023-10-29
[QuestionListing] Category filter


thisislawatts:
It should be possible for someone to filter the Question listings for items which based on the available categories.

Look at the implementation for the how to listing for reference.

davehakkens:
Image as reference 
![category](https://github.com/ONEARMY/community-platform/assets/13672737/ac13a786-c80d-4697-85f7-1928b57bea5d)

zweertsk:
I'm not sure why there is a common `SortFilterHeader` component used in the `HowtoListing` that is circling back depending on a `CategoriesSelect` observable from Howto scope.

It looks like The `CategoriesSelect` observable should be abstracted as common functionality as well.

thisislawatts:
Sounds good @zweertsk, it definitely seems like this area of the code would benefit from a refactor.

zweertsk:
> Sounds good @zweertsk, it definitely seems like this area of the code would benefit from a refactor.

@thisislawatts I would like to start on the questions page but have questions my self 😅. Where can I have a discussion with the business about requirements?

thisislawatts:
Happy to provide clarity if I can, either here or on the Issue which best relates to your questions. Alternatively come on over to the Discord

thisislawatts:
Marking as complete

PR #3151, opened 2024-01-04
feat: add header (search, filter, sort) to question's page

### LINKED ISSUE #2969: [QuestionListing] Search
It should be possible for someone to search the Question listings for items which match based on the contents of the Question title or description.

Look at the implementation for the how to listing for reference.

Shamauk:
PR Checklist

PR Type

# Description
- Updated SortFilterHeader to take the type question and render category filter, sort dropdown and search box. 
- Updated FilterSorterDecorator to include as ItemSortingOption:  LeastComments and LatestComments. 
- Updated question.store to be able to search, sort and filter.
- Added to QuestionListing the SortFilterHeader. 
- Created new file src/stores/utils.ts with a function that checks if a store has author functionality (if not then SortFilterHeader does not render Filter by author dropdown)

# Git issues
closes #2969, #2968, #2967


src/stores/common/FilterSorterDecorator/FilterSorterDecorator.ts
```diff
@@ -141,6 +143,29 @@ export class FilterSorterDecorator<T extends IItem> {
     })
   }
 
+  private sortByLeastComments(listItems: T[]) {
+    return this.sortByComments(listItems).reverse()
```
AlfonsoGhislieri, line 147: **Issue**: This new feature doesn't seem to be relevant to the linked issues which are adding the already existing filter options to the new questions module. So lets keep the focus narrowed.

Also @davehakkens  can also maybe jump in here as to whether or not we want to also in general add a filter for Least and Latest comments? As this should be tackled as a separate issue.

src/stores/common/FilterSorterDecorator/FilterSorterDecorator.ts
```diff
@@ -141,6 +143,29 @@ export class FilterSorterDecorator<T extends IItem> {
...
+  }
+
+  private sortByLatestComments(listItems: T[]) {
+    return [...listItems].sort((a, b) => {
```
AlfonsoGhislieri, line 151: Same as above here, let's narrow the scope this PR.

src/pages/common/SortFilterHeader/SortFilterHeader.tsx
```diff
@@ -80,12 +82,13 @@ export const SortFilterHeader = ({
...
-
-  const authorsOptions = getAuthorOptions(items)
+      : type == 'research'
+      ? (currentStore as ResearchStore).filteredResearches
+      : (currentStore as QuestionStore).filteredQuestions
 
   const urlSelectedAuthor = getQueryParam(window.location.href, 'author', null)
-  if (urlSelectedAuthor) currentStore.updateSelectedAuthor(urlSelectedAuthor)
+  if (urlSelectedAuthor && storeHasAuthorFunctionality(currentStore))
+    currentStore.updateSelectedAuthor(urlSelectedAuthor)
```
AlfonsoGhislieri, line 92: **Question**: I appreciate the addition of a guard clause here, but shouldn't the questions also have an author/creator realistically? I have been out of the loop in regards to the new module, but I would assume we want to allow sorting of questions by author/creator. @thisislawatts?

src/stores/utils.ts
```diff
@@ -0,0 +1,12 @@
...
+import type { ResearchStore } from './Research/research.store'
+
+export const storeHasAuthorFunctionality = (
+  store: any,
+): store is HowtoStore | ResearchStore => {
+  return (
+    store &&
+    typeof store.selectedAuthor !== 'undefined' &&
+    typeof store.updateSelectedAuthor === 'function'
+  )
```
AlfonsoGhislieri, line 11: **Praise/Issue**: thanks for adding this, not sure if it will be needed if the questions store is updated to also have the same functionality as the howto and research store. But an extra guard clause is always nice to prevent errors! :)

However, it would also be good practice when adding a function like this one to make it more extensible and maintainable without requiring constant updates for new store types (if we were to ever add/modify them). So something like creating an interface and performing the validation check could be a more flexible approach.

src/stores/Question/question.store.tsx
```diff
@@ -20,15 +24,74 @@ const COLLECTION_NAME = 'questions'
...
+
+      runInAction(() => {
+        this.activeSorter = ItemSortingOption.Newest
+        this.filterSorterDecorator = new FilterSorterDecorator()
+        this.allQuestionItems = this.filterSorterDecorator.sort(
+          this.activeSorter,
+          activeItems,
+        )
+        if (docs.length) {
+          this.isFetching = false
```
AlfonsoGhislieri, line 78: Issue: The handling of the readiness of a component should be handled by the component itself ideally rather than separating away into the store, will also make things less readable and harder to maintain.

src/pages/Question/QuestionListing.tsx
```diff
@@ -1,27 +1,15 @@
...
-
-  useEffect(() => {
-    const fetchQuestions = async () => {
-      const questions = await store.fetchQuestions()
-      setQuestionList(questions || [])
-      setIsLoading(false)
-    }
-
-    fetchQuestions()
-  }, [])
```
AlfonsoGhislieri, line 24: I would keep the management of readiness of a component inside the actual component.

AlfonsoGhislieri (changes requested):
Thanks for this @Shamauk, some issues that have to be addressed, let me know if anything doesn't make sense!

src/stores/common/FilterSorterDecorator/FilterSorterDecorator.ts
Shamauk: Took the inspiration from here:

https://github.com/ONEARMY/community-platform/issues/2968#issuecomment-1829520087

AlfonsoGhislieri: Lovely, then if you're still interested in adding those could you please make a seperate PR and add some tests as well to make sure they are all in order :-)

Shamauk: Yes, I'll do it.

Shamauk: Just a question:

From _sortByComments_ we have:
`const totalCommentsA = a.comments?.length || calculateTotalComments(a)`

It will get the length either from the comments property or if not there it will calculateTotalComments which is a bit misleading since it counts the number of comments under the updates property. 

I just want to ask if it is safe to assume this – that comments are _either_ under comments or updates->comments.
Just want to know for future proofness and for my testing.

src/stores/utils.ts
thisislawatts: @Shamauk a small ask. We already have a few overloaded `utils.ts` files floating around. Instead please see the src/stores/common directory for an example of how we share functionality between the stores at the moment.

src/pages/Question/QuestionListing.tsx
Shamauk: Thanks again for all the responses, this is for this reply and the previous: I essentially followed the same methodology as Research which you can see in `ResearchList.tsx` and `research.store.tsx`. Not sure if I am missing something from Question that would make it differ?

src/stores/common/FilterSorterDecorator/FilterSorterDecorator.ts
AlfonsoGhislieri: That is correct! That is because for some items the comments are for the different updates rather than directly across the entire item

src/pages/Question/QuestionListing.tsx
AlfonsoGhislieri: Yeah I just had a look at it does seem to be a bit of discrepancy as the research store is using this but the other stores like Howto's are not. I would personally prefer to keep it simple and in the component, but this approach does has the added benefit of not making unnecessary fetches every time the page is loaded.

So disregard my last comment :-)

AlfonsoGhislieri (approved):
Thanks for the changes! Just have to resolve the conflicted files :-)

thisislawatts:
@all-contributors add @Shamauk for code.

PR #3165, opened 2024-01-06
refactor: convert discussion store to a service

mariojsnunes:
@thisislawatts the goal of this PR is to understand how hard it is to migrate stores to services.
And it was straightforward... fetch and add are working. Maybe you could use this code on https://github.com/ONEARMY/community-platform/pull/3102?

Some notes:
- need to decide what's the best approach to access the activeUser, here I passed it as an argument from the main component.
- unsure if this is the best way to invoke firestore

About `<DiscussionContainer />` I think we should refactor it to contain all the discussion related CRUD methods, to avoid all this:
image
But I understand it would only be possible when research and howto modules are also migrated.

### DIFF
--- a/src/pages/Question/QuestionPage.tsx
+++ b/src/pages/Question/QuestionPage.tsx
@@ -1,31 +1,36 @@
 import { useEffect, useState } from 'react'
 import { Link, useParams } from 'react-router-dom'
 import {
+  DiscussionContainer,
   FollowButton,
   Loader,
   ModerationStatus,
   UsefulStatsButton,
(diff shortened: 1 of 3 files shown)


thisislawatts:
Thanks for sharing @mariojsnunes, would you be able to define what you mean by `service` here?

mariojsnunes:
> Thanks for sharing @mariojsnunes, would you be able to define what you mean by `service` here?

Just a collection of methods for doing API requests (firestore in this case).
Unlike a store, it doesn't keep state.

PR #3166, opened 2024-01-08
Fix: Follow button test

benfurber:
PR Checklist

PR Type

## Description

This seems to get the cypress test for the follow button on research articles back to a point where the test is flaky rather than failing.


iSCJT (commented):
Are we sure we want to drop the unfollow test?  I'll leave this open a few days as the follow spec seems to have been less problematic the last couple.  If it starts being a pita again then I'll approve.

benfurber:
@iSCJT That step is duplication. That unfollowing works is covered earlier in the test.

thisislawatts:
Got caught out with flakes on this test again, so merging. Thanks for digging into this @benfurber 🌮 ✨

PR #3167, opened 2024-01-08
feat: add user messaging block to contact form

benfurber:
PR Checklist

PR Type

## Description

This adds the ability to block a user from contacting another.


functions/src/emailNotifications/utils.test.ts
```diff
@@ -0,0 +1,142 @@
...
+}))
+
+describe('isBelowMessageLimit', () => {
+  it("returns true if user hasn't reached message cap", async () => {
+    messageDocs.length = 99
+
+    await expect(isBelowMessageLimit('jeff@email.com')).resolves.toEqual(true)
+  })
+
+  it('errors when user has reached message cap', async () => {
```
evakill, line 47: guessing this is to protect from spam, but what was the reasoning behind this cap? and should we have a way to monitor when users hit it?

functions/src/emailNotifications/utils.ts
```diff
@@ -42,35 +52,57 @@ export const getUserAndEmail = async (userName: string) => {
...
 
+export const isUserAllowedToMessage = async (uid) => {
+  const { toUser } = await getUserAndEmail(uid)
+  if (toUser.isBlockedFromMessaging) {
+    throw new Error(errors.USER_BLOCKED)
+  }
   return true
 }
 
+export const isValidEmailCreationRequest = async ({
```
evakill, line 89: nit: since this util file is shared across all email services, can we rename this `isValidMessageRequest` to clarify that it pertains to the message feature

functions/src/emailNotifications/utils.ts
```diff
@@ -42,35 +52,57 @@ export const getUserAndEmail = async (userName: string) => {
...
+  const userDoc = await firebaseAuth.getUserByEmail(email)
+
+  try {
+    isSameEmail(userDoc, email)
+    await isUserAllowedToMessage(userDoc.uid)
+    await isBelowMessageLimit(email)
+    await isReceiverContactable(toUserName)
+    return true
+  } catch {
+    return Promise.reject(false)
```
evakill, line 102: we never expose the error messages attached to each fn! two proposed alternatives: 
1. catch the error here and log it
2. (preferred) let the error be thrown and it will be caught and logged through the `withErrorAlerting` hook

src/stores/Message/message.store.tsx
```diff
@@ -27,6 +28,14 @@ export class MessageStore extends ModuleStore {
       return Promise.reject(error)
     }
 
+    const isUserBlocked = isUserBlockedFromMessaging(this.activeUser)
+    if (isUserBlocked) {
+      logger.error('User blocked from messaging', values)
+
+      const error = new Error('Blocked from messaging')
+      return Promise.reject(error)
```
evakill, line 36: is there any UI indication for blocked users?

evakill (approved):
all of my comments are either nits or can be addressed in a followup. great backend work! 🥳

functions/src/emailNotifications/utils.test.ts
benfurber: What what I remember it was thinking about a user who decided to message EVERYONE about something. Though putting that cap at 100 currently was entirely arbitrary. Good idea on the monitoring. I'll add on the retool side.

benfurber:
Thanks @evakill. To confirm, as we're trying to get the contact feature out for testing asap I will cover your points in a follow-up PR later in the month.

src/stores/Message/message.store.tsx
benfurber: This (not very special) error is shown on the UI.

PR #3175, opened 2024-01-09
refactor: model level string literal types to enums

iSCJT:
PR Checklist

PR Type

## Description

Model level string literal types refactored to enums

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?


iSCJT:
I tried doing an interactive rebase on this to tidy it up but was getting into all kinds of conflicts.

src/models/user.models.tsx
```diff
@@ -1,16 +1,13 @@
 /* eslint-disable @typescript-eslint/naming-convention*/
```
benfurber, line 1: Could this be removed now? I think it might be related to the changes you're making?

benfurber (approved):
Such a great change, nice one @iSCJT. 

One stylistic question and one request please: 
1. Should the enums be imported as types e.g. `import type { IThing } from 'oa-shared'`?
2. Extract child vales from the types. 
e.g. on `src/pages/Howto/Content/Howto/HowtoDescription/HowtoDescription.tsx`:
Do this at the top level:
```const { ACCEPTED } = IModerationStatus```
Then later on you can do:
```howto.moderation === ACCEPTED && (```

src/models/user.models.tsx
iSCJT: Yea I think we can.

iSCJT:
> 2\. const { ACCEPTED } = IModerationStatus

Re: 

1. if you want to use the enum as a value then you can't import it as a type.
2. I've had a look and I don't think this is valid syntax for using enums, just gives you a pointer to the enum.  You could do something like

`const ACCEPTED = IModerationStatus.ACCEPTED`
`const REJECTED = IModerationStatus.REJECTED`
`const AWAITING_MODERATION = IModerationStatus.AWAITING_MODERATION`

etc etc but I don't think it's worth the refactor time.

PR #3205, opened 2024-01-20
chore: update date-fns

mariojsnunes:
PR Checklist

PR Type

## Description

Updates date-fns to the latest version.

### DIFF
--- a/src/pages/common/UserNameTag/UserNameTag.tsx
+++ b/src/pages/common/UserNameTag/UserNameTag.tsx
@@ -1,7 +1,7 @@
-import { format } from 'date-fns'
 import { Username } from 'oa-components'
 import { isUserVerifiedWithStore } from 'src/common/isUserVerified'
 import { useCommonStores } from 'src/index'
+import { formatDate } from 'src/utils/date'
 import { Flex, Text } from 'theme-ui'
 
 interface UserNameTagProps {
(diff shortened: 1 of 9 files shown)


src/pages/common/UserNameTag/UserNameTag.tsx
```diff
@@ -19,7 +19,7 @@ export const UserNameTag = ({
 }: UserNameTagProps) => {
   const { aggregationsStore } = useCommonStores().stores
 
-  const dateText = `| ${action} on ${format(new Date(created), 'DD-MM-YYYY')}`
+  const dateText = `| ${action} on ${format(new Date(created), 'dd-MM-yyyy')}`
```
thisislawatts, line 22: **question:** Whilst we are in here @mariojsnunes, what do you think about refactoring all of these format calls through to a shared `formatTimestampForDisplay` helper which we use everywhere. It would make it much more straight forward if we wanted to change in the future to relative date stamps.

mariojsnunes: hmm I think that would be too much abstraction here.
maybe instead we could have constants for common formats like `const DATE_ONLY = 'dd-MM-yyyy'`?

thisislawatts: If they are all sharing the same constant it seems preferable for it to be encapsulated into a single module which also reduces the number of imports for `date-fns`, otherwise all of these files are importing both `format` and the constant. Ultimately though I defer to the person writing the PR here, its not a blocker from my side 😉

mariojsnunes: okay I'm convinced :)

PR #3214, opened 2024-01-22
fix: user settings so that impact isn't overwritten

benfurber:
PR Checklist

PR Type

## Description

There have been issues around the impact data not appearing to persist after a user saves it. The biggest reason for this is when a user updates the impact data and then saves their profile, the impact data was being overwritten by the old impact data state being stored.


iSCJT (approved):
Happy to approve this as it gets around one of the problems Marcella found in testing.  I think we need to look at hoisting all of the save logic up to the settings form though and have the main save also save the impact data.

PR #3218, opened 2024-01-22
feat: most relevant sort option only present if search value entered

iSCJT:
PR Checklist

PR Type

## Description

Most relevant sort option only present if search value entered.  If search value deleted then it will return to the sort option chosen before the search value was entered.

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/src/pages/common/SortFilterHeader/SortFilterHeader.tsx
+++ b/src/pages/common/SortFilterHeader/SortFilterHeader.tsx
@@ -74,7 +74,7 @@ export const SortFilterHeader = ({
       value: label,
     }))
     .filter((option) => option.value !== activeSorter)
-    .filter((option) => option.value !== ItemSortingOption.MostRelevant)
+    .filter((option) => option.value !== ItemSortingOption.SearchResults)
 
   const defaultSortingOption =
     Array.isArray(allSortingOptions) && allSortingOptions.length > 0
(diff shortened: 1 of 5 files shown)


iSCJT:
@davehakkens this should address your concerns.

davehakkens:
thanks for the quick "fix"
This seems more logical to me.
But still not sure if I fully get:

Because if you search "upload images" it shows 3 results
afbeelding

From here I could sort those results on "most useful"
afbeelding

Or "Most updates"
afbeelding

That works pretty neat. Dont see how/when you would use the "Most relevant"
That said happy to merge it. Just trying to make sure I understand it for future use.

iSCJT:
Go onto precious plastic how to's and search for something like climbing brush or solar. Should demonstrate the problem a bit better.

https://discord.com/channels/586676777334865928/1192524295172264086

davehakkens:
I see what you mean @iSCJT 
Is it right to describe the function of "Most relevant" as "Search results"?
Or am I missing another function it does?

iSCJT:
@davehakkens "Most Relevant" made sense to me at the time as it orders by the most relevant items to your search based on the weighting we use for the fields we search. In the background every item gets assigned a score from 0.00 - 1.00. If you think "Search Results" or another label would make more sense I'm happy to change it over.

I could also look at maybe cutting off the results so any items with a score below x (would need to test to see what looked sensible) didn't get displayed.

davehakkens:
I would think search always implies its sorted by "most relevant" since that is its job.
Would be up for changing the label to "Search Results" to avoid confusement. 

Cut of sounds nice. But that function might already be somewhat there? 
At least it looks like that on PK research. Like the example above only shows 3 results when searching "upload images"

benfurber (approved):
LGTM. One thought, as preSearchSorter is going into most of the stores, should it go into the module store instead?

benfurber:
Oh and, no tests around `src/pages/common/SortFilterHeader/SortFilterHeader.tsx`...?

iSCJT:
> LGTM. One thought, as preSearchSorter is going into most of the stores, should it go into the module store instead?

Good point, there's quite a lot of duplication across the stores that could be refactored out I think.  I'll look into it under a separate pr along with some tests.

PR #3222, opened 2024-01-24
Feat/question filters

### LINKED ISSUE #2968: [QuestionListing] Sorting filter

mariojsnunes:
PR Checklist

PR Type

## Description

Followup on #3121

Queries firebase directly for fetching questions.
It's possible to filter, sort, and load more items.

This is the new approach, ready for server-side rendering, which doesn't use mobx.
Items are always requested from the database on each search/filter.
Fetches only 10 (configurable) items at a time, instead of the whole collection.

Using the new, modular API from firestore. https://firebase.google.com/docs/firestore/query-data/queries#or_queries
(required for `or` queries)

Deprecated Random filter as I don't think it's possible with firestore queries.

Remaining work:
- [ ] Create firestore index for category filter (requested in the comments below)
- [ ] Test/fix category filter
- [ ] Discuss the migration to populate the keywords array

## Git Issues

Closes #2968


mariojsnunes:
Hit a roadblock with text search. Firestore is really a pain to work with 🥲
Currently the text search has 3 limitations:
1. Search is case-sensitive
2. Can only search from the beginning of a field
3. Can only search on 1 field at a time

https://stackoverflow.com/questions/46568142/google-firestore-query-on-substring-of-a-property-value-text-search

The solution is to have a "keywords" array field where we insert all the words, converted to lower-case, from fields we want to search (title and description)
Then we could use array-contains-any to search.
@thisislawatts do you have an opinion on this?

iSCJT:
> Hit a roadblock with text search. Firestore is really a pain to work with 🥲

That's a polite way of putting it, I hope we can start the process of moving to a proper database at some point this year.

> Then we could use array-contains-any to search.

Any idea what the performance of this will look like? I guess one thing in our favour is we limit most text entry fields to a relatively small amount of characters.

mariojsnunes:
> Any idea what the performance of this will look like? I guess one thing in our favour is we limit most text entry fields to a relatively small amount of characters.

Couldn't find much on the topic, but shouldn't be a concern... don't see any other option anyway...

- https://www.reddit.com/r/Firebase/comments/gxguo8/do_queries_using_arraycontainsany_or_in_take/
- https://stackoverflow.com/questions/60470865/how-much-does-the-size-of-the-array-in-an-array-contains-firestore-query-impact (but this is for array-contains not array-contains-any)

thisislawatts:
@mariojsnunes my preference here would be to stick with the existing Fuse pattern and look to iterate on the search in way which could benefit the How-to and Research content types as well.

mariojsnunes:
> @mariojsnunes my preference here would be to stick with the existing Fuse pattern and look to iterate on the search in way which could benefit the How-to and Research content types as well.

we can't move to ssr with fuse. fuse requires getting the entire collection and there is no store to cache on the server.

thisislawatts:
Any replacement search mechanism needs to operate on an entire collection. As you have found here Firestore is not optimised for Search so we would likely need a separate data store to meet this use case. 

I would rather tackle search in isolation than couple Questions rollout to this work. Particularly as search functionality only becomes relevant once we reach a critical volume of content. 

I appreciate you flagging this as a blocker to SSR though.

mariojsnunes:
I think we can be optimistic about it, this "workaround" wasn't that difficult to implement. 
While Firestore search is not ideal, I'd argue it's still worth it as we transition to a better solution.

For this PR I just need help creating a firestore index.
Getting this error when filtering categories:
>  FirebaseError  The query requires an index. You can create it here:
> https://console.firebase.google.com/v1/r/project/precious-plastics-v4-dev/firestore/indexes?create_composite=CmZwcm9qZWN0cy9wcmVjaW91cy1wbGFzdGljcy12NC1kZXYvZGF0YWJhc2VzLyhkZWZhdWx0KS9jb2xsZWN0aW9uR3JvdXBzL3F1ZXN0aW9uc19yZXYyMDIzMDkyNi9pbmRleGVzL18QARoMCghjYXRlZ29yeRABGgwKCF9jcmVhdGVkEAIaDAoIX19uYW1lX18QAg
> 

Everything else should be working and can already be reviewed and tested :)

mariojsnunes:
> For this PR I just need help creating a firestore index. Getting this error when filtering categories:
> 
> > FirebaseError  The query requires an index. You can create it here:
> > https://console.firebase.google.com/v1/r/project/precious-plastics-v4-dev/firestore/indexes?create_composite=CmZwcm9qZWN0cy9wcmVjaW91cy1wbGFzdGljcy12NC1kZXYvZGF0YWJhc2VzLyhkZWZhdWx0KS9jb2xsZWN0aW9uR3JvdXBzL3F1ZXN0aW9uc19yZXYyMDIzMDkyNi9pbmRleGVzL18QARoMCghjYXRlZ29yeRABGgwKCF9jcmVhdGVkEAIaDAoIX19uYW1lX18QAg

Need a maintainer's help to create that pesky index please :)

thisislawatts:
@mariojsnunes apologies for the delay getting back to you on this. As mentioned in my earlier comment there is a preference at this time is for consistency across our 3 content types and tackle search holistically rather than apply a workaround for Questions. The decision here is based on the existing differences between our modules which carry outsized maintenance costs and slow down feature delivery. 

In addition manually adding indexes to our Firestore DB creates further fragility to our systems. We should aim to manage indexes through automation rather than rely on manual processes. Its not something we have at the moment but looks like we could set this up: https://firebase.google.com/docs/reference/firestore/indexes/

Introducing a new search mechanism is out of scope for this Question module.

mariojsnunes:
That's a shame, cause this is a much better solution than the current one.

I please ask you to reconsider because:

- It scales better and is faster
- This enables SSR today!
- The code is better, less buggy and easier to maintain
- We can easily change Database in the future, just update the service calls
- This is the solution we have to implement anyway on other modules, so consider this the first step of the migration and I'll be happy to work on other 2 modules

davehakkens:
Hi @mariojsnunes. Just some additional background information. 
Many of our decisions are based on the availabilty and capacity or the maintainers. At the moment many are occupied wrapping up our new "nested comments" feature and "questions module". Plus the platform continues to grow in complexcity but there is only so much we can hold with a small team. This is good to hear though` and I'll be happy to work on other 2 modules` 

Perhaps a good topic to bring to the upcoming maintainer call to see whats a good way forward to improve the platform without stretching our capacity. What do you think @thisislawatts?

thisislawatts:
Hey both, I've raised a new Issue to capture migrating search away from the in memory Store. https://github.com/ONEARMY/community-platform/issues/3307

PR #3229, opened 2024-01-29
Feat(contact-form): email verified check

### LINKED ISSUE #3075: Feature: Block users from being able to send contact messages
## Description

Currently under beta, but soon to be released, is a contact form. It will allow logged in users to messages workspaces. Right now if a user abuses the messaging feature, the only option is for admins to delete the user account.

We want to change it so that a user must be authenticated through the sign-up journey AND is not blocked from sending messages.

## Build suggestion

Add a utility for both `UserContactForm` on the frontend and the `isValidEmailCreationRequest` function that checks a user has verified their email address and does not have the `isBlockedFromMessaging` set to true on their account.

For the form, it's fine just to keep the form hidden if the conditions aren't met.

## Testing

At a minimum please include these new scenarios in the `UserContactForm` test script. But you'll probably also create a little util so add a test for that too. Plus I'd personally love it if a cypress test is added for this too.

benfurber:
PR Checklist

PR Type

## Description

This requires user accounts to have their email address verified before they can send messages.

## Git Issues

Closes #3075


davehakkens:
Nice. But shouldn't we first make the email confirmation work properly?

benfurber:
@davehakkens Email verification issues are low now. Worth me updating this branch and merging in?

PR #3269, opened 2024-02-12
Feat: Add crop image component

benfurber:
PR Checklist

PR Type

## Description

This begins to add the functionality to crop images before upload. Users (members especially) upload profile images that are the wrong shape for how they're used, this will fix this by having the user set the crop before uploading it. 

## Screenshots/Videos

Screenshot 2024-02-12 at 15 13 48


benfurber:
@davehakkens Is this the kind of crop feature you had in mind or have I over designed it?

davehakkens:
> @davehakkens Is this the kind of crop feature you had in mind or have I over designed it?

@benfurber I dont see it it the preview link. Works just as before.
Am i looking at the wrong place?

afbeelding

benfurber:
@davehakkens I'm trying to really break down the feature into small chunks so right now the component is only in the component library... 

Tomorrow I'll temporarily add it somewhere random on the app so you can see it.

davehakkens:
Sounds good @benfurber 
Hard to judge from a screenshot, so better to check it out somewhere on the app

benfurber:
@davehakkens I've added it to the member settings page below the profile image. 

The idea when the implementation is finished:
1. The new component is hidden by default.
2. User selects an image from their device as per usual
3. This component appears
4. User selects the crop they want and selects 'upload'
5. Loading state appears while the image uploads
6. Component disappears when the image upload is finished.

davehakkens:
ok nice! Don't think the component is over-designed since it will be used by many.
I'd be more worried about the robustness of it. Making sure that it functions well.
For instance i'm not sure how much of the full functionaility is there atm, but perhaps thats a deeper test for later?


**Other note:**
It could use a bit of alignment with the other component.
Now it says twice "upload image"
1. We could change the top one from "upload image" to "select image"
2. Or if we want to keep the changes in the component. Change the bottom one from "upload image" to "submit image"
afbeelding

benfurber:
@davehakkens Nice. Thanks. Good spot, for when this component is implemented I think - I'll remember.

thisislawatts (approved):

PR #3281, opened 2024-02-18
chore: refactor to adopt firestore modular pattern

thisislawatts:
PR Checklist

PR Type

## Description

https://firebase.google.com/docs/web/modular-upgrade

- [x] Verify that emulators work
- [x] Dig into failing Cypress tests
- [x] Restore limit functionality


src/stores/databaseV2/DatabaseV2.tsx
```diff
@@ -15,15 +15,16 @@ export class DatabaseV2 {
   private _clients: DBClients
 
   constructor() {
-    this.clients = this._getDefaultClients()
+    const useBrowserCacheDb =
+      !window.location.search.includes('no-cache') ||
+      process.env.NODE_ENV !== 'test'
```
thisislawatts, line 20: Move existing lookup to constructor and extended check to prevent IndexDB being used in Jest env as it is not available. This should ultimately move to a higher level so that we remove knowledge of `window` and `process` from this file.

src/stores/databaseV2/DatabaseV2.ts
```diff
@@ -15,15 +15,16 @@ export class DatabaseV2 {
   private _clients: DBClients
 
   constructor() {
-    this.clients = this._getDefaultClients()
+    const useBrowserCacheDb =
+      !window.location.search.includes('no-cache') ||
+      process.env.NODE_ENV !== 'test'
```
benfurber, line 20: Should you be calling the process directly? I thought we'd been trying to get rid of that.

benfurber (approved):
Other than that call to process, looks good to me.

src/stores/databaseV2/DatabaseV2.ts
benfurber: (Just noticed you commented on this already, @thisislawatts!) Sounds like you think there's too much work for this PR to make the extra change and we should live with it for the moment?

Issue #3307, opened 2024-02-26
[feature request] Refactor search to remove dependency on all documents being loaded with fuse.js


thisislawatts:
**Is your feature request related to a problem? Please describe.**
The current search functionality for How-tos, Research and Questions relies on loading all documents using fuse.js, which can cause performance issues and inefficiencies, particularly with large datasets. This dependence on loading all documents poses a scalability challenge and can hinder the overall user experience.

**Describe the solution you'd like**
I propose refactoring the search functionality to eliminate the dependency on loading all documents with [fuse.js](https://github.com/krisk/Fuse). Instead, we should implement a more efficient and scalable search algorithm that allows for dynamic loading of documents based on user queries. This approach would improve performance and user experience, especially for larger datasets.

**Describe alternatives you've considered**
One alternative is to optimize the existing fuse.js implementation by fine-tuning its configuration and indexing process. However, this may only provide marginal improvements and still not address the fundamental issue of scalability. Another alternative is to explore different search libraries or algorithms that offer better support for dynamic loading and scalability.

**Additional context**
It's important to prioritize this feature request as it directly impacts the performance and usability of the search functionality within our application. By refactoring the search to remove the dependency on loading all documents with fuse.js, we can enhance the overall efficiency and scalability of our system, leading to a better user experience. For example this would unblock using SSR. 

@mariojsnunes has previously [done some work around introducing this for Questions](https://github.com/ONEARMY/community-platform/pull/3222/files#diff-589fd2579db34140e38e0174872e669c13c446abba95139b4ac65ad34dc648b2R24-R73). Ideally we want a single solution that can be used for How-tos, Research and Questions. 

Consider that Firebase charges for each read or query operation. So after paying the cost of fetching all documents for a collection, all user searches are "free" because the data has already been fetched from the origin. 
https://firebase.google.com/docs/firestore/pricing

mariojsnunes:
Thanks for raising this!

Currently the code for How-tos and Research is tied up with FilterSorterDecorator. But each module has specific needs which lead to some hardcoded conditions.

For example:
hardcoded properties:
![image](https://github.com/ONEARMY/community-platform/assets/8073622/d1d1b91f-3eae-41c2-bdd7-0cd35a7698e9)

mariojsnunes:
As of now, I haven't found a blocker for using the firestore api for our search.
It's not ideal, but I haven't found a better solution that wouldn't either cost more or take significant more effort.

[...]

PR #3349, opened 2024-03-14
feat: question filters

mariojsnunes:
PR Checklist

PR Type

## Description

Followup on https://github.com/ONEARMY/community-platform/pull/3121, #3222

Queries firebase directly for fetching questions, with pagination.
It's possible to filter, sort, and load more items.

This is the new approach, ready for server-side rendering, which doesn't use mobx/fuse.js.
Items are always requested from the database on each search/filter.
Fetches only 10 (configurable) items at a time, instead of the whole collection.

Using the new, modular API from firestore. https://firebase.google.com/docs/firestore/query-data/queries#or_queries
(required for `or` queries)

Deprecated Random filter as they are not possible with firestore queries.
Search no longer supports fuzziness, but I don't think it's that useful or a blocker.

The required indexes are defined on a firebase config file and automatically deployed.

## Git Issues

Closes https://github.com/ONEARMY/community-platform/issues/2968
Solves https://github.com/ONEARMY/community-platform/issues/3307 for Questions Module


src/stores/Question/question.store.tsx
```diff
@@ -169,12 +170,14 @@ export class QuestionStore extends ModuleStore {
...
     const user = this.activeUser as IUser
     const creatorCountry = this.getCreatorCountry(user, values)
+    const keywords = getKeywords(values.title + ' ' + values.description)
 
     await dbRef.set({
       ...(values as any),
       creatorCountry,
       _createdBy: values._createdBy ?? this.activeUser?.userName,
       slug,
+      keywords: Array.from(keywords),
```
benfurber, line 180: The tests should be updated to cover this.

src/utils/searchHelper.ts
```diff
@@ -0,0 +1,7 @@
+import { stopwords } from './stopwords'
```
benfurber, line 1: Would love a little test or two for this util please. Perfect low hanging fruit to keep the coverage OK.

src/utils/stopwords.ts
```diff
@@ -0,0 +1,129 @@
+export const stopwords = new Set([
```
benfurber, line 1: I assume this list of words came from somewhere? Worth referencing in a comment in the file?

benfurber (changes requested):
That service is so nice to read @mariojsnunes. Love all this. And the questions load so much faster!

src/pages/Question/QuestionListing.tsx
```diff
@@ -1,23 +1,84 @@
...
+      }
+      setSearchParams(params)
+    } else {
+      // search only when sort is set (avoids duplicate requests)
+      fetchQuestions()
+    }
+  }, [q, category, sort])
+
+  const fetchQuestions = async (
+    skipFrom?: QueryDocumentSnapshot<DocumentData, DocumentData>,
```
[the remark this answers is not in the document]
mariojsnunes: Fair point, ideally firestore would support passing the item id, but I could only make it work passing the document snapshot. Maybe we could store it in memory at the service, but that could lead to other issues. I think this is the best option for now.

src/stores/Question/question.store.tsx
mariojsnunes: Can we do this on another PR? For SSR, this store will have to be removed. I could work on that next.

src/pages/Question/question.service.ts
```diff
@@ -0,0 +1,106 @@
+import {
```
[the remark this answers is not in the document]
mariojsnunes: done!

src/utils/searchHelper.ts
mariojsnunes: done!

mariojsnunes:
I've added ts-jest package to make it easier to run tests locally

src/pages/Question/labels.ts
```diff
@@ -7,8 +7,9 @@ export const buttons = {
 }
 
 export const headings = {
-  create: 'Ask your question to the community',
-  edit: 'Edit your question to the community',
+  create: 'Ask a question',
+  edit: 'Edit your question',
+  list: 'Ask your questions and help others out',
```
benfurber, line 12: This undoes some copy changes I made based on user feedback.

benfurber: If there's only little things like this that need changing, I hope you don't mind me going to the branch myself so I can get this all merged in asap.

benfurber (approved):

src/pages/Question/labels.ts
mariojsnunes: not at all, please go ahead

PR #3321, opened 2024-03-02
fix: account settings forms

benfurber:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

I used the change to tidy up the forms in align with approaches I've taken on other forms (form messaging, labels, etc). I'm betting the way I've written the cypress test will become a pain. Would be delighted with ideas for a better approach.

## Git Issues

Closes #3320 

## Screenshots/Videos

Email:
https://github.com/ONEARMY/community-platform/assets/16688508/f62d31b8-03e5-44fd-905f-849ac973cec6

Password:
https://github.com/ONEARMY/community-platform/assets/16688508/84526e18-f4dc-49dc-9c8b-2495918a4df6

Issue #3221, opened 2024-01-23
[QuestionPage] Add Image Upload


davehakkens:
### Description

Add the ability to attach images when posting a question.
Component and behaviour can be copied from the research module

### Page related

Will be used in :
- [Question page](https://dev.onearmy.world/questions/how-come-the-edit-button-disappears-when-commenting)
- [Question create ](https://dev.onearmy.world/questions/create)

### Mockup
![questions_images_added](https://github.com/ONEARMY/community-platform/assets/13672737/44313341-edfa-4c78-bf1c-2210f0e02a1a)
![questions_images](https://github.com/ONEARMY/community-platform/assets/13672737/3fd96002-4d02-4b31-b52e-7bcc423f22e4)

Robert-LC:
Hello first time contributing, wanted to throw my hat in to work on this before someone else grabs it.
![image](https://github.com/ONEARMY/community-platform/assets/72999492/96bad0ab-418d-498c-af73-909342e49446)

I have the image upload working on the front end, just need to figure out how to get the form to submit to the back end if images are in it.

Thanks.

PR #3366, opened 2024-03-16
feat: add image upload to questions

Robert-LC:
PR Checklist

PR Type

## Description

Add image upload to questions

## Git Issues

Closes [#3321](https://github.com/ONEARMY/community-platform/issues/3221)

## Screenshots/Videos

image
image

Also as a side note for a separate issue I think refactors might of introduced a styling bug, because I don't believe it did this when I was adding this code before the refactors.  Tags drop down messes up footer

image


Robert-LC:
It was my code that caused it, fixing now

src/pages/Question/Content/Common/FormFields/QuestionImage.field.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Flex, Label } from 'theme-ui'
+
+const ImageInputFieldWrapper = styled.div`
+  width: 150px;
+  height: 100px;
+  margin-right: 10px;
+  margin-bottom: 6px;
+`
+
+export const QuestionImagesField = ({ inputsAvail }) => {
```
benfurber, line 15: This should definitely be typed please and keep to the pattern across the codebase naming the interface IProps.

src/pages/Question/Content/Common/FormFields/QuestionImage.field.tsx
```diff
@@ -0,0 +1,38 @@
...
+  margin-bottom: 6px;
+`
+
+export const QuestionImagesField = ({ inputsAvail }) => { 
+    return (
+        <>
+            <Flex sx={{ flexDirection: 'row', flexWrap: 'wrap', alignItems: 'center',}} mb={3}>
+                <Label htmlFor={ `images`} sx={{ marginBottom: 2 }}>
+                    {LABELS.fields.images.title}
+                </Label>
```
benfurber, line 21: All this should be possible to replace with `FormFieldWrapper`.

src/pages/Question/Content/Common/FormFields/QuestionImage.field.tsx
```diff
@@ -0,0 +1,38 @@
...
+                {[...Array(inputsAvail)].map((_, i) => (
+                    <ImageInputFieldWrapper
+                    key={`image-${i}`}
+                    data-cy={`image-${i}`}
```
benfurber, line 25: I think `image-upload-{i}` would be a slightly better label for the tests please.

src/pages/Question/Content/Common/FormFields/QuestionImage.field.tsx
```diff
@@ -0,0 +1,38 @@
+import { Field } from 'react-final-form'
+import styled from '@emotion/styled'
+import { ImageInputField } from 'src/common/Form/ImageInput.field'
+import * as LABELS from 'src/pages/Question/labels'
```
benfurber, line 4: I'm not a fan of this style of import. I'd prefer `import { fields } from ...` please.

src/pages/Question/Content/Common/FormFields/QuestionImage.field.tsx
benfurber: Oh and full words please, so `inputsAvailable`.

src/pages/Question/Content/Common/QuestionForm.tsx
```diff
@@ -55,8 +56,13 @@ export const QuestionForm = (props: IProps) => {
       }}
       mutators={{ setAllowDraftSaveFalse }}
       initialValues={props.formValues}
-      render={({ submitting, handleSubmit, form }) => (
-        <Flex mx={-2} bg={'inherit'} sx={{ flexWrap: 'wrap' }}>
+      render={({ submitting, handleSubmit, form, values }) => {
+        // Make sure max only 4 images
+        const numImgInputsAvail = values?.images
+        ? Math.min(values.images.reduce((count, image) => image ? count + 1 : count, 0) + 1, 4)
+        : 1;
```
benfurber, line 63: I'm pondering this. And I can appreciate why you might have ended up with this and `numImgInputsAvail` in QuestionForm but I find the end result difficult to read... 

I'd suggest starting with creating a constant in `src/pages/Question/constants.ts` which sets the max image count and that this file calls that constant.

Is there a reason it needs to be more complex than `numberOfImageInputsAvailable` on `ResearchUpdateForm`?

src/pages/Question/Content/Common/QuestionForm.tsx
```diff
@@ -55,8 +56,13 @@ export const QuestionForm = (props: IProps) => {
       }}
       mutators={{ setAllowDraftSaveFalse }}
       initialValues={props.formValues}
-      render={({ submitting, handleSubmit, form }) => (
-        <Flex mx={-2} bg={'inherit'} sx={{ flexWrap: 'wrap' }}>
+      render={({ submitting, handleSubmit, form, values }) => {
+        // Make sure max only 4 images
```
benfurber, line 60: With the constant set, this comment can be removed.

src/stores/Question/question.store.tsx
```diff
@@ -170,11 +171,24 @@ export class QuestionStore extends ModuleStore {
...
+    // Check for images
+    const updateWithMeta = {...values}
+    if(values.images.length > 0) {
+      const imgMeta = await this.uploadCollectionBatch(
+        values.images.filter((img) => !!img) as IConvertedFileMeta[],
+        COLLECTION_NAME,
+        dbRef.id, 
+      )
+      updateWithMeta.images = imgMeta
+      logger.debug('upload iamges ok')
```
benfurber, line 183: typo

benfurber, line 183: I think this would be nice as as private function then this is ```const images = await this.loadImages(values.images, dbRef.id)```.

src/stores/Question/question.store.tsx
```diff
@@ -170,11 +171,24 @@ export class QuestionStore extends ModuleStore {
...
+    }
+
     await dbRef.set({
       ...(values as any),
       creatorCountry,
       _createdBy: values._createdBy ?? this.activeUser?.userName,
       slug,
+      images: updateWithMeta.images,
```
benfurber, line 191: Deffo need to update the store tests to check this happens.

src/pages/Question/Content/Common/FormFields/QuestionImage.field.tsx
Robert-LC: Not sure what you mean by this, I called a Field Component and then simply used control + . and chose to import that way.

src/pages/Question/Content/Common/QuestionForm.tsx
Robert-LC: I simplified it a little bit in my new code, but the reason I have to do this differently that ResearchUpdateForm is because values?.images is null until a first img is uploaded then it works properly.

So I have to initially set it to 1, and then it will start counting properly to a max of 4.

Vs in ReasrchUpdateForm something is different about that component where values.images is not null even if its at 0 on load.


Both this and images in research will need a refactor later on, because while they work when adding images, on removal of images the counts get messed up on both and they do not decrease. But with the way the components are parents of each other and how data is passed it will take a while to figure out.

Robert-LC: The wording will also need to be changed, as inputsAvailable isnt a good word.

There should always be one InputAvailable until unless there are 4 images, but with the way things are passed down to child I see why its worded this way.

My confusion is because the amount of images is stored in QuestionForm, but QuestionImagesField is what controls if another input button is shown or not

src/pages/Question/Content/Common/FormFields/QuestionImage.field.tsx
benfurber: Rather than:
```
import * as LABELS from 'src/pages/Question/labels'
```

Let's have:
```
import { fields } from 'src/pages/Question/labels'

File: src/pages/Question/constants.ts [hunk 9]
```diff
@@ -1,3 +1,5 @@
 export const QUESTION_MIN_TITLE_LENGTH = 10
 export const QUESTION_MAX_TITLE_LENGTH = 60
 export const QUESTION_MAX_DESCRIPTION_LENGTH = 1000
+export const QUESTION_MIN_IMAGE_INPUTS = 1
```
benfurber, line 4: Images shouldn't be a required field so no need for this.

Robert-LC:
So I found a partial fix for the tag dropdown moving the footer down but not a true fix.

If I move tags to be above images input it won't have that problem. 

I tried to investigate CSS, and setting the tags dropdown to 'position absolute' breaks its placing in the card and its length, but it fixes the dropdown issue.


Couldn't find in the code where to alter the CSS for the dropdown list itself in active state.

benfurber (approved):

benfurber:
@all-contributors please add @Robert-LC for code

PR #3370, opened 2024-03-17
feat: add support badge Username component

Robert-LC:
PR Checklist

PR Type

## Description

Adds a support badge to the username component, with some conditions.
If the user is verified and a supporter, it shows a verified badge.
If they are only a supporter, it shows a supporter badge.

## Screenshots/Videos
Supporter

image

Verified Supporter

image


benfurber (changes requested):
Nice one, thanks @Robert-LC. Could we add/check in any of the cypress tests that the data is flowing through and displaying this correctly?

Robert-LC:
I went through and fixed some errors that popped because I didn't reference the new IsSupporter prop on existing Username component calls. Still working on the Cypress tests.

davehakkens:
How are the Cypress tests going @Robert-LC?
Would be a nice one to merge before we start here in Project Kamp :)

Robert-LC:
@davehakkens, I'll have to get with @benfurber and pair program this one out, as I've never really done anything like this before, and I need help connecting the dots.

I dove into the Patreon and OAuth connection flow, and not sure what the best way to conduct these tests are.

My understanding of the flow is:
 - We send them to the oauth2 page with a redirect to our `/patreon` page once they connect their account
 - With that we get back `access_token` 
 - use that token to query Patreon API to get back a `user`
 - Build data into a `PatreonUser` object
 - If their a supporter, set them as a supporter
 - Save the user and all its updated properties into the db

I'm having trouble on where to put the test, how in-depth I should go, and where/what I should intercept.

Right now, I've added the beginnings of a test to `settings.spec.ts`, where I do the following

- Login
- Navigate to settings
- Click the connect to Patreon button

Here is where I get confused. 

Should I `cy.intercept()` the initial GET request that sends them to the oauth2 page?

Or should I actually login with an account we have for testing(not sure if we do), get back a real response, and then from there redirect to /patreon, let it call the POST to `/oauth2/token`, get back the `access_token`, and finally intercept the `/api/ouath2/v2/identity` call?

Then another part of me thought, "Should this even be a test in `settings.spec.ts`?" 

Maybe, its better to find a user in the db that is a supporter (or create one), and simply make a e2e test of them logging in, making a comment, making a post, etc, and checking if their badge is visible.

benfurber:
@Robert-LC Let's sync and give the testing one more go in a follow-up PR. As the feature seems to work, I'll tidy-up in the branch to get CI passing so we can merge now this in as is.

PR #3410, opened 2024-04-02
chore: deal with jest console warnings

### LINKED ISSUE #3303: Deal with all the jest console warnings
There's loads of console warnings/errors that make reading through the test suite annoying and difficult to find real test errors, especially on CI.

This error is common:
```
Warning: An update to observerComponent inside a test was not wrapped in act(...).
```
These errors should be dealt with as advised by the message.

Another log that's typical is a console log we're choosing to call, in which case the tests should be calling expectations, etc. in a block so that the actual log isn't visible when running the tests.

Robert-LC:
PR Checklist

PR Type

## Description

I dealt with most of the Jest console warnings as well as the unnecessary logging appearing.

- For `logger` itself, I expected the test log entries of debug, info, warn, fatal instead of displaying them.

- For `Message.store.tsx` I removed the logger.error calls because the test looks for Errors thrown so no need to cloud the logging, if the tests will fail in that regard.

- For files that called logger.info, I switched to logger.debug instead to hide the calls in the tests, but keep them around for debugging. Let me know if you think this what the right choice for those items. 

- There were a few tests like in `user.store.test.tsx` where it was testing for if two users were returned form the same id and it would return a warning, but if it wouldn't cause a failure of the test. 

So for those type of scenarios, I decided to mock the warning to hide from the test suite, since its not failing on expect. Let me know if you want something different in that scenario.

Lastly, while trying to figure out the best approach, I noticed that in `functions/src/test/firebase/logger.ts` there was at one point a top level mock that spied on all firebase logger methods

## Git Issues

Closes #3303

### DIFF
--- a/src/stores/Message/message.store.tsx
+++ b/src/stores/Message/message.store.tsx
@@ -1,6 +1,5 @@
 import { createContext, useContext } from 'react'
 import { action, makeObservable } from 'mobx'
-import { logger } from 'src/logger'
 import { isUserBlockedFromMessaging } from 'src/utils/helpers'
 
 import { ModuleStore } from '../common/module.store'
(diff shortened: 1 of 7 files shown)


src/stores/Message/message.store.tsx
benfurber, line 1: Disagree with this change. logger is doing what it should be here.

src/utils/cdnImageUrl.test.ts
benfurber, line 1: Rather than mocking logger, the tests should be anticipating the call to logger happening.
https://stackoverflow.com/questions/49096093/how-do-i-test-a-jest-console-log

src/stores/User/user.store.test.tsx
benfurber, line 1: Rather than mocking logger, the tests should be anticipating the call to logger happening.
https://stackoverflow.com/questions/49096093/how-do-i-test-a-jest-console-log

benfurber (changes requested):
Thanks for your patience waiting to hear back on this one. Does it solve all the console logging stuff?

benfurber:
Hey @Robert-LC hope you're well, will you have a chance to have a look at this again soon?

PR #3429, opened 2024-04-06
feat: add guard clause to capitalizeFirstLetter

benfurber:
PR Checklist

PR Type

## Description

Somehow an entirely empty how-to got it's way into the database. That caused the main how-to page to entirely crash for admins who could see the post (even though it wasn't published for everyone else).

### DIFF
--- a/src/utils/helpers.test.ts
+++ b/src/utils/helpers.test.ts
@@ -42,6 +42,10 @@ describe('src/utils/helpers', () => {
     expect(capitalizeFirstLetter('hello world')).toBe('Hello world')
   })
 
+  it('capitalizeFirstLetter should return an empty string when a wrong type provided', () => {
+    expect(capitalizeFirstLetter(undefined)).toBe('')
+  })
+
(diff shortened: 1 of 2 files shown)


mariojsnunes:
hmm ideally there should be a conditional when calling the method `capitalizeFirstLetter(word || '')` as it already only accepts `string`, not `undefined` or `null`.
Improving our usage of typescript should help with this kind of bugs.

benfurber:
@mariojsnunes Have added. How-to's shouldn't have an empty title though, I'd say it's typed correctly otherwise everything would have to be optional?

mariojsnunes:
Sorry I wasn't clear. I meant to say that your first changes shouldn't be necessary, because
`capitalizeFirstLetter` only accepts a string already, not null/undefined.
The root cause is likely somewhere higher up where an assignment to `howto.title` is undefined.

benfurber:
I had been leaving this open to remind myself to do the bigger fix but I need it now to trigger a release candidate(!) so going to merge it in.

PR #3460, opened 2024-04-20
Remove totalUseful aggregations

mariojsnunes:
PR Checklist

PR Type

## Description

before:
- totalUseful exists in an aggregation `{ username: number }`
- each store (research, howto, question) updates the aggregation
- need to download the whole aggregation
  - 1 read per profile page view
  - bandwidth will increase with #users

after:
- `totalUseful` is a user profile property
- each store (research, howto, question) updates the user `totalUseful` property
- displaying totalUseful is free (comes with the other user properties)
- removed `research.aggregations`, `howto.aggregations` (were only used for `totalUseful`)
- simplified `aggregations.store` since its only purpose now is `verified_users`
- simplified `userNotifications.aggregations` as it's no longer concerned with `totalUseful`

half the file changes were to fix/update tests

## Git Issues

Closes #


mariojsnunes:
Still need to fix tests... but is ready for review

mariojsnunes:
Okay this might be it, test are fixed

benfurber:
Thanks @mariojsnunes. I'm starting to look through this. 

@thisislawatts - I know general guidance in the past has been to move onwards aggregations? Thoughts on this approach?

benfurber (commented):
Some initial comments: 

1. Is our e2e coverage really good enough that no changes are needed?
2. I've tried to use `FieldValue.increment` a couple of times and found it unreliable at low number changes. I had issues that lead me to this: https://github.com/invertase/react-native-firebase/discussions/5914
3. There's still a test commented out in `src/pages/Howto/Content/Howto/Howto.test.tsx`.
4. Lots of changes aren't paired with unit test changes, which I'm assuming means gaps in coverage, eeerrr, fancy adding the coverage please? At least behaviour linked to the PR.

mariojsnunes:
Thanks for the notes @benfurber
The main reason to remove this aggregation is that we can build the same functionality without it, just need an extra profile field which is free in terms of reads and bandwidth.

thisislawatts:
There is a lot going on in that PR @mariojsnunes , which makes it a bit harder to reason with. My preference would be for distinct changes to be split across multiple commits and/or PRs to reduce the number of changes here. For example a preparatory refactoring PR/commit that then makes the change you want to make, easier](https://x.com/KentBeck/status/250733358307500032).

Otherwise the ambition of this PR to remove the TotalUseful Aggregation seems reasonable, I am missing the context here on what problem we were trying to solve with this aggregation.

thisislawatts:
@benfurber did you get this from some documentation?
> I know general guidance in the past has been to move onwards aggregations?
My understanding is that aggregations aim at solving a very specific data access pattern rather than something we should default to. 

The firebase product has also evolved since we introduced our aggregations, so we may want to review our usage of aggregations to see if there are alternatives in Firestore which would simplify the code and reduce maintenance costs. I found this useful entry point to the conversation here:
https://stackoverflow.com/questions/77461961/how-should-i-handle-aggregated-values-in-firestore

mariojsnunes:
I've updated the description to better reflect what this PR does.

> 1. Is our e2e coverage really good enough that no changes are needed?
> 2. I've tried to use `FieldValue.increment` a couple of times and found it unreliable at low number changes. I had issues that lead me to this: [Firestore Fieldvalue.increment and duplicate writes invertase/react-native-firebase#5914](https://github.com/invertase/react-native-firebase/discussions/5914)
> 3. There's still a test commented out in `src/pages/Howto/Content/Howto/Howto.test.tsx`.
> 4. Lots of changes aren't paired with unit test changes, which I'm assuming means gaps in coverage, eeerrr, fancy adding the coverage please? At least behaviour linked to the PR.

Thank you @benfurber 

1. e2e test coverage is unchanged.
2. From what I understand it shouldn't be an issue unless there are concurrent requests, which isn't the case here
3. Good catch! Uncommented
4. The coverage is the same, we had duplicated tests at howto and research, it is now ensured at `toggleDocUsefulByUser.test`

benfurber (approved):
Like it all. Nice one.

PR #3466, opened 2024-04-21
feat: remove map streaming

mariojsnunes:
PR Checklist

PR Type

## Description

Removes streaming from map pins. Should reduce database reads.

BUT, even though is better, this is solution also doesn't scale with an increasing number of users and pins.
we should evaluate some kind of caching, maybe a cloud function?

## Git Issues

Closes #

### DIFF
--- a/src/pages/Maps/Maps.tsx
+++ b/src/pages/Maps/Maps.tsx
@@ -32,7 +32,7 @@ const MapsPage = observer(() => {
   })
 
   useEffect(() => {
-    retrieveMapPins()
+    mapsStore.retrieveMapPins(MAP_PROFILE_TYPE_HIDDEN_BY_DEFAULT)
     mapsStore.retrievePinFilters()
 
     const showPin = async () => {
@@ -45,7 +45,6 @@ const MapsPage = observer(() => {
     showPin()
 
     return () => {
-      mapsStore.removeSubscriptions()
       mapsStore.setActivePin(undefined)
     }
   }, [])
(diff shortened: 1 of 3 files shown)


chrismclarke:
@thisislawatts @mariojsnunes 
Just taking a closer look at some of the DB issues, and I'm not sure if this fix is going to help. The problem with using the `getWhere` request is that it's always operated against live server data, which means any time any user loads the map they will make a full read for the 3.7k map pins. Given the past month had around 33K users visiting the map that is now 122M reads (around $40 worth)

The stream operation works differently, because it caches the result of previous queries and only searches for updates. It also reads from a separate cache for users visiting for the first time, meaning that new users only make 1 read request for first batch of data and then just stream docs as updated. 

I've opened a PR to revert at #3472

mariojsnunes:
What if we have a function that updates the cache server? And always load pins from there

chrismclarke:
> What if we have a function that updates the cache server? And always load pins from there, without client-side caching

Yeah I definitely think that's a good direction to head in, I know Luke had been thinking along similar lines and starting to put together a couple proposals (maybe using firebase functions). There is also the possibility of using something like [firestore data bundles](https://firebase.google.com/docs/firestore/bundles) which aim to do similar.

But any work on the cache-server implementation would also probably benefit from a total rewrite, given that the current cache-server is based on firebase realtime db which was the very first DB we used back in 2017 (I think), and never really had much attention. So we might need to rethink where the cache data is stored - if using databundles then the cache is stored in firebase itself (so only 1 DB to worry about), or if we want to use a 3rd party/external db then could look into alternative platforms that are designed more for these types of operations (e.g. redis/memcached/dragonfly).

Issue #3467, opened 2024-04-22
[Discussions] Should only fetch comments when discussion is expanded 


mariojsnunes:
**Is your feature request related to a problem? Please describe.**
Currently we are fetching all comments as soon as the page is loaded.

**Describe the solution you'd like**
Since Research Update discussions are initially collapsed, we could display only the count and load the comments only when the user expands the comment section.

UPDATE: This not a big deal as the comments are stored as an array. So we only have 1 read per research update.
If we want to optimize further, we would need to store the commentCount for research updates.

~~For example, a page that has 3 research updates each one with 10 comments, is currently costing us 3*10=30 reads on the page load.
If we only display the count, it would cost only 3 reads (1 per update).
Although... if the user loads all comment sections, we would end up with 30+3 reads. But that should be the exception.~~

**Additional context**
Documentation on aggregation query pricing:
https://firebase.google.com/docs/firestore/pricing#aggregation_queries

thisislawatts:
This sounds good to me, originally I loaded all of the comments to align with the original implementation but now that we have split the discussions out to a separate collection this makes less sense.

Although I am not to sure about the pricing model you have outlined there, my understanding is that we are charged per document.

A Research document consists of an array of Research Updates, but these are not considered separate documents under the Firestore pricing so the Read operations for a Research page with 1000 updates and a Research page with 1 update is equal as 1. 

It is the same for the Discussions, we are currently using arrays for storing the comments. So there is a single Read cost for loading a Discussion regardless of how many comments are in the document.

mariojsnunes:
Thanks, I've updated accordingly.

benfurber:
I'm going to pick up the commentCount for research updates issue as we're still getting in a situation where the total count for research article as a whole is different to what the combined count of research updates.

PR #3470, opened 2024-04-22
feat: do not load discussions if it's a bot

### LINKED ISSUE #3468: [Optimization] Should not load Discussions if page is loaded by a bot/crawler
Loading discussions, consumes a lot of unnecessary database reads.

We can detect if page is being accessed by a bot and not load certain features, like discussions.

**Additional context**
Maybe this could be applied to map pins too?
Potential code solution: https://stackoverflow.com/a/33941034

mariojsnunes:
PR Checklist

PR Type

## Description

Stops discussions from loading if the page is requested by a bot/crawler.

## Git Issues

Closes #3468

## Screenshots/Videos

---

## What happens next?


src/utils/isBot.ts
```diff
@@ -0,0 +1,6 @@
+// Regex from here: https://stackoverflow.com/a/33941034
+export const isBot = () => {
```
thisislawatts, line 2: **suggestion:** Rather than relying on the `navigator.userAgent` can we pass this through as a param for the function? This would make it possible to test the method without stubbing globals.

mariojsnunes: hmm I'm conflicted about this... I understand what you mean, but...
- userAgent is just a string, so no way to enforce what is passed as a prop with typescript
- I don't think we need to test this function
- For unit tests, we can mock the function result
- For cypress, we can set the use-agent header https://stackoverflow.com/a/54745402

thisislawatts: I think there is usually value in testing around methods like this. Regular expressions can be hard to reason with, so tests are useful if only to document the intended use cases.

thisislawatts:
**question:** I might be missing the intention of this change, but won't this limit the ability for comments to be indexed by automated crawlers? This seems counter intuitive when we are looking to create a shared repository of knowledge, sure the primary articles are Research/How-to posts, but each additional comment contributes to that original item.

mariojsnunes:
> **question:** I might be missing the intention of this change, but won't this limit the ability for comments to be indexed by automated crawlers? This seems counter intuitive when we are looking to create a shared repository of knowledge, sure the primary articles are Research/How-to posts, but each additional comment contributes to that original item.

Yes it would. A few points to think about:

- We could allow some search index crawlers like google, but most of them are not great at waiting/running javascript, or even need comment information... I'd guess most crawlers that access our pages are for link previews.
- All these extra reads are expensive... need to at least paginate them, and with pagination, the first page comments you index today could be different tomorrow.

benfurber:
My comment is more about the issue than the code, so left a response there: https://github.com/ONEARMY/community-platform/issues/3468#issuecomment-2088162894

mariojsnunes:
Dropped in favor of https://github.com/ONEARMY/community-platform/issues/3467

PR #3471, opened 2024-04-22
feat: deep link into impact section in settings page

paposeco:
PR Checklist

PR Type

## Description

It is now possible to link to the Impact section on the Settings page, for a given year. By adding ?section=impact."wantedyear" to the /settings url, the Impact section is expanded and the user is taken to the provided year. If impact data is missing from a user's profile, by clicking on the "Edit data" from the profile, the user will also be taken straight to the right place on the settings page. 

(for example: /settings?section=impact.2022 will take you to the 2022 Impact section on the settings page)

## Git Issues

Closes # 3055


src/pages/User/impact/ImpactMissing.tsx
```diff
@@ -60,7 +60,7 @@ export const ImpactMissing = observer((props: Props) => {
         </>
       )}
       {isPageOwner && (
-        <a href="/settings">
```
mariojsnunes, line 63: I think we should use the # as it's the standard for this kind of behavior:
```<a href={`/settings#impact_${year}`}>```
and at [ImpactYear.section.tsx](https://github.com/ONEARMY/community-platform/pull/3471/files#diff-f70ea2e69bb16c980f0f06f3a55704a9e5e6c48d281dc633ad6bdec42394d907), the `<Heading>` would have ```id="{`/settings#impact_${year}`}"```

File: src/pages/UserSettings/content/formSections/Impact/ImpactYear.section.tsx [hunk 2]
```diff
@@ -63,13 +66,25 @@ export const ImpactYearSection = observer(({ year }: Props) => {
...
+      }
+    }
+    if (target === year) {
+      scrollToElement()
+    }
+  }, [target])
+
   return (
     <Box sx={sx}>
       <a id="impact"></a>
```
mariojsnunes, line 83: I know this line was here before, but it doesn't make much sense to me... can be removed if you confirm it's not needed

src/pages/UserSettings/content/formSections/Impact/ImpactYear.section.tsx
```diff
@@ -19,15 +19,18 @@ import type { SubmitResults } from 'src/pages/User/contact/UserContactError'
 
 interface Props {
   year: IImpactYear
+  target: number | null
 }
 
-export const ImpactYearSection = observer(({ year }: Props) => {
+export const ImpactYearSection = observer(({ year, target }: Props) => {
```
mariojsnunes, line 25: I think we can simplify this PR a lot if we move all the logic to this component.
`const { hash } = useLocation();`
```
useEffect(() => {
// logic to check if hash is valid
// scroll to ImpactYearSection
}, [hash])
```
This way we avoid passing 3 layers of props

mariojsnunes (changes requested):

src/pages/UserSettings/content/formSections/Impact/ImpactYear.section.tsx
paposeco: Thanks for the feedback and I'm sorry for the delay in getting back to you. I've made the changes you suggested, but haven't submitted them yet. Should I just make a new commit to this branch? (I'm a newbie in case you couldn't tell).

benfurber:
Awesome! Thanks @paposeco!

PR #3472, opened 2024-04-23
fix: map pin caching

chrismclarke:
PR Checklist

PR Type

## Description
Revert changes in #3466 to allow caching of map pins (see comment in PR for details)

Unfortunately the `getWhere` query always executes against server, whereas the `stream` query includes ability to cache results and retrieve from cache

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/src/stores/Maps/maps.store.ts
+++ b/src/stores/Maps/maps.store.ts
@@ -99,10 +99,17 @@ export class MapsStore extends ModuleStore {
     }
 
     // TODO: the client-side filtering done at `processDBMapPins` should be here
-    this.db.collection<IMapPin>(COLLECTION_NAME).stream((update) => {
-      const pins = update.filter((v) => !v._deleted)
-      this.processDBMapPins(pins, filterToRemove)
-    })
+    let updateCount = 0
+    const subscription = this.db
+      .collection<IMapPin>(COLLECTION_NAME)
+      .stream((update) => {
+        updateCount++
+        this.processDBMapPins(update, filterToRemove)
+        // hack - unsubscribe after receiving max 2 updates (1 cache + 1 server)
+        if (updateCount >= 2) {
+          subscription.unsubscribe()
+        }
+      })
   }
 
   @action
--- a/src/stores/Maps/maps.store.ts
+++ b/src/stores/Maps/maps.store.ts
@@ -99,11 +99,10 @@ export class MapsStore extends ModuleStore {
     }
 
     // TODO: the client-side filtering done at `processDBMapPins` should be here
-    const pins = await this.db
-      .collection<IMapPin>(COLLECTION_NAME)
-      .getWhere('_deleted', '!=', true)
-
-    this.processDBMapPins(pins, filterToRemove)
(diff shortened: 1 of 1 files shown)


mariojsnunes:
But by using streaming, doesn't that mean than whenever a pin changes, all users will get an update?

chrismclarke:
> But by using streaming, doesn't that mean than whenever a pin changes, all users will get an update?

It does mean that yes, although querying we will also always pull all pins regardless of whether they have been updated or not on every page load. So I'm working on the assumption that a user streaming map pin updates will still result in fewer reads than a user who loads 3.7k every page load - but can definitely still be optimised a lot more (just not using the `getWhere` operation which unfortunately only queries against the server)

mariojsnunes:
> It also reads from a separate cache for users visiting for the first time, meaning that new users only make 1 read request for first batch of data and then just stream docs as updated.

Can you explain this in more detail? How does streaming differ from getWhere on the first page load?
(I understand there is caching after the first page load)

chrismclarke:
> > It also reads from a separate cache for users visiting for the first time, meaning that new users only make 1 read request for first batch of data and then just stream docs as updated.
> 
> Can you explain this in more detail? How does streaming differ from getWhere on the first page load? (I understand there is caching after the first page load)

The logic of `getWhere` is:

1. retrieve all docs from server (unless limit specified) 
2. retrieve docs from cache (if none on server, e.g. offline) 
3. return docs (no cache update)

Whereas the logic for `stream` is:

1a. retrieve all docs from cache
1b. if no docs exist (new user) use special server-cache which stores daily copy of all docs as single request and update cache
1c. emit list of docs retrieved from cache
2a. check all cache docs for most recent `_modified` timestamp
2b. update cache with updated server docs
2c. emit combined list
3a. subscribe to any new doc changes

So getWhere always forces a large number of read ops on server whereas stream is usually much smaller initially and only increases if lots of docs are getting actively modified. 

So really we want something between the 2, where we use cache-first with server update but without streaming

chrismclarke:
@mariojsnunes 
I've just pushed [a05fd9b](https://github.com/ONEARMY/community-platform/pull/3472/commits/a05fd9bc64bb50fc2b9004fe7d75d62cba609d5e) which manually stops the stream operation after the initial docs have been retrieved from the cache and server. A bit of a hacky/temp workaround but hopefully should provide an alternative option

mariojsnunes:
Thank you!
I agree with the changes. And I understand now, that that PR was naive and assumed way less Pins than exist in reality. And didn't know about the cache db.
Still something is causing lots of pin updates, need to understand what.

chrismclarke:
> Thank you! I agree with the changes. And I understand now, that that PR was naive and assumed way less Pins than exist in reality. And didn't know about the cache db. Still something is causing lots of pin updates, need to understand what.

No worries at all, nothing about this is very intuitive - ordinarily faking a single request makes way more sense than multiple, but unfortunately there's a lot of layers of compilation abstracted away which works like `stream` and `query` don't represent very well.

chrismclarke:
I've added #3474 to try and clarify the code a little more and add general support for not keeping stream alive if not necessary to do so

PR #3499, opened 2024-05-06
fix: user related e2e tests

benfurber:
PR Checklist

PR Type

## Description

Deals with some of our recent cypress test issues. Adds a sign-up util, tidies up a few tests and drops a couple that are duplicates of already tested behaviour.


packages/cypress/src/utils/TestUtils.ts
tuliobluz, line 1: Best move `fillSignupForm` and `signUpNewUser` as a [custom command](https://docs.cypress.io/api/cypress-api/custom-commands) to `cypress/src/support/commands.ts`.

packages/cypress/src/utils/TestUtils.ts
```diff
@@ -19,3 +19,32 @@ export enum DbCollectionName {
...
+  cy.get('[data-cy=submit]').click()
+  cy.url().should('include', 'sign-up-message')
+}
+
+export const generateNewUserDetails = () => {
+  return {
+    username: `CI_${generateAlphaNumeric(5)}`.toLocaleLowerCase(),
+    email: `CI_${generateAlphaNumeric(5)}@test.com`.toLocaleLowerCase(),
+    password: 'test1234',
+    confirmPassword: 'test1234',
```
tuliobluz, line 48: We can reuse the `password` to `confirmPassword` field.

benfurber:
As this is needed for other branches I'll merge it in shortly unless some objects.

benfurber:
I'm getting this flaky fail:
```
The following error originated from your application code, not from Cypress. It was caused by an unhandled promise rejection.

> Cannot read properties of null (reading 'email')

When Cypress detects uncaught errors originating from your application it will automatically fail the current test.

This behavior is configurable, and you can choose to turn this off by listening to the uncaught:exception event.

https://on.cypress.io/uncaught-exception-from-application

Because this error occurred during a after all hook we are skipping all of the remaining tests.

Although you have test retries enabled, we do not retry tests when before all or after all hooks fail
```

Something weird in the teardown?

benfurber:
Sometimes the settings changes weren't persisting... waiting a second fixed that...

benfurber:
Updated error on the test now:
```
  > No document to update: projects/onearmy-test-ci/databases/(default)/documents/KLoGC_v3_users/ci_mwcow
```

`ci_mwcow` is the user name of the account undertest...

mariojsnunes (approved):

PR #3519, opened 2024-05-09
refactor: common discussion approach

benfurber:
## Description

Until last week I hadn't really appreciated how differently the implementation for discussions on research is from the implementation for discussions on questions. 

On questions, the store and component responsibilities are very seperate while for research, the responibilities are highly coupled.

Screenshot 2024-05-20 at 14 31 16

This is a starting point to have a generalised approach to discussions which should make adding discussions to how-tos so much easier.

BREAKING CHANGE: No comment mentions for research. Will add a generalised approach later on. Discussed with Dave.

## Now:
Screenshot 2024-05-20 at 14 31 26

To-do:
- [x] Fix the show/hide button logic
- [x] Research notifications
- [x] Fix all the tests, obvs
- [x] Fix the minor design differences
- [x] Add unit test coverage for `DiscussionWrapper`, `HideDiscussionContainer`, `ResearchDisscussion`, `QuestionDiscussion`

Before releasing:
- [ ] Ensure all the prod comments data is in the right places!
- [ ] Add `primaryContentId` to prod DBs

Write-up tickets:
- [ ] Simplify `DiscussionContainer` props
- [ ] Implement for how-tos
- [ ] Add custom assertations for comments, replies, notifications
- [ ] Add mentions logic to discussion store
- [ ] Change discussion store to service

### DIFF
--- a/packages/components/src/Breadcrumbs/BreadcrumbsItem.tsx
+++ b/packages/components/src/Breadcrumbs/BreadcrumbsItem.tsx
@@ -28,6 +28,7 @@ export const BreadcrumbItem = ({ text, link, isLast }: BreadcrumbItemProps) => (
   <Box
     style={{ display: 'inline-flex', marginRight: '10px' }}
     data-testid="breadcrumbsItem"
+    data-cy="breadcrumbsItem"
   >
     {!isLast ? (
       <BreadcrumbButton link={link} text={text} />
--- a/packages/cypress/src/integration/howto/comments.spec.ts
+++ b/packages/cypress/src/integration/howto/comments.spec.ts
@@ -12,20 +12,23 @@ describe('[How To Commenting]', () => {
 
   describe('[By Authenticated]', () => {
     it('[Commenting]', () => {
+      const commentText = 'A short string intended to test commenting'
       cy.signUpNewUser()
+
       cy.visit(specificHowtoUrl)
 
       cy.step("Cannot edit others' comments")
       cy.get('[data-cy="howto-comments"]').should('exist')
       cy.get('[data-cy="CommentItem: edit button"]').should('not.exist')
 
-      const commentText = 'A short string intended to test commenting'
-
       cy.step('Can add comment')
+      cy.contains('1 Comment')
       cy.get(`[data-cy="comments-login-prompt"]`).should('not.exist')
       cy.get(`[data-cy="comments-form"]`).should('be.exist')
       cy.get('[data-cy="comments-form"]').type(commentText)
       cy.get('[data-cy="comment-submit"]').click()
+      cy.wait(1000)
+      cy.contains('2 Comments')
       cy.get('[data-cy="comment-text"]').should('contain.text', commentText)
 
       cy.step('Can edit comment')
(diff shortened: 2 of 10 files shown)


mariojsnunes:
https://github.com/ONEARMY/community-platform/pull/3165
Might want to consider the service approach

src/common/DiscussionWrapper.tsx
```diff
@@ -0,0 +1,134 @@
...
+          comment={comment}
+          onChange={setComment}
+          onMoreComments={() => {}}
+          handleEdit={handleEdit}
+          handleEditRequest={handleEditRequest}
+          handleDelete={handleDelete}
+          highlightedCommentId={highlightedCommentId}
+          onSubmit={onSubmit}
+          onSubmitReply={handleSubmitReply}
+          isLoggedIn={!!discussionStore.activeUser}
```
codisart, line 129: There are a lot of props for the container. Could it be possible that certains props could be put in custom hooks or even context hooks and be called in or in the children of the `DiscussionContainer` ?

src/stores/Discussions/discussions.store.tsx
```diff
@@ -201,7 +209,7 @@ export class DiscussionStore extends ModuleStore {
     }
   }
 
-  private async _addNotification(
+  private async _addNotifications(
```
benfurber, line 212: The plan is to move this to a cloud function anyway, so the mess in this function is temporary.

src/common/DiscussionWrapper.tsx
```diff
@@ -0,0 +1,156 @@
...
+  } = props
+
+  const [comment, setComment] = useState('')
+  const [discussion, setDiscussion] = useState<IDiscussion | null>(null)
+  const [isLoading, setIsLoading] = useState(true)
+
+  const { discussionStore } = useCommonStores().stores
+  const highlightedCommentId = window.location.hash.replace('#comment:', '')
+
+  const transformComments = (discussion) => {
```
mariojsnunes, line 41: maybe there could be a `const comments = useMemo(() =>  ,[discussion, activeUser])` that automatically transforms the comments when the `discussion` or `activeUser` changes.
So when there is a discussion update, you simply update the discussion state, the useMemo takes care of the rest and can pass the `comments` to the `discussionProps`.

mariojsnunes (commented):
Some overall comments:
1. The `discussionsStore` doesn't really store anything related do discussions, it just calls firebase and the data is stored in the `DiscussionWrapper`. Other than calling firebase, the only use of the store is to access the `activeUser` from the `RootStore`.
If we call the userStore at the `DiscussionWrapper` to access the `activeUser`, we could remove the dependency on RootStore and mobx and rename the class to `DiscussionService`.
2. `QuestionDiscussion` and `ResearchComments` don't do much now, maybe can be removed in favor of using `DiscussionWrapper` directly?
3. Maybe the `#update-` logic could be simplified by just using the commentId and if the comment/discussion access the location hash directly.
4.

benfurber:
@mariojsnunes From your comments:

1. Moving `QuestionStore` to `QuestionService`. I'm with you on that approach. I'll do a ticket for it so this large branch doesn't have to get larger.
2. On the size of `QuestionDiscussion` and `ResearchDiscussion`. Maybe. But I'd rather keep moving in the direction of getting components smaller before deciding whether some are too small.
3. The `#update-` logic. Hadn't engaged with it at all in this PR.
4. `useMemo` Would you be up for applying that to `DiscussionWrapper` in a follow-up PR? I spend some time on it yesterday and just made a mess.

PR #3528, opened 2024-05-10
feat: enable research creation on pp

benfurber:
PR Checklist

PR Type

## Description

This allows any logged in user on Precious Plastic only to create their own research. The create research button is there for all with the standard link to login if they aren't already.

Screenshot 2024-05-10 at 17 42 57

### DIFF
--- a/src/pages/Research/Content/ResearchList.tsx
+++ b/src/pages/Research/Content/ResearchList.tsx
@@ -5,6 +5,7 @@ import { observer } from 'mobx-react'
 import { Button, Loader } from 'oa-components'
 import { AuthWrapper } from 'src/common/AuthWrapper'
 import { useCommonStores } from 'src/common/hooks/useCommonStores'
+import { isPreciousPlastic } from 'src/config/config'
 import { logger } from 'src/logger'
 import { Box, Flex, Heading } from 'theme-ui'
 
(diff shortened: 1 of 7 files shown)


src/pages/Research/Content/ResearchList.tsx
```diff
@@ -116,13 +117,21 @@ const ResearchList = observer(() => {
...
-            <AuthWrapper roleRequired={RESEARCH_EDITOR_ROLES}>
+            {isPreciousPlastic() ? (
               <Link to={userStore.activeUser ? '/research/create' : '/sign-up'}>
                 <Button variant={'primary'} data-cy="create">
                   {listing.create}
                 </Button>
               </Link>
-            </AuthWrapper>
+            ) : (
+              <AuthWrapper roleRequired={RESEARCH_EDITOR_ROLES}>
```
codisart, line 127: suggestion: Why not use the same condition than in `src/pages/Research/research.routes.tsx` and put an empty array in roleRequired if this is "preciousPlastic" ? I could even be a method in the same place as `isPreciousPlastic` like `researcherRoles` who would return the correct array.

benfurber: The button is also currently hidden for non-pp instances and I wanted to display button too. Or have I misunderstood your suggestion?

codisart: no It seems I was wrong on the comprehension of this code. I did not realize that there was a difference in the condition of `AuthWrapper` and `AuthRoute`.

PR #3530, opened 2024-05-10
feat(seo): update headings for user profiles

### LINKED ISSUE #3424: Quick SEO wins
There's a few quick wins we can do from a search and social prospective.

1. Use heading tags correctly.
2. Ensure all images have alt tags.
3. Set more/better open graph tags.

### Heading tags
The research, how-to and question titles should definitely be h1s rather than h2s. There might also be more semantic heading considerations.

### Image alt tags
At the moment we don't ask users to provide alt tags for the images that upload. As a short-term fix, we should apply alt tags to research updates, how-to steps and questions which follow some kind of dynamtic structure, e.g. '<research update title> - <image index>'

### Better open graph tags
At a minimum we need 'og:type' set to website in the `postCraBuild.ts` script, but there's probably [more improvements we should make](https://docs.prerender.io/docs/open-graph)

codisart:
PR Checklist

PR Type

- [?] Bug fix (non-breaking change which fixes an issue)
- [?] New feature (non-breaking change which adds functionality)

## Description

_What this PR does_
This PR update all Heading tags in the User (members and others) profile to specific html heading tags from the default h2 to h1, h2 or h3. It should respect a correct hierarchical structure. To sum up, all tabs are now confirmed h2, and every other heading in the card is now a h3 and the name of the user is now a h1.

## Git Issues

Helps to close #3424

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/src/pages/User/content/MemberProfile.tsx
+++ b/src/pages/User/content/MemberProfile.tsx
@@ -127,6 +127,7 @@ export const MemberProfile = ({ user, docs }: IProps) => {
             </Flex>
             <Box sx={{ flexDirection: 'column' }} mb={3}>
               <Heading
+                as="h1"
                 color={'black'}
                 style={{ wordWrap: 'break-word' }}
                 data-cy="userDisplayName"
--- a/src/pages/User/contact/UserContactForm.tsx
+++ b/src/pages/User/contact/UserContactForm.tsx
@@ -66,7 +66,7 @@ export const UserContactForm = observer(({ user }: Props) => {
 
   return (
     <Flex my={2} sx={{ flexDirection: 'column' }}>
-      <Heading variant="small" my={2}>
+      <Heading as="h3" variant="small" my={2}>
         {`${title} ${user.userName}`}
       </Heading>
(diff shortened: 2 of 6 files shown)


src/pages/User/content/MemberProfile.tsx
```diff
@@ -127,6 +127,7 @@ export const MemberProfile = ({ user, docs }: IProps) => {
             </Flex>
             <Box sx={{ flexDirection: 'column' }} mb={3}>
               <Heading
+                as="h1"
```
thisislawatts, line 130: **question:** Do we want to introduce any tests that verify this document structure? This would prevent us from accidentally introducing a regression

codisart: This is a good idea but there is different ways of testing. We could test each component implementation to validate the html tags used. We could also test the hierarchy of a full page. That would be better IMO for a semantic sense and it removes coupling between tests and code. The ideal way would be to test / assert the semantic of the html of the app but I don't know any way of doing that.

PR #3537, opened 2024-05-13
fix: useful button click spam

mariojsnunes:
PR Checklist

PR Type

## Description

Fixes useful vote click spam by awaiting for the action to complete.

## Git Issues

Closes #

### DIFF
--- a/packages/components/src/UsefulStatsButton/UsefulStatsButton.stories.tsx
+++ b/packages/components/src/UsefulStatsButton/UsefulStatsButton.stories.tsx
@@ -1,3 +1,5 @@
+import { useState } from 'react'
+
 import { UsefulStatsButton } from './UsefulStatsButton'
 
 import type { Meta, StoryFn } from '@storybook/react'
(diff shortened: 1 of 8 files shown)


benfurber:
I agree with the change but wonder if it should be paired with a visual update for the user while waiting for the action to complete?

mariojsnunes:
I felt like it was enough to make it disabled (there is a small animation because of it). have you tried it?

benfurber:
Yeah fair enough. 

Can we make the story book stories more meaningful please. We only need one logged out with some numbers (i.e. delete the first one) and can you add meaningful enough props that the number changes (quick setStates) and put a two second wait on the promise to update the number.

packages/components/src/UsefulStatsButton/UsefulStatsButton.stories.tsx
```diff
@@ -39,6 +39,6 @@ export const CurrentUserHasVoted: StoryFn<typeof UsefulStatsButton> = () => (
     votedUsefulCount={99}
     hasUserVotedUseful={true}
     isLoggedIn={true}
-    onUsefulClick={() => null}
+    onUsefulClick={() => new Promise(() => {})}
```
pizzaisdavid, line 42: In all the places, I think you could do `onUsefulClick={() => Promise.resolve())}` instead.

One could say it might be confusing for people but it is already done that way in other places:
https://github.com/search?q=repo%3AONEARMY%2Fcommunity-platform+Promise.resolve&type=code

But also, feel free to ignore this feedback.

benfurber (approved):
Nice.

PR #3552, opened 2024-05-16
fix(security-rules): unblock v3 categories

thisislawatts:
PR Checklist

PR Type

## Description

_What this PR does_

## Git Issues

Closes #

## Screenshots/Videos

---

## What happens next?

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -82,6 +82,11 @@ service cloud.firestore {
       allow write: if isPublicWritable();
     }
 
+    match /v3_categories/{document=**} {
+      allow read: if isPublicReadable();
+      allow write: if isPublicWritable();
+    }
+
     match /v3_tags/{document=**} {
       allow read: if isPublicReadable();
       allow write: if isPublicWritable();


firestore.rules
```diff
@@ -82,6 +82,11 @@ service cloud.firestore {
       allow write: if isPublicWritable();
     }
 
+    match /v3_categories/{document=**} {
+      allow read: if isPublicReadable();
+      allow write: if isPublicWritable();
```
mariojsnunes, line 87: the write should at least be restrict by authenticated user.

note that currently it's not possible to create categories via our UI.
but unsure how to make a rule here that would deny via our UI but allow wherever they are created now (maybe retool?)

mariojsnunes (changes requested):

firestore.rules
mariojsnunes: and the same for
question_categories_rev20231130, research_categories_rev20221224, v3_tags

mariojsnunes:
to fix all the erros:
```
match /v3_categories/{document=**} {
  allow read: if isPublicReadable();
  allow write: if request.auth != null;
}
    
match /_archived/v3_categories/{document=**} {
  allow read: if isPublicReadable();
  allow write: if request.auth != null;
}

match /_archived/research_categories_rev20221224/{document=**} {
  allow read: if isPublicReadable();
  allow write: if request.auth != null;
}

match /_archived/question_categories_rev20231130/{document=**} {
  allow read: if isPublicReadable();
  allow write: if request.auth != null;
}

match /_archived/v3_tags/{document=**} {
  allow read: if isPublicReadable();
  allow write: if request.auth != null;
}
```

firestore.rules
thisislawatts: My current aim is to introduce rules for each collection used in our app, that aligns with the previous wide open permissions. From this baseline we can then iterate over each collection, introducing permissions that better match the intended access patterns. 

I want to make these changes in an isolated way. The aim here is that we can easily isolate side affects and roll back as necessary. 

There will be a follow up PR that locks down these 'admin-only' category and tag collections. These documents should only be modifiable by moderators using Retool, so I think we can entirely block write access in firestore.rules.

Retool uses a service account for auth when accessing the database which bypasses these controls.

PR #3574, opened 2024-05-22
feat: connect figma to storybook

benfurber:
Setting up chromatic and their figma connected app for the design team.


benfurber:
There's a part of this - showing a github action for chromatic - that isn't working yet. But I won't get time to add that for a while so I think should get merged in as is.

mariojsnunes (approved):

PR #3583, opened 2024-05-23
feat: still display replies for deleted comment

benfurber:
PR Checklist

PR Type

## Description

Still show the nest comments (replies) when a parent comment is deleted. 

## Git Issues

Closes #

## Screenshots/Videos

https://github.com/ONEARMY/community-platform/assets/16688508/02b19d97-57ce-447b-92eb-963e9f22718b

### DIFF
--- a/packages/components/src/CommentItem/CommentItem.tsx
+++ b/packages/components/src/CommentItem/CommentItem.tsx
@@ -93,7 +93,7 @@ export const CommentItem = (props: IProps) => {
       <Flex sx={{ gap: 2 }}>
         {_deleted && (
           <Box sx={{ marginBottom: 2 }}>
-            <Text sx={{ color: 'grey' }}>{DELETED_COMMENT}</Text>
+            <Text sx={{ color: 'grey' }}>[{DELETED_COMMENT}]</Text>
           </Box>
         )}
 
--- a/packages/components/src/CommentItem/CommentItem.tsx
+++ b/packages/components/src/CommentItem/CommentItem.tsx
@@ -11,6 +11,7 @@ import { Username } from '../Username/Username'
 import type { IComment } from './types'
 
 const SHORT_COMMENT = 129
+const DELETED_COMMENT = 'The original comment got deleted'
 
 export interface IProps {
   comment: IComment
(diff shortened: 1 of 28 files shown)


benfurber:
@davehakkens Still working on this but wanted to show you the user display I was thinking. 

When a comment (with replies) is marked `_deleted`:
1. The comment and all details are removed.
2. The existing replies are still visible.
3. No new replies can be added.

benfurber:
Happy with this approach @davehakkens?

packages/components/src/CommentItem/CommentItem.tsx
```diff
@@ -83,119 +86,123 @@ export const CommentItem = (props: IProps) => {
...
+        )}
       </Flex>
 
       <Modal width={600} isOpen={showEditModal}>
         <EditComment
           comment={text}
-          handleSubmit={(commentText) => {
-            handleEdit && handleEdit(_id, commentText)
+          handleSubmit={async (commentText) => {
+            handleEdit && (await handleEdit(_id, commentText))
```
mariojsnunes, line 205: the `?` can be removed from all the `IProps` functions so here we don't need to check if the function is defined

src/stores/Discussions/discussions.store.tsx
```diff
@@ -297,15 +301,23 @@ export class DiscussionStore extends ModuleStore {
     })
   }
 
+  private _showDiscussion(
```
mariojsnunes, line 304: I think this logic doesn't need to be extracted from `_updateDiscussion`. As a reader, I need to jump somewhere else to understand what's going on.

src/stores/Discussions/discussions.store.tsx
```diff
@@ -40,21 +41,18 @@ export class DiscussionStore extends ModuleStore {
       )[0] || null
 
     if (foundDiscussion) {
-      return foundDiscussion
+      return this._showDiscussion(foundDiscussion)
```
mariojsnunes, line 44: I think this change isn't needed, due to the if above, `_showDiscussion` would always return the same

PR #3587, opened 2024-05-24
feat(discussions): add avatars

benfurber:
PR Checklist

PR Type

## Description

Add option user profile image to comments (and tidies up various responsive layout stuff).

## Screenshots/Videos

Screenshot 2024-05-24 at 16 10 34

### DIFF
--- a/functions/src/userUpdates/updateDiscussionComments.ts
+++ b/functions/src/userUpdates/updateDiscussionComments.ts
@@ -18,7 +18,7 @@ export const updateDiscussionComments = async (
 
   if (!snapshot.empty) {
     const { _id, badges, coverImages, location } = user
-    const creatorImage = setCreatorImage(coverImages)
+    const creatorImage = getCreatorImage(coverImages)
 
     const userDetails = {
       creatorCountry: location?.countryCode || '',
@@ -48,7 +48,7 @@ export const updateDiscussionComments = async (
   }
 }
 
-const setCreatorImage = (coverImages) => {
+const getCreatorImage = (coverImages) => {
   if (coverImages && coverImages[0] && coverImages[0].downloadUrl) {
     return coverImages[0].downloadUrl
   }
--- a/functions/src/userUpdates/updateDiscussionComments.ts
+++ b/functions/src/userUpdates/updateDiscussionComments.ts
@@ -11,34 +11,46 @@ export const updateDiscussionComments = async (
 ) => {
   if (!hasKeyDetailsChanged(prevUser, user)) return
 
-  let updatedCommentCount = 0
-
-  const { _id, badges, location } = user
-  const userDetails = {
-    creatorCountry: location?.countryCode || '',
-    isUserVerified: !!badges?.verified,
-    isUserSupporter: !!badges?.supporter,
-  }
-
   const snapshot = await db
     .collection(DB_ENDPOINTS.discussions)
-    .where('contributorIds', 'array-contains', _id)
+    .where('contributorIds', 'array-contains', user._id)
     .get()
 
-  for (const doc of snapshot.docs) {
-    const discussion = doc.data() as IDiscussion
+  if (!snapshot.empty) {
+    const { _id, badges, coverImages, location } = user
+    const creatorImage = setCreatorImage(coverImages)
 
-    const comments = discussion.comments.map((comment) => {
-      if (comment._creatorId !== _id) return comment
+    const userDetails = {
+      creatorCountry: location?.countryCode || '',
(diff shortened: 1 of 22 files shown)


benfurber:
@davehakkens, et al. I'm not finished on this branch so doesn't need a review yet.

davehakkens:
well i'm looking anyway 🤓

benfurber:
> Niceeee! Could we display this functionality only to beta-testers first? Since it might bring some weird UI changes

Sorry @davehakkens, I missed this comment yesterday. I think based on the current beta tester wrapper component and approach, a fair amount of extra re-coding would be needed to enable them to evaulate this change on prod.

Maybe instead the preview link/branch could be left open for longer to get tested by a broader group before merging?

benfurber:
@davehakkens I take all your points.

It's more work, because:
1. We add the beta tester wrapper at the app code level while the avatar addition code is nested five(ish) components deep in the component library.
2. The beta tester wrapper layout stylings is a little clunky for small elements.
3. The wrapper currently adds a 4px border on each side. So that's 8 per axis. As currently coded the avatar is only 30x30 for the mobile display. So if I put the wrapper litterally right around the image, it would changes the experience of feature.

So I'll:
1. Add a property from the DiscussionContainer on down to the CommentItem to set whether to show the avatar or not.
2. Add the beta tester wrapper at the DiscussionContainer level so that the wrapper border doesn't get in the way of experiencing the feature.

benfurber:
# On Mobile

## Before
IMG_1342

## For reg users
IMG_1343

## When no images present
IMG_1344


IMG_1341

pizzaisdavid:
General remark about what I am seeing in the codebase: it seems that it is a common case that something changes about a user, for example image (or location) and then their contributions need to be updated (via `handleUserUpdates`, I think.)

My first thought is that it would be less of a headache to only store the id of a user and the content of their comment. And then, when we want to show the avatar (or location) the id is used to lookup the user and then get their current avatar, merging the two together (either on the backend when requesting discussion data or on the frontend.)

So it would just be that the user can change their avatar as much as they want and only the data in the user collection would change, discussions would not get updated at all.

Maybe there is some Firebase limitation/justification or some other good reason. 😄

mariojsnunes:
> Maybe there is some Firebase limitation/justification or some other good reason. 😄

Yes its a common practice for a NoSql database.
If we eventually [migrate to supabase](https://github.com/ONEARMY/community-platform/discussions/1441) which uses postgres, we would implement the approach you described

pizzaisdavid:
I worked at a company that used Mongodb (which is NoSQL) and they were doing the method I suggested. But on some further research, it is kinda like they were treating NoSQL like SQL (strict validation on fields, foreign keys, data only stored in one place.)

But I have found this video series, which uses a very similar example to the usercase here (user submissions and wanting to show profile pictures.) https://youtu.be/v_hR4K4auoQ?list=PLl-K7zZEsYLluG5MCVEzXAQ7ACZBCuZgZ&t=380 (timstamped for convenience) I guess it is a general trend of NoSQL and the approach optimizes reads over edits, which makes sense. As the edits don't happen often.

Consider my comment resolved. Thanks for the push to learn more @mariojsnunes 😄

benfurber:
Thanks @mariojsnunes and @pizzaisdavid. Any more code comments?

functions/src/userUpdates/updateDiscussionComments.ts
```diff
@@ -11,34 +11,46 @@ export const updateDiscussionComments = async (
...
+        updatedCommentCount++
+        return {
+          ...comment,
+          ...userDetails,
+        }
+      })
+
+      await doc.ref.update({ comments })
+    }
+    return console.log(`Updated ${updatedCommentCount} discussion comments`)
```
pizzaisdavid, line 47: Is there an actual logger object/thingy you can use?

functions/src/userUpdates/updateDiscussionComments.ts
```diff
@@ -11,34 +11,46 @@ export const updateDiscussionComments = async (
...
+  }
+}
 
-    await doc.ref.update({ comments })
+const setCreatorImage = (coverImages) => {
```
pizzaisdavid, line 51: this should be called getCreatorImage or something, I expect a setX function to modify some state and return nothing. :wink:

functions/src/userUpdates/updateDiscussionComments.ts
benfurber: Yes? These do make there way to our cloud logger.

PR #3609, opened 2024-05-29
test: exception for when firebase can't find doc

### LINKED ISSUE #3572: [bug?] 'No document to update' cypress failure reason
Failure reasons like this come up every so often and more recently since trying to isolate more test cases:
```
The following error originated from your application code, not from Cypress. It was caused by an unhandled promise rejection.

  > No document to update: projects/onearmy-test-ci/databases/(default)/documents/vqFdX_v3_users/ci_qxrog
```

Basically there's a firebase error as it's trying to delete a user account that's already(?) been removed.

This problem is probably due to some ordering issue somewhere in the afterhooks but I'm also inclined to just add an exception for that external event, like we've already done for database indexing issues (`packages/cypress/src/support/rules.ts`)

What do others think?

benfurber:
PR Checklist

PR Type

## Description

We've started to get a cypress test failure during teardown for when firebase can't find a document (normally a user record) to 'update'. It's unrelated to the test suite so adding an expection.

## Git Issues

Closes #3572

### DIFF
--- a/packages/cypress/src/support/rules.ts
+++ b/packages/cypress/src/support/rules.ts
@@ -4,6 +4,9 @@ Cypress.on('uncaught:exception', (err) => {
   if (err.message.includes('The query requires an index.')) {
     return false
   }
+  if (err.message.includes('No document to update')) {
+    return false
+  }
   // we still want to ensure there are no other unexpected
   // errors, so we let them fail the test
 })


benfurber:
I'm going to merge this in as it'll reduce the false fails we're getting atm. Easy to remove if we change our minds later.

PR #3600, opened 2024-05-27
feat: add unique title check to questions

benfurber:
PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)

PR Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Developer experience (improves developer workflows for contributing to the project)

## Description

Addes a UI form check and store check for the uniqueness of question titles.

## Git Issues

Closes #3573 

## Screenshots/Videos

Screenshot 2024-05-27 at 15 08 15

PR #3621, opened 2024-06-04
fix: question form submitting not instantly set

benfurber:
PR Checklist

PR Type

## Description

Even after #3600 got merged in, it's happened again that a user has submitted duplicate questions (just this time changing the title to do it...). I did some more testing and it turns out that the submitting button isn't instantly disabled if you're on a low connection (I set to poor-3G for testing), thus maybe giving the user the view that the submission request has been handled. 

This change ensures the button is more quickly set.


mariojsnunes:
As a further sanity check, could have a useState that tracks submits, blocks if any in progress

mariojsnunes:
https://github.com/ONEARMY/community-platform/blob/9e5582a3b28a7ebb218257ea047d9783a278c430/src/stores/Question/question.store.tsx#L86

I think I found the issue, that ! should be before the await

mariojsnunes (approved):

PR #3624, opened 2024-06-05
[WiP][e2e] Cypress seeding before all test instead beforeEach

tuliobluz:
PR Checklist

PR Type

## Description

I found the following ToDo:

[File](https://github.com/ONEARMY/community-platform/blob/master/packages/cypress/scripts/start.ts#L28)

` TODO: CC - 2021-02-24
DB seeding happens inbetween test suites, but really should happen before/after test
scripts start and end (particularly teardown, as it won't be called if tests fail).
Possibly could be done with a Cypress.task or similar
Temp cli function to wipe hanging db: firebase use ci; firebase firestore:delete --all-collections`

This PR will change how the seed is done in our e2e tests, instead of seeding during the `beforeEach` of every scenario, it will run only once and this DB_PREFIX is shared among the scenarios.

I also had to fix some queries because before it was getting the first item from the collections(as every run the db was clean), and I had to add another parameter to some queries.

This will save running time because now it will be seeding only at once.

**WiP:** 

- Test the impact of this change in the current CI,
- Decide how the deleteDatabase will be done. I created a function to do it, but I am afraid to test it now, so I need to make sure it will not delete anything from PRD, now I am just testing the seed.


benfurber:
Exciting start, thanks @tuliobluz.

tuliobluz:
I had to create the env script to append a `DB_PREFIX` on the `.env` file then, I could access this env var to run the `deleteDatabase` for that `DB_PREFIX`. 

@benfurber 

- I could not find a way to delete the collection created when it runs locally. 
- The name of the scripts I set up 
`"test:manual"` This is the script to run locally, but if it is not okay we can the ci to sth like `test:ci`

tuliobluz:
Due to some import problems with the new changes, I will close this PR. Maybe in the future, we can try to implement it again.

PR #3635, opened 2024-06-08
Feat/start admin dashboard

pizzaisdavid:
PR Type

## Description
Add an admin page... for admin things.

Go to:
http://localhost:3000/admin

gif demo:
admin2

### DIFF
--- a/src/pages/Admin/Content/Dashboard/index.tsx
+++ b/src/pages/Admin/Content/Dashboard/index.tsx
@@ -15,7 +15,19 @@ const Dashboard = observer(() => {
       </Flex>
     )
   }
-  return <div>placeholder</div>
+
+  const descriptionElement = document.querySelector('meta[name="description"]')
+  const descriptionContent =
+    descriptionElement === null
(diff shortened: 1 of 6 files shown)


davehakkens:
Thanks for looking into this.
Altough it makes total sense to have this, we already used to have this but it got removed
https://github.com/ONEARMY/community-platform/pull/2689
The PR also has some background information on the reason why

pizzaisdavid:
Hmm.. Thanks for the quick feedback.

davehakkens:
sorry for being the party pooper.

pizzaisdavid:
Ha. It's no problem. Not sure how the integration of academy would happen without an admin interface. But maybe it would need to be more scoped to the academy specifically, Referring to: https://github.com/ONEARMY/community-platform/discussions/3233

But maybe also the academy thing is low priority at the moment / still being defined.

davehakkens:
Good point. Yeh i imagine the admin panel coming back at some point when we're done prototyping it in Retool.

PR #3639, opened 2024-06-10
refactor: cy change shouldExist to beVisible

tuliobluz:
PR Checklist

PR Type

## Description

This MR is an improvement in some asserts as they used `.should('exist')` instead of `.should('be.visible')`. When you use `should('exist')` cypress checks if the element exists in DOM, then cypress will NOT check if the element is visible.


mariojsnunes (approved):
nice one!

PR #3690, opened 2024-06-24
fix: spacing on profile

benfurber:
PR Checklist

PR Type

## Description

Fixing a few spacing issues on the profile page - a recent regression.

## Screenshots/Videos

Before: 
image-profile

After:
Screenshot 2024-06-24 at 17 06 00
Screenshot 2024-06-24 at 17 05 43


src/pages/User/content/MemberProfile.tsx
```diff
@@ -32,107 +32,94 @@ export const MemberProfile = ({ user, docs }: IProps) => {
...
+          <Avatar
+            loading="lazy"
+            src={memberPictureSource}
+            sx={{
+              objectFit: 'cover',
+              width: '120px',
+              height: '120px',
+            }}
+          />
+          <UserStatistics
```
mariojsnunes, line 85: IMO UserStatistics should read directly from the store. But that would cause more changes and maybe harder to test?
As a half-way improvement, maybe just pass the user object instead of most of those props?

Issue #3651, opened 2024-06-12
[feature request] Improve the display of the edit and delete buttons next to user's own comment


dalibormrska:
> This ticket is part of series of 7 tickets concluding a design task focusing on improving design and usability of comments feature. Here you can find the [Figma file](https://www.figma.com/board/GfQPMs7PyC7jbkymDTco2D/CP-%2F-Nested-Comments-Improvements?node-id=0-1&t=7sgMA4ZPYXAMb03A-1) for more context.

### **2/7 - Improve the display of the edit and delete buttons next to user's own comment**
---

**Is your feature request related to a problem? Please describe.**
The problem here is, that the height of the edit and delete buttons is increasing the spacing between the username and the text. This creates an inconsistency of spacing between user's own comments and other comments.

![image](https://github.com/ONEARMY/community-platform/assets/35503298/7d86a6ab-8ffb-416c-8412-423793ca6d1f)

**Describe the solution you'd like**
A solution with consistent height of the wrapping div, that accounts for the size of the buttons regardless if they are there or not. The username should have `justify-content: flex-end;` 

<img width="738" alt="image" src="https://github.com/ONEARMY/community-platform/assets/35503298/056efec1-4691-48bc-bd33-564d4fc8a387">

Additionally, it would be good to have a less dominant version of the buttons as shown here. The issue of creating less dominant variant of the button component is prominent on several places throughout the platform and is discussed in this ticket #3655 . 

![image](https://github.com/ONEARMY/community-platform/assets/35503298/e90ff3ea-1e03-4c7d-b013-94842f5c9357)

**Describe alternatives you've considered**
I first considered, making the buttons the size of the username component which is 25px, but that would be too small for mobile.

A menu with three dots is standard in other platforms, but I am not sure how big and complex of an implementation that would be, so I just propose this quick adjustment for now. However, the three dots menu was a consideration before and could be a good option, since the options will likely grow in the future (adding report function). The three dots would also perform better on mobile, not creating problems with stacking below.

**Additional context**
See the milestone

elideg:
@dalibormrska I can pick this ticket up.

pizzaisdavid:
If other people look at this ticket, I have made a pull request that makes the buttons into the "less dominant version."

pizzaisdavid:
> The issue of creating less dominant variant of the button component is prominent on several places throughout the platform and is discussed in this ticket https://github.com/ONEARMY/community-platform/issues/3655 .

[...]

PR #3700, opened 2024-06-26
chore: make edit and deletes use less dominate version

pizzaisdavid:
PR Type
- New feature (non-breaking change which adds functionality)

## Description
change variants, done during the socialize and program event. 

## Git Issues
For https://github.com/ONEARMY/community-platform/issues/3651 but doesnt close it.

## Screenshots/Videos
Screenshot from 2024-06-26 18-32-59

### DIFF
--- a/packages/components/src/CommentItem/CommentItem.tsx
+++ b/packages/components/src/CommentItem/CommentItem.tsx
@@ -152,6 +152,7 @@ export const CommentItem = (props: IProps) => {
                     variant="outline"
                     small={true}
                     icon="edit"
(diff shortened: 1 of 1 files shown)


mariojsnunes (approved):

mariojsnunes (changes requested):
I think you can use the 'subtle' variant for this.
Check `Button.stories.tsx` file for other options.

Should also remove:
```
opacity: 0.5,
':hover': { opacity: 1 },
```
they look disabled because of that.

PR #3717, opened 2024-07-01
feat: change discord link to feedback form

### LINKED ISSUE #3716: [feature request] Get more direct user feedback
From @davehakkens:

Thinking of changing the current 'chat button' to 'Feedback button' 
* Less traffic going to Discord (good for questions module)
* We can gather more feedback from users (knowing that our tests don't catch everything)
* We have a steady dev and mod team to implement (didn't have this before)
* Still need to figure out how the filtering goes.

From:
image

To:
image

benfurber:
PR Checklist

PR Type

## Description

This changes the current button (that's on every page) that goes to discord to a quick feedback form using retool.

I think it's best to keep it super simple atm, see if we 

## Git Issues

Closes #3716

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -5,7 +5,7 @@ import { Analytics } from 'src/common/Analytics'
 import { getSupportedModules, isModuleSupported, MODULE } from 'src/modules'
 import Main from 'src/pages/common/Layout/Main'
 import { SeoTagsUpdateComponent } from 'src/utils/seo'
-import { Box, Flex } from 'theme-ui'
+import { Box, Flex, Text } from 'theme-ui'
 
 import { ScrollToTop } from '../common/ScrollToTop'
 import { NotFoundPage } from './NotFound/NotFound'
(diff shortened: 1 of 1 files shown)


benfurber:
@davehakkens The form can be editted here: https://onearmy.retool.com/forms/editor/Reporting/CP%20-%20Feedback%20Form

benfurber:
Also. Un-hide the button on mobile/tablet - worth checking it looks ok so have added a preview to review.

davehakkens:
Nice to use retool for this @benfurber to keep it simple
- Who do you imagine screening/sorting the feedback? dev team/moderators?
- Would it be possible to add a label per instance? So we know from which instance the feedback comes in
- Agree with @pizzaisdavid in the issue that 'Report a problem" would be better wording
- Not sure sure about mobile. The button is quite prominent there, but not a strong opinion on it. Also up for trying.

benfurber:
@davehakkens 

* Screening: Dev team
* Label per instance: I've changed url to auto-populate, do you think that does the trick?
* Updated wording
* Mobile: Trying a smaller versioin, what do you think?

davehakkens:
Nice. Yes all good to me, the auto url approuch is even better!
But doesnt work here `unknown`, but perhaps due to test environment.
afbeelding

benfurber:
> Nice. Yes all good to me, the auto url approuch is even better! But doesnt work here `unknown`, but perhaps due to test environment. afbeelding

Hhmmm, I was hoping that the preview branch would have that set the config field correctly... I'll have another go.

benfurber:
@davehakkens In fact I'm confident enough I've got the full url that I've locked the field (if the page loads with a url provided as a param).

Issue #3724, opened 2024-07-02
[feat] Update settings form(s)


benfurber:
This is a whole epic that I'll break down into smaller issues but I wanted to add a big one for what I'm working on right now.

## Big steps

1. [x] Add new form wrapper component into library.
2. [x] Add current form wholesale into the new wrapper.
3. [x] Move chunks of the form into seperate tabs - refactoring the below as I go.
4. [ ] The updated the design for individual fields/field groups - again, moving what I can into the component library.

Refactoring chunks:
1. The form fields themselves - as much into the component library as I can
2. Add mini-store functions for each seperate form.
3. Improve the test expectations (the actual app rather than checking the db).

[Main designs here](https://www.figma.com/design/x4DcaUjwiigjCySAt5omxB/CP-%2F-Signup%2C-Settings-and-Gatekeeping?node-id=95-832&t=mcnNSKlTtr9JN0RC-1)

benfurber:
I created this bug: 
<img width="558" alt="Screenshot 2024-07-18 at 14 21 37" src="https://github.com/user-attachments/assets/00c21fe4-27aa-4bab-98e8-ea02de0b37a9">

benfurber:
To add to the to-do list - add avatars for pages. Or more to the point, move the banner images to a seperate field and move the data.

benfurber:
<img width="521" alt="Screenshot 2024-07-31 at 11 17 27" src="https://github.com/user-attachments/assets/91845685-c111-410a-ae4e-04dd6a60f6df">

benfurber:
[...]

PR #3729, opened 2024-07-05
feat: add wrapper to settings form

benfurber:
PR Checklist

PR Type

## Description

So this is a slightly clunky stepping stone. This uses the new settings form wrapper, but with the existing form setting/structure. It's essentially the fundations to then move out sections of the form into other tabs.

## Git Issues

Part of #3724

## Screenshots/Videos

Screenshot 2024-07-05 at 13 18 09


benfurber:
@davehakkens Please have a play. I've tweaked a couple of bits that felt too weird for this interim step. Of course happy to change things in this PR but setting up the tabs won't take long (a few days next week).

davehakkens:
Seems like it needs a max-width. But perhaps it will be better when sidebar is there
afbeelding

Regarding the merge, would be nice to only merge if we know the other bits will be picked up soon because
- The design doesnt look very finished (specially messy when signing up as a new user)
- Some functions are gone (like the sidebar with community program)
- Could see some bugs coming up in this transition period.

benfurber:
Good idea about a max width - would be a quick fix for instant less messiness.

benfurber:
@davehakkens Have another look now? Looks like 850px is the sweet spot.

> Some functions are gone (like the sidebar with community program)
Of course the main sidebar is gone as per the design, what do you mean about the community program sidebar?

davehakkens:
> Have another look now? Looks like 850px is the sweet spot.

Looks good to me


> Of course the main sidebar is gone as per the design, what do you mean about the community program sidebar?

wasnt sure if the items on the sidebar are covered somewere else (if we delete it)
but looks like it actually is: 
- Profile guideliness are on top
- Community Program guidelines are still there when creating a map pin for 'spaces"
- Contact is now there in the new 'problems?' button

So nevermind my orignal comment on this 🫣

PR #3733, opened 2024-07-06
feat: add comment count to research updates

### LINKED ISSUE #3467: [Discussions] Should only fetch comments when discussion is expanded

benfurber:
PR Checklist

PR Type

## Description

It gets suprisingly difficult to know the right comment count of a research article. By adding the comment count to updates it's easier to then have a correct total on the parent. It turns out that some discussions don't have the `primaryContentId` set which was necessary for the new comment count to work. 

MUST BE FOLLOWED UP WITH A DATA UPDATE. LEAVE TO ME TO MERGE IN PLEASE. (Would still love a review, obvs.

## Git Issues

Closes #3467


benfurber:
One challenge I'm having is that not all discussions for research updates have `primaryContentId` set as they should. I could do a one-off DB update but I'm thinking through how to ensure it's set through the store.

src/stores/Discussions/discussionEvents.ts
```diff
@@ -45,37 +62,33 @@ export const updateDiscussionMetadata = async (
...
+
+      if (!research || !research.updates) return
+
+      const updates = research.updates.map((update) => {
+        return update._id === sourceId ? { ...update, commentCount } : update
+      })
+
+      const totalCommentCount = liveResearchUpdatesCommentCounts(updates)
+
+      return await researchRef.update({
```
mariojsnunes, line 85:
```
const dataToUpdate = {
  totalCommentCount,
  updates,
};

if (latestCommentDate) {
  dataToUpdate.latestCommentDate = latestCommentDate;
}

return await researchRef.update(dataToUpdate);
```

I think this would be more clear

mariojsnunes (approved):
LGTM

PR #3739, opened 2024-07-07
chore: make the functions watching happen in the Docker container

pizzaisdavid:
PR Type
- chore

## Description
Makes the Docker container responsible for building the functions. More-or-less this is an in-between step to some improvements that I want to make. "Just trust me, bro"

It does make the improvement that stopping the `backend:emulator:watch` does stop the container, and we don't need the stop command anymore.


pizzaisdavid:
bonk-bonk unrelated test failed

PR #3762, opened 2024-07-11
style: increase border radius of everything

benfurber:


benfurber:
@davehakkens

benfurber:
@davehakkens Have you had a look around on this yet? Seems pretty good to me.

There's this that I've noticed. Which I can fix, obvs, if you think this is worth continuing?
Screenshot 2024-07-17 at 14 15 04

davehakkens:
Not 100% convinced yet. Some areas yes, some no...
need to sleep longer on it ;)

PR #3763, opened 2024-07-11
feat: start adding tabs to settings form

benfurber:
PR Checklist

PR Type

## Description

First set of tabs for the updated settings page. Uses routing to different tab displays to ensure warnings about leaving unsaved changes is triggered when trying to change tabs.

## Screenshots
Screenshot 2024-07-17 at 09 09 37
Screenshot 2024-07-17 at 09 09 42
Screenshot 2024-07-17 at 09 09 57


benfurber:
@davehakkens Still working on this but for the branch to be finished I'll need icons please for:
- Profile
- Impact
- Notifications

What's already in the system:
Screenshot 2024-07-11 at 13 42 32

benfurber:
Ready for a sneak peak...

davehakkens:
c'mon preview link 🐌

davehakkens:
@benfurber 

Looking at that iconset it's a bit.. inconsistent. Many mixed styles in there.
We will clean up the entire icon set at a later point
For now I suggest we use these icons
[Profile Icons.zip](https://github.com/user-attachments/files/16182822/Profile.Icons.zip)
afbeelding

davehakkens:
I also noticed the Notifications in the set  👆
afbeelding

Is different from the Icon we actually use (dont see this one in the set)
afbeelding

We could also align them.
And replace them by this new version
[Notifications.zip](https://github.com/user-attachments/files/16182875/Notifications.zip)


afbeelding

benfurber:
@davehakkens Preview previewing now - there was a build error.

The last thing I want to workout is if it's possible to warn the user if they try to change tabs with unsaved form changes.

davehakkens:
Nice! 💪

1. Would be nice to still do a few UI tweaks so it looks more like the mockup, reccomend:
- Font-family: Varela round;
- Font-size: 18px;
- Icon size 20px
afbeelding

2. On my mobile I have a straight black border (not when I resize on desktop)
IMG_7373


3. Mobile functionality menu feels a bit unstyled. But could also put that in a different Issue/PR since it might be a big thing?
IMG_7372

4. Padding on the sides on mobile is very large, could be reduced to make the page more readable 

IMG_7374

benfurber:
@davehakkens 
1 and 4 done.

I can't replicate 2 but I tried a thing, did it do anything?

I agree about 3 being a follow-up.

Still not sure about the form blocking thing. Will ponder more on it Monday.

davehakkens:
@benfurber nice much better! 2 is also fixed here.

benfurber:
@davehakkens Mobile uses the right select now and the scrolling to the top should work now when trying to submit with errors (before it was only working on successful save).

packages/components/src/Select/Select.tsx
```diff
@@ -10,18 +10,22 @@ import type {
...
+type IOption = {
+  label: string
+  value: string
+}
+
 export interface Props extends ReactSelectProps {
   options: OptionsOrGroups<any, any>
   value?: any
-  onChange?: any
+  onChange?: (arg: any) => void
```
mariojsnunes, line 21: nice!

packages/components/src/SettingsFormWrapper/SettingsFormWrapper.tsx
```diff
@@ -1,55 +1,73 @@
...
-
-  const handleChange = (
-    _: React.SyntheticEvent<Element, Event> | null,
-    value: string | number | null,
-  ) => {
-    typeof value === 'number' && setValue(value)
+const useRouteMatch = (patterns: readonly string[]) => {
+  const { pathname } = useLocation()
+
+  for (let i = 0; i < patterns.length; i += 1) {
```
mariojsnunes, line 22: looks like the `i` isn't needed for anything other than accessing the index, so this could be used instead:
```
for (const pattern of patterns) {
...
}
```

packages/components/src/SettingsFormWrapper/SettingsFormWrapper.tsx
```diff
@@ -1,55 +1,73 @@
...
-            }}
-          >
-            {tabs.map((tab, index) => {
-              return <SettingsFormTab key={index} value={index} tab={tab} />
-            })}
-          </Flex>
+          {tabs.map((tab, index) => {
+            return (
+              <SettingsFormTab
+                key={index}
```
mariojsnunes, line 63: could use the tab title/path for key instead of index, should be unique. although it won't matter here as we don't change the tabs order

packages/components/src/index.ts
```diff
@@ -61,4 +61,4 @@ export { Username } from './Username/Username'
 export { UserStatistics } from './UserStatistics/UserStatistics'
 export { VideoPlayer } from './VideoPlayer/VideoPlayer'
 // export { IImageGalleryItem } from './ImageGallery/ImageGallery'
-// export { availableGlyphs } from './Icon/types'
+export type { availableGlyphs } from './Icon/types'
```
mariojsnunes, line 64: I remember commenting this out because there were some weird issues with esm imports. hope it works now!

mariojsnunes (approved):

packages/components/src/index.ts
benfurber: Can you say more about 'weird issues'... Seems to be fine...?

mariojsnunes: I think it was on the build step, so should be fine

davehakkens:
> @davehakkens Mobile uses the right select now and the scrolling to the top should work now when trying to submit with errors (before it was only working on successful save).

@benfurber Works for now indeed on error.
But when its successful it goes to the top, but no green banner anymore.

benfurber:
@davehakkens Sorry! Fixed in #3785.

PR #3781, opened 2024-07-17
[e2e] Refactor discussions

tuliobluz:
## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

- There are a lot of duplication codes in discussion scenarios that could be refactored in the `commandsUi.ts`
- Missing assertion to deleteComments amount

## What is the new behavior?

- Create addLastComment, editLast, deleteLastCommentOrReply, addReplytoFirstComment to avoid duplicate code
- Add `cy.get('[data-cy=edit-comment]').should('not.exist')` to cypress does not take the text from the edit modal, making sure the test checks the new comment in the right place, instead of the modal.
- Add a last assertion to validate that deleted comments are absent in the total amount.

- [ ] Yes
- [x] No


packages/cypress/src/support/commandsUi.ts
```diff
@@ -34,7 +34,12 @@ interface IOpeningTime {
 declare global {
   namespace Cypress {
     interface Chainable {
+      addLastComment(newComment: string): Chainable<void>
+      addReplytoFirstComment(reply: string): Chainable<void>
       clickMenuItem(menuItem: UserMenuItem): Chainable<void>
+      checkCommentItem(firstComment: string, length: number): Chainable<void>
```
benfurber, line 40: `firstComment` -> `comment`

packages/cypress/src/integration/howto/discussions.spec.ts
```diff
@@ -17,50 +17,39 @@ describe('[Howto.Discussions]', () => {
...
   })
 
   it('allows authenticated users to contribute to discussions', () => {
     const newComment = 'An interesting howto. The answer must be...'
     const updatedNewComment =
       'An interesting howto. The answer must be that when the sky is red, the apocalypse _might_ be on the way.'
-    const newReply = 'Thanks Dave and Ben. What does everyone else think?'
-    const updatedNewReply = 'Anyone else?'
+    const reply = 'Thanks Dave and Ben. What does everyone else think?'
+    const updatedReply = 'Anyone else?'
```
benfurber, line 28: Either comment and reply should be prefixed 'new' or not. Ending up `newComment` and `reply` is my least favourite end case. 

My preference would be to keep the new prefix. My intent was to highlight that comments were already present before the spec adds them. Plus not changing `newReply` to `reply` keeps the diff smaller.

packages/cypress/src/integration/howto/discussions.spec.ts
```diff
@@ -17,50 +17,39 @@ describe('[Howto.Discussions]', () => {
 
     cy.signUpNewUser()
     cy.visit(`/how-to/${item.slug}#comment:${firstComment._id}`)
-    cy.get('[data-cy="CommentItem"]').should('have.length.gte', 2)
-    cy.get('[data-cy="CommentItem"]')
-      .first()
-      .scrollIntoView()
-      .should('be.inViewport', 10)
-    cy.contains(firstComment.text)
+    cy.checkCommentItem(firstComment.text, 2)
```
benfurber, line 20: I don't think I like this command as is...

* It's checking two things, the scroll view and the @ mention comment content.
* It's more an expectation than a ui command, so it should probably be in commands rather than commands ui.

So let's go with two commands in the other file.

benfurber (changes requested):
Nice one @tuliobluz. 

My main feedback is around the naming. Rather than specifying last or first I had two thoughts first was having last as a default value like `selectTag` does. But then I also thought as we're not using that method in a different context, that's building functionality we won't use yet. So if you want to do that, go for it, otherwise suggested names below:

* `addLastComment` -> `addComment`
* `addReplytoFirstComment` -> `addReply`
* `deleteLastCommentOrReply` -> `deleteDiscussionItem`
* `editLast` -> `editDiscussionItem`

Another thought. I think all the commands should start with the cy step included (generally that's the pattern we've gone with).

The doc queries for latestCommentDate and notifications could deffo be commands too.

Oh and obvs my comments on the howto spec apply to the other two.

tuliobluz:
@benfurber Addressed all the comments, please review them again. I am not being able to squash the commits 

When I am trying I get this error message

`refactor: yarn format
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]
⚠   body must have leading blank line [body-leading-blank]
`

packages/cypress/src/integration/research/discussions.spec.ts
```diff
@@ -71,17 +58,20 @@ describe('[Research.Discussions]', () => {
...
 
     cy.step('Can delete their reply')
-    cy.get('[data-cy="ReplyItem: delete button"]:first').click()
-    cy.get('[data-cy="Confirm.modal: Confirm"]:first').click()
-    cy.contains(updatedReply).should('not.exist')
+    cy.deleteDiscussionItem('ReplyItem')
+    cy.contains(updatedNewReply).should('not.exist')
     cy.contains(`1 Comment`)
+
+    cy.step('Check comments number after deletation')
```
benfurber, line 70: Typo: deletion

benfurber (changes requested):
Nice.

I think the new commands are still missing a `cy.step()`?

What command with the squash issue were you running? I find using fixup in an interactive rebase works best.

tuliobluz:
I did not add the steps because the test files have `cy.step` then it is easy to read the file, however, I can move to there `commandsUi`

benfurber:
@tuliobluz The error message suggests it's more a problem with the commit message than the squashing.

PR #3787, opened 2024-07-19
feat: updated settings - move map

benfurber:
## PR Checklist

## PR Type

- [x] Bugfix (fixes an issue)

Map pins are nested in the main user profile settings tab.

## What is the new behavior?

1. Moves the map pin setting to its own tab
2. Seperate user and map pin stores
3. Increase e2e coverage
4. Fixed impact tab anchoring

- [ ] Yes
- [x] No

## Screenshot
Screenshot 2024-07-24 at 13 31 07


benfurber:
Still very much a WIP.

davehakkens:
Nice. But would be nice to align a bit more to the design from @dalibormrska which is a bit more organised. Not sure if you need specific explanation or you know what I refer to @benfurber?

Also "location data saved" might be nicer to replace by "Your current map pin is here:"

src/pages/UserSettings/content/formSections/SettingsMapPin.section.tsx
```diff
@@ -0,0 +1,328 @@
...
+const LocationDataTextDisplay = ({ location }: ILocationProps) => {
+  if (!location)
+    return (
+      <Text variant="paragraph" data-cy="LocationDataTextDisplay">
+        {mapForm.noLocationLabel}
+      </Text>
+    )
+
+  return (
+    <>
```
mariojsnunes, line 48: no need the fragment

mariojsnunes (approved):

benfurber:
@davehakkens Yes, sorry, went to write a note with my proposed changes. 

When editing a pin, due to how the map loads I realised it's super hard to understand what is currently saved. So I thought displaying it was the easiest option, made good end-to-end testing simply and probably just good practice. Very happy for feedback/improving that, so will update the copy as you suggested and happy for more.

Two fairly held strong opinions on a couple of UX points that I didn't go through with @dalibormrska yet, but happy to be disagreed with:
1. Using a delete icon over the map isn't obvious enough and doesn't make clear it would also clear the description field. I like the icon itself though, could change the button text to include it?
2. The explaination of how to use the map - if you want to help someone use a thing, you have to explain BEFORE the field, not after.

davehakkens:
@benfurber ok that makes sense to me. But there are still a fe additional UI changes.
> 1. Using a delete icon over the map isn't obvious enough and doesn't make clear it would also clear the description field. I like the icon itself though, could change the button text to include it?

Yes adding icon would be nice

2. Page headings and paragraphs are styled different on each page. Different font-size for titles and different font used for the paragraph text. Would be good to align this more, as in the figma file.
image
image
image


3. Would be nice to merge these two text blocks together and remove the blue backround. As in the figma file.
image

benfurber:
Map specific points 1 and 3 in #3801. The consistent point I'll do in a follow-up.

PR #3798, opened 2024-07-25
feat: move email notification setting to own tab

benfurber:
## PR Checklist

## PR Type

Email notification nested in the user profile tab.

## What is the new behavior?

In it's own tab with own store method. 

Screenshot 2024-07-25 at 14 31 50

### DIFF
--- a/src/pages/UserSettings/content/formSections/SettingsNotifications.section.tsx
+++ b/src/pages/UserSettings/content/formSections/SettingsNotifications.section.tsx
@@ -1,0 +1,122 @@
+import { useState } from 'react'
+import { Form } from 'react-final-form'
+import { Loader } from 'oa-components'
+import { useCommonStores } from 'src/common/hooks/useCommonStores'
+import {
+  buttons,
+  fields,
+  notificationForm,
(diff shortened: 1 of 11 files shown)


src/pages/UserSettings/content/formSections/SettingsNotifications.section.tsx
```diff
@@ -0,0 +1,122 @@
...
+    IFormNotification | undefined
+  >(undefined)
+
+  const { userStore } = useCommonStores().stores
+  const user = userStore.activeUser
+  const { description, title } = fields.emailNotifications
+
+  const formId = 'Notifications'
+
+  if (!user) return
```
mariojsnunes, line 29: I'd move this after the onSubmit function. It would become an issue if  we converted the onSubmit to a useCallback hook.
also, it's better to return null for react components (instead of a fragment or undefined)

mariojsnunes (changes requested):

benfurber:
Ah! Click merge when I meant to rebase to incorporate your feedback. D'oh. Will do in follow-up.

PR #3816, opened 2024-08-06
fix: improve comments performance

mariojsnunes:
## PR Checklist

## PR Type

- [x] Bugfix (fixes an issue)

Adding comments is very slow (13 seconds in a particular case!).
Due to notifications, the more contributors it has, the worse it gets!

Lack of loading indicators makes it worse.

## What is the new behavior?

Does not await discussion metadata updates and user notifications.
Locally it reduces the time from 4s to 1s.
1 second should be constant now, so for prod it would reduce from 13s to 1s.

- [ ] Yes
- [x] No

## Git Issues

Closes #

### DIFF
--- a/packages/components/src/CreateComment/CreateComment.tsx
+++ b/packages/components/src/CreateComment/CreateComment.tsx
@@ -50,7 +50,9 @@ export const CreateComment = (props: Props) => {
             },
           }}
         >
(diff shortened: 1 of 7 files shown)


davehakkens:
Seems to work well, also nice loading indication

benfurber (approved):

PR #3827, opened 2024-08-19
feat: add generic map component

benfurber:
## PR Checklist

- [x] - Unit and e2e tests for the changes that have been added (for bug fixes / features) (some)

## PR Type

- [x] Bugfix (fixes an issue)

The main map and the map generated for settings, were seperate instances of the map, which could lead to different behaviour.

## What is the new behavior?

Both now call the same map in the component library. Also fixes not loading from the map api locally.


benfurber:
Manually tested locally and on preview branch. I don't believe this creates any user changes so will merge in once checks pass.

mariojsnunes:
instead of moving the css to packages/components/src/Map/index.css, which might cause issues with the remix migration. Could we have that class inline?

benfurber:
@mariojsnunes Would happily move the tiny contents of `packages/components/src/Map/index.css` to line, but the main leaflet css file is 600+ lines. With remix/vite I really should be able to import a css file!

mariojsnunes:
Let's not make the migration harder than it needs to be... for now we know how to import from the main app.
To import from a package, we need to invest more time hacking around to figure it out.

benfurber:
@mariojsnunes Good point.

PR #3832, opened 2024-08-28
feat: add more detailed cards with creator details

benfurber:
## PR Checklist

## PR Type

- [x] Bugfix (fixes an issue)

## What is the new behavior?

https://github.com/user-attachments/assets/013b746b-2978-4d73-9b3f-a07eda701923


benfurber:
I might need to deploy the api updates before this is really testable.

mariojsnunes (approved):

benfurber:
@dalibormrska @davehakkens Preview branch working with the updated api data.

davehakkens:
Nice. I see a few small design things but leave that up to @dalibormrska 

Visually it does look very pink on there now..
And the "want to get started" is a group, shown in the icon
However now its also shows it as a "Label", not sure if that is correct for consistency.

But also happy to iterate on this on the go
Curious to test on live!

dalibormrska:
It looks very nice overall, great work! I already see that the experience is very smooth, much less laggy.

3 things I had in mind now when playing around with the [current build](https://onearmy-next--pr3833-feat-new-map-add-tab-muckcbgc.web.app/map).

It might be very nice to show the relation between the cards and the map in a way that when you hover through the cards, the pin gets highlighted. It could be a simple hover animation for highlight such as [this](https://codepen.io/Dalby/pen/eYwMovv). The map pins themselves might not need the hover animation though.

Then I thought about the username tags. We could just make a variant of that component, where we remove the flag and change the username for the displayname. The component is already developed well with the hover effect and the text folding to multiple lines. There might be still some improvements to do to the component, I would maybe try to make the verified icon as an inline-block element so that it can snap to the end of the last word and also not get shrinked in size.
image


Lastly the buttons for showing the map or the list. I thought about only showing them for tablet and mobile where the screen size forces you to see only one of them at a time. But if you want to, we can try to experiment with showing them on desktop as well. In that case however, the wording might be a bit confusing, since we see the map or the list already. The action in this case would be to expand the map or the list. Other than that, it would be nice to position the button absolutely, centered and to think about the structure of the containers because on mobile they will need to get pushed up, as you can see in the screenshot.
image

dalibormrska:
Oh yeah. I also see the different display of cards. I'm surprised that it doesn't actually bother me so much that they are not aligned further down. I set the small member card to 130px height and the larger card for spaces to 280px. That made it so that they align nicely, but feel free to experiment. For now I think it can stay as it is. I would love to also see how they feel later on with more data in the cards and also live...

I see your point Dave about the pinkyness, however, let's see when we have more tags also for the spaces, I think it will actually help to differentiate the cards even better, since the color-coded icon (WS,MS,CC,Member) are a bit subtle, and the color differentiation gets a bit lost between the colorfulness of the user's image.

PR #3841, opened 2024-09-04
feat: add profile types filtering to new map

benfurber:
## PR Checklist

## PR Type

No filtering on the new map.

## What is the new behavior?

Adds basic filtering by profile type to the new map.

Screenshot 2024-09-05 at 13 47 08


benfurber:
@davehakkens Early view for you of the basic interaction.

davehakkens:
Cool!
Few things
1. Would we still merge workspaces together and replace the specifics for a label?
2. Verified + Want to get started + machine shop doesnt seem to filter. (0 results)
3. If you search 'shredder" you also see a "want to start" and "community point". I think this is data problem created when people toggle between different types or deleting their accounts but not updating map data. Seems like this might need a cleanup? (as the inaccuracy gets more visible now)

benfurber:
@davehakkens I just used the existing filter set hense the issues with 1 and 2. 

Totally agree about 3.

PR #3843, opened 2024-09-05
feat: updated pin card

benfurber:
## PR Checklist

## PR Type

New map used old pins.

## What is the new behavior?

Uses the new card style.
Screenshot 2024-09-05 at 16 43 36
Screenshot 2024-09-05 at 16 29 30
Screenshot 2024-09-05 at 16 29 43
Screenshot 2024-09-05 at 16 31 21


davehakkens:
Needs a @dalibormrska in here :)

dalibormrska:
Uuuu, juicy!!! Overall looks goooood, there were two things that popped into my mind... 

1. Looking at the closing button, I see the problem with the button component in this case. I see that as it is it's hard to customise it and make it round and without border as in the screenshot. Currently I think it is too prominent and takes too much attention if it has the primary color and the border. image image
Later on, if needed, I could see myself visiting the button component and designing all the variants we might need so that we can use the button component consistently across the platform in suitable variants. What do you think could be the solution for now? I can see that the filter in the old map had the closing button made quite simply. Could we maybe replicate that? Having the closing SVG (maybe a bit bigger than the one) with a white circle behind it? image
The svg could be scaled to 32x32px as well as the circle.image

dalibormrska:
The second thing would be that the card does not need to have the hover effect in the map. But awesome progress nevertheless!  🎉

benfurber:
I agree about the button prominence and card hover (which is partly because I don't want to duplicate some work in #3841). I'll have a play with both.

PR #3850, opened 2024-09-11
feat: add location search to new map

benfurber:
## PR Checklist

## PR Type

No location search on the new map.

## What is the new behavior?

Existing location search added to the new map:
https://github.com/user-attachments/assets/4822df7f-5062-4a54-8d1b-b8a3606d659c


mariojsnunes (approved):

benfurber:
@davehakkens @dalibormrska Worth having a look. Uses the same business logic of the existing search.

benfurber:
@mariojsnunes Do you mind if this gets merged in now, pretty please!

dalibormrska:
image

image


Nice one! Great to see the search incorporated! I would be advocating here for having a search icon on the right side of the input field for easier recognition. Nowadays it is not needed to have a separate button beside, but the icon would be a great help, should be clickable, but not a must (in edge cases for example on mobile after you type and the keyboard gets buggy and does not open, it is good for the user to have the option to run the search with the button).

@benfurber Should I make a new issue, since I believe it is related to adjusting the component itself?

benfurber:
@dalibormrska Good idea! Yes, add as an issue please.

PR #3852, opened 2024-09-17
feat: add function for pins after user updates

benfurber:
## PR Checklist

## PR Type

- [x] Bugfix (fixes an issue)

User details on map pins not updated when user updates them.

## What is the new behavior?

Changes to user data now copied over to map pin creator details.


functions/src/userUpdates/updateMapPins.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const updateMapPins = async (prevUser: IUserDB, user: IUserDB) => {
+  if (!hasDetailsForMapPinChanged(prevUser, user)) return
+
+  const snapshot = await db
+    .collection(DB_ENDPOINTS.mappins)
+    .where('_id', '==', user._id)
+    .get()
+
+  if (!snapshot.empty) {
```
jsonDoge, line 20: Could be written as `if(snapshot.empty) { return; }` to avoid wrapping large blocks in ifs. Similar to above `hasDetailsForMapPinChanged`.

functions/src/userUpdates/utils.test.ts
```diff
@@ -1,12 +1,80 @@
...
+  hasLocationDetailsChanged,
   hasUserImageChanged,
-} from './hasKeyDetailsChanged'
+} from './utils'
 
 import type { IUserDB } from '../../../src/models'
 
-describe('hasKeyDetailsChanged', () => {
+describe('hasDetailsChanged', () => {
+  it("returns false for every field that's the same", () => {
```
jsonDoge, line 12: Mixed quote usage, prettier seems to indicate single quote priority

functions/src/userUpdates/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+    ...hasDetailsChanged(prevUser, user),
+    ...hasLocationDetailsChanged(prevUser, user),
+  ]
+  return !!detailsChanged.find((detail) => detail === true)
+}
+
+export const hasUserImageChanged = (
+  prevUser: IUserDB,
+  user: IUserDB,
+): boolean => {
```
jsonDoge, line 58: Function logic could also be written as:
`
  if (!prevUser.userImage && !user.userImage) return false

  if (!prevUser.userImage || !user.userImage) return true

  return !valuesAreDeepEqual(prevUser.userImage, user.userImage)
`
But current may be easier to read with quick glance. Apart from the unreachable `return false` at the end.

functions/src/userUpdates/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+    return !valuesAreDeepEqual(prevUser.userImage, user.userImage)
+  }
+
+  if (prevUser.userImage && !user.userImage) return true
+  if (!prevUser.userImage && user.userImage) return true
+
+  return false
+}
+
+export const getCreatorImage = (userImage: IUserDB['userImage']) => {
```
jsonDoge, line 71: How about?
`=> userImage?.downloadUrl || null`

functions/src/userUpdates/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  if (userImage && userImage.downloadUrl) {
+    return userImage.downloadUrl
+  }
+  return null
+}
+
+export const getFirstCoverImage = (coverImages: IUserDB['coverImages']) => {
```
jsonDoge, line 78: Similar technique can be used here:
`=> coverImages?.[0]?.downloadUrl || null`

functions/src/userUpdates/utils.test.ts
benfurber: This is because of the single quote use in the string.

functions/src/userUpdates/utils.ts
benfurber: I think I prefer the longer readable version. Good point about the unreachable false though!

benfurber:
Thanks for the review @jsonDoge!

functions/src/userUpdates/utils.test.ts
jsonDoge: You are right, checked with prettier auto formatting seems that this is the current set preference instead of escaping 👍

PR #3861, opened 2024-09-24
tests: cy howto read flaky

tuliobluz:
## PR Checklist

## PR Type

- [x] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Currently, the scenario **[Read a How-to][By Everyone][See all info]** has flaky steps when trying to assert the data on the how-to. The cause is a rerender from a cypress that loses track.

Example:

`assert expected <div.howto-description-container.css-703r4> to contain product`

[cypress cloud with the issue](https://cloud.cypress.io/projects/4s5zgo/runs/6249/test-results/25283bbf-c706-4201-96ac-f3bd951796d4/replay?actions=%5B%5D&att=1&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%7B%22value%22%3Atrue%2C%22label%22%3A%22Flaky%22%7D%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&pc=log-http%3A%2F%2Flocalhost%3A3456-130__command-logs&specs=%5B%5D&statuses=%5B%5D&testingTypesEnum=%5B%5D&ts=1727121987362.1)

## What is the new behavior?

To avoid the page being rerendered I used the [aliases](https://docs.cypress.io/guides/core-concepts/retry-ability#Correctly-ending-chains-after-an-action)

- [ ] Yes
- [x] No

### DIFF
--- a/packages/cypress/src/integration/howto/read.spec.ts
+++ b/packages/cypress/src/integration/howto/read.spec.ts
@@ -50,13 +50,14 @@ describe('[How To]', () => {
 
     beforeEach(() => {
       cy.visit('/how-to')
-      // Hack to get the tags store to finish populating before the expectations
-      cy.wait(1000)
     })
 
     describe('[By Everyone]', () => {
       it('[See all info]', () => {
         const howto = howtos[0]
+        // Hack to avoid flaky test as the tags are not being loaded on time
+        cy.queryDocuments('howtos', '_id', '==', howto._id)
+
         cy.visit(specificHowtoUrl)
         cy.step('Edit button is not available')
         cy.get('[data-cy=edit]').should('not.exist')


tuliobluz:
Before merging this PR, can you retry the e2e jobs to see if the flaky is still happening?

mariojsnunes:
we haven't had flaky tests for a while

tuliobluz:
For the previous runs 9/15 happened this flaky as you can see [here](https://cloud.cypress.io/projects/4s5zgo/analytics/flaky-tests).

mariojsnunes:
Ah you are right, I was thinking about failed runs

tuliobluz:
The test is still flaky, I will continue to investigate it.

benfurber:
Also looks like the question discussion spec is more of a pain atm. I wonder if it's because of changes I made [to the discussion data mocks recently](https://github.com/benfurber/community-platform/commit/7bcd566b7bb56dd4a66afd07d75ee98e4716c9d9)

tuliobluz:
If I remember correctly, this issue has been happening for several months. I tried to investigate before. I attempted different approaches with Cypress but had no success. It might be related to how `howto` populate the `tags`.

tuliobluz:
I found a workaround that might fix the flaky regarding tags not being displayed, I ran it several times locally and could not reproduce the flaky anymore.

@benfurber can you please retry few times the e2e jobs to see if it still has any flakiness? If not we can merge the PR, if agree with the solution. I don't like the solution because it does not make sense at all.

PR #3862, opened 2024-09-25
Update map setting display

### LINKED ISSUE #3856: feat: update map setting display
### Description

The setting to add/update a map pin (at `/settings/map`) does it's thing but more context could be given to the user so they understand how their info will be displayed on the map.

### Design requirements and description by @dalibormrska 

The design proposal for now is this. First of all, the user would have the style of the pin based on the type of profile they have.

The preview of the card would be done just as it would be on the real map. The pin would be always in the open state and draggable. On click we would want to move the pin to the bottom of the map container so that both the pin and the card would be visible. I thought this would save space, layout decisions and introducing another element into the pool. In this way we would just utilise the existing elements.

image

Make sure that the height of the map is enough for the card of a space profile to show.

### Build suggestion

While the necessary user data will need to be passed through to `MapWithPin` on `SettingsPageMapPin`, hopefully `MapPin` (`packages/components/src/MapWithPin/MapPin.tsx`) is the only component that needs much work. And what needs implementing is probably very close to what's used on `Clusters` (`src/pages/Maps/Content/MapView/Cluster.tsx`) with the `Popup` component. 

As `MapPin` is in the component library already (which is what we want), moving `Popup` (`src/pages/Maps/Content/MapView/Popup.tsx`) to the component library too would be the best path forward.

LahuenGR:
## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Standard pin showing in the Map Settings and without relevant information

2024-09-25 04_16_23-Window

## What is the new behavior?

1) Pin is now showing the profile type design and, when clicked, shows the popup with relevant information
2) Removed one-click option to move the pin and setted it to double-click, as it was not working well with the click-n-popup idea

2024-09-25 04_20_16-Window

_Describe the new behaviour_

- [ ] Yes
- [x] No

## Git Issues

Closes #3856

## What happens next?


packages/components/src/MapWithPin/MapWithPin.stories.tsx
```diff
@@ -1,6 +1,9 @@
+import React from 'react'
```
benfurber, line 1: style preference:
```
import { useRef } from 'react'
```

benfurber (changes requested):
Thanks @LahuenGR. I've added on quick style thing I could see but it's a little hard to review fully as you're pulling in unrelated commits from your other branch. Plus maybe some unrelated model changes?

Also. Have you looked into our cypress tests yet. Any ideas on how to test your change within the existing settings test(s)?

LahuenGR:
@benfurber I'm not sure why it dragged the commit from the other branch 🤔 let me try to fix it

What model changes are you referring to?

As for the cypress tool... nope, never used it hahaha, i'm going to need some help with that <3

benfurber:
From QA.

1. Definitely need to increase the map height. If you change your profile type to something other than a member you'll see what I mean.
2. Add onblur to hide the pin when click/tapping the map. This is especially important for mobile.
3. Move the +/- controls to the bottom left, on mobile they nearly overlap with the search for address.
4. For pin onhover change the cursor to grab.
5. I don't think the explainer text is right because it is possible still to grab to move your pin. Let's go with: "To move your pin, grab it to move it or double click where you want it to go. Tap on your pin to see how it'll look on the map.

LahuenGR:
Did i understand correctly?

https://github.com/user-attachments/assets/c45fdf03-289c-401b-8d6e-f18120beb14f

src/pages/UserSettings/SettingsPageMapPin.tsx
```diff
@@ -223,7 +245,7 @@ export const SettingsPageMapPin = () => {
     >
       <Flex sx={{ flexDirection: 'column', gap: 1 }}>
         <Heading as="h2" id="your-map-pin">
-          {mapPin ? addPinTitle : yourPinTitle}
+          {mapPin ? yourPinTitle : addPinTitle}
```
benfurber, line 248: Why this switch in logic? I think it's wrong and why the unit test is failing.

benfurber (changes requested):

src/models/index.ts
```diff
@@ -1,5 +1,3 @@
-import type { IComment } from './discussion.models'
-
```
benfurber, line 2: Always appreciate some tidy-up! But maybe better in a seperate PR as it's unrelated to the scope of this PR.

src/pages/UserSettings/SettingsPageMapPin.tsx
```diff
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react'
+import React, { useEffect, useState } from 'react'
```
benfurber, line 1: `import { useEffect, useRef, useState } from 'react'

benfurber:
Hey @LahuenGR I've sorted out the complex rebase that was needed on this branch after we merged in a big refactor.

I've run the branch locally and I think there's a couple more bits to finish. If you won't have time to do it, no worries, just let me know and I can finish it off.

Screenshot 2024-10-08 at 15 03 37

1. Let's move those controls (+/- and location search) to the top.
2. For some reason the zoom out button doesn't work. Not sure if that relates to your branch or not but if you could take a look that would be awesome.

You're using `ondragstart` and `ondragend` for props with the map. On MapWithList, etc. we use the `onclick` prop to close a pin. Might be worth trying that?

benfurber:
Hey @LahuenGR, hope you're well. How are you getting on with this? :)

benfurber:
Thanks for your work on this @LahuenGR!

PR #3868, opened 2024-09-26
fix: setting countryCode when updating country

benfurber:
## PR Checklist

## PR Type

- [x] Bugfix (fixes an issue)

Setting your country (on the infos settings tab) doesn't set/update your flag icon on comments, etc.

## What is the new behavior?

https://github.com/user-attachments/assets/1045eb4d-c167-4b67-99ac-f168916e66ca

Also adds emojis and a little preview to show users how their location will be shown:
Screenshot 2024-09-26 at 15 31 12
Screenshot 2024-09-26 at 15 31 22


davehakkens:
oe nice visuals flags! 💫
Seems to work well here!
Would you still have in mind at some point to decouple the flag from the map location, or leave it how it currently functions?

src/stores/User/user.store.ts
```diff
@@ -218,6 +219,17 @@ export class UserStore extends ModuleStore {
       )
     }
 
+    if (location && location.country) {
```
jsonDoge, line 222: can be shortened to `location?.country`  :)

src/stores/User/user.store.ts
```diff
@@ -218,6 +219,17 @@ export class UserStore extends ModuleStore {
...
+      const countryCode = Object.keys(countries).find(
+        (key) => countries[key].name === location.country,
+      )
+      updatedUserProfile.location = {
+        ...location,
+        country: location.country,
```
jsonDoge, line 228: If the object is being populated by `...location`, the extra assignment `country: location.country` is redundant :)

PR #3890, opened 2024-10-03
feat: profile views counter

iSCJT:
## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Profile views were not logged / displayed.

## What is the new behavior?

When a profile is viewed for the first time in a 'session' then a view is logged against the user profile.  Uses the same logic as for the howto, research and question logging.  One thing to note is that the first time someone views a profile within the session their own view won't count towards the displayed value.  This is consistent with the other document types.

image

- [ ] Yes
- [x] No

## Git Issues

Closes #

## What happens next?


davehakkens:
Oe I like it!
I hope @benfurber or @mariojsnunes approves it 🫣
One remark: Could we add it in a beta-tester authwrapper first?

mariojsnunes (approved):
Looks good. We can merge this as a quick improvement.
But in the short term, these view counter increments will move serve-side.

iSCJT:
I've added the beta feature wrapper.

benfurber:
One risk from a product prospective. We're obvs only starting to count now. So seeing a low number of profile views for established users might be demoralising (plus give first time users a misunderstanding of a profile's reach and the depth of the community.

src/assets/icons/eye.svg
benfurber, line 1: There's an icon for this already.

src/pages/User/content/MemberProfile.tsx
```diff
@@ -31,6 +37,17 @@ export const MemberProfile = ({ docs, user }: IProps) => {
     ? cdnImageUrl(userImage.downloadUrl)
     : DefaultMemberImage
 
+  const total_views = user.total_views || 1
```
benfurber, line 40: I think if for some reason a total can't be called from the DB, it would be better to show nothing?

src/pages/User/content/SpaceProfile.tsx
```diff
@@ -302,6 +316,7 @@ export const SpaceProfile = ({ user, docs }: IProps) => {
                       howtoCount={docs?.howtos.length || 0}
                       usefulCount={user.totalUseful || 0}
                       researchCount={docs?.research.length || 0}
+                      total_views={renderTotalViews()}
```
benfurber, line 319: totalViews

packages/components/src/UserStatistics/UserStatistics.tsx
```diff
@@ -18,6 +19,7 @@ export interface UserStatisticsProps {
   howtoCount: number
   usefulCount: number
   researchCount: number
+  total_views?: EmotionJSX.Element
```
benfurber, line 22: Calling the element isn't the style of this component, all the rest are numbers. So I think the approach should be the same.

shared/mocks/data/users.ts
```diff
@@ -209,6 +209,7 @@ export const users = {
     machineBuilderXp: null,
     _authID: 'vWAbQvq21UbvhGldakIy1x4FpeF2',
     about: null,
+    total_views: 10,
```
benfurber, line 212: totalViews

benfurber (changes requested):
Feature test please! At least to prevent regression of losing the display.

packages/components/src/UserStatistics/UserStatistics.tsx
iSCJT: It's like this because we're using the beta feature wrapper.  Think it's the simplest implementation that keeps the UserStatistics component purely as a display component.  Open to pointers on another way of doing it though?

src/pages/User/content/MemberProfile.tsx
iSCJT: The reason I defaulted to 1 is because the only reason we shouldn't return anything is if it's their first view (or I guess a db error but then I think the whole component doesn't render) but happy to turn off the default.

shared/mocks/data/users.ts
iSCJT: I've only used total_views to keep it consistent across all the document types so that we can use the same incrementing function and logic for all of them.  We could refactor everywhere?

src/assets/icons/eye.svg
iSCJT: I did look for one but didn't think to search for 'view' 🤦

src/pages/User/content/MemberProfile.tsx
```diff
@@ -92,6 +109,7 @@ export const MemberProfile = ({ docs, user }: IProps) => {
               researchCount={docs?.research.length || 0}
               usefulCount={user.totalUseful || 0}
               sx={{ alignSelf: 'stretch' }}
+              total_views={renderTotalViews()}
```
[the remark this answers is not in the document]
iSCJT: I used render to keep consistent with the space profile naming for other components but will change.

packages/components/src/UserStatistics/UserStatistics.tsx
benfurber: Oh got it, makes sense...

I guess something like adding another props to this component like `showViews` would work. Then the beta wrapper could call that prop but the public one doesn't.

shared/models/user.ts
```diff
@@ -204,6 +204,7 @@ export interface IUser {
   machineBuilderXp?: IMAchineBuilderXp[] | null
   isExpert?: boolean | null
   isV4Member?: boolean | null
+  total_views?: number
```
benfurber, line 207: I take your point about consistency, which is why I think it should be totalViews everywhere.

src/assets/icons/eye.svg
benfurber, line 1: Missed removing this file now you're using the `Icon` component?

shared/models/user.ts
iSCJT: Just been thinking about this and personally don't think it's worth the effort to update every how to, research, question doc to migrate to the new property name.  I think it would be better to pick this up as part of full db migration, there's a number of doc properties that need renaming.

packages/components/src/UserStatistics/UserStatistics.tsx
iSCJT: Just double checked to make sure... we can't use the auth wrapper for individual props.  Think the options are:

1. leave as is
2. not use the standard auth wrapper beta feature and write separate logic that checks whether the user is a beta user and then send the views prop as undefined if not.

iSCJT:
> One risk from a product prospective. We're obvs only starting to count now. So seeing a low number of profile views for established users might be demoralising (plus give first time users a misunderstanding of a profile's reach and the depth of the community.

This is a good point by the way, we could:

- accept the risk as is (possibly with some indication of when view logging started from)
- build things out more so we can display time bound views, e.g. 'views in last 30 days' although that requires some kind of indication of the time range / less simple screen design
- do without publicly displayed profile views but continue to log them

@davehakkens

davehakkens:
This is one of the reasons to add it now and start counting in the (beta) background.

We still have time to think about how to display this in a more impactful way

packages/components/src/UserStatistics/UserStatistics.tsx
benfurber:
```
<AuthWrapper 
  roleRequired={UserRole.BETA_TESTER}
  fallback={<Component oneSetofProps />}
>
  <Component anotherSetofProps />
</AuthWrapper>

Issue #3908, opened 2024-10-18
feat: add profile tags to all profile types


benfurber:
Many elements of the Precious Plastic specific profile types (workspaces, machine shop, etc.) are highly specifc. This makes extension for other projects more complex.

The main is to remove all the space fields for the PP specific profile types and replace them with a more generic profile tags field which can then also be added to the member and space profile types.

To-do:
- [x] Add tags to typing
- [x] Add tags to the user update functions
- [x] Add tags to machine shops profiletype setting
- [x] Add tags to machine shops profile page
- [x] Update machine shop data on prod DB 
- [x] Add tags to workspaces profiletype setting
- [x] Add tags to workspaces shops profile page
- [x] Update workspaces data on prod DB 
- [x] Add tags to collection point profiletype setting
- [x] Add tags to collection point shops profile page
- [x] Update collection point data on prod DB 
- [x] Add tags to community profiletype setting
- [x] Add tags to community profile page
- [x] Add tags to spaced profiletype setting
- [x] Add tags to spaced profile page
- [x] Update spaces tags data to theme level
- [x] Update member tags data to theme level
- [x] Add tags to member profiletype setting
- [x] Add tags to member profile page
- [ ] Update design of select list field (for this field and other tag fields)
- [x] Add rule so that five really is the max number that can be selected

benfurber:
@davehakkens Do you mean have a seperate list of tag options?

davehakkens:
> @davehakkens Do you mean have a seperate list of tag options?

Could be. But not sure if that is worth the added complexity at this point? What do you think?

I do know that at minimum we needs tags for spaces

benfurber:
@davehakkens As community and members don't have any custom fields atm it's super easy to add them.

Can you please provide a list of labels for:
1. PP members, PK members and FF members.
2. PP community.
3. PK space and FF space.

Happy receive 1, 2 and 3 whenever. Need all parts of 1 or 3 at the same time to make releasing simple.

benfurber:
> > @davehakkens Do you mean have a seperate list of tag options?
> 
> Could be. But not sure if that is worth the added complexity at this point? What do you think?
> 
> I do know that at minimum we needs tags for spaces
> 
> Members is a nice to have

Sorry, @davehakkens I missed that the first setence was a question. I think it's worth doing all of this. I see the value from a product prospect, it's not that much more work to do it all and then the code is simplier at the end as we have one way of doing it all.

[...]

PR #3938, opened 2024-10-23
Feat: add profile labels to collection point profile type

benfurber:
## PR Checklist

## PR Type

Collection points use two custom fields for setting and displaying data.

## What is the new behavior?

Uses only the new generic profile tags field.

Screenshot 2024-10-23 at 14 21 02

- [ ] Yes
- [x] No

## Git Issues

Part of #3908


benfurber:
@davehakkens Thinking about the screen shot. There's seven plastic types but we're asking users to select a max of five tags. Not a big deal just thought I'd note it.

benfurber:
Will start on the data update now. 

As there's two fields to change the plan is to set the plastic types as tags and add the open times to user's about field. There's a risk the open times are already on the about not from a random sample of live profiles, they weren't mentioned.

mariojsnunes (approved):

PR #4002, opened 2024-11-12
feat: update profile tags display

benfurber:
## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## What is the new behavior?

Screenshot 2024-11-12 at 15 02 14


benfurber:
@davehakkens This is what I'm able to do quickly with the multi-select field to add some more interest.

Screenshot 2024-11-12 at 15 35 59

I could implement the full design but it would require 1-3 days of dev as I'd need to use a completely different component for the field and I'm not sure it's worth doing that at this stage. Thoughts?

davehakkens:
> @davehakkens This is what I'm able to do quickly with the multi-select field to add some more interest.

Cool. But I thought we skipped using the colors for this? (because they can overlap between different spaces and don't align with the academy colors)

Or what logic do you apply to choose a color?

davehakkens:
> @davehakkens Oh OK, I thought we decided to prioritise adding extra styling to this and so I went off what was on the final design tag. I'm processing now your comments.

Agree we should do some final styling comments, (round borders etc)
But I dont think the mixed color scheme design was fully ready to implement.

However also dont mind if the feature is there. Just dont think PP or PK are ready to properly use it at this point.
But perhaps in the near future

davehakkens:
And what will happen with these old labels and styles, will this data be moved into the new 'system'?
image

mariojsnunes (approved):

PR #4005, opened 2024-11-14
feat: drop old map

benfurber:
## PR Checklist

## PR Type

The new map was a very slow render and had a few little bugs.

## What is the new behavior?

Fixes the speed issues (and bugs) to a point where the old map can be removed.


packages/components/src/CardList/CardList.tsx
```diff
@@ -3,36 +3,34 @@ import { Flex, Text } from 'theme-ui'
...
+const DEFAULT_BREAKPOINTS = { 600: 1, 1100: 2, 1600: 3 }
 
 export const CardList = (props: IProps) => {
-  const { columnsCountBreakPoints, dataCy, filteredList, list } = props
+  const { dataCy, list } = props
 
-  const listToShow = filteredList === null ? list : filteredList
-  const displayItems = listToShow
+  const displayItems = list
+    .splice(0, 30)
```
mariojsnunes, line 22: do not use `splice` as it mudates the original array. Can use `slice` or `toSpliced`

src/pages/Maps/Content/MapView/MapContainer.tsx
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+
+  const onBlur = () => {
+    navigate(`/map`, { replace: true })
+  }
+
+  const onPinClick = (pin) => {
+    navigate(`/map#${pin._id}`, { replace: true })
+  }
```
mariojsnunes, line 73: do we want replace `true` here? that means the user can't navigate back, which could be useful, no?

src/pages/Maps/Content/MapView/MapList.tsx
```diff
@@ -0,0 +1,117 @@
...
+    showMobileList,
+  } = props
+
+  const onFilterChange = (changedOption: MapFilterOption) => {
+    const isFilterPresent = !!activePinFilters.find(
+      (pinFilter) => pinFilter.label == changedOption.label,
+    )
+
+    if (isFilterPresent) {
+      return setActivePinFilters((filter) =>
```
mariojsnunes, line 43: should return after

src/pages/Maps/Content/MapView/MapList.tsx
```diff
@@ -0,0 +1,117 @@
...
+        filter.filter(
+          (existingOption) => existingOption.label !== changedOption.label,
+        ),
+      )
+    }
+
+    return setActivePinFilters((activePinFilters) => [
```
mariojsnunes, line 50: no need to return

src/pages/Maps/Maps.client.tsx
```diff
@@ -61,116 +34,32 @@ const MapsPage = observer(() => {
...
-  const onBlur = () => {
-    navigate('/map')
-    setSelectedPin(undefined)
-  }
-
-  const onPinClick = (pin) => {
-    navigate(`/map#${pin._id}`)
-    if (pin._id !== selectedPin?._id) {
-      selectPinByUserId(pin._id)
+    if (allPins) {
```
mariojsnunes, line 37: I think this would be more performant on `useMemo`. If I understand correctly, here we want to change pin filters based on the current pins. That means a `useMemo` with an `[allPins]` dependency, should be enough, instead of this effect + the allToggleFilters useState.

packages/components/src/CardList/CardList.tsx
```diff
@@ -1,60 +1,68 @@
...
-        <CardListItem
-          item={item}
-          key={item._id}
-          isSelectedPin={isSelectedPin}
-          onPinClick={isSelectedPin ? onBlur : onPinClick}
-          viewport={viewport}
-        />
+  useEffect(() => {
+    setRenderCount(ITEMS_PER_RENDER)
+  }, [list])
```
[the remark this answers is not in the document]
benfurber: Yes, it's important. The list can change a lot so it needs to reset the count back to 20.

src/pages/Maps/Content/MapView/MapContainer.tsx
benfurber: If you click around a lot though you'll get the opposite annoying problem and as the page title doesn't change it doesn't really create a navigable history. Happy to change if you feel strongly.

src/pages/Maps/Maps.client.tsx
benfurber: Will change for dev performance but for prod users, this _should_ only run once.

PR #4011, opened 2024-11-20
feat: update active pin interaction for new map

benfurber:
## PR Checklist

## PR Type

## What is the new behavior?

When clicking on a card on the left, the pin opens on the right. When you click on a profile link on an open pin, it opens the profile in a new tab.


packages/components/src/CardList/CardList.tsx
```diff
@@ -26,7 +37,17 @@ export const CardList = (props: IProps) => {
         Date.parse(b.creator?._lastActive || '0') -
         Date.parse(a.creator?._lastActive || '0'),
     )
-    .map((item) => <CardListItem item={item} key={item._id} />)
+    .map((item) => {
+      const isSelectedPin = item === selectedPin
```
mariojsnunes, line 41: would be better to compare by the id

src/pages/Maps/Maps.client.tsx
```diff
@@ -123,8 +117,15 @@ const MapsPage = observer(() => {
...
 
   const onBlur = () => {
-    setSelectedPin(null)
     navigate('/map')
+    setSelectedPin(undefined)
+  }
+
+  const onPinClick = (pin) => {
+    navigate(`/map#${pin._id}`)
+    if (pin !== selectedPin) {
```
mariojsnunes, line 126: compare by id

mariojsnunes (changes requested):

PR #4020, opened 2024-11-26
feat: Improve CreateComment and CreateReply components (ONEARMY#3656)…

### LINKED ISSUE #3656: [feature request] Improve CreateComment and CreateReply components
> This ticket is part of series of 7 tickets concluding a design task focusing on improving design and usability of comments feature. Here you can find the [Figma file](https://www.figma.com/board/GfQPMs7PyC7jbkymDTco2D/CP-%2F-Nested-Comments-Improvements?node-id=0-1&t=7sgMA4ZPYXAMb03A-1) for more context.

### **7/7 - Improve CreateComment and CreateReply components**
---

The component is currently taking a lot of vertical space, since the button is below. The button is also too big visually on mobile. The textarea is not increasing the height automatically, so the user has to drag the bottom-right corner to re-size it. Additionally, its default height is too big for the text size and is making the text not centered. 

image

[Figma link](https://www.figma.com/design/x4DcaUjwiigjCySAt5omxB/CP-%2F-Signup%2C-Settings-and-Gatekeeping?node-id=543-14229&t=8B7R1EvPTUd7KZSh-4) to the file where you can see the measurements. Click on the element and hold option mac key to see the measurements.

image

> Ticket scope to include (notes from Ben):
> 
> 1. Update button component to choose to hide text on mobile viewpoint. 
> 2. Changing our textarea component (and implement for all instances) to include the character limit count.
> - Questions - title, description
> - how-to - title - Short description
> - how-to steps - title, description
> - research - title, description
> - research update - title, description
> - settings - Tell us a bit about yourself, Short description of your pin
> - Error handling
> - Rules around colours to use? e.g. red for under/near limit?

I think there are simpler alternatives, just changing the flex-direction so that the textarea and the button are next to each other, but I think this component deserves some love and nice styling. <3

**Additional context**
See the milestone

johannes-ross:
… Create CSS & JS logic for size change

## PR Checklist

## PR Type

- [X] Feature (adds functionality)

Site just breaks or doesn't look pretty when you try to comment on a mobile device.
When you type, the textarea never changes size. You can change it's size manually.

## What is the new behavior?
Site now looks pretty when you try to comment on a mobile device or a different screensize.
When you type, the textarea now dynamicly  changes it's size.

- [X] No

## Git Issues

Closes #3656

### DIFF
--- a/packages/components/src/CreateComment/CreateComment.css
+++ b/packages/components/src/CreateComment/CreateComment.css
@@ -1,0 +1,44 @@
+/* https://css-tricks.com/the-cleanest-trick-for-autogrowing-textareas/ */
+.grow-wrap {
+    /* easy way to plop the elements on top of each other and have them both sized based on the tallest one's height */
+    display: grid;
+}
+.grow-wrap::after {
+    /* Note the weird space! Needed to preventy jumpy behavior */
+    content: attr(data-replicated-value) " ";
+    
+    /* This is how textarea text behaves */
+    white-space: pre-wrap;
+    
+    /* Hidden from view, clicks, and screen readers */
+    visibility: hidden;
+}
+.grow-wrap > textarea {
+    /* You could leave this, but after a user resizes, then it ruins the auto sizing */
+    resize: none;
+    
+    /* Firefox shows scrollbar on growth, you can hide like this. */
+    overflow: hidden;
+}
+.grow-wrap > textarea,
+.grow-wrap::after {
+    /* Identical styling required!! */
+    background: none;
+    resize: none;
+    padding: 15px;
+    word-wrap: anywhere;
+    border-color: transparent;
+    font-size: 12px;
+    
+    /* Place on top of each other */
+    grid-area: 1 / 1 / 2 / 2;
+}
+.grow-wrap > textarea:focus {
+    border-color: transparent;
+}
+
+.grow-wrap.value-set  > textarea,
+.grow-wrap.value-set::after {
+     /* Identical styling required!! */
+    padding-bottom: 27px !important;
+}
--- a/packages/components/src/CreateComment/CreateComment.test.tsx
+++ b/packages/components/src/CreateComment/CreateComment.test.tsx
@@ -37,7 +37,9 @@ describe('CreateComment Component', () => {
         onChange={mockOnChange}
       />,
     )
-    expect(screen.getByText('Login')).toBeInTheDocument()
+    expect(
+      screen.getByText('Hi there! Login to leave a comment'),
+    ).toBeInTheDocument()
     expect(
       screen.queryByPlaceholderText('Leave your questions or feedback...'),
     ).toBeNull()
--- a/packages/components/src/CreateComment/CreateComment.test.tsx
+++ b/packages/components/src/CreateComment/CreateComment.test.tsx
@@ -53,7 +53,7 @@ describe('CreateComment Component', () => {
         onChange={mockOnChange}
       />,
     )
-    expect(screen.getByText('Leave a comment')).not.toBeDisabled()
+    expect(screen.getByTestId('send-comment-button')).not.toBeDisabled()
   })
 
   it('disables submit button when no comment is entered', () => {
@@ -66,7 +66,7 @@ describe('CreateComment Component', () => {
         onChange={mockOnChange}
       />,
     )
-    expect(screen.getByText('Leave a comment')).toBeDisabled()
+    expect(screen.getByTestId('send-comment-button')).toBeDisabled()
   })
 
   it('handles user input in textarea', () => {
@@ -96,7 +96,7 @@ describe('CreateComment Component', () => {
         onChange={mockOnChange}
       />,
     )
-    const button = screen.getByText('Leave a comment')
+    const button = screen.getByTestId('send-comment-button')
     fireEvent.click(button)
     expect(mockOnSubmit).toHaveBeenCalledWith('Test comment')
   })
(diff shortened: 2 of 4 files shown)


johannes-ross:
I would like to add, that I haven't added new Unit Tests, since this is also just a visual change.

mariojsnunes:
Thanks @johannes-ross!
My first impression is that it would be a bit hard to maintain. Could it be done with css only?

johannes-ross:
I agree and also think a bit hard is understated. The values I used for the breakpoints are all coming from testing. I don't even understand why they work. Definition of magic numbers.

I've tried to use a CSS only approach. I used the following sources and variations and combinations of it:
- https://css-tricks.com/auto-growing-inputs-textareas/
- https://css-tricks.com/the-cleanest-trick-for-autogrowing-textareas/
- https://davidwalsh.name/autogrow-textareas
- https://til.simonwillison.net/css/resizing-textarea

The Problem that I always encountered was flickering when getting near the line end. Here is what I've found:
- The hoped for behaviour is, when you type the first line, we want the character count to be inside the element, but when you reach it, it should snap to the bottom and let you finish that first line. All other lines should be within the new, bigger Line. That caused flickering when using CSS, which I've tried to get rid of, but I never managed to find a solution, except ditching all CSS.
- When you press enter there should be a new line. Not a problem, can easily be done with CSS. However, it is a problem when trying to combine CSS and JS, because JS needs the values from CSS to calculate the new values. (It's a bit more complicated then that)
- When you use CSS you cannot widen the input dynamically, because you are trying to use the word-break as an indicator, that you need to add a new line, which isn't a problem except for that very first line.
- Since we always want to have the same look, another problem was detecting, when we only have one line left, when previously having multiple, since having multiple lines has a bigger bottom padding then the single line. So You need to know, when you reach a character width, that will cause the character counter to go down **and** increase the bottom padding **and** know when you've crossed that line again, so you can reduce the bottom padding and snap back into the single line.
- You also can't just set a character count limit and then break because 5 Ws are way bigger than five is: 
"WWWWW"
"iiiii"

So using pure CSS doesn't work, because we have to change the width of the text area when we reach the first character count.
Using a combination of CSS and JavaScript doesn't work, because it's just to complex.
Which only leaves the approach of using JS to fix it.

A simple Idea to get rid of all the complex JS and CSS combination would be to just get rid of this:
image

and always have the character count on the bottom line, like in the second one.

I've just noticed, I missed that when the Input is empty, the count shouldn't be displayed. I have to fix that, but I'm going to wait until we've discussed this and found a useful solution.

Our greatest hope is, that the future will fix it:
- https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing

dalibormrska:
Wow, such a cool contribution @johannes-ross ! Thank you for all the effort, I see now the problem of complexity of that trick and maintenance which I didn't foresee from the perspective of a designer. 

I agree that we can simplify a bit. The point with the moving character count was so that the component has only the necessary height in the beginning when it's not active, in case that the user sees several of them at once, so that they don't occupy a big portion of the screen.

This point could be achieved also simply like this:
image

The char count is not shown when the textarea is not active, but once the user clicks into the text area (or types the first character, or whatever makes sense to code) the char count shows below. 

What do you think? Would that help?
Cheers!

dalibormrska:
Oh and one more thing... Would it be possible to set the font-size of the text-area to be equal to the font-size of the comment? Side note: In the near future it might happen that for mobile, the comments' font size might be lower (14px) than on desktop. But there is no decision yet, so I don't know if it's good to hard-code it.

The current difference between font-sizes:
image

benfurber (changes requested):
Thanks so much @johannes-ross for such a thorough dive into this! 

So yes, definitely agree with @mariojsnunes about the maintenance issue.

How's the below for a tighter technical scope?
1. Go as far as can be achieved with css (i.e. if functionality isn't possible without custom js, that it's not possible.)
2. There's a readability challenge due to how much styling has to be changed between mobile/desktop so I think it's OK to use the responsive display none appoach at a slightly higher level to have seperate components for mobile/desktop like how it's done on `src/pages/Maps/Content/MapView/MapWithList.tsx`. Then 
3. As it's available on chrome, I'm very happy for field-sizing to be tried (with a code comment). 

A more detailed point...
* I think the use of `Grid` can be avoided using 2 above. Then for desktop it's just the badge and button using the space they need and the textbox flexing in the middle and for mobile the icon (which should be done using oa-components' `Button` with the `showIconOnly` prop

It's funny how much back and forth can happen on this type of relatively small issue so if there's still changes needed after your next draft maybe we can pair to finish it off?

johannes-ross:
I've tried to use as much CSS as possible. The only thing I've used JS for is the focus.

I can remove it, because it adds 2 new Renders, but I don't think it's actually that significant.

benfurber:
@mariojsnunes Looks like all the preview branches are failing as they need `SUPABASE_PROJECT_ID`

benfurber:
Thanks @johannes-ross! Rather than ask you to do a bunch of minior clean-up that's more subjective so I thought I'd just do it. Merging in now.

dalibormrska:
Heyy, sorry for the bump again. I just noticed that the change is implemented across HowTo and Research modules, but not in the Questions module. Could we do a quick fix and implement it there as well? 

Thankssss, great job @johannes-ross once again! Works like a charm. :))

johannes-ross:
Hey, I've looked at the Question Code and I can't quickly find a reason, why it would behave the way you described. @dalibormrska I think we should open a new Ticket for further investigation.

mariojsnunes:
> Heyy, sorry for the bump again. I just noticed that the change is implemented across HowTo and Research modules, but not in the Questions module. Could we do a quick fix and implement it there as well?
> 
> Thankssss, great job @johannes-ross once again! Works like a charm. :))

I'll take care of that

Issue #3989, opened 2024-11-06
[feature request] Change name 'how-to' to 'library'


davehakkens:
**Is your feature request related to a problem? Please describe.**
Throughout the years how-to's have evolved into much more then guides as it also shares 3D models and files to download.
it's finally time for the change

**Describe the solution you'd like**
We want to change the name from 'how-to' 'to Library" so it represents better what it is. 
And we call items inside the library 'projects'

**Describe alternatives you've considered**
More about it in this discussion https://github.com/ONEARMY/community-platform/discussions/3241

# **Changes we need:**
### **Navbar**
<img width="360" alt="Screenshot 2024-08-26 at 09 56 42" src="https://github.com/user-attachments/assets/f8764844-9a5f-463f-8fb7-53088038a225">

### **Create button**
<img width="359" alt="Screenshot 2024-08-26 at 09 56 46" src="https://github.com/user-attachments/assets/f1e4b5c2-269a-4e5e-8b16-a2cd9955d7dd">

### **Search bar**
<img width="345" alt="Screenshot 2024-08-26 at 09 56 48" src="https://github.com/user-attachments/assets/65e439d0-913c-422e-8b1e-466f9b093465">

### **Footer**
Contribute to the Precious Plastic Library, Share your project
<img width="1076" alt="image" src="https://github.com/user-attachments/assets/71c0d169-f6d6-488e-b2be-55c34b208282">

### **Create a howto**
- Add a project
- Title of your project
- 4 . Create your project
https://community.preciousplastic.com/how-to/create
<img width="1091" alt="image" src="https://github.com/user-attachments/assets/d128904c-b9cf-4330-8cb2-e75dc8d93b72">

### **Profiles**
- Projects in Library
- Library
<img width="1088" alt="image" src="https://github.com/user-attachments/assets/a57cb9f9-71cd-4526-a268-4340df49da9d">
<img width="856" alt="image" src="https://github.com/user-attachments/assets/c5e2e874-618a-42cd-994a-3307ca3cba4c">

### ⚠️ Important
Make sure that all existing urls continue to work and are redirected accordingly

benfurber:
I thinking about initially doing just routing and user label changes. Then the DB migration later on can take care of renaming everything in the system. What do you think @mariojsnunes?

benfurber:
@davehakkens The academy guides will need updating too.

benfurber:
[...]

PR #4027, opened 2024-12-06
feat: change labels from howtos to projects

benfurber:
## PR Checklist

## PR Type

User things are named 'How to', 'how-tos', etc.

## What is the new behavior?

Top level is called 'Library', children are called 'projects'. This change is for labels and urls only. Developer things are still called howtos until we migrate the DB.

- [ ] Yes
- [x] No

## Git Issues

Part of #3989


src/routes/_.how-to.$slug._index.tsx
```diff
@@ -1,50 +1,16 @@
...
-  if (!howto) {
-    return []
-  }
-
-  const title = `${howto.title} - How-to - ${import.meta.env.VITE_SITE_NAME}`
-
-  return generateTags(title, howto.description, howto.cover_image?.downloadUrl)
-})
-
 export default function Index() {
```
mariojsnunes, line 14: I think this can be removed

davehakkens:
thought I left a comment before but dont see it anymore.
Question: Is there a conscious reason on why you choose different wording from the issue?

benfurber:
> thought I left a comment before but dont see it anymore. Question: Is there a conscious reason on why you choose different wording from the issue?

Yes. Definitely went with what seemed most natural to me personally. The principle I tried to go with wast a `project` is the main unit, the main 'thing' that should be labelled and so only went with library when talking about all the projects collectively. But very happy to debate the grammar.

benfurber:
@davehakkens See your point. I'll update. Applying your approach but with some suggestions: 

Create button
"Add your how-to" -> "Add to the library" or "Add your project to the library"

Create page
"Create a how-to" -> "Add to the library" or "Add your project to the library"

Profiles/Contribution tab: 
"How-to's" -> "Library Projects"


Thoughts?

davehakkens:
Tried that before but adding 'project' to the wording often makes it to long for the layout making it look strange

benfurber:
@davehakkens Have a look [at this now](https://github.com/ONEARMY/community-platform/pull/4027/commits/dd5d3d956aec497e6f3c3472459931a7a314d1aa).

davehakkens:
The url doest go anywhere. And the preview doesnt show any changes at first glance, not sure if i'm seeing the latest?
image

benfurber:
@davehakkens Actually freshly deployed now: https://community-platform-pr-4027.fly.dev/library?sort=Newest

davehakkens:
I see the same as the screenshot above. Is that the latest?

benfurber:
@davehakkens If you refresh and everything you don't see this? 
Screenshot 2024-12-13 at 10 06 15

davehakkens:
Nice works for me, two changes that would be nice:

1. Change text in searchbar
image

2. Change text in the header
image

benfurber:
One done. The second is an environement variable so I'll change the default option but the actual prod variables will need changing when this is deployed.

PR #4091, opened 2025-01-20
fix: typing and fields for categories

benfurber:
## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## What is the new behavior?

Adds a missing field for categories and updates the typing so it's now shared.

- [x] Yes
- [ ] No


supabase/migrations/20250120115148_add_deleted_to_categories_table.sql
```diff
@@ -0,0 +1,2 @@
+alter table
+if exists public.categories add deleted boolean default false;
```
mariojsnunes, line 2: do we really need a deleted field? what is the use case? asking because it adds complexity...
currently, we have a cascade which sets the category to null (on the questions table) whenever it is deleted

benfurber: I guess I was defaulting to it being a content driven table and thus should have a publishing status. But I'm fine to remove.

src/routes/api.categories.$type.ts
```diff
@@ -1,23 +1,49 @@
...
       { headers, status: 400, statusText: 'type is required' },
     )
   }
 
-  const categoriesResult = await client
+  const cachedCategories = await cache.get('categories')
+
+  if (cachedCategories && isProductionEnvironment()) {
+    return Response.json(
+      { categories: filterByType(cachedCategories, type) },
```
mariojsnunes, line 32: this won't work as below it is filtering by type.
I would be fine with either removing the cache or query all

mariojsnunes (approved):

PR #4119, opened 2025-02-11
feat: pages with comments more mobile friendly

benfurber:
## PR Checklist

## PR Type

### Questions
Screenshot 2025-02-11 at 13 29 23

### Library
Screenshot 2025-02-11 at 13 27 14
Screenshot 2025-02-11 at 13 27 06

### Research
Screenshot 2025-02-11 at 13 28 49
Screenshot 2025-02-11 at 13 29 02

### DIFF
--- a/packages/components/src/Breadcrumbs/BreadcrumbsItem.tsx
+++ b/packages/components/src/Breadcrumbs/BreadcrumbsItem.tsx
@@ -41,7 +41,7 @@ export const BreadcrumbItem = ({ text, link, isLast }: BreadcrumbItemProps) => (
         sx={{
           display: 'block',
           color: 'black',
-          fontSize: 15,
+          fontSize: [2, 3],
           textOverflow: 'ellipsis',
           whiteSpace: 'nowrap',
           overflow: 'hidden',
--- a/packages/components/src/CommentBody/CommentBody.tsx
+++ b/packages/components/src/CommentBody/CommentBody.tsx
@@ -1,0 +1,62 @@
+import { createRef, useEffect, useState } from 'react'
+import { Text } from 'theme-ui'
+
+import { LinkifyText } from '../LinkifyText/LinkifyText'
+
+interface IProps {
+  body: string
+}
+const SHORT_COMMENT = 129
+
+export const CommentBody = ({ body }: IProps) => {
+  const textRef = createRef<any>()
+
+  const [textHeight, setTextHeight] = useState(0)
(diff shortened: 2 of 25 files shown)


mariojsnunes (approved):

benfurber:
@dalibormrska @davehakkens I've added a preview branch, worth having a play.

davehakkens:
This seems like a PR part of a larger process of optimising the Discussion UI. It's hard to understand what I would specifically have to check for this one. But I believe me and Dalibor are aligned on what the endresult should look like. 
Could I leave it up to you @dalibormrska to test, give feedback, create issues until its how we want it?

dalibormrska:
Thanks a lot! I played around with the preview branch, tested on mobile and looked through the screenshots. I will only address the spacings, since that is the point of this PR.

Reflecting back to the problem / goal with this change - Text and elements were squeezing on mobile, due to width. So the goal was to give a bit more space to the comments on mobile, but let's elaborate on that saying, to all the elements that are squeezed on mobile. Which then applies also to the original Library/Research/Question post or the list of Research/Question items that could benefit from some more pixels of width.

I noticed that in here, the paddings are also deleted. That does not work well with the filters and library items.
image

If we want to apply this problem statement logic to the Research and Question items (I would say Library items don't need this change, there information in the cards are not squeezed), I would propose it like this (It might become a bit tricky, since the sliding icon filter should not have the padding, but the heading and the sort/filter/search should). I would propose keeping the 10px paddings, as it is the default now:
image

Notice that in my designs I try to utilise a simple rule of 15px spacing from the side. I would apply these to the single item pages (specific L/R/Q item).
image

My argument for 15px here is that there would be a lot of text too close to the edge, which is problematic aesthetically, but also due to readability on phones with cracked screens for example.

In the Research, as seen on your screenshot as well, there is some content with 10px and the text with 20px padding, which creates a bit of inconsistency.

Phone with cracked protective glass:
image

So to conclude:
- In the Library main page, keep the layout as it was
- In the R/Q main pages, we CAN utilise the changes as proposed above
- In the single-item pages, utilise 15px side padding instead of 10px, also on the research page

benfurber:
Further changes:
Screenshot 2025-02-17 at 10 06 07
Screenshot 2025-02-17 at 10 09 05
Screenshot 2025-02-17 at 10 05 42

dalibormrska:
@benfurber 

A small detail that I found could be fixed. The heading lacks the 10px side margin.

Quiestions h1 margin

PR #4094, opened 2025-01-22
refactor: comment display to the component library

benfurber:
## PR Checklist

- [x] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)
- [x] - Unit and e2e tests for the changes that have been added (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [x] Refactoring (no functional changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes
- [ ] Other... Please describe:

## What is the new behavior?

1. Move a couple of UI focused components to the component library. (For easier contributor tickets about to be put up as issues)
2. Rename V2 appended component names to Supabase appended component names.
3. Move/tidy-up the new comment typing into shared.

## Does this PR introduce a DB Schema Change or Migration?

- [ ] Yes
- [x] No

PR #4124, opened 2025-02-16
feat: add new author badge

### LINKED ISSUE #4096: [ui] Prototype adding an 'author' effect on comments
### Description

On commenting, it can be super useful to know you're responding to the original author of the content. We have an idea for an effect to give comments from the original author and would like to see it in action.

### Page related

As we're changing DBs at the moment, there's two sets of commenting components, only update `CommentAvatar` and `CommentDisplay` which is used on questions as the old components won't be around for long.

### Mockup

A simple SVG, that you can find below in the comments should be positioned with the profile picture approximately like this.

Image

There is 30x30px avatar and 50x50px avatar used in the comments based on the mobile/desktop. So make sure that the positioning of this badge works with both.

Image

### Build suggestion

Update `CommentAvatar` with an optional prop via `CommentDisplay`

joseh29:
## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [X] Feature (adds functionality)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

image

## What is the new behavior?

Author of the question, will have a badge representing he his the author, both in the comment itself and on replies.
image

- [ ] Yes
- [X] No

## Git Issues

Closes #4096

## What happens next?


packages/components/src/CommentDisplay/CommentDisplay.tsx
```diff
@@ -69,10 +72,22 @@ export const CommentDisplay = (props: IProps) => {
...
+            display: 'inline-block',
+          }}
+        >
           <CommentAvatar
             name={comment.createdBy?.name}
             photoUrl={comment.createdBy?.photoUrl}
+            isAuthor={
+              comment.createdBy?.firebaseAuthId ===
+              userStore.activeUser?._authID
+            }
```
joseh29, line 90: I highly doubt this is the way to go when comparing the author of the comment with the current user. But I gave it my best :laughing:

joseh29 (commented):
.

packages/components/src/CommentDisplay/CommentDisplay.tsx
```diff
@@ -69,10 +69,19 @@ export const CommentDisplay = (props: IProps) => {
...
+          sx={{
+            flexDirection: 'column',
+            position: 'relative',
+            display: 'inline-block',
+          }}
+        >
           <CommentAvatar
             name={comment.createdBy?.name}
             photoUrl={comment.createdBy?.photoUrl}
+            isAuthor={true}
```
joseh29, line 84: I'm not sure how to get the author of the question. I didn't have time to explore.
After implementing that, I can work on the tests.

benfurber:
Thanks for openning @joseh29.

As I'm sure you think as well, the badge is too big right now. Can you play around with making it as small as is useful?

benfurber:
For anyone wanting a quick look, this discussion thread show it in action: https://community-platform-pr-4124.fly.dev/questions/can-i-ask-a-question-through-the-new-auth-studd-ok

packages/components/src/CommentDisplay/CommentDisplay.tsx
joseh29: @benfurber do you know top-of-your-head how I can check the author of the question?

joseh29:
> nice @joseh29
> But seems like everyone gets the author badge. not just the author?
> 
>   image

Yes, there is a comment above about that 👆

packages/components/src/CommentDisplay/CommentDisplay.tsx
joseh29: I've done some digging and we have a reference to the question by using `comment.sourceId`. Now I just need to figure out how we can retrieve the author, by querying the question by id.

benfurber: @joseh29 Easiest is probably to go higher. If you look at the models for research (update), library and question you'll find a value related to author for the parent content, i.e. the three content types which have comments. Probably via `DiscussionWrapper` for research/library and `CommentSectionSupabase` for questions.

mariojsnunes: I'm make this change only for questions, the other components will be removed soon

joseh29: I would like to avoid prop drilling. I think we should be able to fetch the question by id through a global state management store. I'm not sure how the state is managed globally in this project, as I haven't had time to read about remix. But seeing other components I believe it is possible?

joseh29: > I'm make this change only for questions, the other components will be removed soon

What do you mean by this? 🤔

mariojsnunes: Currently question comments use a different set of components `CommentSectionSupabase` as it was migrated to supabase.
Research and Library are still on firebase and using the old components `DiscussionWrapper `. We are migrating them soon.
So all these comment UI/UX improvements can temporarily be done only for new Discussion components used in the questions module (CommentSectionSupabase).

shared/providers/authorsContext.ts
```diff
@@ -0,0 +1,9 @@
+import { createContext } from 'react'
+
+interface AuthorsContextType {
+  authors: Array<string>
```
joseh29, line 4: This probably should be an actual Actor class, that should be extended for all types of Author's:
`QuestionAuthor`, etc

For now it's only an array of author id's.

joseh29: Also, I don't think we'll need a setter method, but let me know your thoughts.

shared/providers/index.ts
```diff
@@ -0,0 +1 @@
+export * from './authorsContext'
```
joseh29, line 1: Followed camelCase convention here.

joseh29 (commented):
Ok guys, I've left a POC using a ContextProvider to get the author in any comment component. Let me know what you think :muscle:

packages/components/src/CommentDisplay/CommentDisplay.tsx
joseh29: As discussed below, I've added a context provider to pass down the authors, instead of using prop drilling.
Resolving this...

joseh29:
What do you think about this @benfurber:
Large viewport
image

Small viewport
image

@davehakkens I've now added the check to see if any of the "authors" of the question (future-proof :crossed_fingers:) is the author of the comment. It should now work as expected (check screenshots).

benfurber:
@mariojsnunes How do you feel about the AuthorsContext pattern? It's a big change from how most data is moved around the app.

mariojsnunes:
> * Move the context to the common folder in the app (component library should just be for UI components)

this would be ideal, but it needs to be imported in the /components lib.
I don't see a better way right now, besides moving all the components used by SupabaseCommentSection to the /src. (undo what you did here #4094)

dalibormrska:
Awesome work, thanks for the implementation and for the preview link!

2 small details...

On the preview link, I cannot really select the text of the comment, or click the reply button, since the image is in the way and is draggable. I would imagine the css property `pointer-events:none` would solve that.


Additionallly, I played around with the values for the margins, so that the baloon fits better. I propose this:

```
margin-left: -8px;
margin-top: -46px;
```

joseh29:
> Update the e2e tests to expect to find the badge.

Would love to, but everytime I run e2e tests my sad 8gb ram pc crashes. 😅

joseh29:
@dalibormrska the preview should not be updated yet, because the build is failing.
Also, I've changed the sizing and added a `pointer-events: none' already, so the issue you are describing should not be happening anymore.

benfurber:
Nice one, thanks @joseh29

PR #4105, opened 2025-01-23
verticalList || ImageGallery buttons issue #4099 solved

sanwalsulehri:
## PR Checklist

- [ ] - Commit [messages are descriptive](https://github.com/ONEARMY/community-platform/blob/master/CONTRIBUTING.md#--commit-style-guide), it will be used in our [Release Notes](https://github.com/ONEARMY/community-platform/releases/)
- [x] - Unit and e2e tests for the changes that have been added (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix (fixes an issue)
- [x] Feature (adds functionality)
- [ ] Code style update
- [ ] Refactoring (no functional changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes
- [ ] Other... Please describe:

## What is the new behavior?

_Describe the new behaviour_
_If useful, provide screenshot or capture to highlight main changes_

## Does this PR introduce a DB Schema Change or Migration?

- [ ] Yes
- [x] No

## Git Issues

Closes #4099

## What happens next?

Thanks for the contribution! We try to make sure all PRs are reviewed ahead of our monthly maintainers call (first Monday of the month)

If the PR is working as intended it'll be merged and included in the next platform release, if not changes will be requested and re-reviewed once updated.

If you need more immediate feedback you can try reaching out on Discord in the [Community Platform `development` channel](https://discord.com/channels/586676777334865928/938781727017558018).

PR #4125, opened 2025-02-17
Arrow styling

sanwalsulehri:
## PR Checklist
Applied arrow styling in verticallist.tsx and in arrrows.tsx


benfurber:
Sorry for the delay getting back to you @sanwalsulehri. I'm afraid that even though we discussed it lots on #4105, this is still far away from the approach I'm looking for. I'm inclined to close this PR and #4105. Thanks for trying.

Issue #4109, opened 2025-01-27
[backend] [data] Data seed for questions


benfurber:
We've already moved questions over to supabase, but we don't have a DB seed for local it yet (and maybe the e2e tests). Please write one.

Ideally uses realistic copy/data, a range of lengths (min, max, etc.) And for the questions adding comments, question categories and users (profiles).

## Suggested steps:

1. Follow the [supabase guide to generate a seed](https://supabase.com/docs/guides/local-development/seeding-your-database).
2. Edit it to be meaningful for us.
3. See what can/should be dropped now that the seed exists.

Avinashguthula:
👀 I’m interested in ...

joseh29:
As discussed in discord, I would like to try and tackle this one.
I've already opened a PR, so you guys can track how's it going :)

joseh29:
As discussed IRL with @mariojsnunes, I will no longer be contributing to this project.
I've left some of what I think it's a solid base and I believe the seeding is at 70%'ish.
Left a PR with some work already done.
Best regards and good luck 💪 It was a pleasure 🙏

PR #4137, opened 2025-02-25
chore: add supabase seed for questions, profiles, tags and categories

joseh29:
## PR Type

Nothing to seed the supabase db.

## What is the new behavior?

Base content can be seeded: 
- Questions
- Users (auth level)
- Profiles (though they're not displaying in the UI as that's still from firebase)
- Categories
- A tag

## Git Issues

Part of #4109


joseh29:
I've decided to move questions to a separate json fle to be easier to track who made the question and what are its comments. Now I'm struggling a little bit with comments be associated with both questions and profiles at the same time, but I'm working on it :muscle: I haven't had much time though.

joseh29:
OAN, I've ran snaplet seeding without `dryRun` to see what the db would look like and it looks promising! At least we already have profiles, questions and some other tables. Working on comments now.

benfurber:
@joseh29 Thank you so much for your work on this and sorry to hear you won't be contributing any more.

PR #4138, opened 2025-02-26
[auth] New account attempt for existing email address

benfurber:
## PR Checklist

## PR Type

For security reasons, the authetication service won't error if an email address already exists on the system.

## What is the new behavior?

We explain this to users.

Screenshot 2025-02-26 at 09 41 55

### DIFF
--- a/packages/themes/src/common/commonStyles.ts
+++ b/packages/themes/src/common/commonStyles.ts
@@ -32,6 +32,7 @@ export const commonStyles = {
     subscribed: 'orange',
     notSubscribed: '#1b1b1b',
     betaGreen: '#98cc98',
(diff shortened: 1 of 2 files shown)


benfurber:
Happy with this approach @davehakkens. It's essentially a UX/security trade-off.

davehakkens:
I would not phrase it like this as they dont aim to join OA. They just want to login. 

More
Something along the lines of:

we're doing a big update to this software to improve the security of the platform. We need to confirm your email address to make sure its the right one

davehakkens:
Also the bottom part is very confusing..

davehakkens:
> 
> 
> How's about this?
> 


Not sure if im following. This is to fix the problem above? ☝️

PR #4153, opened 2025-03-10
feat: migrate patreon to supabase

mariojsnunes:
## PR Checklist

## PR Type

## How does it work?

- It now saves patreon data on supabase in addition to firebase.
  - Still saves in firebase due to map pins.
- Migrated firebase function code to a remix api route and removed the firebase function.
  - Disconnect is also executed server-side in a remix api route.
  - Both endpoints require a logged in user.
- The Profile page PatreonIntegration component gets the patreon data via an API call to /api/patreon (supabase) instead of relying on the firebase profile data.

- [x] Yes

## Git Issues

Closes #


benfurber (approved):
Nice. LGTM!

Not sure patreon needs it's own page - could just redirect to the existing settings page with the response, but I don't think it's important enough to change as this has been implemented.

benfurber:
@mariojsnunes Assume you'll merge in as data needs migrating?

src/models/profile.model.ts
```diff
@@ -3,11 +3,71 @@ export class DBProfile {
   username: string
   display_name: string
   is_verified: boolean
+  is_supporter: boolean
```
benfurber, line 6: I think this should match the current is support model which is `user.badges.isSupporter`

benfurber:
@mariojsnunes Changed my mind! I think the data structure change is important. Left a comment. I think the badges structure is important as the top level of user doesn't get too blutted - been trying to move stuff away from that top level (e.g. all the custom profile values).

src/models/profile.model.ts
mariojsnunes: I'd rather not store is_verified and is_supporter in a json column, it will be harder to query.
maybe we could have a badges table? not sure it's worth the work now for just 2 "badges"

benfurber: Feature creep.

I don't see the query argument. These are fields we always want as a group (e.g. there's no time we want to know if a user is verified but not a supporter) and having them as a group makes extending much easier in the future.

Reminder. The argument I'm making is exactly the one you used earlier in the project when we were doing top level schemas - json is fine for stuff.

mariojsnunes: I'm not completely against storing as json, but to explain the query argument...
Let's say we want to have a page to highlight patreon supporters.
`select * from profiles where is_supporter = 1` is much faster than `select * from profiles where badges->is_supporter = 1`. As the latter cannot have an index and every row in the table needs to be parsed.
Even for internal analytics, if you want to know which/how many users are verified/supporters.

Also, my earlier argument was about storing things like website links or other profile query that doesn't need to be queried.

Currently AFAIK these badges don't need to be queried either. But I see some use cases for it.

benfurber: Had a LONG debate elsewhere, Mario convinced me.

PR #4155, opened 2025-03-10
fix: flex for library steps

benfurber:
## PR Checklist

## PR Type

- [x] Bugfix (fixes an issue)

Screenshot 2025-03-10 at 12 05 30

## What is the new behavior?

Screenshot 2025-03-10 at 13 03 52

### DIFF
--- a/src/pages/Library/Content/Page/LibraryStep.tsx
+++ b/src/pages/Library/Content/Page/LibraryStep.tsx
@@ -38,12 +38,12 @@ const Step = (props: IProps) => {
 
       <Flex
         sx={{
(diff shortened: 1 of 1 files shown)


davehakkens:
@benfurber not sure if it happend after this PR.
But some steps are weird
https://community.preciousplastic.com/library/shredder-pro-by-weima-10
image

benfurber:
@davehakkens Oh wait. The fix wasn't deployed. That was a waste of half an hour...

Deploying now. I've confirmed locally that the fix works for the above step.

PR #4156, opened 2025-03-10
feat: messaging for all users unless config set

benfurber:
## PR Checklist

## PR Type

Only spaces can be messaged

## What is the new behavior?

All users can be messaged unless `VITE_NO_MESSAGING` is set to true.


.circleci/config.yml
```diff
@@ -361,6 +361,7 @@ jobs:
               --build-secret VITE_HOWTOS_HEADING="$VITE_HOWTOS_HEADING" \
               --build-secret VITE_COMMUNITY_PROGRAM_URL="$VITE_COMMUNITY_PROGRAM_URL" \
               --build-secret VITE_QUESTIONS_GUIDELINES_URL="$VITE_QUESTIONS_GUIDELINES_URL" \
+              --build-secret VITE_NO_MESSAGING="$VITE_NO_MESSAGING" \
```
mariojsnunes, line 364: I believe you need to add the secret right below as well

src/config/config.ts
```diff
@@ -183,3 +183,5 @@ export const isPreciousPlastic = (): boolean => {
 export const MAP_PROFILE_TYPE_HIDDEN_BY_DEFAULT = isPreciousPlastic()
   ? 'member'
   : undefined
+
+export const NO_MESSAGING = getFromLocalStorage('VITE_NO_MESSAGING')
```
mariojsnunes, line 187: why get it from localstorage? it should be available through import.meta.env or process.env

.circleci/config.yml
benfurber: Do you mean for supabase? Supabase doesn't need it.

src/utils/helpers.ts
```diff
@@ -190,11 +190,19 @@ export const isUserBlockedFromMessaging = (user: IUser | null | undefined) => {
   return user.isBlockedFromMessaging
 }
 
+export const isMessagingBlocked = () => {
+  return NO_MESSAGING === 'true' ? true : false
```
mariojsnunes, line 194: just `return NO_MESSAGING === 'true'`

src/config/config.ts
mariojsnunes: hmm still think having that localstorage just because of cypress is not great. is it possible to stub the `isMessagingBlocked` function instead?

benfurber: > hmm still think having that localstorage just because of cypress is not great.

Why?

I don't really see the problem, in fact I like it. Just a form of dependency injection. Makes the tests (as a form of documentation) super clear about what needs to change and easy to change and has enabled me to increase coverage.

benfurber: Update: Discussed and working on a new approach. But merging this in for now.

PR #4175, opened 2025-03-25
feat: add news

benfurber:
## PR Checklist

## PR Type

- [x] Bugfix (fixes an issue)

No news!

## What is the new behavior?

All the news.

Screenshot 2025-04-02 at 17 27 32
Screenshot 2025-04-02 at 17 25 32
Screenshot 2025-04-02 at 17 24 33

- [x] Yes


benfurber:
In addition to a markdown field, other key things left are continuing to reduce duplication, unit testing and e2e testing. Then the final UI once everyone is happy with it.

davehakkens:
cool nice quick setup! curious to test the rich editor/markdown

shared/models/document.ts
```diff
@@ -1,17 +1,142 @@
-export abstract class Doc {
+import type { Author, DBAuthor } from './author'
+import type { ContentType } from './common'
+
+// Base level for all content on supabase
+export interface DBDocSB {
```
mariojsnunes, line 5: does it need to have the SB sufix?

src/pages/News/Content/Common/FormFields/NewsBody.field.tsx
```diff
@@ -0,0 +1,24 @@
+import { Field } from 'react-final-form'
```
mariojsnunes, line 1: I know we use this `*.field.tsx` in multiple places, but I don't think it's a common convention, I'd just use `NewsBodyField.tsx` instead.

src/pages/News/Content/Common/NewsForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+            guidelines={<NewsPostingGuidelines />}
+            handleSubmit={handleSubmit}
+            heading={LABELS.headings[parentType]}
+            icon={IconHeaderHowto}
+            parentType={parentType}
+            saveError={saveError}
+            submitting={submitting}
+            valid={valid}
+          >
+            <>
```
mariojsnunes, line 114: unnecessary fragment

shared/models/document.ts
benfurber: For the moment yes. `DBDoc` is used for firebase typing.

src/pages/News/NewsEdit.tsx
```diff
@@ -0,0 +1,31 @@
...
+import type { News } from 'oa-shared'
+
+interface IProps {
+  news: News
+}
+
+export const NewsEdit = ({ news }: IProps) => {
+  const { slug } = useParams()
+  const navigate = useNavigate()
+  const { userStore } = useCommonStores().stores
```
mariojsnunes, line 16: thinking it would be worth to use the user store directly instead of `useCommonStores()`

benfurber: Example please? `const { userStore } = useCommonStores().stores` is what I've always done.

src/pages/common/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -15,43 +16,62 @@ interface BreadcrumbsProps {
...
+        })
+      }
+
+      steps.push({ text: project.title })
+      break
+    case 'news':
+      const news = content as News
+      steps.push({ text: 'News', link: '/news' })
+
+      // if (news.category) {
```
mariojsnunes, line 64: can remove?

src/pages/common/FormFields/Tags.field.tsx
```diff
@@ -1,20 +1,23 @@
...
   return (
-    <FormFieldWrapper htmlFor={name} text={fields.tags.title}>
+    <FormFieldWrapper htmlFor={name} text={title}>
       <Field
         data-cy={`field-${name}`}
         name={name}
         id={name}
         component={TagsSelectFieldV2}
-        category="question"
+        category="news"
```
mariojsnunes, line 20: should it be "news"?

src/routes/_.news.$slug.edit.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  return Response.json({ news }, { headers })
+}
+
+export default function Index() {
+  const data = useLoaderData<typeof loader>()
+  const news = data.news as News
+
+  return (
+    <AuthRoute>
```
mariojsnunes, line 41: No need to use `<AuthRoute>`, validate the user auth server-side.

src/routes/_.news.create.tsx
```diff
@@ -0,0 +1,37 @@
+import { UserAction } from 'src/common/UserAction'
+import { AuthRoute } from 'src/pages/common/AuthRoute'
+import { listing } from 'src/pages/News/labels'
+import { NewsCreate } from 'src/pages/News/NewsCreate'
+import { Box } from 'theme-ui'
+
+export async function clientLoader() {
```
mariojsnunes, line 7: use loader instead and validate user auth.
could redirect to an "incomplete profile" route if we want to keep that behavior.

src/routes/_.news.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const env = useContext(EnvironmentContext)
+  const rootStore = useCommonStores()
+
+  if (!isModuleSupported(env?.VITE_SUPPORTED_MODULES || '', MODULE.NEWS)) {
+    return null
+  }
+
+  return (
+    <Main style={{ flex: 1 }}>
+      <DiscussionStoreContext.Provider value={new DiscussionStore(rootStore)}>
```
mariojsnunes, line 27: we no longer use the discussion store with supabase, so `DiscussionStoreContext` is unnecessary.

mariojsnunes: seems like it can be removed from questions too

src/routes/_.questions.$slug._index.tsx
```diff
@@ -23,40 +25,22 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
-  if (tagIds?.length > 0) {
-    const tagsResult = await client
-      .from('tags')
-      .select('id,name')
-      .in('id', tagIds)
-
-    if (tagsResult.data) {
-      tags = tagsResult.data.map((x) => Tag.fromDB(x))
-    }
+    await incrementViewCount(
```
mariojsnunes, line 28: could you test with a `.then()` instead of await?
the idea is to not block the page.
but I know it won't write to the database without `await`. maybe the `.then()` will do the trick.

src/routes/_.questions.$slug.edit.tsx
```diff
@@ -26,7 +26,7 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
   const dbQuestion = result.data as unknown as DBQuestion
 
   const images = dbQuestion.images
-    ? storageServiceServer.getImagesPublicUrls(
+    ? await storageServiceServer.getImagesPublicUrls(
```
mariojsnunes, line 29: is this await needed?

src/routes/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+      .from('useful_votes')
+      .select('*', { count: 'exact' })
+      .eq('content_id', id)
+      .eq('content_type', 'news'),
+    client
+      .from('subscribers')
+      .select('user_id', { count: 'exact' })
+      .eq('content_id', id)
+      .eq('content_type', 'news'),
+    await tagsService.getTags(client, tagIds),
```
mariojsnunes, line 27: do not use await here

src/services/storageService.server.ts
```diff
@@ -4,24 +4,27 @@ import type { SupabaseClient } from '@supabase/supabase-js'
...
-  return images?.map((x) => {
-    const { data } = client.storage
+  if (!images || images.length === 0) {
+    return []
+  }
+
+  const imageList: Image[] = []
+
+  for await (const image of images) {
+    const response = await client.storage
```
mariojsnunes, line 19: `getPublicUrl` doesn't return a promise, at least in my local code version

src/services/tagsService.ts
```diff
@@ -9,6 +11,25 @@ const getTags = async () => {
     return []
   }
 }
+
+const getTags = async (client: SupabaseClient, tagIds: number[]) => {
```
mariojsnunes, line 15: ⚠️supabase client must only be used in server-side code. ⚠️
move this method to a `tagsService.server.ts` file, to ensure that.

src/routes/_.questions.$slug.edit.tsx
benfurber: Yes. In case of slow connections. Want the files uploads and linked to the question before the page loads.

mariojsnunes: check the other comment at storageServiceServer, the base method doesn't return a promise

src/routes/_.news.create.tsx
benfurber: Fancy coming up with this pattern in branch?

src/routes/_.questions.$slug._index.tsx
benfurber: Fancy coming up with this pattern in branch?

davehakkens:
I would remove (or hide) all of these components on release as its irrelevant for users

image

And could we (pre) release with this item not in the navigation bar, either hidden or for beta-testers.
This allows us to deploy on live with a space to test

image

mariojsnunes:
image
I think this doesn't make sense to appear for logged out users.
Should only appear for logged in users which are admins.

benfurber:
Feedback:
- Image qual
- Wrong body font
- Limiting headers?
- Read more button - use outline version

supabase/migrations/20250331134409_add_news.sql
```diff
@@ -0,0 +1,106 @@
+alter type "public"."content_types" ADD VALUE 'news';
+
+create table "public"."news" (
```
mariojsnunes, line 3: is it possible to have just 1 script for all the migrations? might be tricky with the preview database

src/utils/helpers.ts
```diff
@@ -248,3 +251,46 @@ export const buildStatisticsLabel = ({
...
+    if (error) {
+      errors.push(error)
+    }
+    continue
+  }
+
+  return { valid: errors.length === 0, errors }
+}
+
+export const getSummaryFromMarkdown = (text: string) => {
```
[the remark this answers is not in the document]
benfurber: I'm using as much of one as I can find. If you can find a simplier one I'd be delighted. I spend 2.5 hours on this already. cc. @davehakkens.

supabase/migrations/20250331134409_add_news.sql
benfurber: I'm fine doing it as one. Reminder, it requires an extra step for preview though as the db has to be reset rather than rolled forward.

PR #4207, opened 2025-04-25
feat: Add visitor policy to profile settings

### LINKED ISSUE #4204: [map] [profile] Open to visitors
Image

This should be a fun one that's important to users, but fairly chunky so please shout if the guidance isn't clear enough.

## Context

Many people on our maps, and Project Kamp themselves, have this issue that people expect them to be open to visitors. So this is to make it clearer to users if a given user on the map is actually open to visitors and if so how/when.

## Component infos

1. The preference setting to show on the main settings tab.
2. How to display the setting on profiles. (The interaction can use the existing `Modal` component)
3. The filtering setting to show on the map filtering modal.

All three UIs can be seen [on the design ticket](https://github.com/ONEARMY/community-platform/issues/4093).

## Data 

This is a new field for users so should be easy to add. Probably typed something like this:
```ts
type UserVisitorPreferenceStatus =  'open' | 'appointment' | 'closed'

type UserVisitorPreference = {
  status: UserVisitorPreferenceStatus,
  details: string
}

export interface IUser {
  ...
  openToVisitors?: UserVisitorPreferenceStatus | undefined
}
```

## Testing

Extend the existing cypress tests, for the settings `/packages/cypress/src/integration/settings.spec.ts` and for the map `/packages/cypress/src/integration/map.spec.ts`.

benkelaar:
WIP, created draft PR to allow potential early feedback and to show progression state.

## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## What is the new behavior?

_Describe the new behaviour_

- [ ] Yes
- [x] No

## Git Issues

Closes #4204 

## What happens next?


src/pages/User/content/UserProfile.tsx
```diff
@@ -41,6 +42,8 @@ export const UserProfile = ({ docs, isViewingOwnProfile, user }: IProps) => {
   const defaultValue =
     useLocationHook?.hash?.slice(1) || (hasProfile ? 'profile' : 'contact')
 
+  const [selectedTab, setSelectedTab] = useState(defaultValue)
```
benfurber, line 45: Why is this needed?

benfurber (changes requested):
Thanks a lot @benkelaar, I like your style. AND some alphabetising/typo fixing! Whoop.

Have you been able to run the cypress tests locally? I'd really like to see this feature tested there, as per the testing info on the ticket.

There's a lot more complexity in `packages/components/src/ProfileTagsList/ProfileTagsList.tsx` now. Do you think it's worth it?

package.json
```diff
@@ -32,6 +32,7 @@
     "build": "yarn build:shared && yarn build:themes && yarn build:components && yarn build:vite",
     "lint": "yarn lint:style && yarn lint:code",
     "lint:code": "eslint . --ext .js,.jsx,.ts,.tsx src --color",
+    "lint:fix": "yarn lint:style --write && yarn lint:code --fix",
```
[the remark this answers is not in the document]
benkelaar: You're absolutely correct, missed that, thanks!

src/pages/User/content/UserProfile.tsx
benkelaar: Because linking to #contacts didn't work (I think since we're already on the profile url the react router doesn't consider it a move at all and tab selection isn't achieved with anchors) I needed a way to programmatically select the contacts tab after clicking the "contact this space" button.  I think the way to do that is by introducing this tab value model, and through the onChangeListener the normal process still works.

Definitely open to alternatives.

benkelaar:
> Thanks a lot @benkelaar, I like your style. AND some alphabetising/typo fixing! Whoop.
> 
> Have you been able to run the cypress tests locally? I'd really like to see this feature tested there, as per the testing info on the ticket.
> 
> There's a lot more complexity in `packages/components/src/ProfileTagsList/ProfileTagsList.tsx` now. Do you think it's worth it?

Thanks for the compliment BTW.  Definitely going to look at the cypress tests.  They were on my todo list still.

@benfurber For the complexity, I'm now doubting between two designs to better separate the concerns in ProfileTagList.

Option 1 (Less work but seems slightly less correct)
- ProfileTagList still creates all normal tags but has a property for extra tagStyles
- It renders its children in the same element as the tags.

This would basically (is pseudoSX a thing?) make the small taglist something like:
`<PTL tags={tags}/>`
While the large one would essentially be
`<PTL tags={tags} tag-sx={font-size: 2}><VisitorTag /></PTL>`
Possible alternative could be `<PTL tags={tags} tag-sx={font-size: 2} additionalTagElements={[<VisitorTag />]}/>`

Option 2 (seems like slightly better separation of concerns, but has a bit more duplication)
- ProfileTagList is essentially just a wrapper that renders its children (not sure if it needs a dedicated component then)
- All logic is moved to the Tag component

This would make the small taglist:
`<PTL>tags.map((t, i) => <Tag key={i} tag={t}/>)</PTL>`
While the large one would be
`<PTL>tags.map((t, i) => <Tag key={i} tag={t} sx={font-size: 2}/>)<VisitorTag /></PTL>`

src/pages/User/content/UserProfile.tsx
benfurber: Got it. Makes sense. 

I'm realising I setup our only two pages with tabs work differently. Profiles and settings. For the setting, I created a whole wrapper that does path matching and doesn't have this problem, while profiles uses anchors and so does. 

So the other approach would be like `SettingsPage` with `SettingsFormWrapper`. 

...Thinking... The tech debt of these different approaches is fine for now.

benfurber:
> > Thanks a lot @benkelaar, I like your style. AND some alphabetising/typo fixing! Whoop.
> > Have you been able to run the cypress tests locally? I'd really like to see this feature tested there, as per the testing info on the ticket.
> > There's a lot more complexity in `packages/components/src/ProfileTagsList/ProfileTagsList.tsx` now. Do you think it's worth it?
> 
> Thanks for the compliment BTW. Definitely going to look at the cypress tests. They were on my todo list still.
> 
> @benfurber For the complexity, I'm now doubting between two designs to better separate the concerns in ProfileTagList.
> 
> Option 1 (Less work but seems slightly less correct)
> 
>     * ProfileTagList still creates all normal tags but has a property for extra tagStyles
> 
>     * It renders its children in the same element as the tags.
> 
> 
> This would basically (is pseudoSX a thing?) make the small taglist something like: `<PTL tags={tags}/>` While the large one would essentially be `<PTL tags={tags} tag-sx={font-size: 2}><VisitorTag /></PTL>` Possible alternative could be `<PTL tags={tags} tag-sx={font-size: 2} additionalTagElements={[<VisitorTag />]}/>`
> 
> Option 2 (seems like slightly better separation of concerns, but has a bit more duplication)
> 
>     * ProfileTagList is essentially just a wrapper that renders its children (not sure if it needs a dedicated component then)
> 
>     * All logic is moved to the Tag component
> 
> 
> This would make the small taglist: `<PTL>tags.map((t, i) => <Tag key={i} tag={t}/>)</PTL>` While the large one would be `<PTL>tags.map((t, i) => <Tag key={i} tag={t} sx={font-size: 2}/>)<VisitorTag /></PTL>`
> 
> Any preference?

Two clear options, thank you. Let's go with option two.

benkelaar:
@dalibormrska Hi Dalibor!  Thanks, I had been wondering about that.  What's supposed to happen when someone has visitors enabled and then changes to a member type profile, should it be turned off or should the visitor state remain?

So we can do:

- Switch to member profile - delete visitor info (risks data loss with accidental switch)
- Switch to member profile - turn off visitor info, keep details (Requires different data architecture)
- Switch to member profile, keep visitor info
- Switch to member profile, keep visitor info, but don't show it.

dalibormrska:
@benkelaar Cool, thanks for such a quick response and the options. @benfurber, do you have a preference?

benfurber:
Just keep the data but hide the display with the super sophisticated `isMember` variable that's already there.

benfurber:
Was going to ask if you're ready for another review yet @benkelaar as the description says you're still drafting. :)

benkelaar:
Nah, have the cypress tests working now so want to get those done first.  Give me a couple more evenings and I should be done.

benfurber:
@benkelaar Sure thing!

I've just run the linter so if all the checks pass I'll merge this in.

PR #4250, opened 2025-05-27
feat: add reply complexity to new notifications

benfurber:
## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## What is the new behavior?

1. Users can now subscribe to new comments only or to new replies (of a single comment).
2. Users are automatically subscribed when they leave a comment or reply.
3. This creates another action at the comment level (to subscribe/unsubscribe) so it was finally time to implement the action box to hide away comment actions.

Screenshot 2025-05-27 at 15 56 20

Screenshot 2025-05-27 at 15 55 59

Screenshot 2025-05-28 at 15 24 00

## Tech debt/still to do

1. Two buttons for same subscribe action (get out of sync):
https://github.com/user-attachments/assets/851a01ba-d7f0-4666-a093-336d875cd6fc

2. Only the creator of a research update gets automatically subscribed to the comments of their new update (i.e. collaborators/original author of the research item aren't subscribed). This is because we save the username string of collaborators so it's messy to get their profile ids. Will do this in a seperate PR.


packages/components/src/ActionSet/ActionSet.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const toDisplay = children.filter((child) => !!child)
+  if (!children || toDisplay.length === 0) {
+    return <></>
+  }
+
+  const onClick = () => setShow((show) => !show)
+
+  return (
+    <>
```
mariojsnunes: unnecessary fragment

davehakkens:
If I leave a comment it doesnt show my avatar. (as you can see I have one)
Not sure if its related to this PR but just noticed now when testing
image

davehakkens:
image in email template could be better (transparant background and no cropped of edges) 
image

PR #4261, opened 2025-06-03
feat: add user settings for notifications

benfurber:
## PR Checklist

## PR Type

Users are sent whatever emails we want to send them.

## What is the new behavior?

Control to opt-out of various types (and have those preferences respected, obvs).

Screenshot 2025-06-04 at 15 49 23

Doesn't implement the checkbox style - will write up a seperate ticket for that.

- [x] Yes
- [ ] No


benfurber:
The flow is working but I need to decide what/how to test and make the UI pretty.

packages/components/src/GridForm/GridForm.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export interface IProps {
+  fields: Fields[]
+  heading: string
+}
+
+export const GridForm = ({ fields, heading }: IProps) => {
+  return (
+    <>
+      <Grid gap={2} columns={2} width={'10%'} sx={{ padding: 2 }}>
```
mariojsnunes: gap and width into sx? same for others below

src/services/notificationsPreferencesService.ts
```diff
@@ -0,0 +1,30 @@
+import type { NotificationsPreferencesFormData } from 'oa-shared'
+
+const getPreferences = async () => {
+  try {
+    const preferencesData = await fetch('/api/notifications-preferences')
```
mariojsnunes, line 8: ideally this would destruct the preferences and return the type instead of any.
or the api would return the object already destructured

src/pages/UserSettings/SupabaseNotifications.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  const { userStore } = useCommonStores().stores
+  const user = userStore.activeUser
+  const formId = 'SupabaseNotifications'
+  const hasMessagingOn =
+    user?.isContactableByPublic === undefined
+      ? true
+      : user?.isContactableByPublic
+
+  if (!user) return null
```
mariojsnunes: returning too early. this will cause issues with the useEffects and other functions below.
move this instead to right before the main return.

src/pages/UserSettings/SupabaseNotifications.tsx
```diff
@@ -0,0 +1,154 @@
...
+export const SupabaseNotifications = () => {
+  const [isLoading, setIsLoading] = useState<boolean>(true)
+  const [initialValues, setInitialValues] = useState<
+    DBNotificationsPreferences | undefined
+  >(undefined)
+  const [submitResults, setSubmitResults] = useState<SubmitResults | null>(null)
+
+  const { userStore } = useCommonStores().stores
+  const user = userStore.activeUser
+  const formId = 'SupabaseNotifications'
```
mariojsnunes: move `formId` outside the component (still in this file). so it's not recreated on every render

src/pages/UserSettings/content/sections/PublicContact.section.tsx
```diff
@@ -23,6 +23,7 @@ export const PublicContactSection = observer((props: Props) => {
         gap: 2,
       }}
     >
+      <a id="public-contact"></a>
```
mariojsnunes, line 26: is this needed? how does it work?

src/routes/api.news.ts
```diff
@@ -183,6 +184,7 @@ export const action = async ({ request }: LoaderFunctionArgs) => {
 
     const news = News.fromDB(newsResult.data[0], [])
 
+    subscribersServiceServer.add('news', news.id, profile.id, client)
```
mariojsnunes: nice!

src/routes/api.notifications-preferences.ts
```diff
@@ -0,0 +1,98 @@
...
+    data: { user },
+  } = await client.auth.getUser()
+
+  if (!user) {
+    return Response.json({}, { status: 401, statusText: 'unauthorized' })
+  }
+
+  const { data } = await client
+    .from('notifications_preferences')
+    .select('*, profiles!inner(id)', { count: 'exact' })
```
mariojsnunes: is the count necessary?

src/routes/api.notifications-preferences.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    const { data, error } = await client
+      .from('profiles')
+      .select('id, auth_id')
+      .eq('auth_id', user?.id)
+      .single()
+
+    if (!data) {
+      throw error
```
mariojsnunes: we shouldn't send database errors to the frontend.
just console.error() here, and do throw Error("user not found") or whatever is more appropriate.

the console.error() is important so we can better monitor the app when it's running live

src/services/notificationEmailService.server.ts
```diff
@@ -50,6 +68,36 @@ const createInstantNotificationEmail = async (
   }
 }
 
+const shouldSendEmail = async (
+  client: SupabaseClient,
+  dbNotification: DBNotification,
+  profileId: number,
+): Promise<boolean> => {
+  const preferenceTypes: PreferenceTypes = {
```
mariojsnunes: make `preferenceTypes` a constant elsewhere, so it's not recreated every function run

src/services/notificationsPreferencesService.server.ts
```diff
@@ -0,0 +1,24 @@
...
+const getPreferences = async (client: SupabaseClient, profileId: number) => {
+  try {
+    const { data } = await client
+      .from('notifications_preferences')
+      .select()
+      .eq('user_id', profileId)
+      .single()
+
+    if (data) {
+      return Response.json({ data }, { status: 200 })
```
mariojsnunes: this should not return a Response.json.
Response.json is only when we want to send data to the client/browser.

here this function runs on the server, and called by other server-side functions.
So just `return data as Preferences` (or wtv type it is)

mariojsnunes (changes requested):

packages/components/src/GridForm/GridForm.tsx
benfurber: Yes to gap. No to width. In this context it's defining the minimum width for the children not itself.

PR #4297, opened 2025-06-18
chore: add banner about library changes

benfurber:
Screenshot 2025-06-18 at 17 07 46

Temp banner for PP


mariojsnunes:
already pushed these changes

PR #4305, opened 2025-06-23
4163 - Update confirm modal, add delete modal and delete user endpoint

ismaelabujadur:
## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## What is the new behavior?

_Describe the new behaviour_

- [ ] Yes
- [ ] No

## Git Issues

Closes #

## What happens next?

### DIFF
--- a/packages/cypress/src/integration/deleteUser.spec.ts
+++ b/packages/cypress/src/integration/deleteUser.spec.ts
@@ -45,9 +45,7 @@ describe('Delete account flow', () => {
   describe('when I type the exact "DELETE" confirmation', () => {
     beforeEach(() => {
       cy.get('[data-cy=delete-account-modal-open-button]').click()
-      cy.get('[data-cy=delete-account-modal-confirmation-input]').type(
-        'DELETE',
-      )
+      cy.get('[data-cy=delete-account-modal-confirmation-input]').type('DELETE')
     })
 
     it('enables and lets me delete my account', () => {


ismaelabujadur:
Hi @benfurber,

The work of this PR is still WIP. I'll summarise here the issues I'm currently facing, but once those are sorted and testing implemented it should be ready to go!

- There was no behaviour specified for when the user had written wrongly or rightly the 'DELETE' word to confirm his account being deleted. I did some research on the most common sites like github, and found out that the button itself it's disabled until the user properly writes the required confirmation word. However that has caused an issue which has been a headache and it's still unresolved.
   - Once we implement the `useState` hook to control the user input and match it with the confirm word 'DELETE', it loses the focus on key down. One possible reason could be the fact that it's being re-rendered by the state being updated after every key press. However, after some debugging I couldn't find that to be an issue, everything seems to be rendering as expected, but focus still won't work!
- Any kind of query to the database should be done from the server side, my first thought was to create an endpoint `profile/delete` and followed through. However, noticed that the logout is also an endpoint. After some thought it might be best to update the delete feature to be a service instead of an endpoint to avoid multiple redirections (screen flashing between redirections) which can look quite ugly. 

I will refactor the current code into a service instead of an endpoint and try to sort out the issue with the focus for the input field. However, any feedback would be welcomed! Thanks!

benfurber:
Nice work so far @ismaelabujadur Rather than using the `ConfirmModal` it might be better to create a new component. Then you can build this as a proper form with a submit button. We build forms with react-final-form and with that you'll see there's validators you can add to fields, so you'd create a new validator and can setup the submit button to be disabled unless the field is valid.

You're definitely right to keep the api files small and put as much business logic into services, in fact you'll want a `service.server` (so that the file only loads on the backend).

benfurber:
@ismaelabujadur We've just merged in a monster PR, apoligies. Very happy to pair on dealing with the merge conflicts for your branch.

mariojsnunes:
Closing this since it has fallen significantly behind the main branch.

PR #4307, opened 2025-06-24
fix: redirect previous slugs to latest content

### LINKED ISSUE #4166: [bug] [questions] previousSlugs not implemented
**Describe the bug**
Functionality to update the previousSlugs field and find/display by previousSlugs not implemented.

**To Reproduce**

1. Go to create a question.
2. Copy the URL.
3. Edit the question (thus changing the question slug.
3. Go to the old url
4. See a 404.

**Expected behaviour**
The old url should redirect to the new one. 

**Additional context**
There's three problems/changes needed:
1. Update `isDuplicateSlug`to check both slug and previousSlugs.
2. The edit route should update the previous slugs array.
3. The display route should check the previous slugs for.

JCSergent:
## PR Checklist

## PR Type

- [x] Bugfix (fixes an issue)

Content (questions, research, news, library) are not storing their previous slugs

When navigating to a contents previous url (using a previous slug), the url gives you a 404 (instead of directing you to the content).

When checking for duplicate slugs on content creation, we are not accounting for previous slugs.

## What is the new behavior?

We now persist all the previous slugs of all content.

When navigating to content using its previous slug it will show you the updated version of that content.
Example: the url is for a previous slug name "testing113", and it is showing the latest version of the question with slug "testing115"
screenshot-2025-06-24_15:45:53

When creating new content or updating existing content, it will now block you if the content's new slug matches a previous slug of another content of the same type. This will not block content from using one of it's own previous slugs.

- [ ] Yes
- [x] No

## Git Issues

Closes #4166

### DIFF
--- a/src/routes/api.questions.$id.ts
+++ b/src/routes/api.questions.$id.ts
@@ -1,5 +1,7 @@
 import { Image, Question } from 'oa-shared'
 import { createSupabaseServerClient } from 'src/repository/supabase.server'
+import { profileServiceServer } from 'src/services/profileService.server'
+import { questionServiceServer } from 'src/services/questionService.server'
 import { hasAdminRightsSupabase, validateImages } from 'src/utils/helpers'
 import { convertToSlug } from 'src/utils/slug'
 
@@ -9,10 +11,12 @@ import { uploadImages } from './api.questions'
 import type { LoaderFunctionArgs } from '@remix-run/node'
 import type { Params } from '@remix-run/react'
 import type { SupabaseClient, User } from '@supabase/supabase-js'
-import type { DBProfile, DBQuestion } from 'oa-shared'
+import type { DBQuestion } from 'oa-shared'
 
 export const action = async ({ request, params }: LoaderFunctionArgs) => {
   try {
+    const id = Number(params.id)
+
     const formData = await request.formData()
     const imagesToKeepIds = formData.getAll('existingImages')
 
@@ -25,6 +29,7 @@ export const action = async ({ request, params }: LoaderFunctionArgs) => {
       tags: formData.has('tags')
         ? formData.getAll('tags').map((x) => Number(x))
         : null,
+      slug: convertToSlug(formData.get('title') as string),
     }
 
     const { client, headers } = createSupabaseServerClient(request)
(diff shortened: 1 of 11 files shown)


src/routes/api.questions.$id.ts
```diff
@@ -92,15 +96,23 @@ export const action = async ({ request, params }: LoaderFunctionArgs) => {
...
 
-    const slug = convertToSlug(data.title)
+    const newSlug = convertToSlug(data.title)
+    let previousSlugs = currentQuestion.previous_slugs
+    if (currentQuestion.slug !== newSlug) {
+      previousSlugs = previousSlugs
+        ? [...previousSlugs, currentQuestion.slug]
+        : [currentQuestion.slug]
+    }
+
```
benfurber: Perfect little function to pull out into a seperate fuction I'd say so that `previousSlugs` doesn't have to be reassigned.

JCSergent: This function was actually repurposed from api.research.$id.ts so totally makes sense to split it out and let the two api's share it.

Will do!

JCSergent: Where would the best place for this function to live? contentService.server.ts? That sees to be all the shared functionality between content types.

JCSergent:
> Fantastic, thank you @JCSergent. This functionality should also be on research, library and news(!), so would you be up for next refactoring this into a more generic implementation. It might be at bits of this work for them now as the change to `src/services/contentService.server.ts` will cover them, but I doubt there's the test coverage is there.

Yeah I'm up for it! I did notice the content types did have differing versions of handling the previousSlug, so I ended up basing questions off of how research did it. It would be good to get them all on the same implementation.

How do you want to break up the work? Should I put it in this PR? Or should we push this through, and put the refactor in a new PR?

src/routes/api.questions.$id.ts
benfurber: Yep. There sounds good to me.

benfurber:
@JCSergent Awesome. Either? :) 

In a single PR is cool if you've got time to do it before, say, Monday? That's when I can review next.

JCSergent:
@benfurber - yeah Monday is perfect. I'll get it done this weekend.

JCSergent:
@benfurber - It's all ready for review when you get the chance!

It ended up being a bit more changes than I expected, but I went through each of the api.{content-type}.$id.ts and standardized how they each validate the request and update the content (including the previous slug now). 

Also added the E2E test for each of them.

benfurber:
Fantastic, thank you @JCSergent! I'll do a quick QA shortly but this looks perfect.

PR #4313, opened 2025-06-25
chore: drop library banner

benfurber:


benfurber:
@mariojsnunes You cool to merge/release this over the weekend?

PR #4320, opened 2025-07-01
transform image steps and add landscape image size

mariojsnunes:
## PR Checklist

## PR Type

- [x] Bugfix (fixes an issue)

## What is the new behavior?
- step images are now transformed
- project cover image is now transformed with the correct size
- research images sizes also reviewed

- [ ] Yes
- [x] No

## Git Issues

Closes #

### DIFF
--- a/src/config/imageTransforms.ts
+++ b/src/config/imageTransforms.ts
@@ -4,12 +4,16 @@ export type ImageSize = {
 }
 
 export const IMAGE_SIZES: { [key: string]: ImageSize } = {
+  LANDSCAPE: {
+    width: 1280,
+    height: 960,
+  },
   GALLERY: {
-    height: 450,
     width: 956,
+    height: 450,
   },
   LIST: {
-    height: 225,
     width: 478,
+    height: 225,
   },
 }
--- a/src/services/libraryService.server.ts
+++ b/src/services/libraryService.server.ts
@@ -4,7 +4,7 @@ import { IMAGE_SIZES } from 'src/config/imageTransforms'
 import { storageServiceServer } from './storageService.server'
 
 import type { SupabaseClient } from '@supabase/supabase-js'
-import type { DBProject, DBProjectStep } from 'oa-shared'
+import type { DBProject, DBProjectStep, Image } from 'oa-shared'
 
 const getBySlug = (client: SupabaseClient, slug: string) => {
   return client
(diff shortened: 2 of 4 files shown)


davehakkens:
Does this mean the image we display is cropped but if you click you still get the original, uncropped?

mariojsnunes:
> Does this mean the image we display is cropped but if you click you still get the original, uncropped?

no, but if you need the original, just remove '?width=956&height=450' from the url.
see
https://wbskztclbriekwpehznv.supabase.co/storage/v1/render/image/public/precious-plastic/projects/366/IMG_3615-1957c4eaede.jpg?width=956&height=450
https://wbskztclbriekwpehznv.supabase.co/storage/v1/render/image/public/precious-plastic/projects/366/IMG_3615-1957c4eaede.jpg

mariojsnunes:
@davehakkens made the changes to not crop the image, only resize

davehakkens:
did a test but two images dont show up (both of them where +5MB)
https://community-platform-pr-4320.fly.dev/library/asdfdasf


image

image

davehakkens:
Also only tested library. Is it also applied to questions, news and research?

mariojsnunes:
Added a loading state for when the forms are submitting.
image



> did a test but two images don't show up (both of them where +5MB)

hmm I was able to upload a 7mb image, could you have navigated out of the page before it finished uploading?

> Also only tested library. Is it also applied to questions, news and research?

yes, applies to all.

davehakkens:
@benfurber can you test this as well.  because I get:
image

mariojsnunes:
@davehakkens I was able to replicate, there were two issues:
1. supabase limit which images larger than 2500px width or height cannot be transformed:
https://supabase.com/docs/guides/storage/serving/image-transformations?queryGroups=language&language=js#limits

To prevent it, added these two validations:
image
image


2. when updating an existing question images, their metadata was persisted in a wrong format (fixed).
(Checked other image uploads, seems like it only affected existing question images)

davehakkens:
Hmm thats quite tight @mariojsnunes 
Many smartphones produce larger files than that.. 

And as far as i am aware resizing on a phone is not very easy.

davehakkens:
Why are we doing these optimization again now? Just to understand the urgency

mariojsnunes:
> Why are we doing these optimization again now? Just to understand the urgency

They were there before, but with the cdn. Large images on the page will make the page load slower, users waste more internet data and affect SEO as well.

davehakkens:
ah ok. so because CDN isnt migrated we need an alternative way? (as before we didnt have upload limit)

https://supabase.com/docs/guides/storage/serving/image-transformations?queryGroups=language&language=js#limits
Also the link you shared above says 25MB limit, but your warning says 5mb. Is 5MB limit the max?

mariojsnunes:
After a chat with @davehakkens, setting a limit to 10MB and dropping the width/height limit.
Also simplified the message:
image

PR #4322, opened 2025-07-01
feat: add logged out email notifications flow

benfurber:
## PR Checklist

## PR Type

## What is the new behavior?

This adds a way for users, who are not logged in already, to update their email preferences.

A new token is added to the emails, if the token is valid, when a user completes the form their preferences will be updated.

For security reasons, if a token is present but in anyway wrong, the form will appear to work, but it won't.

Screenshot 2025-07-01 at 16 19 48


.circleci/config.yml
```diff
@@ -351,6 +351,7 @@ jobs:
               --build-secret VITE_NO_MESSAGING="$VITE_NO_MESSAGING" \
               --build-secret VITE_HIDE_MEMBER_PINS_BY_DEFAULT="$VITE_HIDE_MEMBER_PINS_BY_DEFAULT" \
               --build-secret TENANT_ID="$TENANT_ID" \
+              --build-secret TOKEN_SECRET="$TOKEN_SECRET" \
```
mariojsnunes: should be added to the "Set Supabase Secrets" instead.

src/pages/UserSettings/SupabaseNotificationsViaEmail.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+import type { DBNotificationsPreferences } from 'oa-shared'
+import type { SubmitResults } from 'src/pages/User/contact/UserContactError'
+
+interface IProps {
+  userCode: string
+}
+
+export const SupabaseNotificationsViaEmail = ({ userCode }: IProps) => {
+  if (!userCode) return null
```
mariojsnunes: do not short circuit before react hook declarations, it will cause all sort of issues.

src/routes/api.notifications-preferences-via-email.$userCode.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    const userId = userData.data?.id
+
+    if (!userId) {
+      return Response.json({}, { status: 401, statusText: 'unauthorized' })
+    }
+
+    const formData = await request.formData()
+
+    const existingPreferences = await client
```
mariojsnunes, line 76: seems like this can be done after `validateRequest`

supabase/functions/send-email/_templates/layout.tsx
```diff
@@ -51,9 +51,14 @@ type LayoutArgs = {
   children: React.ReactNode
   preview: string
   settings: TenantSettings
+  code?: string
 }
 
-export const Layout = ({ children, preview, settings }: LayoutArgs) => {
+export const Layout = ({ children, code, preview, settings }: LayoutArgs) => {
```
mariojsnunes: adding `code` here makes it more coupled. A nicer approach would be to have a `url` param instead. (same reasoning could apply to other params as well)

davehakkens:
Just a question/idea that popped my mind. 
Its currently not possible to (quickly) 'mute this topic' you got the email notification from right?


Does 'mute this topic' from email adds lots of complexity, as we already use the token?

As that seems like the fastest and least impactful way to stop being annoyed by an active thread and prevents people unsubscribing of everything.

benfurber:
@davehakkens Muting sounds like unsubscribing? Can be done. Not hard but not quick either as we'll need a new api route.

davehakkens:
yes indeed unsubscribing, thanks good to know.

PR #4330, opened 2025-07-08
feat: update icons

benfurber:
## PR Checklist

## PR Type

- [x] Build related changes

## What is the new behavior?

Screenshot 2025-07-08 at 13 42 29

Screenshot 2025-07-08 at 14 15 47


benfurber:
All the current icons:
Screenshot 2025-07-08 at 14 13 50

davehakkens:
Is there a logic in color usage? 


Also would make the one in the stats-div greyed out like the other ones


And i think the 'following' for research and comments still uses the thunder icon. The megaphone was introduced to separate it from 'all notifications' right @dalibormrska?

benfurber:
I was trying to go for consistent implementation and blue icons was what I was given.

Subscribing and following are the same thing, so it increasingly seems wrong to have different wording and iconagraphy?


Screenshot 2025-07-10 at 10 56 27

davehakkens:
> I was trying to go for consistent implementation and blue icons was what I was given.

Yeh I figured. But I was mainly wondering what the logic of colors is because these two are not blue.
image
image

davehakkens:
> Subscribing and following are the same thing, so it increasingly seems wrong to have different wording and iconagraphy?

Not sure if we are talking about the same thing. I was referring to this icon
image
Which is used to display 'top level notifications', but it seems that whatever we use to show 'follow' or 'subscribe' an specific piece of content should be used here?

davehakkens:
image
Was there ever a design made to see this implemented? As it raises some questions to me about which icon will be shown were..

davehakkens:
Also dont mind to just merge whatever we have open for notifications and do a proper UI review when all the pieces are up and running on live. Makes the review and testing easier as all the components are in place with real data.

dalibormrska:
Sorry that I'm only getting to this now.

Well, some miscommunication happened somewhere during the process, most likely from my side by not communicating clearly. First I was contemplating about tackling the iconography as a whole to improve consistency, but we agreed that it is not the time yet and I would draw the icons that will be used only for the notifications.

However I see now that they've been implemented across the whole platform.
image
image
image

That's not what I intended to do. I thought the icons would only be used in the notifications panel and settings, that's why I went with the colored variant, since in other contexts such as these below, the colored version really does not fit well.
image
image

However, we use the logic of 2 icon versions - colored and B/W - for lightning bolt for example, so we can utilise the same logic here. I will prepare the icons.

It is also true that I missed the icon for direct message - this is still a placeholder.
image

Currently there is no clear logic in color.

I agree with Dave that there is a difference between Following (lightning bolt) and Notifications (megaphone), so the icon here should be lightning bolt in my opinion.
image

And there were no icons drawn for the modules at that point. I also remembered that we have [these from the CP's page](https://platform.onearmy.earth/roadmap.html), but they are not suitable for smaller sizes and now we also have the news module to be added.

---

Now as Dave suggested, I am preparing an overview of the notifications with the final designs, after that I will update this thread with specific points to change.

PR #4331, opened 2025-07-09
test: add unit tests

benfurber:
## PR Checklist

## PR Type

- [x] CI related changes

## What is the new behavior?

Unit tests provided for the recent notifications work. Heavily reliant on Claude for this.

### DIFF
--- a/shared/models/notificationsPreferences.ts
+++ b/shared/models/notificationsPreferences.ts
@@ -7,13 +7,16 @@ export class NotificationsPreferences {
   isUnsubscribed: boolean
 }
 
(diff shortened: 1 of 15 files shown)


src/services/notificationEmailService.serve.test.ts
```diff
@@ -0,0 +1,547 @@
+import { transformNotification } from 'src/routes/api.notifications'
```
mariojsnunes: route name serve instead of server.

also please double check this file isn't included in the bundle, otherwise there might be a security leak.

PR #4359, opened 2025-07-22
fix: auto content subscribing

benfurber:
## PR Checklist

## PR Type

- [x] Bugfix (fixes an issue)

## What is the new behavior?

1. Missed implementation to auto-subscribe creaters of projects to the discussion.
2. Fix bug that was wrongly subscribing research update creator/collaboraters.


src/services/subscribersService.server.ts
```diff
@@ -41,9 +41,9 @@ const addResearchSubscribers = async (
     client,
   )
 
-  subscribers.forEach((subscriber) => {
+  for (const subscriber of subscribers) {
     addFunction('research', research.id, subscriber, client)
```
mariojsnunes: this should return an await Promise.all([...
same for all the others.

If whoever calls this function doesn't want to await it, then it's their responsability, not this function's

PR #4361, opened 2025-07-22
feat: email footer service message or preferences

benfurber:
For notification emails:
Screenshot 2025-07-22 at 14 50 19

Service emails say: "This is a service email" instead.


benfurber:
Looking for thumbs up on this approach before deploying. @davehakkens

davehakkens:
text works for me

PR #4363, opened 2025-07-23
feat: switch not-following reply wording to action

benfurber:
## PR Checklist

Screenshot 2025-07-23 at 10 04 23

## What is the new behavior?

Screenshot 2025-07-29 at 09 23 28

### DIFF
--- a/src/pages/common/CommentsSupabase/CommentItemSupabase.tsx
+++ b/src/pages/common/CommentsSupabase/CommentItemSupabase.tsx
@@ -102,8 +102,8 @@ export const CommentItemSupabase = observer((props: ICommentItemProps) => {
                   iconFollow="discussionFollow"
                   iconUnfollow="discussionUnfollow"
                   itemId={comment.id}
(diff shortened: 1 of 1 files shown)


benfurber:
I'll wait for some thumbs up this time.

davehakkens:
Dont think the 'following replies' text works for the active state
- the visual difference between the following and not following is low (very small detailed change in icon)
- Clicking 'following replies' doesnt remove the thunder icon in the comment. Which makes the action confusing.

If the above functionally stays the same I would go for
'Follow replies'
'unfollow replies'

benfurber:
@davehakkens

Currently:
"Mark as useful" <--> "Marked as useful"
"Follow comments" <--> "Following comments"

So changing this to "Follow replies" <--> "Unfollow replies" is inconsistent with those two. Just confirming you're fine with that?

davehakkens:
yes.
Not ideal but given the current functionality it seems most logical.

PR #4380, opened 2025-08-14
Feat/comments/add useful button

### LINKED ISSUE #4342: [comments] Add useful button to comments
Image
[Full figma file here](https://www.figma.com/design/XYV3fr8UVOdkNyBmKHHR7f/CP---Comments-Useful---Mobile?node-id=488-20613&m=dev)

For some other content types (projects, questions, research), content can be marked as useful. We want to be able to do the same for comments and replies. 

So this feature requires a lot:
- the useful field being added to the DB
- link up the DB to the app (hopefully the existing useful api only needs minor updates)
- New useful UI component (and stories/tests)
- Adding the new component to CommentDisplay (probably)
- Update the end-to-end tests

matteobu:
## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

No useful button for comments

## What is the new behavior?

Adding a useful button for comments

_Describe the new behaviour_

This PR is part of the issue #4342, is not intended to be fully working now what is missing is:

- Final UI for the `UsefulButtonLite`
- Stories for the  `UsefulButtonLite`
- The end-to-end tests

What I did so far:
The component is a smaller and slightly different version of `usefulStatsButton`.  
Since in the Figma design the count appears on the left, I needed to adjust the layout accordingly.

When the user clicks the button, it triggers the `onUsefulClick` function, which exists in both `CommentItemSupabase` and `CommentReplySupabase`.  
Before this PR, there were 5 separate instances of `onUsefulClick` across different files. I took the opportunity to export it into a shared location and have each component import and use the same function.

The `onUsefulClick` function calls either `usefulService.add` or `usefulService.remove`, depending on whether the goal is to add or remove a "like".  
From there, the process follows the existing logic until it reaches the database, where I had to update the `content_type` type to include `"comment"`.  
*(Is this change only in my local DB instance? How can the production DB be updated? @benfurber @mariojsnunes)*

To retrieve the useful count, I added a function in `CommentSectionSupabase`.  
When the initial `useEffect` runs and fetches all comments, I also call `usefulService.getVoteCount` (which is a new service) to retrieve the vote counts for both the comments and their replies.  
These counts are then added to the comment objects in the same `useEffect`.  
As a result, when passing the comments to `<DisplayComments />`, each comment already contains its count, and this data flows through to the new `usefulStatsButton`.

Pushing this now so we can start the discussion on whether these changes need anything else added.

I’ll keep working on the UI and the remaining parts. In the meantime, have a look at the UI progress so far.

https://github.com/user-attachments/assets/ab1bdc92-57cc-4bc8-a288-b57a5b8c74f1

Screenshot 2025-08-14 at 10 11 46

Bug or what? @benfurber @mariojsnunes 

When clicking “useful” right after leaving a comment, I get NaN instead of the count.
Can someone point me to where I should check for this issue?
Could it be a database retrieval problem (no comment actually stored yet) or something else?
In the console, I see "user not found", but after refreshing, everything works fine.

- [x] Yes
- [ ] No

## Git Issues

Closes #4342

### DIFF
--- a/packages/components/src/CommentDisplay/CommentDisplay.tsx
+++ b/packages/components/src/CommentDisplay/CommentDisplay.tsx
@@ -7,6 +7,7 @@ import { CommentAvatar } from '../CommentAvatar/CommentAvatar'
 import { CommentBody } from '../CommentBody/CommentBody'
 import { DisplayDate } from '../DisplayDate/DisplayDate'
 import { AuthorsContext } from '../providers/AuthorsContext'
+import { UsefulButtonLite } from '../UsefulStatsButton/UsefulButtonLite'
 import { Username } from '../Username/Username'
 
 import type { Comment } from 'oa-shared'
@@ -14,10 +15,17 @@ import type { ReactNode } from 'react'
 
 export interface IProps {
   comment: Comment
-  isEditable: boolean | undefined
   itemType: 'ReplyItem' | 'CommentItem'
+  isLoggedIn: boolean
+  isEditable: boolean | undefined
   setShowDeleteModal: (arg: boolean) => void
   setShowEditModal: (arg: boolean) => void
+  votedUsefulCount: number
+  hasUserVotedUseful: boolean
+  onUsefulClick: (
+    vote: 'add' | 'delete',
+    eventCategory?: string,
+  ) => Promise<void>
   followButton?: ReactNode
   followButtonIcon?: ReactNode
 }
@@ -27,12 +35,16 @@ const DELETED_COMMENT = 'The original comment got deleted'
 export const CommentDisplay = (props: IProps) => {
   const {
     comment,
-    isEditable,
     itemType,
-    setShowDeleteModal,
-    setShowEditModal,
+    isLoggedIn,
+    isEditable,
     followButton,
     followButtonIcon,
+    hasUserVotedUseful,
+    votedUsefulCount,
+    setShowDeleteModal,
+    setShowEditModal,
+    onUsefulClick,
   } = props
 
   const { authors } = useContext(AuthorsContext)
@@ -147,7 +159,20 @@ export const CommentDisplay = (props: IProps) => {
                 </ActionSet>
               </Flex>
             </Flex>
-            <CommentBody body={comment.comment} />
+            <Flex
+              sx={{
+                flexDirection: 'column',
+                gap: 2,
+              }}
+            >
+              <CommentBody body={comment.comment} />
+              <UsefulButtonLite
+                votedUsefulCount={votedUsefulCount}
+                hasUserVotedUseful={hasUserVotedUseful}
+                isLoggedIn={!!isLoggedIn}
+                onUsefulClick={onUsefulClick}
+              />
+            </Flex>
           </Flex>
         </Flex>
       </Flex>
--- a/src/pages/common/CommentsSupabase/CommentReplySupabase.tsx
+++ b/src/pages/common/CommentsSupabase/CommentReplySupabase.tsx
@@ -8,12 +8,12 @@ import {
   Modal,
 } from 'oa-components'
 import { UserRole } from 'oa-shared'
-import { trackEvent } from 'src/common/Analytics'
 import { useCommonStores } from 'src/common/hooks/useCommonStores'
 import { usefulService } from 'src/services/usefulService'
+import { onUsefulClick } from 'src/utils/onUsefulClick'
 import { Box, Flex, Text } from 'theme-ui'
 
-import type { Reply } from 'oa-shared'
+import type { ContentType, Reply } from 'oa-shared'
 
 const DELETED_COMMENT = 'The original comment got deleted'
 
@@ -64,33 +64,23 @@ export const CommentReply = observer(
     const item = 'ReplyItem'
     const loggedInUser = userStore.activeUser
 
-    const onUsefulClick = async (
+    const configOnUsefulClick = {
+      contentType: 'comment' as ContentType,
+      contentId: comment.id,
+      eventCategory: 'Comment',
+      slug: `${comment}+${comment.id}`,
+      setVoted,
+      setUsefulCount,
+      loggedInUser: loggedInUser,
+    }
+
+    const handleUsefulClick = async (
       vote: 'add' | 'delete',
       eventCategory = 'Comment',
     ) => {
-      console.log('Reply Comment Test', vote, eventCategory)
-
-      if (!loggedInUser?.userName) {
-        return
-      }
-
-      // Trigger update without waiting
-      if (vote === 'add') {
-        await usefulService.add('comment', comment.id)
-      } else {
-        await usefulService.remove('comment', comment.id)
-      }
-
-      setVoted((prev) => !prev)
-
-      setUsefulCount((prev) => {
-        return vote === 'add' ? prev + 1 : prev - 1
-      })
-
-      trackEvent({
-        category: eventCategory,
-        action: vote === 'add' ? 'CommentUseful' : 'CommentUsefulRemoved',
-        label: `comment-${comment.id}`,
+      await onUsefulClick({
+        vote,
+        config: { ...configOnUsefulClick, eventCategory },
       })
     }
 
@@ -127,7 +117,9 @@ export const CommentReply = observer(
                 comment={comment}
                 setShowDeleteModal={setShowDeleteModal}
                 setShowEditModal={setShowEditModal}
-                onUsefulClick={onUsefulClick}
+                onUsefulClick={() =>
+                  handleUsefulClick(voted ? 'delete' : 'add')
+                }
                 hasUserVotedUseful={voted}
                 votedUsefulCount={usefulCount}
                 isLoggedIn={!!loggedInUser}
--- a/src/pages/common/CommentsSupabase/CommentReplySupabase.tsx
+++ b/src/pages/common/CommentsSupabase/CommentReplySupabase.tsx
@@ -8,7 +8,9 @@ import {
   Modal,
 } from 'oa-components'
 import { UserRole } from 'oa-shared'
+import { trackEvent } from 'src/common/Analytics'
 import { useCommonStores } from 'src/common/hooks/useCommonStores'
+import { usefulService } from 'src/services/usefulService'
 import { Box, Flex, Text } from 'theme-ui'
 
 import type { Reply } from 'oa-shared'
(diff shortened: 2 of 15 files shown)


matteobu:
UI update preview, let me know if this is ok.


https://github.com/user-attachments/assets/984f9dcd-2cec-497c-9ef7-f4945be52fbc

src/pages/common/CommentsSupabase/CommentReplySupabase.tsx
```diff
@@ -37,6 +41,17 @@ export const CommentReply = observer(
       )
     }, [activeUser, comment])
 
+    useEffect(() => {
+      const getVoted = async () => {
+        const voted = await usefulService.hasVoted('comment', comment.id)
```
mariojsnunes: ok this will be problematic. it will cause N calls to the API, N being the number of comments, which could be hundreds.
Unsure of the best solution, something to think about.

mariojsnunes: I see two options:
- pass in the current user to the "get comments" api call, comments have a property if the current user voted on them
- somewhere above, after retrieving the comments, get the voted state for all of them in 1 call.

src/routes/api.useful.$contentType.$id.count.ts
```diff
@@ -0,0 +1,46 @@
+import { createSupabaseServerClient } from 'src/repository/supabase.server'
+
+import type { LoaderFunctionArgs } from '@remix-run/node'
+import type { ContentType } from 'shared/lib/models/common'
```
mariojsnunes: this import should be 'oa-shared'

src/utils/onUsefulClick.ts
```diff
@@ -0,0 +1,63 @@
+import { trackEvent } from 'src/common/Analytics'
+import { usefulService } from 'src/services/usefulService'
+
+import type { ContentType, IUserDB } from 'shared/lib'
```
mariojsnunes: 'oa-shared'

packages/components/src/CommentDisplay/CommentDisplay.tsx
```diff
@@ -7,17 +7,25 @@ import { CommentAvatar } from '../CommentAvatar/CommentAvatar'
...
 import { Username } from '../Username/Username'
 
 import type { Comment } from 'oa-shared'
 import type { ReactNode } from 'react'
 
 export interface IProps {
   comment: Comment
-  isEditable: boolean | undefined
   itemType: 'ReplyItem' | 'CommentItem'
+  isLoggedIn: boolean
```
mariojsnunes: to avoid prop drilling maybe passing the whole 
```
 <UsefulButtonLite
                votedUsefulCount={votedUsefulCount}
                hasUserVotedUseful={hasUserVotedUseful}
                isLoggedIn={!!isLoggedIn}
                onUsefulClick={onUsefulClick}
              />
```
as a param would be better.

This is just a suggestion, the component would need to be created above, unsure if it's cleaner.

mariojsnunes:
for the db changes you can generate a new migration with
`supabase db diff --file [migration_name]`
review the file to ensure you only have the related changes for this PR.

src/pages/common/CommentsSupabase/CommentReplySupabase.tsx
matteobu: So would it be better to have a call about `voted` here as well? 
In [`src/pages/common/CommentsSupabase/CommentSectionSupabase.tsx` line 77](https://github.com/ONEARMY/community-platform/blob/2b1197f4ce13833bb1e112f988570cb716ab70de/src/pages/common/CommentsSupabase/CommentSectionSupabase.tsx#L43)

matteobu: ### Solution proposal:

#### 1. Backend API (`/api/discussions/${sourceType}/${sourceId}/comments`)
- Add a LEFT JOIN to the `useful_votes` table in the existing query
- Filter votes by current user's auth_id 
- Include vote data in the response so each comment comes back with `hasVoted: boolean`

#### 2. Frontend (`useEffect` in comment component)
- **Remove** the separate `useEffect` that calls `usefulService.hasVoted()` for each comment
- The vote status will now come directly from the comments API response
- This eliminates the N+1 API calls entirely

### BEFORE (N+1 problem):
```
GET /api/discussions/questions/123/comments → returns comments
GET /api/useful/comment/1/voted → check vote for comment 1
GET /api/useful/comment/2/voted → check vote for comment 2
... (N more calls for N comments)
```
### AFTER (single call):
```
GET /api/discussions/questions/123/comments → returns comments WITH vote status
```

**Note**: We'll still need the individual vote count calls (`getVoteCount`) for now, but eliminating the `hasVoted` calls will significantly reduce the API overhead.

benfurber:
@matteobu We've just merged in a monster PR, apoligies. Very happy to pair on dealing with the merge conflicts for your branch.

matteobu:
> @matteobu We've just merged in a monster PR, apoligies. Very happy to pair on dealing with the merge conflicts for your branch.

Fine, but first, can you please have a look at this proposal?

https://github.com/ONEARMY/community-platform/pull/4380#discussion_r2276497832

and this help I need:

> When clicking “useful” right after leaving a comment, I get NaN instead of the count.
Can someone point me to where I should check for this issue?
Could it be a database retrieval problem (no comment actually stored yet) or something else?
In the console, I see "user not found", but after refreshing, everything works fine.

src/pages/common/CommentsSupabase/CommentReplySupabase.tsx
mariojsnunes: yep sounds good to me

mariojsnunes:
> When clicking “useful” right after leaving a comment, I get NaN instead of the count.

I'd say first, deal with conflicts, in case it fixes itself.

It probably won't, so I recommend debugging the server code and see what's happening.
I recently added the config for debug, so you should be able to see this, and click play:
image
otherwise, hit me up on discord and I'll try to help

matteobu:
In the meantime, I did CommentDisplay stories, CommentDisplay tests with UsefulButtonLite, and UsefulButtonLite unit tests.

src/pages/common/CommentsSupabase/CommentReplySupabase.tsx
matteobu: @mariojsnunes, please have a look at the new solution.

matteobu:
@mariojsnunes @benfurber 
Conflicts should be solved now.
Please review this PR again.

davehakkens:
Looking nice!
Would suggest a small design tweak to make it visually a bit less prominent and keep focus on the text
1. Make the default state without background
2. Align it with the 'reply' button
3. Dont show when 0


useful button

matteobu:
> Looking nice! Would suggest a small design tweak to make it visually a bit less prominent and keep focus on the text
> 
> 1. Make the default state without background
> 2. Align it with the 'reply' button
> 

I can align the Btn with the 'reply' btn, this involves moving the UsfBtnLite out of the `CommentDisplay`, redoing the tests, and redoing the stories.

matteobu:
> Looking nice! Would suggest a small design tweak to make it visually a bit less prominent and keep focus on the text
> 
> 1. Make the default state without background
> 2. Align it with the 'reply' button
> 3. Dont show when 0


Hey @davehakkens , @mariojsnunes , @benfurber

If I align the `UsefulButtonLite` to the `ButtonShowReplies`, then when a user opens the replies, the button goes down.


See the picture for reference:
Screenshot 2025-08-25 at 10 10 15

Screenshot 2025-08-25 at 10 12 37

davehakkens:
> If I align the `UsefulButtonLite` to the `ButtonShowReplies`, then when a user opens the replies, the button goes down.


AH I see. thanks for flagging. If it behaves like that I would not align them as it gets a bit confusing.
Could just stick to the previous position for now

matteobu:
Ok, so this is the final UI. Let me know!

Screenshot 2025-08-25 at 14 59 18
Screenshot 2025-08-25 at 14 58 48

https://github.com/user-attachments/assets/9d52f014-fb3e-4a2d-bc7c-c8653602d9e4

davehakkens:
Nice! Small last details

1. Can we make the dark grey background of this state `comment useful by others` in the same light grey as the `comment "useful" by the user`  image

2. Could we release it for beta testers first. In the auth wrapper?

matteobu:
Background color change 
Screenshot 2025-08-26 at 15 36 54
Screenshot 2025-08-26 at 15 36 46

dalibormrska:
Wow, really awesome work @matteobu tackling this issue! I am really impressed and it's so cool to play around with the implementation. Really important feature for the future of the platform!
Thanks @davehakkens for stepping in with QA to align it with design.

There would be 3 details from my side to align it with design as intended.

1. Scale - Currently the button is a bit too small, problematic to comfortably tap on a mobile. Yes as Dave pointed out, the intention was to have it in the same line as the reply button. The reply button is 32px tall and this useful button should be in that sense consistent with it.  image

To take the measurements from Figma more precisely: 
image

The square button with the icon is 32px32px. The icon itself is 24px24px. The font-size of the numeral is 14px and with some padding on top and bottom, the total height of the grey container is 25px. 
image

2. Transitions - Have a look at the [Figma prototype](https://www.figma.com/proto/XYV3fr8UVOdkNyBmKHHR7f/CP---Comments-Useful---Mobile?node-id=488-20613&t=pNZHjW8Ei6NdTWog-1&scaling=min-zoom&content-scaling=fixed&page-id=48%3A79&starting-point-node-id=488%3A20613&show-proto-sidebar=1) and try to replicate the hover and the transition behaviour. To describe it, there is a hover effect on the useful button that makes the button blue and the icon scales down slightly with transform: scale() property. Currently in the preview branch there is also a hint of yellow in the transition, which is not desired. Other than these 2 remarks, the transition looks great!

3. Alignment with Reply Button - Dave already mentioned this and I see the issue at hand with the button moving to the bottom when expanded. However, the alignment with the reply button is really really desired and it would bring a great deal of visual harmony for the comments. Would you guys be up for one more shot of trying to see if there would be a good solution, so that the useful button is aligned, but when expanded it stays with the comment such as in the [Figma examples](https://www.figma.com/proto/XYV3fr8UVOdkNyBmKHHR7f/CP---Comments-Useful---Mobile?node-id=488-20613&t=N1CtBYo1usIAPjFP-0&scaling=min-zoom&content-scaling=fixed&page-id=48%3A79&starting-point-node-id=488%3A20613&show-proto-sidebar=1)?


Once again, really awesome work, really appreciated! <3

dalibormrska:
Awesome work @matteobu. Such a smooth implementation - and thanks @mariojsnunes for finishing it off!

davehakkens:
yeh thanks! Nice feature

matteobu:
Thanks everyone and thanks @mariojsnunes for finishing it!

PR #4381, opened 2025-08-15
feat: update modified_at when deleting content

benfurber:
When useful when debugging, like I am now. It would be useful to know when content was marked as deleted.


benfurber:
Tests going to fail as the DB is setup for profiles. Will wait for that to be merged in.

PR #4382, opened 2025-08-18
[for discussion] refactor: research update crud flow

benfurber:
## PR Checklist

## PR Type

## What is the new behavior?

Recently a reserarch update was created that didn't generate any notifications, even though the previous research update on the same research item generated the expected notifications. From the DB it looks like the content creator was having issues creating the update with multiple versions of the same update created.

One hypothesis for all is that there was an error between the research item being created and the broadcast service (that would have generated the notifications) being called - probably the file update.

This lead me to understand better the flow of events when creating and updating research updates.

Main ideas with this PR:
1. If there's an issue with any stage of creating a research update, there should be an error and any previous events should be rolled back.
2. The order of events and design patterns should be consistent and as easy to read as possible across create and update.


mariojsnunes:
@benfurber could this be closed?

PR #4410, opened 2025-09-05
eslint warnings

jproberson:
## PR Checklist

## PR Type

- [x] Eslint warnings that are not console log warnings

191 warnings 

## What is the new behavior?
109 warnings left - All console logs

- [ ] Yes
- [x] No


jproberson:
With my local setup I wasn't able to tell if the cypress test changes are going to cause problems. So hopefully the branch shows us if they will be as consistent/inconsistent as prior.

PR #4448, opened 2025-10-20
feat: update user activity

### LINKED ISSUE #4399: [feature] Update last_active on more activities
As can be seen on the loader on `src/routes/api.profile.ts`, there's a field for profile.last_active. But it's not updates for all the actions a user could take:
* creating content
* updating content
* marking content useful
* following content
* signing in
* (probably more)

Add updating last_active to these actions, as well as thinking about a testing approach for them all. Probably lots of minor updates to the existing e2e cypress tests.

Considering how many times this new feature is going to be called, a function in `profileService` should do it.

mariojsnunes:
## PR Checklist

## PR Type

User last active is only updated when viewing own profile and on sign-in.

## What is the new behavior?

Last active is now updated on all api actions (POST, PUT, DELETE).

- [ ] Yes
- [x] No

## Git Issues

Closes #4399


davehakkens:
What does on `sign-in `mean?
Each time the person logs in with credentials?

Or also when already logged in and you are browsing around while signed in?

mariojsnunes:
> What does on `sign-in `mean? Each time the person logs in with credentials?
> 
> Or also when already logged in and you are browsing around while signed in?

Yes first one.
But that was the old behavior. Now it updated when you take some action, like posting a comment, creating content, etc
Does not update when just browsing.

davehakkens:
ok cool sounds good!
And should it be live now? as the map still looks the same..

PR #4455, opened 2025-10-22
feat: auto-zoom map on location search and profile selection

### LINKED ISSUE #4451: [investigation] Zooming level for map
At the moment, if you go to the map and either, search for a location (that exists and you select) or select a profile. The map won't zoom in to a more appropriate level - it will just stay at the zoom level already selected.

It would be really useful if someone could investigate a straight forward solution to this problem.

Interlinks with #4225.

jproberson:
## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

When users search for a location or select a profile on the map (by clicking a profile card, clicking a pin on the map, or visiting a URL with a hash like  `/map#username`), the map pans to the location but does not zoom in. Users must manually zoom in to see the area clearly.

## What is the new behavior?

The map now automatically zooms when:
  - Searching for a location via the search bar
  - Clicking a profile card in the list
  - Clicking a profile pin on the map
  - Loading a page with a profile hash in the URL (e.g., `/map#user`)

This uses Leaflet's native `setView()` method to update both center and zoom, preventing race conditions. A constant `PROFILE_ZOOM_LEVEL = 12` is defined in `MapContext.ts`. This value can be adjusted, for the locations / test data I have, it felt okay. Just let me know and I can update it accordingly.

- [ ] Yes
- [x] No

## Git Issues

Closes #4451


mariojsnunes:
it is zooming too much if you search for a country.
it's ok for a city though.
maybe we could detect it and adjust the zoom?

davehakkens:
this is a very pleasant upgrade!
Agree country zoom is too much/deep.

jproberson:
@mariojsnunes @davehakkens Thanks for the feedback on that. I have an idea.

It looks like there are a couple ways to approach it. There is a type field on the response that can describe what was returned, but some searching around seems like it could be inconsistent depending on how places are marked.

A simpler approach might be to just consider the bounding box returned and set an appropriate zoom based on that.

src/pages/Maps/utils/adaptiveZoom.ts
```diff
@@ -0,0 +1,66 @@
...
+  SMALL_COUNTRY: 2,
+  LARGE_STATE: 1,
+  STATE: 0.5,
+  CITY: 0.2,
+  TOWN: 0.1,
+  NEIGHBORHOOD: 0.05,
+  STREET: 0.01,
+}
+
+export const getAdaptiveZoomLevel = (
```
mariojsnunes: I understand the constants, but wondering how it would look like with just a math.round and a min/max calculation

jproberson:
Yeah, I like the idea of simplifying. It worked well but maintaining it didn't seem fun as some of the chosen values were just arbitrary. Let's try just using the fitbounds functionality with the bounds that come back, let me know how that feels. It seems close enough for most searches I tried.

mariojsnunes:
it's perfect now!

benfurber:
@jproberson Such nice work, thank you.

PR #4456, opened 2025-10-22
feat: add followed button display to content list views

### LINKED ISSUE #4442: [ui] Show what content a user follows
Image

Currently it is not possible to keep track of what items you chose to follow. This feature would allow the user to have a quick overview when scrolling of which items are familiar (which ones they already interacted with)

- Adding the yellow follow icon to the items that the user follows to Questions, Library and Research modules.
- Find a good solution also for mobile
- If needed, squeezing the group of metrics and this icon a bit tighter together on desktop, so that there is still enough space for the title and the Category label.
- Test it on all the screens - big desktop, small desktop and mobile.

## Tech notes

This should be straightforward to implement as `FollowButtonAction` exists and should be used.

There's some feature tests for the following functionality, e.g. `/packages/cypress/src/integration/research/follow.spec.ts`. Please implement a consistent set of follow tests for the three content types.

jproberson:
## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Doesn't show what you are following in the list views.

## What is the new behavior?

After following a project/article/question, the follow icon is shown in their respective list views for the item.
If there are icons shown in mobile / tablet view this icon is also shown. 

image

The styling to make the FollowButtonAction feel similar to the other icons was a little awkward. I had to override some upstream Button styling that conflicted with the icon-only display we needed downstream. It might be good to eventually refactor this into its own "display-only" variant. However, for now this felt like an okay first pass.

- [ ] Yes
- [x] No

## Git Issues

Closes #4442

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -98,7 +98,7 @@ export const Button = (props: BtnProps) => {
       {...sanitizedProps(props, ['small', 'large', 'showIconOnly', 'iconColor', 'iconFilter'])}
       sx={{
         ...getSizeProps(size, !!props.icon),
-        ...(props.showIconOnly ? { pr: 0 } : {}),
+        ...(props.showIconOnly ? { pr: 0, WebkitTapHighlightColor: 'transparent' } : {}),
         ...props.sx,
       }}
     >


davehakkens:
Nice PR. Sorry for the slow review on this one! Having an internal discussion around this one.
Will get back to it soon

davehakkens:
hey @jproberson 
Little update on this.
As the 'follow' got a bit mixed with following content vs comments (as it was written like that in the issue.)
However it was only intended for following content updates from the author, not comments.

-----
So for now we want to reduce the scope only release this for the following content. 
Which means it will only be visible on the research module. 
Meaning: If person clicks the 'follow' button. They follow the research topic and it should be visible.
image


We might latest extend the feature to discussion/comments as well.
But this would require some more design work first.

dalibormrska:
Great work @jproberson. I don't see any flaws with this one, spacing looks great.

If I were to look really hard for things to pinpoint, it would be that on mobile view there is this small blue hover rectangle that appears on tap. It doesn't really have a function. So the improvement would be not to show this blue rectangle, and instead we could open the tooltip on tap?

But if it's a hassle to do, I'm fine with letting it pass as it is. :))

image

dalibormrska:
@jproberson just letting you know that the blue background still persists on live site. 🤔 
image

jproberson:
@dalibormrska  Interesting, I don't  see this if I use the dev tools and using the `Responsive Design Mode` but I do if I just manually shrink my browser down / actually on my phone. What a silly thing lol. 
I'll take a look at fixing it.

PR #4478, opened 2025-11-05
feat: just pan on pin click

### LINKED ISSUE #4225: [ui] Adjusting zoom level on the map when clicked on a profile
In the map when clicking on a profile, the zoom level is not adjusted. This results in not really clear, since it is somewhere in the cluster.
Image

This happens whenever the url of the map has the profile in it (https://community.preciousplastic.com/map/#precious-plastic), so in the map page but also when the user clicks on the country button in the profile.
Image

Adjust the zoom level when clicking on a profile, so that the pin is not in the cluster. My suggestion would be to have also a minimum zoom so that it is clear to see the region and the country, even if it is not in a cluster by default.

So instead of seeing the whole world even if the pin is alone,
Image

have a minimal zoom level to see the country more up-close.
Image

**Should have**
Could the zoom transition be done smoothly? Or is Leaflet not made for that?

**Additional context**

jproberson:
## PR Checklist

## PR Type

- [ ] Bugfix (fixes an issue)
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

We do the big zoom on pin click.

## What is the new behavior?

We pan on pin click.

- [ ] Yes
- [x] No

## Git Issues

Closes #4225

### DIFF
--- a/src/pages/Maps/Maps.client.tsx
+++ b/src/pages/Maps/Maps.client.tsx
@@ -111,7 +111,7 @@ const MapsPage = () => {
   ])
 
   useEffect(() => {
-    if (selectedPin && allPins && allPins.length > 0) {
+    if (selectedPin && allPins && allPins.length > 0 && boundaries) {
       const isPinStillVisible = filteredPins.some(
         (pin) => pin.id === selectedPin.id,
       )
@@ -119,7 +119,7 @@ const MapsPage = () => {
         selectPin(null)
       }
     }
-  }, [filteredPins, selectedPin, allPins])
+  }, [filteredPins, selectedPin, allPins, boundaries])
 
   useEffect(() => {
     const init = async () => {
@@ -220,14 +220,15 @@ const MapsPage = () => {
     if (allPins && username) {
       const foundPin = allPins.find((pin) => pin.profile!.username === username)
       if (foundPin) {
-        if (selectedPin?.profile?.username !== username) {
+        const isPinVisible = filteredPins.some((pin) => pin.id === foundPin.id)
+        if (isPinVisible && selectedPin?.profile?.username !== username) {
           selectPinAndHandleCluster(foundPin)
         }
       } else {
         selectPin(foundPin)
       }
     }
-  }, [location.hash, allPins])
+  }, [location.hash, allPins, filteredPins])
 
   return (
     <MapContext.Provider


davehakkens:
@dalibormrska can be reviewed

dalibormrska:
Wow, kudos for such a fast implementation @jproberson ! The experience is much smoother now, nice!

One more thing I found while testing... So currently clicking on the pins works really nice, but when I click on the card from the list, we still go to the max zoom level. Would it be possible to apply the same logic also when clicking on a profile card from the list and also when clicking on the button from the profile (the common denominator is that it adds the `#user-name` to the URL)?

Attaching a GIF recording.

Pin feedback

jproberson:
@dalibormrska No problem lets give it a shot. Let me know if this does what you were looking for.
What you were seeing was a little bit of a relic I added recently. Previously we landed on using a default zoom value, but that was before we started doing other things to fit to the bounds that were searched. So hopefully this is closer.

@davehakkens One thing it's not doing at the moment is trying to _center_ (we can 100% center if you want)
It just felt a bit natural the way it's staying fit the the current view. (It is also closer to the experience you posted in the gif from WWOOF)

jproberson:
Looks like I may have a test to update still, but I'd like to confirm the functionality is as we expect. 
Interestingly, the preview branch is behaving differently for me than if I run the code locally.
Do the preview branches go stale on new builds, or do the rebuild everything?

dalibormrska:
@jproberson thanks for the fast response! Indeed it's weird if the preview branch behaves differently than your local env. And so for me if I open the deployed preview, I can't see the changed behaviour.

Maybe you can record the a gif or a short video of the behaviour? :)) 

EDIT: Thanks Mario, just needed to remove and add the label.

jproberson:
Thank you @mariojsnunes the recycling did the trick.
Let me know @dalibormrska and if we're happy I'll start digging into fixing that integration test.

dalibormrska:
Awesome! Yeah the behaviour is spot on I think. No need for centering, this works really well. From my side there is no more blocking feedback, but when I was testing I came across a bug/glitch that happens when i drag the map with a bit faster. You can see it in the gif. If it is easy to fix - great, if not, it's not a big deal. :))

Pin feedback 2

jproberson:
I retract the non-centering comment I made earlier. I just tried out a change to fix some odd behavior that happens in the live version today, where if you pan the current selection off the screen, it tries to bounce back. I think centering is the safest way to ensure the pop-up is on screen when selected.

dalibormrska:
Looks perfect to the last detail!
# 🥹

mariojsnunes:
awesome work!

PR #4497, opened 2025-11-19
feat:comment sort component with tests

### LINKED ISSUE #4494: [feature request] Sort comments by useful
## Description and problem statement

We have recently introduced the marking as useful feature for comments, and we want to elaborate on it. Making it possible to sort the comments by which ones were marked as most useful would be a great feature to have.

## Proposed solution

- Utilising the sort dropdown component that is used on other modules, but without white background
    - Shrinking its width on desktop
    - On mobile positioning it below the Comment amount and the button, the sort gets full width here
- The gap is 10px between the buttons
- The options would be newest, oldest and most useful
    - We should decide on the default between newest and oldest
- We would only sort the top comments, no need to sort the nested comments.
    - (We think it's a good decision for now, I don't think it will be problematic until we have many comments in a reply thread and some of them being really useful. When we start to observe that this is the case, we can think of other ways.)

Image

jproberson:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

Ability to sort comments by newest,oldest, most useful

- [ ] Yes
- [x] No

## Git Issues

Closes #4494

### DIFF
--- a/packages/cypress/src/integration/research/discussions.spec.ts
+++ b/packages/cypress/src/integration/research/discussions.spec.ts
@@ -27,7 +27,8 @@ describe('[Research.Discussions]', () => {
     cy.step('Create admin user session and add three comments');
     cy.signIn(admin.email, admin.password);
     cy.visit(researchPath);
-    cy.get('[data-cy="HideDiscussionContainer:button"]').first().click();
+    //cy.get('[data-cy="HideDiscussionContainer:button"]').first().click();
+    cy.get('[data-cy="HideDiscussionContainer:button"]').eq(1).click();
 
     cy.addComment(comment1);
     cy.wait(1000);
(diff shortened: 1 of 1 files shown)


davehakkens:
some bugs on this @mariojsnunes @jproberson 
1. dropdown not visible
image

2. Button overlap
IMG_0658
image 




3. More of a feature.
Would be nice to hide this dropdown if comments are less then 5 (?). As it takes quite some space on mobile but is a bit useless when there are only a few comments.

jproberson:
@davehakkens I expected the first one, as it's its more of a existing thing with the existing styles, however I do think your idea here would probably be the easiest way to avoid it. I can get a fix for that.

> Would be nice to hide this dropdown if comments are less then 5 (?). As it takes quite some space on mobile but is a bit useless when there are only a few comments.

For the second point you had that's interesting! My phone views / my phone itself doesn't have that issue.
But I can play around with more break points to try and clean that up. 
What screen size or break points did you have to see that? This was my phone this morning.

davehakkens:
Iphone SE. but also also see the problem on this topic. Perhaps because it has a 3 digits comment count? 

https://community.preciousplastic.com/news/precious-plastic-is-in-trouble

jproberson:
Interesting, I did not test it on that many of comments. I am going to change it wrap when needed on my branch instead of what it's doing today.

PR #4502, opened 2025-11-21
feat: Show who marked content as useful

### LINKED ISSUE #4443: [front end] Show who marked content as useful
Image

Image

It's nice for content creators especially to see specifically who has marked their content as useful and more generally adds to the life of the platform.

## Tech notes

`ContentStatistics` needs to be updated to accomidate the action. Ideally though an optional prop added to the existing statistics field, then if present, it opens the component provided to that field. Therefore the modal functionality is added to `ContentStatistics` but what component is rendered is provided to `ContentStatistics`

With this extra complexity being added, it might be worth pulling out the statistics.map render to a new component.

New db query will be needed as we only get the useful count at the moment.

Yes to feature testing please! Might be worth pulling out into a new spec file.

[Design file here](https://www.figma.com/design/uqu0iljGbuNdMbY0pE3H00/-PP--Library?node-id=25-11709&t=ydJMjkpq0h8JqT4f-1)

chris-staunton:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

- Created a new component that is a list of profiles.
- If a piece of content has 1 or more useful votes, the stat count will be clickable
- Once clicked, it will fetch the users that have 'useful voted' for this piece of content.
- Users will see the list of profiles.
- Note: UI is slightly different from design files because recent changes replaced the stat label with just a stat number.

image

image

- [ ] Yes
- [ ] No

## Git Issues

Closes #4443

### DIFF
--- a/src/pages/Library/Content/List/ProjectCard.tsx
+++ b/src/pages/Library/Content/List/ProjectCard.tsx
@@ -77,7 +77,7 @@ export const ProjectCard = ({ item, query }: ProjectCardProps) => {
           <Flex sx={{ justifyContent: 'flex-end' }}>
             {item.category && (
               <Flex sx={{ flex: 1 }}>
-                <Category category={item.category} />
+                <Category category={item.category} sx={{ color: 'black' }} />
               </Flex>
             )}
 


davehakkens:
Nice one @chris-staunton!
@analordello @dalibormrska this can be tested 🍿

dalibormrska:
Awesome work @chris-staunton. One small thing. Could we make the whole div with the cursor:pointer? So that it is a bit more obvious that it is clickable. :))

image

chris-staunton:
> Awesome work @chris-staunton. One small thing. Could we make the whole div with the cursor:pointer? So that it is a bit more obvious that it is clickable. :))
> 
> image

For sure! Just pushed an update now.

packages/components/src/ProfileList/ProfileList.tsx
```diff
@@ -0,0 +1,123 @@
...
+      onClick={onClose}
+    >
+      <Box
+        sx={{
+          bg: 'background',
+          borderRadius: '10px',
+          width: ['80%', '23%'],
+          height: 'auto',
+          maxHeight: ['40%', '50%'],
+          border: '2px solid',
```
mariojsnunes, line 37: need `overflow: hidden;` here so that the scroll bar doesn't overflow the border radius.

chris-staunton: Nice catch, fixed now!

davehakkens:
Nice feature and fun to check! But noticed on livesite it doesnt work for the research module
image

chris-staunton:
> Nice feature and fun to check! But noticed on livesite it doesnt work for the research module image

Will look into this ASAP, I think it’s a data issue cause it only happens to some. Thanks for the catch!

analordello:
@mariojsnunes Could you change the gap between the buttons and the counters to 15px on mobile for a larger space between clickable areas?
image

Also, I noticed you added the underline in the number, which is fine. could you also add the following hover interaction of adding color to the icon? Let me know if it is technically challenging
image

mariojsnunes:
@analordello added the gap.
But the color on hover would require a refactor of that statistics component.

mariojsnunes:
@chris-staunton I found the issue, it's related to the useful table data on prod. Deploying the fix soon.

PR #4505, opened 2025-11-25
feat: donate button

### LINKED ISSUE #4474: [feature request] Donate Button for spaces
Adding donate button on the following pages:
- Space profile (workspace, machine builder, community point, collection point)
- Library posts

When clicking on the donate button, a Donorbox modal will be opened

[Figma link](https://www.figma.com/design/uqu0iljGbuNdMbY0pE3H00/Library---Research-Tidy-up?node-id=145-4056&p=f&t=FTws9UXSDXLWeS6j-0)

[Userflow link](https://www.figma.com/board/o2R1ihNF2MLswPtXuL2yKa/CP-Next-Roadmap?node-id=1127-2723&t=AVMAlqyh4XK0BPZZ-0)

Profile:
Image

Library / research posts:

Image

mariojsnunes:
## PR Checklist

## PR Type

- [x] Feature

## What is the new behavior?

Allows space profiles to display a donorbox donate button

image
image
image

- [x] Yes

## Git Issues

Closes #4474


packages/components/src/Icon/DonateIcon.tsx
```diff
@@ -0,0 +1,18 @@
+export const DonateIcon = () => (
```
benfurber, line 1: This pattern isn't quite right. This should be a svg file...

packages/components/src/Icon/Icon.tsx
```diff
@@ -64,6 +65,7 @@ export const glyphs: IGlyphs = {
   category: iconMap.category,
   comment: iconMap.comment,
   'comment-outline': iconMap.commentOutline,
+  donate: <DonateIcon />,
```
benfurber, line 68: ...Then this `donate: iconMap.donate,`...

server.js
```diff
@@ -89,6 +89,7 @@ app.use(
         '*.analytics.google.com',
         '*.google-analytics.com',
         'www.youtube.com',
+        'donorbox.org',
```
benfurber, line 92: I'm suprised this is needed and/or wasn't there before?

src/common/DonationRequestModalContainer.tsx
```diff
@@ -0,0 +1,53 @@
+import { useEffect, useState } from 'react';
```
benfurber, line 1: Annoying question time. Is this a container or a wrapper?

src/common/DonationRequestModalContainer.tsx
```diff
@@ -0,0 +1,53 @@
...
+    if (props.isOpen && !settings) {
+      fetchSettings();
+    }
+  }, [props.isOpen, settings]);
+
+  return (
+    <>
+      {settings?.campaignId && (
+        <DonationRequestModal
+          spaceName={settings.spaceName}
```
benfurber, line 41: As the data model type is Author. Should this be `authorName` rather than `spaceName`?

src/pages/Library/Content/Page/ProjectPage.tsx
```diff
@@ -180,11 +165,34 @@ export const ProjectPage = observer(({ item }: ProjectPageProps) => {
...
+                        category: 'projects',
+                        label: item.author?.username || '',
+                      });
+                      setIsDonationModalOpen(true);
+                    }}
+                  >
+                    Support the author
+                  </Button>
+                </>
+              )}
```
benfurber, line 193: Even though it's only in two places currently, I think most of this should be a common component to prevent future regressions.

benfurber (changes requested):
Deffo needs some feature tests please:
1. Adding a donation link, 
2. Opening the donation modal on each content type.
3. Removign the donation link.
4. Donation button gone on each content type.

packages/components/src/Icon/DonateIcon.tsx
mariojsnunes: That was my first approach, didn't work for some reason.
I can try it again.

src/pages/Library/Content/Page/ProjectPage.tsx
mariojsnunes: The "Container" has a direct dependency on the api, so it stays on  the main app.
The DonationRequestModal is on the components library.

server.js
mariojsnunes: my findings:
- I saw the error on the console, so this fixes that.
- Seems to work without the widgets script - I assume that's why it worked before.
- All examples from donorbox include the script (even those with iframe like we do) so better keep it in case it's needed for something

src/common/DonationRequestModalContainer.tsx
mariojsnunes: this should only be used by spaces, so I rather keep the naming more specific

src/common/DonationRequestModalContainer.tsx
mariojsnunes: I don't like any of them 😅
if you have a reason for either, or some other naming, let me know!

mariojsnunes:
> Deffo needs some feature tests please:
> 
> 1. Adding a donation link,
> 2. Opening the donation modal on each content type.
> 3. Removign the donation link.
> 4. Donation button gone on each content type.

I've been thinking about how to test it over the last few days. I'm favoring only testing the component itself with vitest.
e2e tests are currently my biggest time sink due to waiting the pipeline and flakyness, so I don't think they are worth it currently.

Already added the vitest tests!

packages/components/src/Icon/DonateIcon.tsx
mariojsnunes: The problem was changing the icon color. I did not figure out how to change it dynamically with the approach you mentioned.

PR #4509, opened 2025-11-25
fix: prevent breadcrumbs overflow on mobile with horizontal scroll

### LINKED ISSUE #4508: [bug] Breadcrumbs are overflowing on mobile
**Describe the bug**
When there is a long title of a question, a research or other piece of content that involves breadcrumbs, and the screen is not so wide, the breadcrumbs overflow, resulting in whole screen that can scroll sideways. Not very pleasing aesthetically and quite unprofessional. Feels broken.

**To Reproduce**

1. Take your phone and go to https://community.projectkamp.com 
2. Go to any research that has a long title, like [this one](https://community.projectkamp.com/research/can-we-renovate-a-3-rooms-old-brick-ruin).
3. Scroll sideways

**Expected behaviour**
See the gif below.
Image

I believe this could be the ideal and future-proof solution. Keep in mind that the names of the categories could be longer, or there can be more nesting of categories in the future.

So the solution would be to truncate the last item in the breadcrumbs chain until some given min-width. In the gif I put 120px min width. The categories would not need to be truncated. When it happens that we hit the truncation minimum, meaning the screen width is now smaller than the component width, we would apply horizontal scrolling.

Is this feasible to implement? If you believe it's too complicated, what would be an alternative fix? 

**Additional context**
It is possible that the truncation was coded before, but for some reason does not work now. Check the code of the component first.

manojalluri:
## Summary

This PR fixes the breadcrumbs overflow issue on mobile devices by adding horizontal scroll support to the Breadcrumbs component.

## Changes Made

- Add `overflowX: 'auto'` to the Breadcrumbs Flex container to enable horizontal scrolling when content exceeds screen width
- Set `minWidth: '120px'` for the last breadcrumb item to ensure readable truncation while maintaining a minimum visible width

Fixes #4508

## PR Checklist

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

When breadcrumbs contain long titles on mobile devices, instead of causing the whole screen to scroll sideways, the breadcrumbs container now scrolls horizontally within itself. The last breadcrumb item maintains a minimum width of 120px for readability.

No

## Other information

This fix was suggested in the issue - truncating the last item with a minimum width and applying horizontal scroll to the container when the screen width is smaller than the component width.


dalibormrska:
Wow @manojalluri, really quick response, thank you for that!
I still see that it only solved the issue partially. Yes, currently the if doesn't overflow on most mobile devices. But...

1.  If the category name would be longer (which it will most likely be in the near future), then we still get the overflow as you can see on the picture.
image

2. If we open it on a tablet, or a wider phone, we see that the truncation is stuck on a certain width, even though it has space to expand.
image

mariojsnunes:
heey @manojalluri, just checking in
do you think you will be able to look into @dalibormrska points above?

dalibormrska:
Closing this PR, this issue got tackled by @jproberson [here](https://github.com/ONEARMY/community-platform/pull/4557)

PR #4517, opened 2025-11-28
use icon only for follow button if showIconOnly

jproberson:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

Removing any hover / click effects from the following icon when it's display only in the research list view.

- [ ] Yes
- [x] No

## Git Issues


jproberson:
@dalibormrska when this preview is up and running can you double check it's behaving as you would expect? 
Thanks again for your detailed testing of this!

dalibormrska:
Awesome! No need to add the tooltip for mobile now, because then we would need to do that also for all the other icons - out of scope for now, maybe in a separate issue. 

So all good, ready to merge! :))

PR #4520, opened 2025-12-01
feat: map settings improvements

### LINKED ISSUE #4516: Improve UX for adding a map pin
**Describe the bug**
We received feedback from users that the process of adding a map pin was unintuitive - they contacted the support, because the process of adding a pin to the map required a doubleclick which was not obvious.

**Try it yourself**

1. Go to https://community.preciousplastic.com/settings/map
2. Or to one of the active pr-previews such as https://community-platform-pr-4502.fly.dev/map
2. Go the the map tab in the settings
3. Try to add the pin
4. See that it requires a doubleclick to add a pin to the map
5. See also that it's not so obvious that the pin is draggable

**Expected behaviour**
1. Single-click instead of a double-click. Wherever I click on the map the pin should move there. Whether it's in a case of adding it first time or moving the existing pin.
2. When hovered on the existing pin, change the cursor to `move`. This will suggest that it's draggable.
3. If the conditions 1. and 2. above are fulfilled, then the experience should be much more intuitive, which means we can remove the blue info box.
4. Change the input placeholder to "Type your address"

**Screenshots**
Image

Image

Image

**Additional context**
I noticed that there are 3 different versions of those map pins, the purple one, the member icon and the new icon. I don't know why this is as it is, but make sure the changes are implemented in all scenarios.

mariojsnunes:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

_Describe the new behaviour_

- [ ] Yes
- [x] No

## Git Issues

Closes #4516

## What happens next?

### DIFF
--- a/src/pages/UserSettings/SettingsPageMapPin.test.tsx
+++ b/src/pages/UserSettings/SettingsPageMapPin.test.tsx
@@ -44,6 +44,8 @@ describe('SettingsPageMapPin', () => {
       profile: mockUser,
       update: vi.fn(),
     });
+    mockGetCurrentUserMapPin.mockResolvedValue(null);
+
     // Act
     let wrapper;
     act(() => {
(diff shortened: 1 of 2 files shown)


mariojsnunes:
@dalibormrska yeah it's better showing both pins below, please check!

dalibormrska:
Awesome. I personally find the text saying "Preview the new pin" a bit confusing. I would change it to something like "Your updated map pin", or "Your chosen location". What do you think?

mariojsnunes:
@dalibormrska yep changed the text and merged

PR #4524, opened 2025-12-02
feat: upgrade badge

### LINKED ISSUE #4482: Add "Go Pro" Button
Objective: Bring more visibility to becoming a pro member.

jproberson:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

image

- [x] Yes
- [ ] No

## Git Issues

Closes #4482

## What happens next?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -209,7 +209,7 @@ jobs:
             docker system prune -a
       - run:
           name: Install fly command
-          command: curl -L https://fly.io/install.sh | sh
+          command: curl -L https://fly.io/install.sh | sh -s -- -y
       - run:
           name: Add fly to PATH
           command: echo "export PATH=\"/home/circleci/.fly/bin:$PATH\"" >> $BASH_ENV
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+storybook.onearmy.earth
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -120,4 +120,4 @@ We always welcome UX/UI design contributions in various shapes and forms. Whethe
 
 The best way to start would be to look at [open design issues](https://github.com/ONEARMY/community-platform/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22Task%3A%20Design%22) here in our GitHub Repository to see if anything sparks your interest. Another way is to get in touch with us [through Discord](https://discord.gg/p4hWHYeG), the introduce-yourself channel is a good place for that. :) Then we can chat about what would be interesting for you to help out with, as well as what we are currently working on.
 
-In the meantime you can check out our [UI Component library in Storybook](https://docs.platform.onearmy.earth/storybook-static/index.html?path=/story/components-icon--default).
+In the meantime you can check out our [UI Component library in Storybook](https://storybook.onearmy.earth).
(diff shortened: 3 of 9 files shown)


shared/models/upgradeBadge.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class DBUpgradeBadge {
+  id: number;
+  action_label: string;
+  badge_id: number;
+  is_space: boolean;
+  action_url: string;
+  tenant_id: string;
+  badge?: DBProfileBadge;
+
```
mariojsnunes, line 13: add readonly to the badge as it is populated via join

shared/models/upgradeBadge.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+export class UpgradeBadge {
+  id: number;
+  actionLabel: string;
+  badgeId: number;
+  isSpace: boolean;
+  actionUrl: string;
+  badgeImageUrl?: string;
```
mariojsnunes: might be worth to have a property `badge: ProfileBadge` if later we need other properties

src/services/upgradeBadgeService.ts
```diff
@@ -0,0 +1,17 @@
...
+  try {
+    const response = await fetch(`/api/upgrade-badges`);
+    return (await response.json()) as UpgradeBadge[];
+  } catch (error) {
+    logger.error('Failed to fetch upgrade badges', { error });
+    return [];
+  }
+};
+
+export const UpgradeBadgeService = {
```
mariojsnunes: `upgradeBadgeService` lowercase

src/services/upgradeBadgeService.ts
```diff
@@ -0,0 +1,17 @@
+import { logger } from 'src/logger';
+
+import type { UpgradeBadge } from 'oa-shared';
+
+const getUpgradeBadges = async () => {
+  try {
+    const response = await fetch(`/api/upgrade-badges`);
```
mariojsnunes: I wonder if we should add a cache here on client-side as well.
Please check when it is being called, if it calls the api on every page navigation, or everytime we open the user menu, might be worth.

`await fetch('/api/upgrade-badges', { headers: new Headers({'Cache-Control': 'max-age=1800'})},`

jproberson: I decided to just go the client side route as you were suggesting. I took out some of the server side caching I was using previously here. So it'll hit the cache when you open the menu or hit the profile page.

mariojsnunes:
@jproberson with the donations PR merged, you can now add the analytics stuff

shared/models/upgradeBadge.ts
mariojsnunes: I meant the readonly on the DBUpgradeBadge badge. That's the one coming from the database. For UpgradeBadge below, is not necessary.

src/services/upgradeBadgeService.ts
mariojsnunes: I think we should always keep the api cache, it will be a significant reduction on database calls.
So both client and server is probably best.

mariojsnunes:
@davehakkens the preview is configured

dalibormrska:
Heyyy! Nice work. I went to test this out and noticed 2 things. :))

1. On mobile navbar we don't have this button.
image

2. On mobile in the profile there is flex-direction:column for the buttons. Why not row? They fit next to each other nicely and it would take less space.
image
image

Issue #4440, opened 2025-10-11
[ui] Nicer contributions tab UI


benfurber:
<img width="2026" height="1224" alt="Image" src="https://github.com/user-attachments/assets/72beb09b-d8b1-4c68-a880-da21e08aa398" />

## Is your feature request related to a problem? Please describe.

Currently the contributions tab in someone’s profile looks boring and is hard to read. There is a lot of text close to each other. This design idea shows how could it be more interesting and readable.

## Describe the solution you'd like

A new card component with a white background that includes:
* Research: Image, title, useful count, comment count, research update count.
* Library: Image, title, useful count, comment count
* Questions: Title, useful count, comment count

And don’t forget about the mobile version. It is not provided in the screenshot, but test and optimise it so that it is usable and optimised for mobiles. This would probably mean utilising a reasonable breakpoint where the layout changes from 2 columns to 1.

You'll notice:
* The screenshot (which is older) we still use the title How-to(s) which is outdated. Now we use Projects instead.
* The icons in the screenshot is out of date too. You'll see on the main list pages the icons to use.

## Extra tech details

I would suggest that we keep the title text to one-line. If it is longer we can apply [[truncating (text-overflow:elipsis)](https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/)](https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/). 

The components for this should go in the component library (`/packages/components`).

## Additional context

Look at how we currently do on a space profile that has published some research or library project. e.g. [Sustainable Design Studio](https://community.preciousplastic.com/u/sustainable-design-studio).

benfurber:
Started this and getting in the first layer of this done in #4566.

Design questions for @analordello:

1. Thinking about the meta fields to display. As we spent time thinking about them for the main page display for each content type, thoughts on displaying all of them here? Feels inconsitent otherwise.
2. Should we go for the black and white icons, again to be consistent with the main page displays.


Tech question for @mariojsnunes. I need to update the RPC for this (e.g. `get_user_projects`), I thought updating the schema file (e.g. `supabase/schemas/projects.sql`) was going to be enough to done it. What are the steps I need to complete?

mariojsnunes:
@benfurber you still need to create a diff migration file, same as before.

PR #4566, opened 2026-01-10
feat: update styling of contributions on profile

benfurber:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

The start of #4440. Basic styling updates while tech support and design questions are pending.

Before:
Screenshot 2026-01-10 at 10 55 53

After:
Screenshot 2026-01-10 at 12 50 51
Screenshot 2026-01-10 at 12 51 19

- [ ] Yes
- [x] No

### DIFF
--- a/src/pages/User/content/UserCreatedDocuments.tsx
+++ b/src/pages/User/content/UserCreatedDocuments.tsx
@@ -1,20 +1,21 @@
-import { Flex, Heading } from 'theme-ui';
+import { Flex, Grid, Heading } from 'theme-ui';
 
 import UserCreatedDocumentsItem from './UserCreatedDocumentsItem';
 
 import type { UserCreatedDocs } from 'oa-shared';
 
 interface IProps {
+  columns: number;
   docs: UserCreatedDocs;
 }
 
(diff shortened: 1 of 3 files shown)


src/pages/User/content/UserCreatedDocuments.tsx
```diff
@@ -1,20 +1,21 @@
-import { Flex, Heading } from 'theme-ui';
+import { Flex, Grid, Heading } from 'theme-ui';
 
 import UserCreatedDocumentsItem from './UserCreatedDocumentsItem';
 
 import type { UserCreatedDocs } from 'oa-shared';
 
 interface IProps {
+  columns: number;
```
mariojsnunes, line 8: would be easier to maintain if there was no `columns` configuration. Is it really needed?
There should be a way to do with `grid`

benfurber: Not sure. It's because the profile page is wider for spaces, etc. so just being explict about it seems worth doing?

PR #4576, opened 2026-01-16
fix: prevent dropdown default value loss when spamming Research link (Issue #4010)

### LINKED ISSUE #4010: [bug] Spamming Clicking "Research" causes the middle Dropdown to loose it's default
**Describe the bug**
On the Research Module of "Project Kamp", the middle (2nd) Dopdown looses it's default value, when clicking the "Research" Link in the top right very fast.
Also when you load the page there is a splitsecond, where the middle Dropdown is empty.

**To Reproduce**

1. Go to [the Project Kamp Website](https://community.projectkamp.com/)
2. Click on "Research" at the Top Right => Browser will navigate to https://community.projectkamp.com/research and after a quick while to https://community.projectkamp.com/research?sort=LatestUpdated (I think, this is where the error occurs: instead of having a default value and replacing it when there is a different value, it checks if there is a different value and only if there isn't one, it sets a default)
3. Repeatedly and quickly click on "Research" at the Top Right => Default value for the middle Dropdown is not set

**Expected behaviour**
Once you are on the research page and you click the "Research" again one of two things should happen:
A) Don't navigate at all, you are already on the research page
B) Default values should always be set when loading the research module

**Screenshots**
Clicking Research quickly and repeatedly empties the middle Dropdown

**Additional context**
For more questions, ask me :)

KostasKoukianakis:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

Fixed a race condition where the sort dropdown would appear empty when rapidly clicking the "Research" link.

**Before:** The dropdown would briefly show empty when clicking the Research link multiple times quickly, because the component rendered before the URL parameter was set.

**After:** The dropdown always shows its default value ("Latest Updated" or "Most Relevant" based on search query) immediately, even before the URL parameter is updated. No more empty state.

**Technical Changes:**
- Compute default sort value synchronously using `useMemo` to avoid race condition
- Use computed default sort when URL parameter is missing
- Ensures dropdown always has a value during component mount
- Applied fix to both `ResearchList.tsx` and `ResearchListHeader.tsx`

- [ ] Yes
- [x] No

## Git Issues

Closes #4010

## What happens next?

If you need more immediate feedback you can reach out to us on Discord in the [Community Platform development channel](https://discord.com/channels/586676777334865928/938781727017558018).

### DIFF
--- a/src/pages/Research/Content/ResearchList.tsx
+++ b/src/pages/Research/Content/ResearchList.tsx
@@ -86,33 +86,6 @@ const ResearchList = () => {
     // sort is always defined (either from URL or default), so this is safe
     fetchResearchItems();
   }, [fetchResearchItems]);
-    setIsFetching(true);
-
-    try {
-      const result = await researchService.search(
-        q?.toLocaleLowerCase(),
-        category,
-        sort,
-        status,
-        skip,
-      );
-
-      if (result) {
-        if (skip) {
-          // if skipFrom is set, means we are requesting another page that should be appended
-          setResearchItems((items) => [...items, ...result.items]);
-        } else {
-          setResearchItems(result.items);
-        }
-
-        setTotal(result.total);
-      }
-    } catch (error) {
-      logger.error('error fetching research items', error);
-    }
-
-    setIsFetching(false);
-  };
 
   const researchItemList = showDrafts ? drafts : researchItems;
 


jproberson:
Hey @KostasKoukianakis ,

Thanks for jumping in and wanting to help out, we really do appreciate it.

That said, we're going to close these PRs for now. I wanted to explain why so it doesn't feel like we're just shutting you down.

The main issue is that each PR has commits from your other branches mixed in. When we try to review one feature, we're seeing changes from unrelated work, which makes it hard to review anything in isolation.

We're a small team, and we just don't have the bandwidth to untangle overlapping changes across 6-7 PRs at once. We'd much rather work with you on one thing at a time, give you real feedback, and help you get integrated in our ecosystem.

If you want to keep contributing (and we hope you do), here's what we'd suggest: start fresh from master with one issue, keep that PR focused on just that work, and let's get it reviewed and merged before moving to the next one.

We're actively updating our contributing docs to make this clearer for everyone going forward.

Hope that all makes sense. Happy to chat if you have questions.

PR #4579, opened 2026-01-17
Feat/4545 improve news module

### LINKED ISSUE #4545: [feature request] Improving the News module
## **Is your feature request related to a problem? Please describe.**  
We plan to send out news articles natively as emails. As a preparation to that we decided to style the emails to just one column so that it is easier to translate into email layout and so that it is consistent in both mediums. 

Even though the current way the content images are enlarged looks nice, it poses more edge cases (youtube embeds as an example), making it harder to maintain.

Also a problem is that sometimes the images are detailed and contain text. In which case it would be nice to see them enlarged.

## **Describe the solution you'd like**

1. Make the content images that are currently enlarged the same width as the content.  Image

2. Introduce a feature for both title image and content images to view the full size image on click. Similarly to other modules.  Image

## **Additional considerations**
Make sure it works well on mobile.

othman-shamla:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other:

### 1. Standardized Content Width (Images & Embeds)

*   **Behavior Change**: Both **images** and **embedded content (iframes/YouTube)** are now constrained to the text container's width.
*   **Implementation**: 
    *   Updated styles in `NewsPage.tsx` for both `img` and `iframe` elements to set `maxWidth: '100%'` and `marginLeft: 0`.
    *   This resolves the specific edge case mentioned in the issue where YouTube embeds could break the layout or look inconsistent.
    *   Ensures a clean, single-column layout that translates better to mobile devices and email newsletters.

### 2. Full-Screen Image Viewer (Lightbox)
*   **Scope**: This applies to both the **Header (Hero) Image** and all **Content Images**.
*   **Interaction**:
    *   Images now display a `zoom-in` cursor on hover.
    *   Clicking an image opens it in a full-screen gallery view (using `PhotoSwipe`).
    *   Users can navigate through all images in the article (header + content) as a single gallery.

- [ ] Yes
- [x] No

## Git Issues

Closes #4545

### DIFF
--- a/src/pages/News/NewsPage.tsx
+++ b/src/pages/News/NewsPage.tsx
@@ -3,12 +3,12 @@ import { Link } from 'react-router';
 import { observer } from 'mobx-react';
 import {
   Category,
+  ContentImageLightbox,
   ContentStatistics,
   DisplayDate,
   ProfileBadgeContentLabel,
   TagList,
 } from 'oa-components';
-import PhotoSwipeLightbox from 'photoswipe/lightbox';
 // eslint-disable-next-line import/no-unresolved
 import { ClientOnly } from 'remix-utils/client-only';
 import { Breadcrumbs } from 'src/pages/common/Breadcrumbs/Breadcrumbs';
@@ -27,8 +27,15 @@ interface IProps {
 
 export const NewsPage = observer(({ news }: IProps) => {
   const [subscribersCount, setSubscribersCount] = useState<number>(news.subscriberCount);
-  const contentRef = useRef<HTMLDivElement>(null);
+  const [heroImage, setHeroImage] = useState<HTMLImageElement | null>(null);
   const heroImageRef = useRef<HTMLImageElement>(null);
+
+  useEffect(() => {
+    if (heroImageRef.current && !heroImage) {
+      // need state to trigger re-render of ContentImageLightbox because ref doesn't trigger re-render
+      setHeroImage(heroImageRef.current);
+    }
+  }, []);
 
   const { profile } = useProfileStore();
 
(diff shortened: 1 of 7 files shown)


shared/mocks/data/news.ts
```diff
@@ -2,7 +2,7 @@ import type { DBNews } from '../../models/news';
 
 export const news: Partial<DBNews>[] = [
   {
-    body: 'Test info with a link to [OneArmy](https://www.onearmy.earth/).',
+    body: 'Test info with a link to [OneArmy](https://www.onearmy.earth/).\n![test-img](https://firebasestorage.googleapis.com/v0/b/precious-plastics-v4-dev.appspot.com/o/uploads%2Fproject%2F0up6oJCPP3M9bDYx34Et%2F1426018318_414579695-17fcd6de5f7.jpg?alt=media&token=9f18315b-a1ad-410b-b31c-2161fb1d7142)',
```
mariojsnunes: don't use a firebase link. We migrated away from firebase.

othman-shamla: By the way, I got that link from other test data. I used it as a temporary solution for the logo because I couldn't find a way to upload images to Supabase staging yet. The test will still pass even if I change the link to a #, as the image doesn't actually need to render for this test.

dalibormrska:
Heyy and thanks! Tested it out, works as expected. :))

I have one request about the zoom-in cursor on hover. In other parts of the platform (like the research, library or questions), if there is an image that is clickable to open the preview we don't use the cursor zoom-in but the classical pointer instead.

So, I would also make the cursor pointer here in the news for the sake of consistency. :))

src/pages/News/NewsPage.tsx
```diff
@@ -102,13 +179,13 @@ export const NewsPage = observer(({ news }: IProps) => {
               h6: { fontSize: 2 },
               img: {
                 borderRadius: 2,
-                maxWidth: ['105%', '105%', '120%'],
-                marginLeft: ['-2.5%', '-2.5%', '-10%'],
+                maxWidth: '100%',
+                marginLeft: 0,
```
benfurber: Are these actually needed? What happens if the maxWidth and marginLeft are just removed?

benfurber:
Hi @othman-shamla, thanks so much for this!

I'm afriad I'd missed previously the lightbox request(!) that us maintainers should have provided more architectural guidance on. 

I'll belatedly do that now if that's OK so it does extend the issue but I don't want to merge in the current approach. Again, totally the mistake of us maintainers not noticing. 

The main thing is that it should be a consistent approach for all images on the platform. So the new image light box should be a seperate component to do the 'useEffect', etc. and isn't on `src/pages/News/NewsPage.tsx`. It should probably share the lightbox code that `ImageGallery` uses.

shared/mocks/data/news.ts
mariojsnunes: yeah we still need to clean those :)

dalibormrska:
Apart from the zoom-in cursor comment I had above, everything seems fine, works well. :)) Thanks!

mariojsnunes:
Thanks @othman-shamla, this was neat contribution!

@all-contributors add @othman-shamla for code

PR #4582, opened 2026-01-18
feat: update profile contributions display

### LINKED ISSUE #4440: [ui] Nicer contributions tab UI

benfurber:
## PR Checklist

## PR Type

- [x] Feature

## What is the new behavior?

Screenshot 2026-01-31 at 11 11 11
Screenshot 2026-01-31 at 11 12 06

- [x] Yes
- [ ] No

## Git Issues

Closes #4440


benfurber:
This doesn't add all the additional metadata fields but I'm concerned this issue has already taken up a fair amount of time so I think we should close it here with this.

supabase/schemas/projects.sql
```diff
@@ -195,28 +195,6 @@ BEGIN
 END;
 $$;
 
-CREATE OR REPLACE FUNCTION "public"."get_user_projects"("username_param" "text") RETURNS TABLE("id" bigint, "title" "text", "slug" "text", "total_useful" bigint)
```
mariojsnunes, line 198: according to the migration file, this should be updated instead of removed

davehakkens:
Nice change@benfurber 
But before we saw the amount of useful. Not amount of comments.
Whats the reason for this change?
image

benfurber:
@davehakkens Useful should be there. Not sure what happened. Looking now.

PR #4583, opened 2026-01-18
chore: refactor notifications send in batch

mariojsnunes:
## PR Checklist

## PR Type

- [x] Bugfix

## What is the new behavior?

Emails are sent in batches of 100.
This should solve some emails not being sent due to hitting resend limits.

The new approach brings the resend code to our server, instead of using the edge function.

- [x] Yes

## Git Issues

Closes #

### DIFF
--- a/docs/technical-decisions.md
+++ b/docs/technical-decisions.md
@@ -83,3 +83,7 @@ Why not merge `newComment` and `newReply`?
 - They were the same at first, but the logic is different enough to warrant the separation. For instance, in a `newReply` the "parent" is a comment, and to obtain extra info (content title), need to "go up a level" twice.
 - Due to this complexity, a decision was made to cut showing the `newReply` respective content title (can be added again later if necessary).
 - This also made the logic of `newComment` more straighforward.
+
+Notification links are redirects: `/redirect?id={content_id}&ct={content_type}`
+where the `content_type` could be comments, questions, research_updates, news, questions, projects
+By using the redirect, we no longer need to generate the direct link in the "send notifications" step, which reduces complexity significantly.


supabase/schemas/notifications.sql
```diff
@@ -66,3 +66,45 @@ ALTER TABLE "public"."notifications_preferences" ENABLE ROW LEVEL SECURITY;
 
 CREATE POLICY "tenant_isolation" ON "public"."notifications" USING (("tenant_id" = ((SELECT current_setting('request.headers'::"text", true))::"json" ->> 'x-tenant-id'::"text")));
 CREATE POLICY "tenant_isolation" ON "public"."notifications_preferences" USING (("tenant_id" = ((SELECT current_setting('request.headers'::"text", true))::"json" ->> 'x-tenant-id'::"text")));
+
+CREATE OR REPLACE FUNCTION "public"."get_subscribed_users_emails_to_notify"(
```
benfurber: Isn't having this in the migration and schema duplication?

mariojsnunes: schema is only for better change review on PRs.
migrations are still needed to actually apply the new schema.

supabase/functions/send-email/index.ts
```diff
@@ -23,10 +23,6 @@ type EmailData = {
   email_action_type: string;
   notification?: NotificationDisplay;
   redirect_to?: string;
-  token?: string;
-  token_hash?: string;
-  token_new?: string;
-  token_hash_new?: string;
```
benfurber, line 29: Aren't these auth focused?

mariojsnunes: They are not being used as far as I can see.

mariojsnunes:
> My worry about the responsibility of the SQL function current is that it's hard to follow. Can't the validating the user email preferences stay a TS service focused?

Before we were doing 1 query per user (and was still an sql function because needs access to auth.users).
Could split the query into 2:
- Get the subscribers profileIds
- Get emails corresponding to those profileIds

But I don't see a benefit to that, the 2nd query especially would be longer: `select email from x where profile_id IN ({all profile ids, which could be hundreds})`

But I'll look into making it more readable.

supabase/functions/send-email/index.ts
benfurber: There's for a new password no?

packages/cypress/src/integration/news/discussions.spec.ts
```diff
@@ -157,7 +157,6 @@ describe('[News.Discussions]', () => {
     cy.expectNewNotification({
       content: updatedNewReply,
       path: newsPath,
-      title: news.title,
```
benfurber, line 77: This shouldn't be removed, it is essential to ensure that the new comment just created has generated THIS notification and we're not just checking for a previously created notification.

mariojsnunes: I removed the title from **reply** notifications due to complexity.

mariojsnunes: there must be another way to make that test then, maybe some uuid

supabase/functions/send-email/index.ts
mariojsnunes: ah yeah the `token_hash` is used there. really should get rid of that `as any`, typescript would have caught it.
anyway just changing the type wouldn't cause a real issue.

PR #4586, opened 2026-01-25
track published_at for questions, research, news, projects

### LINKED ISSUE #4479: [bug] Show Publish date not draft date
When a news article is published it shows the date that the draft created as the 'published' date
Both on the news overview ([example](https://community.preciousplastic.com/news?sort=Newest)) and single article ([example](https://community.preciousplastic.com/news/route-update-to-version-5))

Image

Image

We want to show the date the article got published.
So after hitting Publish here 

Image

jproberson:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

Now providing both dates to have full context. News page still only shows published as their may be small edits.

image

- [x] Yes
- [ ] No

## Git Issues

Closes #4479

## What happens next?


packages/components/src/DisplayDate/DisplayDate.tsx
```diff
@@ -7,39 +7,67 @@ type DateType = string | number | Date;
 
 export interface IProps {
   createdAt: DateType;
-  action?: string;
+  publishedAction?: string;
```
mariojsnunes: if this only allows 'Published' and 'Created', restrict the type.

mariojsnunes: another option would be to deduce it by comparing the dates (if reliable, I'd prefer it)

mariojsnunes (commented):
looks good to me, added a minor comment

packages/components/src/DisplayDate/DisplayDate.tsx
jproberson: @mariojsnunes we have a couple options depending on the page, Published, Started, Asked. But we can def just add a type to limit it. I'll update that. Also, I'll add a screenshot but an ask from @davehakkens  was to actually show both, minus the news page.

davehakkens:
Looks good to me!
Only remark would be that 'last edit date' is not needed to show at questions overview
image

PR #4587, opened 2026-01-27
Remove username from raw_user_meta_data

jproberson:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

_Describe the new behaviour_

- [ ] Yes
- [ ] No

## Git Issues

Closes #

## What happens next?

### DIFF
--- a/packages/cypress/src/integration/SignUp.spec.ts
+++ b/packages/cypress/src/integration/SignUp.spec.ts
@@ -9,12 +9,6 @@ describe('[User sign-up]', () => {
 
   describe('[New user]', () => {
     it('Validate sign-up form', () => {
-      cy.step('Username is too short');
-      cy.get('[data-cy=username]').click();
-      cy.get('[data-cy=username]').clear().type('a');
-      cy.get('[data-cy=consent]').uncheck().check();
-      cy.contains(FRIENDLY_MESSAGES['sign-up/username-short']).should('be.visible');
-
       cy.step('Email is invalid');
       cy.get('[data-cy=email]').click();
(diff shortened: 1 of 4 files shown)


benfurber:
@jproberson Can I get a description for this please to understand why this change is being proposed?

mariojsnunes:
@benfurber currently we store the username on the auth.users object, it was essential for the firebase migration.
But it causes 2 issues:
- in case an account exists in multiple tenant, the username must be the same across them
- harder to change the username

It is also one more thing adding complexity now that we are also looking into creating accounts without profiles (in a first step, for the memberships flow)

This PR is to investigate if we can safely get rid of it, or there is some dependency.

mariojsnunes:
@jproberson there are still some references to `user_metadata`

`ensureProfile()` is how we allow seamless login to each tenant if the user has already registered once.
Need an alternative to that, either:
- have a UI to prompt the user for a username
- create the profile without the username - need to see what are the edge cases here
- get the username from another tenant (via supabase admin rpc)

`send-email` edge function:
- fetch the profile via auth_id, or
- change the email message to not include the `username`

fetch a research item with updates - it's required due to `collaborators` being stored as a list of usernames:
- just make an additional query to the current user profile to get the username, or
- migrate the collaborators to use `ids` instead - might be worth doing eventually

search projects - to search by author:
- just make an additional query to the current user profile to get the username

jproberson:
remade the branch to avoid some rather interesting conflicts. closing.
#4684

PR #4608, opened 2026-02-08
fix: notifications settings checkbox sizes

mariojsnunes:
## PR Type

- [x] Bugfix

## What is the new behavior?

before:
image

after:
- bigger, easier to click
- on hover pointer
image


davehakkens:
Nice

PR #4609, opened 2026-02-08
feat: improve update copy clipboard

mariojsnunes:
## PR Checklist

## PR Type

- [x] Feature

## What is the new behavior?

The current behavior actually sets the page url to the link and scrolls to it.
It defeats the purpose of having a "Link copied" message.

The new behavior:
- no longer "navigates" to the update link.
- changes the icon to a green checkbox for 2 seconds for better user feedback.

https://github.com/user-attachments/assets/3fd5edfc-864b-4054-8854-f402cd2b3d4e

Notes:
- There is still a small flicker of the tooltip changing, I couldn't solve it.
- Could explore other approaches: having a snackbar (new component for the platform) or modal with a readonly input with the link.

### DIFF
--- a/src/pages/Research/Content/ResearchLinkToUpdate.tsx
+++ b/src/pages/Research/Content/ResearchLinkToUpdate.tsx
@@ -15,7 +15,10 @@ export const ResearchLinkToUpdate = ({ research, update }: IProps) => {
   const [showCheck, setShowCheck] = useState(false);
 
   const copyURLtoClipboard = async (slug: string, id: number) => {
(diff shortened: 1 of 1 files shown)


mariojsnunes:
@dalibormrska opinions on this?
If a different solution would take a while, I think we should still merge it, as the current behavior doesn't make sense to me.

dalibormrska:
Looks good to me as a quick fix.

Long-term I see that the snackbar component could be introduced, since now there is no feedback for example for copying the comment link.

PR #4615, opened 2026-02-14
feat: add news notifications

benfurber:
## PR Checklist

## PR Type

- [x] Feature

## What is the new behavior?

Screenshot 2026-03-30 at 10 52 19

Screenshot 2026-03-30 at 10 56 24

This adds the generating of all app notifications to **all users**.

- [x] Yes
- [ ] No

## (At least) Two design considerations

1. The layout of the notification - uses the existing pattern that it's the user that generated a notification, but news appears as if it's been generated by the platform. How should this be layed out?
2. I've included the email preference already. Can either hide it for the moment, or change the text to read something like "we don't email about news yet, but will be soon."

### DIFF
--- a/src/services/notificationsSupabaseService.server.ts
+++ b/src/services/notificationsSupabaseService.server.ts
@@ -1,6 +1,7 @@
 import type { SupabaseClient } from '@supabase/supabase-js';
 import type {
   DBComment,
+  DBNews,
   DBProfile,
   DBResearchItem,
   ResearchUpdate,
@@ -12,19 +13,34 @@ import { NotificationEmailServiceServer } from './notificationEmailService.serve
 
 export class NotificationsSupabaseServiceServer {
   constructor(private client: SupabaseClient) {}
-
+  
   async getSubscribedUsers(
     contentId: number,
     contentType: SubscribableContentTypes,
   ): Promise<SubscribedUser[]> {
     try {
-      const { error, data } = await this.client.rpc('get_subscribed_users_emails_to_notify', {
-        p_content_id: contentId,
-        p_content_type: contentType,
-      });
+      let data;
 
-      if (error || data.length === 0) {
-        throw error || new Error('No emails to send');
+      if (contentType === 'news') {
+        const response = await this.client.from('profiles').select(
+          `id,
+          roles,
+          badges:profile_badges_relations(
+            profile_badges(id)
+          )
+          `
+        );
+        data = response.data && response.data.map(({ id, roles, badges }) => ({ 
+          badge_ids: badges ? badges.map(({profile_badges}) => profile_badges).flat().map((badge) => badge.id) : [],
+          profile_id: id,
+          roles,
+        }));
+      } else {
+        const response = await this.client.rpc('get_subscribed_users_emails_to_notify', {
+          p_content_id: contentId,
+          p_content_type: contentType,
+        });
+        data = response.data;
       }
 
       return data as SubscribedUser[];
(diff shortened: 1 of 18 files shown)


mariojsnunes:
cool! make sure this uses the batch send

benfurber:
@mariojsnunes How will I know if I am/am not using the batch function? Though I don't really need to worry about that for a while.

benfurber:
I realising too that we're going to want to batch in the SQL updates too...

mariojsnunes:
> @mariojsnunes How will I know if I am/am not using the batch function? Though I don't really need to worry about that for a while.

Make sure you use the src/.server/resend sendBatchEmails

mariojsnunes:
> I realising too that we're going to want to batch in the SQL updates too...

Not sure what you mean by batching SQL updates

benfurber:
@mariojsnunes As for PP, creating a new news article would generate ~40k new notification records, it seems sensible to ensure the insert command is efficient. And very much not my thing(!) Looks like [this](https://medium.com/@mcansener/efficiently-inserting-large-datasets-into-sql-server-cc944d167589) has some good ideas.

benfurber:
@davehakkens and @analordello - This is the branch where I've started to create the core functionality to send users notifications. This is the first step where the database creates the core notification and shows it to users on the app. Ready for your playing please.

src/services/notificationsSupabaseService.server.ts
```diff
@@ -112,14 +128,32 @@ export class NotificationsSupabaseServiceServer {
...
-          status: 500,
-          statusText: 'Error creating notifications: Comments',
-        },
-      );
+  async createNotificationsNews(news: DBNews) {
+    try {
+      const contentId = news.id;
+      const contentType: SubscribableContentTypes = 'news';
+
+      let subscribers = await this.getSubscribedUsers(contentId, contentType);
```
mariojsnunes: small thing, I'd pass the `'news'` directly `await this.getSubscribedUsers(contentId, 'news');` because the input is already typed.

src/stores/Subscription/subscription.store.tsx
```diff
@@ -200,7 +214,11 @@ export const SubscriptionStoreProvider = ({ children }: { children: React.ReactN
     }
   }, [profile]);
 
-  return <SubscriptionStoreContext.Provider value={subscriptionStore}>{children}</SubscriptionStoreContext.Provider>;
```
mariojsnunes, line 203: are you using the Biome formatter?

CTRL+SHIFT+P

Image

Image

supabase/migrations/20260328093904_add_news_notifications.sql
```diff
@@ -0,0 +1,5 @@
+alter table "public"."notifications" alter column "action_type" set data type text using "action_type"::text;
```
mariojsnunes, line 1: changing this to text? because it's hard to maintain?
Having types gives a lot of safety...

mariojsnunes (commented):
Looking good!

supabase/migrations/20260328093904_add_news_notifications.sql
benfurber: I get the safety argument but it's a pain changing them everytime and it feels like overkill for us atm. Hopefully we end up in a regular flow of adding notification types for a while.

src/stores/Subscription/subscription.store.tsx
benfurber: Set as my default formatter now.

PR #4629, opened 2026-02-27
fix: truncate long usernames on map user cards

### LINKED ISSUE #4598: [bug] Usernames on user cards in Map module are overflowing
**Describe the bug**
In the Map module on the left side we have the user cards. If an user has a long username, it is sometimes overflowing, not properly truncated with ... and importantly the PRO badge is therefore not showing.
Image

**To Reproduce**

1. Go to 'https://community.preciousplastic.com/map'
2. Find long username on the list on the left side
3. See error

**Expected behaviour**

The expected behaviour is that if the username is longer than the card, it gets truncated with ... but the PRO badge would always be visible. As is the case with the image below.
Image

GMetaxakis:
## PR Checklist

CSS-only fix (adding `minWidth: 0` and `flexShrink: 0` to existing flex containers). Existing tests pass. Verified visually.

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

Long usernames on map user cards are now properly truncated with `...` and the PRO badge always remains visible.

The `<Text>` element already had `text-overflow: ellipsis` but flex parent containers had the default `min-width: auto`, preventing them from shrinking below content size. Fixed by:

- Adding `minWidth: 0` to the Username flex container and its InternalLink wrapper
- Adding `minWidth: 0` to CardDetailsMemberProfile and CardDetailsSpaceProfile parent containers
- Adding `flexShrink: 0` to UserBadge so badge icons are never compressed

- [ ] Yes
- [x] No

## Git Issues

Closes #4598


mariojsnunes:
@all-contributors add @GMetaxakis for code

dalibormrska:
@GMetaxakis Legend! Thank you! 🎊

PR #4641, opened 2026-03-01
fix: enable partial word search in research and projects

### LINKED ISSUE #4589: [bug] No search results for unfinished words
**Describe the bug**

Image

For example if I write "shred" in the search, nothing shows up, even though there are items with Shredder in the title. 

As far as I know, I found this bug in Research and Library

**To Reproduce**

1. Go to 'https://community.preciousplastic.com/research'
2. Search for 'shredd'
3. See no results
4. If you spell the whole word shredder, then it retrieves results

**Expected behaviour**
The search functionality should discover items also when you write the word partially

GMetaxakis:
## PR Checklist

## PR Type

- [x] Bugfix

## What is the new behavior?

Replace `to_tsquery` with `plainto_tsquery` + prefix matching (`':*'`) in `get_research`, `get_research_count`, `get_projects`, and `get_projects_count` PostgreSQL functions.

- Searching "shred" now matches "shredder"
- Special characters no longer cause syntax errors
- Multi-word queries continue to work correctly

- [x] Yes

## Git Issues

Closes #4589

### DIFF
--- a/packages/cypress/src/integration/research/list.spec.ts
+++ b/packages/cypress/src/integration/research/list.spec.ts
@@ -22,6 +22,15 @@ describe('[Research - List Articles]', () => {
     cy.get('[data-cy=ResearchListItem]').should('have.length.at.least', 1);
   });
 
+  it('[Search Functionality - Partial word search]', () => {
+    cy.wait(2000);
+    cy.step('Search with a partial word');
+    cy.get('[data-cy=research-search-box]').clear().type('tes');
+
+    cy.step('Verify partial match returns results');
+    cy.get('[data-cy=ResearchListItem]').should('have.length.at.least', 1);
+  });
+
   it('[Search Functionality - Sorts by Latest Updated]', () => {
     cy.visit(researchPageUrl + '?sort=LatestUpdated');
     cy.step('Verify sorted results are displayed');


GMetaxakis:
Schema files updated in d6552cbd — both `supabase/schemas/research.sql` and `supabase/schemas/projects.sql` now match the migration.

dalibormrska:
Exciting improvement @GMetaxakis !

benfurber:
Really great user enhancement, thank you. Can we get a feature test updated to show this change working too please?

GMetaxakis:
Added an e2e test for the partial word search — it searches for "tes" and verifies it matches results containing "test", confirming the prefix matching works. Thanks for the suggestion!

dalibormrska:
Amazing @GMetaxakis, tested it out, already a big improvement for Library and Research. In the future, some sorting based on relevancy of the match would be cool...

However, I looked again at the search for questions module, which behaves a bit differently and weirdly. See gif. Do you have any idea why this is? And should I create a separate issue for this @mariojsnunes ?
search questions

GMetaxakis:
@dalibormrska Good catch on the questions search\! I've opened #4663 to fix it — questions now use the same prefix matching RPC approach as research and library.

mariojsnunes:
for visibility, this solution may have broken searching by multiple works. (not 100% sure if it was working before)
I'm working on a fix for that now.

PR #4643, opened 2026-03-01
feat: add delete button to question and news forms

### LINKED ISSUE #4265: [feature] Delete button for all forms
On `ResearchUpdateForm` it's possible to delete an item the user creates. This should be possible for questions, research items and news (and library too but that's still in the old system).

It would be great to move the delete UI flow into `FormWrapper`.

GMetaxakis:
## PR Checklist

## PR Type

- [x] Feature

## What is the new behavior?

- Add DELETE API endpoints to `api.questions.$id` and `api.news.$id` with creator/admin authorization (soft-delete via `deleted: true`)
- Add delete button with confirmation modal to QuestionForm and NewsForm (visible only in edit mode), matching the existing ResearchUpdateForm pattern
- Add client-side `deleteQuestion` and `deleteNews` service functions
- Add Cypress e2e tests for deleting questions and news

- [x] No

## Git Issues

Closes #4265


GMetaxakis:
The E2E test failures are caused by the new `published_at` column not existing on the CI test Supabase instance.

The migration (`20260301100000_add_news_published_at.sql`) needs to be applied to the test database for the news API queries to work:

```sql
ALTER TABLE news ADD COLUMN published_at timestamptz;
UPDATE news SET published_at = created_at WHERE (is_draft = false OR is_draft IS NULL);
```

Once the migration is applied, the 4 failing news tests should pass.

benfurber:
Hey @GMetaxakis, thanks for working on so many items atm. I'm confused by the branch name and some some of the changes in this branch. Are they all related to the creation date change?

GMetaxakis:
Hey @benfurber, you're right — sorry about the confusion! I accidentally pushed the publish date changes onto this branch instead of keeping them on their own. I've now cleaned this up:

- **This PR** is back to being purely the delete feature for questions and news (closing #4265) — title and description updated to match
- **The publish date fix** has been moved to its own PR: #4654 (closing #4479)

Should be much clearer now. Let me know if you have any questions!

src/pages/News/Content/Common/NewsForm.tsx
```diff
@@ -54,8 +55,19 @@ export const NewsForm = (props: IProps) => {
...
   const id = news?.id || null;
+  const isEdit = parentType === 'edit';
+
+  const handleDelete = async () => {
+    if (!id) {
+      return;
+    }
+    setShowDeleteModal(false);
+    await newsService.deleteNews(id);
+    window.location.assign('/news');
```
mariojsnunes: is there a good reason to use this instead of `useNavigate`?

GMetaxakis: Following the same pattern used in ResearchUpdateForm — after a delete, `window.location.assign` forces a full page reload to clear any cached state for the deleted item. A client-side `useNavigate` could show stale data.

mariojsnunes:
@GMetaxakis seems like both cypress tests for deleting news and questions are failing

mariojsnunes:
Fixed the conflicts on this branch

mariojsnunes:
@dalibormrska need your input here.
we have 2 ways of deleting an item currently.

For research appears on the item itself.
image

With this PR questions and library will have it on the form (need to edit first):
image

Technically it's simpler to **not** have it on the form, but we can do it however is best for UX.

dalibormrska:
Judging now the logic decision and not necessarily the current layout, I lean towards the option of having just the edit button and the delete inside of the form.

src/services/newsService.ts
```diff
@@ -36,6 +36,17 @@ const upsert = async (id: number | null, form: NewsFormData) => {
   return news;
 };
 
+const deleteNews = async (id: number) => {
+  const response = await fetch(`/api/news/${id}`, {
+    method: 'DELETE',
+  });
+
+  if (response.status !== 200 && response.status !== 201) {
+    throw new Error('Error deleting news', { cause: 500 });
```
mariojsnunes: Should handle the error with the message sent from the api, see above.

src/services/questionService.ts
```diff
@@ -34,6 +34,17 @@ const upsert = async (id: number | null, question: QuestionFormData) => {
   return Question.fromDB(new DBQuestion(newQuestion.question), []);
 };
 
+const deleteQuestion = async (id: number) => {
+  const response = await fetch(`/api/questions/${id}`, {
+    method: 'DELETE',
+  });
+
+  if (response.status !== 200 && response.status !== 201) {
+    throw new Error('Error deleting question', { cause: 500 });
```
mariojsnunes: (same as on newsService)
Should handle the error with the message sent from the api, see above.

src/pages/News/Content/Common/NewsForm.tsx
```diff
@@ -31,7 +32,19 @@ export const NewsForm = (props: IProps) => {
...
 
+  const id = props?.id || null;
+  const isEdit = props.formAction === 'edit';
+
+  const handleDelete = async () => {
+    if (!id) {
+      return;
+    }
+    setShowDeleteModal(false);
+    await newsService.deleteNews(id);
```
mariojsnunes: should handle the error here, maybe show an `<Alert />` component

src/pages/Question/Content/Common/QuestionForm.tsx
```diff
@@ -31,7 +32,18 @@ export const QuestionForm = (props: IProps) => {
...
+  const [showDeleteModal, setShowDeleteModal] = useState(false);
   const id = props?.id || null;
+  const isEdit = props.formAction === 'edit';
+
+  const handleDelete = async () => {
+    if (!id) {
+      return;
+    }
+    setShowDeleteModal(false);
+    await questionService.deleteQuestion(id);
```
mariojsnunes: should handle the error here, maybe show an `<Alert />` component

mariojsnunes (changes requested):
just a minor thing about error handling.

we are now sending the error messages correctly from the api. so we should handle them correctly on the frontend as well.

Also please ensure my conflict resolution didn't break something :)

mariojsnunes:
@GMetaxakis the news delete test is failing

mariojsnunes:
This became a bigger priority now due to more spam accounts.
I will apply these changes on #4773. Thanks @GMetaxakis !

PR #4646, opened 2026-03-01
fix: reduce flakiness in research write e2e tests

GMetaxakis:
## PR Checklist

## PR Type

- [x] Bugfix

## What is the new behavior?

Replaces arbitrary `cy.wait()` calls in `research/write.spec.ts` with assertion-based waits:

- Wait for `[data-cy=delete-image]` to appear after image uploads instead of `cy.wait(2000)`
- Wait for Uppy upload button to disappear after file upload before submitting
- Wait for elements to be visible before interacting (`should('be.visible').click()`)
- Use increased timeout for slow post-submit assertions instead of fixed waits
- Wait for form fields to be visible before filling them after navigation

- [x] No


mariojsnunes:
Thanks for starting this, found another issue, merged some of these changes to #4649

PR #4650, opened 2026-03-03
feat: add self-service account deletion to settings page

### LINKED ISSUE #4163: [feature] Add Profile delete button
Currently deleting a profile has to be requested via discord, it's hard both for users and moderators.

Enable users to delete their own profile via the Settings Account tab.

1. In the profile Account tab, have a delete button.
2. The button should be red and open a confirmation popup.
3. On confirm, the account is deleted, user logged out and redirected to the homepage.

The account deletion API:
1. Use the current authenticated user
2. Delete map pin from firebase
3. Delete profile from firebase
4. Delete profile from supabase
5. Delete user account from supabase (automatically deletes the profile due to cascade constraint)

GMetaxakis:
## PR Checklist

## PR Type

- [x] Feature

## What is the new behavior?

Adds a "Delete account" section to Settings > Account, replacing the old Discord link. Users can now delete their own account by entering their password and confirming via a modal.

- Accordion-based form (collapsed by default) with password confirmation
- Destructive (red) button + confirmation modal with clear warning
- Server-side API route verifies password, cleans up blocking FK references (notifications, notification preferences), then deletes the auth user via admin API — cascade constraints handle the rest
- Session is cleared and user is redirected to `/`

- [x] No

## Git Issues

Closes #4163


packages/cypress/src/integration/SignUp.spec.ts
```diff
@@ -102,4 +102,51 @@ describe('[User sign-up]', () => {
         .should('be.visible');
     });
   });
+
+  describe('[Delete account]', () => {
+    it('Rejects wrong password', () => {
```
benfurber: Not sure this needs a seperate test. Add as a step to the main test.

packages/cypress/src/integration/SignUp.spec.ts
```diff
@@ -102,4 +102,51 @@ describe('[User sign-up]', () => {
...
+
+      cy.step('Submit with correct password');
+      cy.get('[data-cy="deleteAccountPassword"]').type(password);
+      cy.get('[data-cy="deleteAccountSubmit"]').click();
+
+      cy.step('Confirm deletion');
+      cy.get('[data-cy="Confirm.modal: Confirm"]').click();
+
+      cy.step('Redirected to homepage');
+      cy.url().should('eq', Cypress.config().baseUrl + '/');
```
benfurber, line 126: Would be good to see what happens to the content of a deleted user:
* Comments (de-author)
* Questions (de-author)
* Research (de-author)
* Projects (de-author)
* News (de-author)
* Map pins (delete)

benfurber (changes requested):

src/pages/UserSettings/services/account.service.ts
```diff
@@ -20,7 +20,18 @@ const changePassword = async (oldPassword: string, newPassword: string) => {
   });
 };
 
+const deleteAccount = async (password: string) => {
+  const data = new FormData();
+  data.append('password', password);
+
+  return await fetch('/api/account/delete', {
```
mariojsnunes, line 27: just to double-check, using a POST instead of DELETE because we need to send FormData?

src/routes/api.account.delete.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const { client, headers } = createSupabaseServerClient(request);
+  const formData = await request.formData();
+
+  const password = formData.get('password') as string;
+
+  const claims = await client.auth.getClaims();
+  const authId = claims.data?.claims?.sub;
+
+  if (!authId) {
+    return Response.json({ error: 'Unauthorized' }, { headers, status: 401 });
```
mariojsnunes: please update the error handling to use the new methods `validationError`, `unauthorizedError` etc.
Also need to add a try-catch. see other api endpoints like `api.questions.$id.ts`

src/routes/api.account.delete.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const profileService = new ProfileServiceServer(client);
+  const profile = await profileService.getByAuthId(authId);
+
+  if (!profile) {
+    return Response.json({ error: 'Profile not found' }, { headers, status: 404 });
+  }
+
+  const adminClient = createSupabaseAdminServerClient();
+
+  await adminClient.from('notifications').delete().eq('owned_by_id', profile.id);
```
mariojsnunes: I'd rather use a delete cascade for all of these, so we need a schema change.

src/routes/api.account.delete.tsx
```diff
@@ -0,0 +1,50 @@
...
+  await adminClient.from('notifications').delete().eq('triggered_by_id', profile.id);
+  await adminClient.from('notifications_preferences').delete().eq('user_id', profile.id);
+
+  const { error } = await adminClient.auth.admin.deleteUser(authId);
```
mariojsnunes: We can't delete the user without checking if the user has a profile on other tenants.

1 user account can have multiple profiles (1 per tenant)

If the user does not have other profiles, we can delete the account, this will cascade delete the profile.
If the user has other profiles, just delete the profile for this tenant (try using the normal client, not the admin client as it is restricted to the current tenant).

packages/cypress/src/integration/SignUp.spec.ts
mariojsnunes: To do that, the best way might be with a seed. Have the profile and associated items which we delete on the test, instead of signing up a new user and creating the content each time.
If that's difficult for some reason, at least this PR must ensure those delete/set null cascades are correct.

mariojsnunes (changes requested):
Great first step! Still a few things to be improved before merging :)

mariojsnunes:
I'm reusing most of this here #4777. Thanks @GMetaxakis !

PR #4630, opened 2026-02-27
test: add admin permission test for questions

GMetaxakis:
## PR Checklist

- [x] - Unit and/or e2e tests for the changes that have been added (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes)
- [ ] Documentation
- [x] Other: E2E test coverage

## What is the new behavior?

Replaces the commented-out `[Admin]` placeholder in `questions/write.spec.ts` with an actual Cypress test that verifies:

- Admin can see the edit button on another user's question
- Admin can access the edit page
- Admin can update the question description and see the changes

This is the first step toward the full admin permission coverage requested in the issue. Further tests for other content types (news, research, library) and discussions can follow.

## Does this PR introduce a DB Schema Change or Migration?

- [ ] Yes
- [x] No

## Git Issues

Ref #4210

PR #4651, opened 2026-03-03
test: add admin permission tests for library, research, and news

### LINKED ISSUE #4210: [Audit] What can an admin actually do?
It's not clearly (especially from the cypress tests) what an admin can actually currently do/can't do.

Basically they should be able to edit anything, it would be great to ensure each cypress write/discussions test covers admin permissions (starting with questions) as a way to see what bugs exist at the moment.

GMetaxakis:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other: E2E tests

## What is the new behavior?

Adds `[By Admin]` Cypress E2E tests to library, research, and news write specs, verifying that admins can edit content created by other users. Follows the same pattern established in the questions `[By Admin]` test (#4630).

Each test:
1. Signs in as `demo_admin`
2. Visits a seeded item not authored by the admin
3. Asserts the edit button is visible and accessible
4. Edits a form field and submits
5. Verifies the edit appears on the detail page

- [ ] Yes
- [x] No

## Git Issues

Closes #4210


mariojsnunes:
I fixed the conflicts, but seems like the tests need to be updated

benfurber:
@GMetaxakis Hope you're well. Any change you'll have time for this soon? :)

GMetaxakis:
Hey @benfurber, @mariojsnunes — sorry for the long silence, thanks for the nudge. Just pushed an update: master is merged in and the `[By Admin]` tests are reworked.

Mário was right that the tests needed updating — the root causes turned out to be more than the conflict:

1. **Admin sign-in wasn't tenant-aware.** The tests hardcoded `demo_admin@example.com`, but the seeder now creates `demo_admin+<tenantId>@example.com`. All three now use `getTenantUser(users.admin)`.

2. **None of the three forms redirect on submit.** They show a toast with an action link (`Project published!` / `Research published!` / `News published`). The old assertions did `submit()` then `cy.url().should('include', slug)`, which passes trivially while still sitting on `/edit`, and then `cy.contains(adminEdit)` failed because the edited text only exists as an input value. All three now assert the toast and follow its action link before checking the detail page.

3. **The seeded fixtures aren't publishable**, so the edits could never be saved:
   - Library: `seedLibrary` seeds `project_steps` without images, and every step needs images or a video URL, plus the API requires 3 steps (`MOCK_DATA.projects[0]` only has 2).
   - Research: `MOCK_DATA.research[0]` has no cover image, and `api.research.$id.ts` rejects publishing without one.
   - News: `hero_image` is commented out in `shared/mocks/data/news.ts`, and `api.news.$id.ts` throws `A hero image is required to publish`. **This was the actual cause of the failing news test.**

So library and research now create the content under a non-admin author (`howto_creator` / `research_creator`) and then edit it as admin — self-contained, and consistent with the neighbouring write tests. News can't do that (only admins and editors may create news, and there's no editor in the mock users), so it still uses a seeded item and uploads the required hero image. I moved it to `MOCK_DATA.news[1]`, because `news[0]` is asserted on exactly by `news/read.spec.ts` (its body is compared against the meta description tags) and specs share one seeded database.

One caveat: I couldn't run Cypress locally (no Docker here, so no local Supabase), so I'm relying on CI for the actual verification. Happy to iterate quickly if anything still fails.

GMetaxakis:
CI is green now — `106 tests passed`, all four e2e jobs passing.

Two corrections to my previous comment, since I got a detail wrong there:

**The news failure was not the hero image.** The actual assertion error was:

```
Timed out retrying after 15000ms: Expected to find element: `[data-cy=Username]`, but never found it.
```

News pages don't render an author byline at all — nothing under `src/pages/News/` uses `<Username>`, unlike `ResearchDescription.tsx` — so `cy.get('[data-cy=Username]').should('not.contain', ...)` could never pass there. That's why research passed and news didn't, and it's the same reason the original run failed back in March. It now asserts the seeded item's title instead. The hero image *is* genuinely required to publish (`api.news.$id.ts:237`), so the upload stays, but it wasn't what was failing.

**The library failure was a race, not a validation gap.** The error was:

```
Timed out retrying after 15000ms: Expected to find element: `a[data-cy=toast-action-link]`, but never found it.
```

The create submit never produced a toast, because `fillStep()` uploads step images without waiting for them to register, and the test clicked submit immediately after the third one. With `steps[i].images` still empty, `validateStepMedia` (`LibraryStep.field.tsx:82`) blocked submission. The neighbouring tests get away with it because they do extra work (move step, add step, delete step) that absorbs the upload latency. The test now waits for each step's image before publishing.

Worth noting for anyone debugging these in future: `cypress.config.ts` sets `reporter: 'junit'` writing to `coverage/out-*.xml`, so failure messages never reach stdout and don't appear in the CircleCI logs. The quickest way to the actual error is the failure screenshots published as CircleCI artifacts, or those XML files locally.

Happy to squash the commits if you'd prefer a tidier history.

mariojsnunes:
Thanks @GMetaxakis !
(we always squash merge, so no need to do it yourself)

PR #4655, opened 2026-03-05
fix: pass COMMIT_SHA build arg to bust Docker cache on preview deploys

### LINKED ISSUE #4127: [bug] Preview website is not updated on PR push
**Describe the bug**
When the label "Review allow-preview ✅" is added to an open PR, a new site instance is created on fly.io with the PR changes. It works well.
When further changes are pushed to the branch, it triggers a new deploy to the same instance, but the code changes are not visible.
The current work-around is to remove the label (which triggers the instance to be deleted) and add it again.

**Expected behaviour**
It should update the deployed code when pushing changes to the PR.

GMetaxakis:
## PR Checklist

## PR Type

- [x] Bugfix

## What is the new behavior?

The `Dockerfile.preview` already defines a cache-busting `ARG COMMIT_SHA` before `COPY . .`, but the deploy workflow never passed it. This meant Fly.io's remote Docker builder reused cached layers on subsequent pushes, so code changes weren't reflected in the preview app.

This PR passes `--build-arg COMMIT_SHA=<head_sha>` to `fly deploy`, which invalidates the Docker cache from that point onwards — ensuring new code is copied and rebuilt on every push.

- [x] No

## Git Issues

Closes #4127

### DIFF
--- a/.github/workflows/pr-preview-fly-deploy.yml
+++ b/.github/workflows/pr-preview-fly-deploy.yml
@@ -71,6 +71,7 @@ jobs:
         with:
           config: fly-preview.toml
           name: community-platform-pr-${{ github.event.number }}
+          args: --build-arg COMMIT_SHA=${{ github.event.pull_request.head.sha }}
 
       - name: Set Fly.io Secrets
         run: |


mariojsnunes:
You are a legend if this fixes the issue !

GMetaxakis:
> You are a legend if this fixes the issue !

Keep me updated

PR #4662, opened 2026-03-08
Adding pagination for Research, News, Questions and Library Page

### LINKED ISSUE #4523: [feature request] Add pagination to content list
## **Is your feature request related to a problem? Please describe.**
Currently when we go to News, Library, Research or Questions page, we get the Load More button at the bottom.
Image

This is not ideal, since we don't get the information about how many batches of items are there, it's hard to get to older items. If the user also gets to an older item (clicks the load more button several times), then click on the item and navigates back, the item list resets and the user needs to click on Load More several times again to get to where they were before.

## **Describe the solution you'd like**

Create and implement the Pagination component.

[SEE THE INTERACTION IN THE FIGMA PREVIEW](https://www.figma.com/proto/1KTllB5BkYiElbaDy6nG7z/CP---Pagination?page-id=0%3A1&node-id=1-1067&viewport=727%2C67%2C0.38&t=PZjrwTsTspTBkqIV-1&scaling=min-zoom&content-scaling=fixed)

[TAKE THE MEASUREMENTS FROM FIGMA DESIGN FILE](https://www.figma.com/design/1KTllB5BkYiElbaDy6nG7z/CP---Pagination?node-id=1-5390&t=GRikxITt0QQKhyJ0-1)

Image

Image

How I created the components in Figma:
Image

**Requirements:**
---
- [ ] Examine the Figma prototype and Figma design file, you can get the measurements, begaviour, colors, etc. from there.
- [ ] See how I created the components in the design file, maybe it can be helpful for creating the components in the code?
---
- [ ] Implement tooltips for <<, <, >, >> buttons.
- [ ] On the first page, we don't need to show the << and < navigation buttons.
- [ ] On the last page, we don't need to show the >> and > navigation buttons.
---
- [ ] The number in the square is an input field, we can type our desired page there. Show type cursor when hovered.
  - [ ] The input should be handled with min-width, so that at minimum it's a 44x44px square, but can expand if necessary.
  - [ ] It should only accept numbers
  - [ ] On mobile when the keyboard opens we would only want to show the number keyboard (there should be a way to force this by HTML as far as I know)
---
- [ ] If there is for example only 5 pages, it should not be possible to go to the page 9.
- [ ] On mobile the component actually does not need a special variant. It should be under 300px wide which should not be a problem for any device.
- [ ] For example, when the user goes to page 3, clicks on some item, then navigates back, they should go to the page 3. My guess is that this would mean keeping the track of the current page in the URL. This would also allow for someone to copy the url of a specific page and send to someone else to see the same items.
---
- [ ] Since it is a crucial navigational element, accessibility needs to be handled correctly.
  - [ ] Make sure that the pagination is usable by keyboard by the TAB button
  - [ ] Correct HTML semantics (probably <nav> element?) and aria-labels.
---

## **Describe alternatives you've considered**
Disabling the buttons <<, <<, >>, >> instead of hiding them when necessary?

## **Additional context**
@mariojsnunes Any tech details?

V24039:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

Added pagination to the research apge

- [ ] Yes
- [x] No

## Git Issues

Closes #4523 

## What happens next?


V24039:
Hi @mariojsnunes, @benfurber can you please review my PR, I have fixed the test cases which were failing due my changes, but not sure why other test cases are failing, can you please help me with this

mariojsnunes:
1. The spacing looks a bit strange to me.
image

2. Are we sure we want the "go to last" button?
3. Maybe make the number input background color white

(This are just suggestions, to be validated by the design team)

dalibormrska:
Will look at this this evening

dalibormrska:
Mario's comment:
1. Agree, examine Figma file for exact sizings. If you'd like any help or support, tag me anytime. :))
2. Let's keep the skip to last button for now as is described in [the issue
](https://github.com/ONEARMY/community-platform/issues/4523)
3. Agree

My comment:
4. 1 out of 1 pages is not needed, no? image
5. Doublecheck also font family and size.

V24039:
Hi @dalibormrska, sure will the changes, and check font family and spacing.
Quick question
1. Do we have any specific fontfamily thats used, shall I pick it from the figma.
2. For single page (1 of 1), shoul I remove the pagination section.

dalibormrska:
@V24039 Sorry about being late.

1. In Figma I used Varela Round Bold. (The font officially does not have bold, but browsers render it with a slight stroke)
2. If we are sure that all content is there, then no need to show "page 1 of 1". I don't know what the best way is to prevent errors or bugs though.

Also, for the numerical input - we don't show the enter or "Go" button, so the interaction is slightly more awkward. It is therefore important to not rely only on Enter key, but also to submit when clicking out of the box. Or any other suggestions? :))

V24039:
> @V24039 Sorry about being late.
> 
> 1. In Figma I used Varela Round Bold. (The font officially does not have bold, but browsers render it with a slight stroke)
> 2. If we are sure that all content is there, then no need to show "page 1 of 1". I don't know what the best way is to prevent errors or bugs though.
> 
> Also, for the numerical input - we don't show the enter or "Go" button, so the interaction is slightly more awkward. It is therefore important to not rely only on Enter key, but also to submit when clicking out of the box. Or any other suggestions? :))

1. The font by default is set Varela Round, but as mentioned its not bold, do I add bold style to the text or not make any changes to it.
2. Understood, I am currently using totalPages = 1 to no show the pagination. This should work fine and not cause any bugs.
3. Nice point, currently there is debounce to control the page, will check for the outside click and add it if not present.
4. As mentioned early, even I don't see the point to have last page icon if user wants to go to last they can directly enter it in the input. Just an opinion.

mariojsnunes:
> 3. Nice point, currently there is debounce to control the page, will check for the outside click and add it if not present.

@V24039 Check the blur behavior, not the click outside

dalibormrska:
fontweight
1. Font weight 600 does do something and is utilised on the platform. I don't know why it works but it does haha.
2. Great!
3. Great!
4. Thanks for the comment, I thought about it, but I still do see the point. Going to the last page, is a common action. With the button it takes 1 click, with the input it takes 2 clicks and 2-3 keystrokes. We have the space available, it doesn't look visually distracting in my opinion, so I would say let's go for it. I don't see a big con.

V24039:
Hi @mariojsnunes, @dalibormrska can you please check the changes

dalibormrska:
Really cool @V24039, getting there!

Some remarks:
1. The arrows have different styled icons.
   image
The single arrow should be rounded style as well. This is probably not your problem, I'm assuming it's the icon already in the codebase. Could we replace the sharp arrow by the rounded arrow @mariojsnunes ? Here is the difference:
   image
And here is the icon in the 27x27 box, same as the original one: -> Single Arrow <-

2. The text box "of 3" is on two lines now. It should be on one line, spanning as much width as needed. Also pay attention to the padding of this text box so that it looks centred between the neighbours.
   image

3. The font of the input seems to be Inter, while the text box is Varela Round. Would it be possible to also have Varela Round Bold in the input?
   image

4. I don't know if this is an issue of the test environment or the pagination, but for me the content does not change when navigating the pages. @mariojsnunes?
nav

mariojsnunes:
@V24039 3 issues:
1. pagination doesn't seem to be working on library, only works on refresh. (ask if you want me to take a look)
2. `&pageNo=0` pagination is currently 0 based, should be 1 based.
3. change param name from `pageNo` to `page`

V24039:
> @V24039 3 issues:
> 
> 1. pagination doesn't seem to be working on library, only works on refresh. (ask if you want me to take a look)
> 2. `&pageNo=0` pagination is currently 0 based, should be 1 based.
> 3. change param name from `pageNo` to `page`

Hi @mariojsnunes I fixed the library issue, it was due the page missing as a dependency in the useEffect responsible for triggering the API call. I also made the changes requested for the URL

dalibormrska:
Looks really cool, and works really well! Great job @V24039!

Two small remarks about spacing.

1. Some weird spacing behaviour between the single left arrow and the input. It seems that the gap there is not 5px but like 2px. It's as if the parent box of that arrow got squished by 2.75px. In the devtools the solution for me was to move the button one level up, out of the parent div.
   image

   pagination

2. Currently the page counter "of 3" does not have spacing around.  image
I would propose margin-left of 10px for visual balance.
   image



So the final spacings would look like this:
image

mariojsnunes:
> I would propose margin-left of 10px for visual balance.

Sorry, but margins are almost always the wrong choice in modern css (spacing should be decided by the parent, not the element itself).
Here I'd propose instead to group the input and the "of x" text on a `<Flex>` with a gap.

V24039:
Hi @dalibormrska Sorry for the delay I was stuck with work, I have made the changes, following @mariojsnunes fixed both the comments, can you please review it once

dalibormrska:
@V24039 Thanks for the update! No need to be sorry about the delay. :))

The point 1. seems to be nicely fixed, but when I launch the preview, the point 2. is still standing.

The spacing between the input and the "of 3" did not increase by 10px and additionally the input is not square anymore, but is squished.
Screenshot 2026-04-20 at 03 15 47

For reference, pasting the image of what we want to achieve:
image

dalibormrska:
Awesome @V24039 thanks!

Let's go with 15px gap instead of 10px here and then it's good to be released from my side!

image

--- 

@mariojsnunes good to go from your side? :)))

V24039:
@dalibormrska I have changed the gap to 15px

packages/components/src/PaginationIcons/PaginationIcons.tsx
```diff
@@ -3,11 +3,18 @@ import { Button } from '../Button/Button';
 interface IProps {
   hidden?: boolean;
   title: string;
-  directionIcon: 'chevron-left' | 'chevron-right' | 'double-arrow-left' | 'double-arrow-right';
+  ariaLabel: string;
+  directionIcon:
+    | 'paginationSingleLeft'
```
mariojsnunes, line 8: why this change to `paginationSingle...`? I believe chevron is the common name for this icon.

V24039: The chevron-left and chevron-right are used in other places and does not match the icon present in the Figma, so I created a new set of svg icons to match the Figma requirements

mariojsnunes:
@V24039 nice one! will deploy it soon :)

V24039:
🎉🎉 finally it paid off. Happy to see the changes in live website.

dalibormrska:
Thanks @V24039, looks awesome live!

PR #4666, opened 2026-03-11
feat: image and file uploads

mariojsnunes:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [x] Technical

## What is the new behavior?

- Images and files are now uploaded right when adding them on a form.
- Forms only send image metadata to the api.
- Image upload api optimizes images automatically
  - Converts png, jpg, jpeg to webp (up to 90% size reduction)
  - Resizes images to a maximum of 2048x2048, keeping the original aspect ratio
- Limits research updates images to 10
- Better error handling, shows useful errors to the user (ex: "Image exceeded size limit" instead of "Bad Request")
- Adds request size limits to the server

- [x] Yes

### DIFF
--- a/packages/cypress/src/integration/library/write.spec.ts
+++ b/packages/cypress/src/integration/library/write.spec.ts
@@ -367,7 +367,7 @@ describe('[Library]', () => {
       cy.step('Replace step image');
       cy.get('[data-cy=step_0]').within(() => {
         cy.get('[data-cy=delete-image]').first().click({force: true});
-        cy.get('[data-cy=image-upload-0]').find(':file').selectFile('src/fixtures/images/howto-step-pic2.jpg', { force: true });
+        cy.get('[data-cy=new-image-upload]').find(':file').selectFile('src/fixtures/images/howto-step-pic2.jpg', { force: true });
         cy.get('[data-cy=delete-image]').should('exist');
       });
       


dalibormrska:
QA:

- The loader sits in the corner instead of center. Feels a bit broken, since we put it in the centre usually. image

- What is the color system for the loaders that some are yellow and some are grey? The yellow would then need to change according to the instance, no? Green for PK. I would say, let's skip the complexity of color system for loaders and just go for grey. The Uploading image text is ok, showing system status is always good.
image  image

- For the upload in library the error message shifts the layout. Can we show the error message under the "Upload images or video" title, as we do in other places?
image  image

- I uploaded a jpg as a[ file attachment to a research update](https://community-platform-pr-4666.fly.dev/research/testtttt#update_219666). Went to download it and got an empty json opened in my browser. Is this just an issue of the preview environment or something to fix? And is it relevant to this issue, or should be investigated separately?

- In library the upload is on the right, while in other cases it is on the left. I would advise to pick one way for consistency.
image  image

dalibormrska:
Cool, thanks for the updates, Mario. I tested on mobile now. 

- Library upload has no vertical gap (in research it has). And the loader in research still in the corner.





- The same picture gets uploaded OK on the news title but 90 degree turned in the md text editor (other upload cases do the 90 degree turn as well)

PR #4684, opened 2026-03-26
removing username from auth table, remade branch

jproberson:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other:

## What is the new behavior?

_Describe the new behaviour_

- [ ] Yes
- [ ] No

## Git Issues

Closes #

## What happens next?

### DIFF
--- a/packages/components/src/Tooltip/Tooltip.tsx
+++ b/packages/components/src/Tooltip/Tooltip.tsx
@@ -15,7 +15,11 @@ type TooltipProps = {
 
 export const Tooltip = ({ children, id }: TooltipProps) => {
   return (
-    <StyledTooltip id={id} openEvents={{ mouseenter: true, focus: true }} closeEvents={{ mouseleave: true, blur: true }}>
+    <StyledTooltip
+      id={id}
+      openEvents={{ mouseenter: true, focus: true }}
+      closeEvents={{ mouseleave: true, blur: true }}
+    >
       {children}
     </StyledTooltip>
   );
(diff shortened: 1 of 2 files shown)


supabase/functions/send-email/index.ts
```diff
@@ -53,6 +53,8 @@ Deno.serve(async (req) => {
       ...email_data,
     } as any;
 
+    const username = user['user_metadata']?.username || 'there';
```
mariojsnunes: maybe we should fetch the username from the profile, if that's possible

shared/models/userEmailData.ts
```diff
@@ -2,7 +2,7 @@ export type UserEmailData = {
   email: string;
   code: string;
   new_email?: string;
-  user_metadata: {
-    username: string;
+  user_metadata?: {
```
mariojsnunes: is this still needed?

src/services/profileService.server.ts
```diff
@@ -191,6 +191,7 @@ export class ProfileServiceServer {
     const { data, error } = await this.client
       .from('profiles')
       .update({
+        username: values.username,
```
mariojsnunes: I wonder if we should have some special endpoint just for the username update.
Some special UI for it as well.

Is it currently being validated for uniqueness on a tenant?

jproberson: it is a constraint but I like the idea of a separate endpoint just to make it obvious and give any extra logic down the road a place to live. Adding it.

shared/models/userEmailData.ts
jproberson: no it is not, updated

supabase/functions/send-email/index.ts
jproberson: sounds good

PR #4690, opened 2026-03-31
feat: Add news notifications

### LINKED ISSUE #4674: [draft] [news] News creators can generate a preview email notification
Sidebar button that, when pressed, saves as a draft the news, generates and sends a preview email notification of the news.

benfurber:

- [ ] 🐛 Bugfix — fixes incorrect behavior without changing functionality
- [ ] ♻️ Refactoring — improves code structure with no functional changes
- [ ] 🔖 Other:

## What is the new behavior?

1. Adds in app notification display
571238807-edb0ede2-bdd2-4c10-9a55-346ca97023c3

2. Adds the ability to set the email visibility of news for badgers
Screenshot 2026-04-16 at 14 29 29

3. Can change news email notifications
Screenshot 2026-04-16 at 14 53 04

4. Adds the ability for news content creators to preview how a news article will look like in an email.
Screenshot 2026-03-31 at 15 37 36

_Turns out the email components package we're using gives us all of this styling for free:_

Screenshot 2026-03-31 at 15 35 31
Screenshot 2026-03-31 at 15 35 40
Screenshot 2026-03-31 at 15 35 43
Screenshot 2026-03-31 at 15 35 48
Screenshot 2026-03-31 at 15 35 53

- [x] Yes

## Git Issues

Closes #4674


src/pages/UserSettings/SupabaseNotifications.tsx
```diff
@@ -1,36 +1,42 @@
...
 
   const { profile } = useProfileStore();
 
   const refreshPreferences = async () => {
-    const preferences = await notificationsPreferencesService.getPreferences();
-    setInitialValues(preferences);
+    const dbPreferences = await notificationsPreferencesService.getPreferences();
+
+    if (!dbPreferences) {
+      return setSubmitResults({ type: 'error', message: 'Error finding preferences' });
```
mariojsnunes: I'm deleting this submitResults thing on another PR. Please use the toast instead (it's already available).

src/routes/api.notifications-preferences.ts
```diff
@@ -19,13 +25,21 @@ export const loader = async ({ request }: LoaderFunctionArgs) => {
...
 
-  const preferences = data || DEFAULT_NOTIFICATION_PREFERENCES;
+  if (data) {
+    preferences = {
+      ...preferences,
+      ...data,
+    };
+  }
 
   return Response.json({ preferences }, { headers, status: 200 });
```
mariojsnunes: I just realized that here we are returning raw database data. Big no!
Please create a `NotificationsPreferences` class, maybe with that static `.fromDb` method we use on other places.
And need to update the save and form as well, since they are expecting currently `snake_case` variables, use `camelCase` instead.

supabase/migrations/20260328093904_add_news_notifications.sql
```diff
@@ -0,0 +1,5 @@
+alter table "public"."notifications" alter column "action_type" set data type text using "action_type"::text;
```
mariojsnunes, line 1: please don't do this, database types give us a lot of safety. less bugs, less tests we need to maintain.

src/utils/isAdmin.ts
```diff
@@ -0,0 +1,5 @@
+import { DBProfile, Profile, UserRole } from 'oa-shared';
+
+export const isUserAdmin = (user: DBProfile | Profile) => {
```
mariojsnunes, line 3: where is this being used? For the frontend we have this method on the profileStore. So assuming this will only be used server-side, I'd accept only `DBProfile` as param.

src/services/notificationsSupabaseService.server.ts
```diff
@@ -13,19 +14,57 @@ import { NotificationEmailServiceServer } from './notificationEmailService.serve
...
+      console.log({ data: response.data });
+      return response.data;
+    } catch (error) {
+      console.error(error);
+      throw new Error(error);
+    }
+  }
+
+  async getProfilesByBadgeId(badgeId: number) {
+    console.log('getProfilesByBadgeId');
```
mariojsnunes: console log

src/services/notificationsSupabaseService.server.ts
```diff
@@ -115,6 +156,66 @@ export class NotificationsSupabaseServiceServer {
...
+      // For in-app
+      let subscribers;
+      if (!news.profile_badge) {
+        subscribers = await this.getAllProfiles();
+      }
+
+      if (news.profile_badge) {
+        subscribers = await this.getProfilesByBadgeId(news.profile_badge.id);
+      }
+      await this.createNotifications(notification, subscribers);
```
mariojsnunes: maybe short circuit here if subscribers is empty

src/.server/templates/components/button.tsx
```diff
@@ -4,24 +4,26 @@ import React from 'react';
 interface IProps {
   children: React.ReactNode;
   href: string;
+  customStyle?: React.CSSProperties;
```
mariojsnunes, line 7: seems like you are just cleaning up, but couldn't this be done via variant?

src/services/emailContentReachService.server.ts
```diff
@@ -0,0 +1,24 @@
+import type { SupabaseClient } from '@supabase/supabase-js';
+import { DBEmailContentReach } from 'oa-shared';
+
+const getAll = async (client: SupabaseClient): Promise<DBEmailContentReach[] | null> => {
```
mariojsnunes: All other server services are classes, where the client is passed via constructor instead of to every method. can you apply it here as well? otherwise I'll have to do it later.

src/routes/api.email-preview.ts
```diff
@@ -0,0 +1,84 @@
...
+  const draftNotificationDisplay = NotificationDisplay.fromNotification(draftNews);
+
+  try {
+    const claims = await client.auth.getClaims();
+
+    if (!claims.data?.claims) {
+      return Response.json({}, { headers, status: 401 });
+    }
+
+    const { valid, status, statusText } = await validateRequest(request);
```
mariojsnunes: We have a better pattern for these validations. please see other api endpoints.
in summary:
- the validateRequest will throw any error it finds.
- the catch block will check for hono/HttpException (copy paste from other endpoints)

src/routes/api.email-content-reach.ts
```diff
@@ -0,0 +1,25 @@
...
+import { createSupabaseServerClient } from 'src/repository/supabase.server';
+
+const cache = new Keyv<EmailContentReach[]>({ ttl: 3600000 }); // ttl: 60 minutes
+
+export const loader = async ({ request }: LoaderFunctionArgs) => {
+  const { client, headers } = createSupabaseServerClient(request);
+
+  const cachedEmailContentReach = await cache.get('email_content_reach');
+
+  if (cachedEmailContentReach && isProductionEnvironment()) {
```
mariojsnunes: this needs to be mapped before saving in cache.
otherwise causes a bug that I fixed recently on another endpoint. (only works on requests where it doesn't hit the cache)

src/pages/common/FormFields/EmailContentReachNewsField.tsx
```diff
@@ -0,0 +1,61 @@
...
+export const EmailContentReachNewsField = (props: IProps) => {
+  const { placeholder, title, shouldDisableEmailContentReach } = props;
+  const [options, setOptions] = useState<SelectValue[]>([]);
+  const name = 'emailContentReach';
+
+  useEffect(() => {
+    const fetchEmailContentReach = async () => {
+      const emailContentReach = await emailContentReachService.getAll();
+      if (emailContentReach) {
+        const fieldOptions = emailContentReach.map((reach) => {
```
mariojsnunes: I'd do this mapping at the `emailContentReachService` level. avoids doing it everywhere it's called.

src/services/notificationEmailService.server.ts
```diff
@@ -1,55 +1,95 @@
...
+  constructor(private client: SupabaseClient) {
+    Object.assign(this);
+  }
+
+  subscribersToEmail(subscribers: SubscribedUser[], dbNotification: DBNotification) {
+    return subscribers.filter((result) => {
+      // if (result.profile_id === dbNotification.triggered_by_id) {
+      //   return false;
+      // }
+      console.log({ result });
```
mariojsnunes: console log

mariojsnunes (changes requested):

benfurber:
@mariojsnunes This should be a draft still.

src/.server/templates/components/button.tsx
benfurber: This is the email templates - not linked to the theming.

src/utils/isAdmin.ts
benfurber: `SupabaseNotificationsForm`

mariojsnunes:
Current progress, for visibility:
- [x] make hero image full width
- [x] text 100% width
- [x] scroll locking when choosing block type
- [x] some images are not contained in the table - overflow
- [x] Styling on quotes is not aligned (blue on webpage)
- [x] header images has a max height on the platform. But emails shows full height
- [x] line height in email/bullets its not correct
- [x] border in email hero image
- [x] quote font size issue
- [x] buttons don't work - links to prod website
- [x] are tags working?
- [x] add email reach 'no email' option = null
- [x] strict database content_reach column type enum
- [x] remove email_content_reach table and refactor everything that used it
- [x] ensure email reach is mandatory
- [x] add 'moderator' role which can view all news posts
- [x] get_news_feed not respecting tenant constraints
- [x] roles 'admin', 'editor' and 'moderator' must be included in the 'batch' of users who get news emails, but still filtered by their notification preferences
- [x] why is "Regular post" appearing after publish, but the options are different beforehand
- [x] add more reach details to admins on the news form
- [x] public news should also send emails to all badged users (currently it's not sending to anyone)
- [x] Investigate: "Tried to 'publish' and 'save as draft' this article. But it says it already exists. That said after reloading the draft did exist. And I could publish it."

mariojsnunes:
@davehakkens I could not replicate this:
"Tried to 'publish' and 'save as draft' this article. But it says it already exists. That said after reloading the draft did exist. And I could publish it."

But I'm not sure the current behavior is intended. Should we allow returning to draft after already being published? The edge case is that it will currently send the notifications again when publishing again.

davehakkens:
> Should we allow returning to draft after already being published? 

Would say no. You can only edit or delete

And regarding the bug we can merge and continue to test. As the bug is for admins and we have a feedback loop

PR #4694, opened 2026-04-02
fix: list header and items spacing

### LINKED ISSUE #4689: [bug] Spacing on the research is off when the "Sign up" button is showing
**Describe the bug**
Spacing on the research is off when the "Sign up" button is showing.

Image

**Expected behavior**
The inputs should be aligned.
Confirm it doesn't break mobile behavior.

**Potential solutions**
- Consider moving the "total items" below the inputs, aligned on the left.
- Consider shrinking the filter inputs.

mariojsnunes:
## PR Checklist

- [x] 🐛 Bugfix — fixes incorrect behavior without changing functionality

## What is the new behavior?

Fixes lists headers spacings.
- Reduced the size of filter and search inputs
- Changed the custom "Search for a Research" placeholder to be only "Search" (same for Library and Questions)
- Moved the "total items" to always show below the filters
- Changed the "Add Research" text to always be "Add Research" even when signed out (still redirects to login page). Same for Library and Questions.
- The mobile filters now appear on larger screens (but at the same point where our Header changes to Hamburger menu)
- Reduced the spacing between the category picker and the other filters
- Research items now always have a gap between them

Before:
image
image

After:
image
image

Also changed list items to have borders again.
Before:
image

After:
image

- [x] No

## Git Issues

Closes #4689

### DIFF
--- a/src/pages/Research/Content/ResearchList.tsx
+++ b/src/pages/Research/Content/ResearchList.tsx
@@ -103,7 +103,7 @@ const ResearchList = () => {
             m: 0,
             display: 'flex',
             flexDirection: 'column',
-            gap: [0, 2],
+            gap: 2,
           }}
         >
           {researchItemList.map((item) => (


dalibormrska:
I would say let's release, looks good to me for now and the fix is urgent, since currently the layout is a bit broken live.

For details and adjustments I think we need @analordello since she put the most thought into this layout. But if some more changes are needed, they could be a separate issue.

PR #4695, opened 2026-04-02
fix: set minWidth on FieldTextarea to fill container

### LINKED ISSUE #4688: [bug] Question description is shrinking
**Describe the bug**

Image

**Expected behaviour**
The textarea should use all of the space available.

**Potential Solution**
I think issue could be solved by adding `minWidth: '100%'` [here](https://github.com/ONEARMY/community-platform/blob/c05dfefb07cdfa9334504f2eda4e8fee442dcd99/packages/components/src/FieldTextarea/FieldTextarea.tsx#L49)
But the textarea is being used in other places across the platform, need to check that by fixing it here, doesn't break somewhere else.
If it breaks somewhere else, propose a solution to fix both.

tmchow:
## Summary

The textarea in the question description form was shrinking instead of using all the available width. Added `minWidth: '100%'` to the `FieldTextarea` component's `sx` prop so it fills its container properly.

## Changes

One-liner in `packages/components/src/FieldTextarea/FieldTextarea.tsx` (line 60): added `minWidth: '100%'` to the Textarea's `sx` prop.

Checked all 9 usages of `FieldTextarea` across the platform (EditComment, ResearchDescription, LibraryDescription, LibraryStep, UserContact, QuestionDescription, VisitorSection, UserInfos). They all render the textarea inside flex containers that should fill available width, so this change is safe across all contexts.

## Testing

Visual verification against the issue screenshot. The textarea now correctly fills its parent container width.

Fixes #4688

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/packages/components/src/FieldTextarea/FieldTextarea.tsx
+++ b/packages/components/src/FieldTextarea/FieldTextarea.tsx
@@ -57,6 +57,7 @@ export const FieldTextarea = ({
         rows={rows ? rows : 5}
         sx={{
           resize: rest?.style?.resize ? rest.style.resize : 'vertical',
+          minWidth: '100%',
         }}
         {...input}
         {...rest}


mariojsnunes:
@all-contributors add @tmchow for code

mariojsnunes:
thanks @tmchow !

tmchow:
Thanks for the review and merge!

PR #4718, opened 2026-04-13
[news] Updated page styling

benfurber:
## What is the new behavior?

As requested by Dave...

Screenshot 2026-04-13 at 14 47 45

Screenshot 2026-04-13 at 14 49 14

Screenshot 2026-04-13 at 14 49 24


davehakkens:
Is the (easy) possible to release something like this for beta-testers?
Normal users see to old style, beta testers see new style.

As this gives the ability to view it on livesite. 
Can imagine some styling errors will pop-up , this would remove the urgency to fix them

PR #4719, opened 2026-04-16
chore: add some map pins to seed.ts

elbotho:

- [x] 🔖 Other:

## What is the new behavior?

This adds the other profile types and some mockup pins to the database so the map is not empty in local dev.

- [ ] Yes
- [x] No


mariojsnunes:
@all-contributors add @elbotho for code

PR #4724, opened 2026-04-18
feat: improve map module list and pin cards

### LINKED ISSUE #4667: [feature request] Optimise the map view
## Problem

Image

Currently in the map view we have a 50/50 split between the List and the Map. The amount of columns of cards in the List is determined by breakpoints (ex. 1120px). But this means that on small desktop (screenshot) the 3 columns are too tight and the cards are unreadable. This is also the case with the viewport width of around 850px where 2 columns are also too tight.

Basically changing the amount of columns by breakpoints (40em, 52em, 70em = 640px, 832px, 1120px) is not working well for the list cards.

## Proposed solution

There is no fixed solution proposed, you can be creative and find the ideal solution for this issue. Yayy! But the possible approaches that came to my mind include:

1. **Setting min and max widths for the List cards.** This is a strong candidate for a solution, because we can confidently set a limit saying - under a certain width the card would be too packed and also that above a certain width it feels unnecessarily large. In my design I tried between 250px and 400px. Image

2. **Not enforcing the 50/50 split between the map and the list.** I can see a solution where the map can be tighter (taking for ex. 40vw) so that the List cards have enough space and are not tightly packed. This could be done maybe with some css clamp() function magic?

3. **Setting min and max width for the List.** Relevant only for the vw of around 900 where the list could hit a min-width to prevent the List cards from getting even more packed. And the max-width of the List could be utilised for really large and wide screens, so that the cards don't get too wide.

This is not a definitive list of possible options I believe. Feel free to have a go and propose your own solution.

### LINKED ISSUE #4668: [feature request] Improve the MAP PIN cards in the map module
## Context - [Map module](https://community.preciousplastic.com/map)

Image  Image

This issue is part of a bigger intention to improve the map module visually and functionally. The map pin cards could be more useful and optimised visually.

Improve the Member Pin Card and the Space Pin Card as shown in Figma. I prepared the step by step improvements that you can follow.
Image

You will find step by step improvements that you can follow in the [FIGMA FILE](https://www.figma.com/design/NjbfiyDnBAZKdeOUlLejg5/CP---Map-Enhancements?node-id=1398-18644&t=PC3tChVASJsHYLIe-4)

Image

## Additional context
There are links in the Figma File step by step improvement that link to any needed additional context.

### LINKED ISSUE #4669: [feature request] Improve the LIST cards in the map module
## Context - [Map module](https://community.preciousplastic.com/map)
Image

This issue is part of a bigger intention to improve the map module visually and functionally. The map list cards could be more useful and optimised visually.

Improve the Member List Card and the Space List Card as shown in Figma. I prepared the step by step improvements that you can follow.
Image

You will find step by step improvements that you can follow in the [FIGMA FILE](https://www.figma.com/design/NjbfiyDnBAZKdeOUlLejg5/CP---Map-Enhancements?node-id=1446-15151&t=PC3tChVASJsHYLIe-4)

Image

## Additional context
There are links in the Figma File step by step improvement that link to any needed additional context.

rsgb:
## PR Checklist

- [ ] 🐛 Bugfix — fixes incorrect behavior without changing functionality
- [x] ♻️ Refactoring — improves code structure with no functional changes
- [ ] 🔖 Other:

## What is the new behavior?

Three connected issues covering the Map view layout and both the map-pin popup cards and the list cards. Putting them together since they share components (`CardProfile`, `DisplayName`, `ProfileTagsList`).

### #4667 — Optimise the map view

The 50/50 Map/List split combined with breakpoint-driven columns (1, 2, 2, 3) was producing cramped cards around 850px and 1120px viewport widths. Replaced the `columnCount` layout in `MapCardList.tsx` with a CSS Grid using `repeat(auto-fill, minmax(240px, 1fr))`, and clamped each card to `minWidth: 230px` / `maxWidth: 400px` in `CardListItem.tsx`. Column count now adapts to available width instead of fixed breakpoints, and cards never get too narrow to read or absurdly wide on ultra-wide screens.

### #4668 — Improve the MAP PIN cards

Followed the Figma step-by-step for the pin popup:

- **Member pin card** (`CardDetailsMemberProfile.tsx`) — switched to a vertical layout: avatar + name/country block at the top, then tags, then a 2-line clamped `about` blurb. Fixed the card to `width: 325px` in `PinProfile.tsx` so the popup has a predictable footprint.
- **Space pin card** (`CardDetailsSpaceProfile.tsx`) — kept the full-bleed cover image, but dropped the old floating badge pattern and replaced it with an 80px avatar that overlaps the cover (`marginTop: -40px`, white border). The member badge now sits as a small overlay on the avatar corner. Same 2-line about clamp as the member card.
- Removed the custom close `ButtonIcon` from `PinProfile.tsx` and flipped `closeButton={true}` on the Leaflet popup in `Popup.client.tsx` — same affordance, less custom code.

### #4669 — Improve the LIST cards

List cards share `CardProfile` with pin cards but need a tighter layout. Added a `variant?: 'pin' | 'list'` prop threaded through `CardListItem.tsx` → `CardProfile.tsx` → `CardDetailsMemberProfile.tsx`. In `list` mode the tags render next to the avatar/name block and the `about` blurb is hidden, matching the Figma.

### Shared component touch-ups

- **`DisplayName.tsx`** — added a secondary row under the name showing the country flag + country label. Affects every `DisplayName` call site, not just the map — the Figma treats this as the new default.
- **`FlagIcon.tsx`** — added optional `width` / `height` props (defaults unchanged) so the smaller in-card flag can be requested.
- **`ProfileTagsList.tsx`** — reduced tag chip `paddingY` from 5px to 2px to match the Figma chip density. Also applies on the profile page; left as a global change since the Figma treats this as the new default.

### Open question for design

The Figma spec shows 13.33 × 10px for the flag. This is a 4:3 ratio, while country flags are typically 3:2 — @dalibormrska can you confirm this is intentional, or should it be 15 × 10?

### Screenshots

| iPhone 13 | iPad | Macbook | Desktop |
|---|---|---|---|
| iPhone 13 | iPad | Macbook | Desktop |

- [ ] Yes
- [x] No

## Git Issues

Closes #4667
Closes #4668
Closes #4669

## What happens next?

If you need more immediate feedback you can reach out to us on Discord in the [Community Platform \`development\` channel](https://discord.com/channels/586676777334865928/938781727017558018).

### DIFF
--- a/packages/components/src/CardProfile/CardProfile.tsx
+++ b/packages/components/src/CardProfile/CardProfile.tsx
@@ -6,9 +6,10 @@ import { CardDetailsSpaceProfile } from './CardDetailsSpaceProfile';
 export interface IProps {
   item: MapPin;
   isLink?: boolean;
+  variant?: 'pin' | 'list';
 }
 
-export const CardProfile = ({ item, isLink = false }: IProps) => {
+export const CardProfile = ({ item, isLink = false, variant }: IProps) => {
   const { profile } = item;
 
   const isWorkspace = profile?.type && profile?.type.isSpace;
@@ -18,7 +19,7 @@ export const CardProfile = ({ item, isLink = false }: IProps) => {
       {isWorkspace ? (
         <CardDetailsSpaceProfile profile={profile} isLink={isLink} />
       ) : (
-        <CardDetailsMemberProfile profile={profile} isLink={isLink} />
+        <CardDetailsMemberProfile profile={profile} isLink={isLink} variant={variant} />
       )}
     </Flex>
   );
--- a/packages/components/src/CardProfile/CardProfile.tsx
+++ b/packages/components/src/CardProfile/CardProfile.tsx
@@ -3,23 +3,36 @@ import { Flex } from 'theme-ui';
 import { CardDetailsMemberProfile } from './CardDetailsMemberProfile';
 import { CardDetailsSpaceProfile } from './CardDetailsSpaceProfile';
 
+export type CardVariant = 'pin' | 'list';
+
 export interface IProps {
+  cardVariant?: CardVariant;
   item: MapPin;
   isLink?: boolean;
-  variant?: 'pin' | 'list';
+  sendMessageButton?: React.ReactNode;
 }
 
-export const CardProfile = ({ item, isLink = false, variant }: IProps) => {
-  const { profile } = item;
-
-  const isWorkspace = profile?.type && profile?.type.isSpace;
+export const CardProfile = ({
+  cardVariant = 'pin',
+  item,
+  isLink = false,
+  sendMessageButton,
+}: IProps) => {
+  const isWorkspace = item.profile?.type && item.profile?.type.isSpace;
+  const isContactable = item.profile?.isContactable !== false && item.profile?.username;
+  const messageButton = isContactable ? sendMessageButton : undefined;
 
   return (
     <Flex sx={{ alignItems: 'stretch', alignContent: 'stretch' }}>
       {isWorkspace ? (
-        <CardDetailsSpaceProfile profile={profile} isLink={isLink} />
+        <CardDetailsSpaceProfile item={item} isLink={isLink} sendMessageButton={messageButton} />
       ) : (
-        <CardDetailsMemberProfile profile={profile} isLink={isLink} variant={variant} />
+        <CardDetailsMemberProfile
+          item={item}
+          isLink={isLink}
+          cardVariant={cardVariant}
+          sendMessageButton={messageButton}
+        />
       )}
     </Flex>
   );
--- a/packages/components/src/CardProfile/CardProfile.tsx
+++ b/packages/components/src/CardProfile/CardProfile.tsx
@@ -6,33 +6,20 @@ import { CardDetailsSpaceProfile } from './CardDetailsSpaceProfile';
 export type CardVariant = 'pin' | 'list';
 
 export interface IProps {
-  cardVariant?: CardVariant;
+  variant?: CardVariant;
   item: MapPin;
   isLink?: boolean;
-  sendMessageButton?: React.ReactNode;
 }
 
-export const CardProfile = ({
-  cardVariant = 'pin',
-  item,
-  isLink = false,
-  sendMessageButton,
-}: IProps) => {
+export const CardProfile = ({ variant = 'pin', item, isLink = false }: IProps) => {
   const isWorkspace = item.profile?.type && item.profile?.type.isSpace;
-  const isContactable = item.profile?.isContactable !== false && item.profile?.username;
-  const messageButton = isContactable ? sendMessageButton : undefined;
 
   return (
     <Flex sx={{ alignItems: 'stretch', alignContent: 'stretch' }}>
       {isWorkspace ? (
-        <CardDetailsSpaceProfile item={item} isLink={isLink} sendMessageButton={messageButton} />
+        <CardDetailsSpaceProfile item={item} isLink={isLink} variant={variant} />
       ) : (
-        <CardDetailsMemberProfile
-          item={item}
-          isLink={isLink}
-          cardVariant={cardVariant}
-          sendMessageButton={messageButton}
-        />
+        <CardDetailsMemberProfile item={item} isLink={isLink} variant={variant} />
       )}
     </Flex>
   );
--- a/packages/components/src/MapCardList/MapCardList.tsx
+++ b/packages/components/src/MapCardList/MapCardList.tsx
@@ -4,11 +4,9 @@ import { useEffect, useState } from 'react';
 import { Box, Flex, Text } from 'theme-ui';
 import { Button } from '../Button/Button';
 import { CardListItem } from '../CardListItem/CardListItem';
-import type { CardVariant } from '../CardProfile/CardProfile';
 import { Icon } from '../Icon/Icon';
 
 export interface IProps {
-  cardVariant?: CardVariant;
   list: MapPin[];
   onPinClick: (arg: MapPin) => void;
   selectedPin?: MapPin | null;
@@ -21,7 +19,7 @@ const ITEMS_PER_RENDER = 20;
 export const MapCardList = (props: IProps) => {
   const [renderCount, setRenderCount] = useState<number>(ITEMS_PER_RENDER);
   const [displayItems, setDisplayItems] = useState<JSX.Element[]>([]);
-  const { cardVariant = 'list', list, onPinClick, selectedPin, viewport } = props;
+  const { list, onPinClick, selectedPin, viewport } = props;
 
   useEffect(() => {
     setRenderCount(ITEMS_PER_RENDER);
@@ -33,7 +31,6 @@ export const MapCardList = (props: IProps) => {
 
       return (
         <CardListItem
-          cardVariant={cardVariant}
           item={item}
           key={item.id}
           isSelectedPin={isSelectedPin}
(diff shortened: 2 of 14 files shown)


dalibormrska:
Cool one @rsgb, thanks! 

1. To answer your question - in Figma we use slightly different flag library, so it's true that it's not matching exactly what is implemented. This inconsistency was not intentional, and I suggest that in this case we don't need to stick to the Figma specs and rather reuse the flags in the sizes that they are used in the platform already. This is my bad and should have been communicated, sorry.  

2. I see, that mariospace has no flag defined. In the username component it would show the grey flag, but in this case it does not show anything and the spacing is off for the component (visible on hover).
   image
   
   Also on list card, the spacing is off
   image

   I did not think of this problem beforehand, great that we discover this edge case. Since the flag was never set, it's not needed to show the grey flag in this context, nor writing undefined or something. Your approach is in my opinion good, not to show the grey flag, but the spacings need fixing. The desired result is that the displayname without the flag is positioned in the same place as if it would be normally. Like so:
   image
   
   And the hover box is without the extra 5px on the bottom. (in the devtools it meant turning off the gap:5px)
   image


3. There is a little x which shows for member profiles. No need for that. I think the x is also present for the space cards, but just invisible. Remove that as well.
   image

4. Brazil seems to be only notated with BR - maybe check the namings of all countries so that they are full names?
   image

rsgb:
Thanks for the review @dalibormrska! 

Pushed a commit addressing three of the four issues you mentioned:
1: Flag size is now reverted to default
2: Spacing issues are now fixed (country row is now conditionally rendered, so no phantom gap for users without a country)
3: X button removed

4: For the country-code-vs-name issue, I couldn't reproduce "BR"... all profiles and map pins store and show full country names on my end. Could this be something in your test environment?

dalibormrska:
image

Cool, thanks for the changes. I see one more thing, which is that the card is off-centered from the pin. From the devtools it seems like the card overflows the box that is centered to the map pin.

With the BR code - interesting. I have no idea, hopefully on prod it would be good as well.

@mariojsnunes can you do a quick QA? And if possible check what's up with the BR code instead of Brazil?

rsgb:
@dalibormrska good point, thank you — the popup wrapper width is now aligned with the pin card! 

As to the country code, I'll wait for Mario's feedback before considering the issue closed.

mariojsnunes:
Ideally the padding would be around the whole white area. In other words, the button should be inside it.
That way, there is no need add padding to the button as well. and it will be automatically spaced with the gap.

image

This will make it easier to maintain.

mariojsnunes:
as for the country, it's because we are passing in profile data to the component which only stores the country code. we need to pass the map pin data as well.

what needs to change:
`CardDetailsSpaceProfile` and `CardDetailsMemberProfile` need to take in the `MapPin` prop instead of `PinProfile`
then when creating the `DisplayName` component, override the `user.country` prop. (if this is confusing, I'm happy to do it right after the merge)

rsgb:
@mariojsnunes thanks for the feedback! I've addressed both issues you mentioned, please let me know if it looks good.

packages/components/src/MapCardList/MapCardList.tsx
```diff
@@ -19,7 +21,7 @@ const ITEMS_PER_RENDER = 20;
 export const MapCardList = (props: IProps) => {
   const [renderCount, setRenderCount] = useState<number>(ITEMS_PER_RENDER);
   const [displayItems, setDisplayItems] = useState<JSX.Element[]>([]);
-  const { list, onPinClick, selectedPin, viewport } = props;
+  const { cardVariant = 'list', list, onPinClick, selectedPin, viewport } = props;
```
mariojsnunes: is there any situation where the variant is passed into this component? I don't see it in this PR.
If it's never passed down, just pass the 'list' value directly into the `CardListItem`.
Same for the `CardProfile`

packages/components/src/MapCardList/MapCardList.tsx
```diff
@@ -4,9 +4,11 @@ import { useEffect, useState } from 'react';
 import { Box, Flex, Text } from 'theme-ui';
 import { Button } from '../Button/Button';
 import { CardListItem } from '../CardListItem/CardListItem';
+import type { CardVariant } from '../CardProfile/CardProfile';
 import { Icon } from '../Icon/Icon';
 
 export interface IProps {
+  cardVariant?: CardVariant;
```
mariojsnunes: usually we just call 'variant' because we are already in the context of a 'card'

packages/components/src/CardProfile/CardProfile.tsx
```diff
@@ -3,22 +3,36 @@ import { Flex } from 'theme-ui';
 import { CardDetailsMemberProfile } from './CardDetailsMemberProfile';
 import { CardDetailsSpaceProfile } from './CardDetailsSpaceProfile';
 
+export type CardVariant = 'pin' | 'list';
```
mariojsnunes, line 6: I don't like that this type is defined here, maybe should be in `CardDetailsMemberProfile` which is where it is used

packages/components/src/CardProfile/CardProfile.tsx
```diff
@@ -3,22 +3,36 @@ import { Flex } from 'theme-ui';
...
-  const isWorkspace = profile?.type && profile?.type.isSpace;
+export const CardProfile = ({
+  cardVariant = 'pin',
+  item,
+  isLink = false,
+  sendMessageButton,
+}: IProps) => {
+  const isWorkspace = item.profile?.type && item.profile?.type.isSpace;
+  const isContactable = item.profile?.isContactable !== false && item.profile?.username;
+  const messageButton = isContactable ? sendMessageButton : undefined;
```
mariojsnunes: if we are now passing down the `item`, then this `isContactable` logic could be done down at the `CardDetailsSpaceProfile` and `CardDetailsMemberProfile`.
Same for the `sendMessageButton` we are currently passing it down 2 levels, I rather that it just exists down there, because it seems like we have the info available down there anyway.

mariojsnunes (changes requested):

packages/components/src/MapCardList/MapCardList.tsx
rsgb: @mariojsnunes thanks for the feedback! Done.

packages/components/src/MapCardList/MapCardList.tsx
rsgb: Also done, but I kept the type to CardVariant. Let me know if that's ok.

packages/components/src/CardProfile/CardProfile.tsx
rsgb: I initially moved CardVariant into CardDetailsMemberProfile as you suggested, as it was the only component consuming it. But after addressing #4 below, the type ended up being consumed by CardDetailsMemberProfile, CardDetailsSpaceProfile, and the new SendMessageButton. So I moved it back to CardProfile.tsx which is the shared owner. Let me know if this makes sense.

packages/components/src/CardProfile/CardProfile.tsx
rsgb: Done — but I now pulled the send-message button into its own component called `SendMessageButton`, which handles the `isContactable` check and the link itself. The member and space cards now call it directly with `{variant === 'pin' && <SendMessageButton item={item} />}`, so CardProfile doesn't have to pass anything down anymore, and PinProfile doesn't build the button either.

mariojsnunes (approved):

mariojsnunes:
thanks @rsgb !

PR #4746, opened 2026-05-16
fix: decouple seed script from build

benfurber:
I was getting the below error when running `bun run db:seed` (specifically the `npx tsx seed.ts > seed.sql` part of it):

```
/Users/benfurber/projects/community-platform/src/utils/helpers.ts:2
import { UserRole } from 'oa-shared';
         ^

SyntaxError: The requested module 'oa-shared' does not provide an export named 'UserRole'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:228:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:335:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:647:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5)

Node.js v22.18.0
```

I spent some time trying to work out an ordering problem or build problem within the flow but in the end decided just to decouple the app helpers from the seed script.

### DIFF
--- a/seed.ts
+++ b/seed.ts
@@ -27,7 +27,16 @@ import libraryJson from './.snaplet/library.json';
 import questionsJson from './.snaplet/questions.json';
 import { profilesSeed } from './seed/profilesSeed';
 import { usersSeed } from './seed/usersSeed';
-import { convertToSlug } from './src/utils/slug';
+
+// Copied from the helper to decouple the seed script from any app builing processes
+export const convertToSlug = (text: string) => {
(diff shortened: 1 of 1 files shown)


jproberson:
Does that change fix the error in the stack trace? It mentions the oa-shared import. 

I do have a pr: https://github.com/ONEARMY/community-platform/pull/4742

I forgot I did not merge this yet, but I think it should solve that error.

PR #4753, opened 2026-05-21
Keep email after failed sign-in attempt

### LINKED ISSUE #4737: [bug] Keep email after using wrong password on login
**Describe the bug**
When you enter your password wrong, you have to enter your email again. If you fail a couple of times it becomes a bit annoying to re-write the email adress again

**To Reproduce**
Create an account and login with the wrong password

Image

**Expected behaviour**
Keep the email you made the attempt with

daphne951207-bit:
## PR Checklist

- [x] Bugfix — fixes incorrect behavior

When a sign-in attempt fails because the password is wrong, the form is re-rendered from the action response without an initial email value. This clears both the email and password fields, so the user needs to type their email again.

Fixes #4737

## What is the new behavior?

The failed sign-in action now returns the submitted email together with the error message. The sign-in form uses that email as its initial value after the failed submission, while the password field is not restored.

This keeps the email field after a wrong-password attempt, clears the password, and leaves the existing successful sign-in path unchanged.

No.

## Testing

- `git diff --check -- src/routes/_.sign-in.tsx packages/cypress/src/integration/SignIn.spec.ts`
- `biome check --verbose src/routes/_.sign-in.tsx`
- `tsc --noEmit --pretty false`
- `cypress verify`

I did not run the full Cypress E2E suite locally because the current E2E startup flow depends on the project test app and Supabase test environment variables. I added a focused Cypress regression spec for the intended behavior.

### DIFF
--- a/packages/cypress/src/integration/SignIn.spec.ts
+++ b/packages/cypress/src/integration/SignIn.spec.ts
@@ -1,4 +1,5 @@
-import { generateNewUserDetails } from '../utils/TestUtils';
+import { MOCK_DATA } from '../data';
+import { generateNewUserDetails, getTenantUser } from '../utils/TestUtils';
 
 describe('[Sign in]', () => {
   it('[By Anonymous]', () => {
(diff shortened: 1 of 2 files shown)


mariojsnunes:
thanks @daphne951207-bit !
@all-contributors add @daphne951207-bit for code

PR #4754, opened 2026-05-21
Fix login email reset after failed password attempt

### LINKED ISSUE #4737: [bug] Keep email after using wrong password on login

chriszlr:
## PR Checklist

- [x] Unit and/or e2e tests added where appropriate
- [x] Code style and type checks reviewed locally

- [x] Bugfix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation

When a login attempt fails because the password is wrong, the sign-in form is re-rendered without the previously entered email address.

## What is the new behavior?

The attempted email is returned with the auth error response and used as the form initial value, so users only need to re-enter their password. The password field still resets after a failed attempt.

- [ ] Yes
- [x] No

## Git Issues

Closes #4737

## Validation

- `bunx biome check src/routes/_.sign-in.tsx`
- `bunx tsc --noEmit --pretty false`
- `bun run build:shared`


mariojsnunes:
heey @chriszlr, thanks for your contribution!

The code looks good, but there is a failing e2e test which seems related to the change.

mariojsnunes:
close as https://github.com/ONEARMY/community-platform/issues/4737 already fixed the issue

PR #4799, opened 2026-07-07
feat: add marker styling for selected pin

### LINKED ISSUE #3854: feat: update map pin icons
### Description

A cute little update to the map pin icons

### Design notes and requirements by @dalibormrska 

This issue talks only about the map pins. The issue of clusters will be in scope of a different issue later on.

**1. First requirement**

This is the behaviour that we would like to achieve:
https://codepen.io/vsync/pen/mdEJMLv

First of all, by default the old map pins were funky circles with a lot of vector points, but since they are being displayed so small, they were practically invisible and my assumption was that they were slowing down the interface by taking computing power to display all the vector points. So I simplified the pins to pure circles. I also simplified the symbols inside the circles to be more lightweight.

image

I based it on a modular idea, since the two circles and the animation of opening the pin would be the same across all the pins. The only thing that would change is the color of the 2 circles and the icon in the center.

image

However, I don't know how implementable this modular idea is. Have a look at what would be a good solution and let me know, I can provide you with any SVGs you might need.

**2. Second requirement**

For the user it would be very useful to see the relation between the cards and the map pins. Therefore we would love to have a simple interaction animation that highlights the hovered map pin. If the pin is in a cluster, I would say to highlight the whole cluster.

image

We DON'T WANT the map pins to be hoverable by themselves, it would create a lot of distraction while just hovering over the map. And we also don't need to have the vice-versa interaction (that on hover of the map pin, the card would get highlighted, since the user would get the same information by just clicking the map pin).

### Build suggestion

* The function to update is `createMarkerIcon` in `src/pages/Maps/Content/MapView/Sprites.tsx`
* How to do the onclick/select effect might require some thought.

benfurber:
## What is the new behavior?

Add basic styling for a selected map pin. Could be enhanced in the future, but even getting it to this point was a suprising amount of work with a lot of trial and error.

https://github.com/user-attachments/assets/85528e47-5965-494f-8934-d499c8330753

- [x] Yes
- [ ] No

## Git Issues

Closes #3854 #3854


supabase/migrations/20260707140410_add_background_color_to_profile_types.sql
```diff
@@ -0,0 +1 @@
+alter table "public"."profile_types" add column "background_color" text;
```
mariojsnunes, line 1: don't forget to change the schema too

davehakkens:
why does your map card chop of the avatar?
image

benfurber:
> why does your map card chop of the avatar? image

Unrelated to anything I've touched.

davehakkens:
what data do you use?
Its a machine shop icon with a member profile layout. Is it special testing data?


image

mariojsnunes:
I wonder if it's possible to do this without having to set the background color for the icon on the database. it will be annoying to maintain.

Edit: Doesn't seem to be possible.

@benfurber if it's not configured, it should behave same as before this PR.

mariojsnunes:
Maybe we could just make the selected pin bigger or highlighted in some other way. To avoid the extra database configuration.

benfurber:
> Maybe we could just make the selected pin bigger or highlighted in some other way. To avoid the extra database configuration.

@davehakkens What do you think? Pro of this is that it's easier to maintain (e.g. it won't really need to be changed again). Con is I've done this work already(!)

davehakkens:
How much work/hours went in it? @benfurber

mariojsnunes:
I think the question should be:

1. design and implement another style that doesn't require the background color db setting
2. or maintain the setting for the foreseeable future

Having work done (although it would feel bad to throw away) is not that relevant and some of it can be reused.

PR #4813, opened 2026-07-27
fix: hide membership prices for archived Stripe products

jproberson:
## PR Checklist

- [x] 🐛 Bugfix — fixes incorrect behavior without changing functionality
- [ ] ✨ Feature — adds new functionality
- [ ] ♻️ Refactoring — improves code structure with no functional changes
- [ ] 🔖 Other:

## What is the new behavior?

Archiving a product in the Stripe dashboard now removes its tiers from the Membership pricing page.

Stripe keeps a product's prices `active: true` even after the product itself is archived, so `getPrices()` was leaking prices for archived products (e.g. the €0.50 test tier — confirmed against live data: `price_active=true` on a `product_active=false` product). We now expand `data.product` on `prices.list` and drop any price whose parent product is archived or deleted. Conservative: prices whose product can't be resolved are kept, so no legitimate tier is hidden.

- [ ] Yes
- [x] No

## Git Issues

Closes #

## What happens next?

If you need more immediate feedback you can reach out to us on Discord in the [Community Platform \`development\` channel](https://discord.com/channels/586676777334865928/938781727017558018).

### DIFF
--- a/src/services/stripeService.server.ts
+++ b/src/services/stripeService.server.ts
@@ -373,13 +373,20 @@ export class StripeServiceServer {
           product: productId,
           active: true,
           limit: 100,
-          expand: ['data.currency_options'],
+          expand: ['data.currency_options', 'data.product'],
         }),
       ),
     );
 
     return allPrices
       .flatMap((result) => result.data)
+      .filter((p) => {
+        const prod = p.product as { active?: boolean; deleted?: boolean } | string;
+        if (typeof prod === 'object' && prod !== null) {
+          return prod.deleted !== true && prod.active !== false;
+        }
+        return true;
+      })
       .filter((p) => p.recurring && p.unit_amount !== null)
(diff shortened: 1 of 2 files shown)


src/services/stripeService.server.ts
```diff
@@ -373,13 +373,20 @@ export class StripeServiceServer {
...
       .flatMap((result) => result.data)
+      .filter((p) => {
+        const prod = p.product as { active?: boolean; deleted?: boolean } | string;
+        if (typeof prod === 'object' && prod !== null) {
+          return prod.deleted !== true && prod.active !== false;
+        }
+        return true;
+      })
       .filter((p) => p.recurring && p.unit_amount !== null)
       .flatMap((p) => {
```
mariojsnunes, line 415: note that this will loop 4x. should be ok since there are only a few items.

Issue #4826, opened 2026-08-10
[platform improvement] Add landmark regions to all pages


benfurber:
At the moment, we don't use html5 landmark regions impacting both accessibility and probably search crawler performance.

This is really an epic to start adding the reasons to all pages on the platform.

To do:

- [ ] Global header, navigation and footer
- [ ] News index
- [ ] News page
- [ ] Projects index
- [ ] Project page
- [ ] Research index
- [ ] Research page
- [ ] Questions index
- [ ] Questions page
- [ ] Map page?
- [ ] Support page

~~ALSO: As components are changed, please move them from the old component library (`packages/components`) into the new component library (`src/components`)~~

mariojsnunes:
Note I already made a few changes towards this here
https://github.com/ONEARMY/community-platform/pull/4819/changes#diff-df8410de8e5fddaa1ba22946746d5da76c83493c91e1e3f8ba1a6b6685075b38

I wanted to try, and since they were so simple, kept them

yulinlina:
Since this is an epic and touches a lot of pages, I think it helps to have one short reference doc so contributors apply the same pattern. I'd add docs/accessibility/landmark-regions.md with the landmark mapping, a checklist for the listed pages, and a reminder to move components from packages/components to src/components when editing them. That won't close the epic, but it gives a safe starting point and review criteria for follow-up PRs.

Iyamokuma:
Hi, can you assign this issue to me?

yulinlina:
Sure — I can’t assign it from my side, but you don’t need to wait for a formal assignment. Just comment here with the specific page/component you’re starting on so others can avoid duplicate work.

Since this is an epic, I’d suggest splitting it into small PRs rather than trying to do all pages at once. A good first PR could be:

1. Add `docs/accessibility/landmark-regions.md`
2. Implement global landmarks for the shared layout:
   - site header → `<header>`
   - primary navigation → `<nav>`
   - page content → `<main>`
   - site footer → `<footer>`

Useful pattern:

```tsx
<body>
  <header role="banner">
    <nav aria-label="Primary">...</nav>
  </header>

  <main id="main-content">
    {children}
  </main>

  <footer role="contentinfo">...</footer>
</body>
```

Notes:
- Usually only one `<main>` per page.
- Give navigation meaningful `aria-label` values, e.g. `Primary`, `Footer`, `Breadcrumbs`.
[...]

PR #4832, opened 2026-08-11
Add landmark regions accessibility guide

yulinlina:
Adds a short reference guide for landmark regions under docs/accessibility. It captures the shared global layout pattern, a page checklist from the epic, and the reminder to move touched components from packages/components to src/components. This gives contributors a consistent starting point without duplicating the merged layout work.

Addresses #4826

### DIFF
--- a/docs/accessibility/landmark-regions.md
+++ b/docs/accessibility/landmark-regions.md
@@ -1,0 +1,51 @@
+# Landmark regions
+
+Use semantic HTML5 landmark regions to help assistive technology and search crawlers understand page structure. Prefer native elements before ARIA roles.
+
+## Global layout
+
+The shared layout should provide:
+
+- Site header: `<header>`
+- Primary navigation: `<nav aria-label='Primary'>`
+- Page content: `<main id='main-content'>`
+- Site footer: `<footer>`
+
+Example:
+
+```tsx
+<header>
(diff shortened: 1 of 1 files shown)


mariojsnunes:
not merging, reasoning here: https://github.com/ONEARMY/community-platform/issues/4826#issuecomment-5250858599

yulinlina:
Thanks for linking the reasoning. Given that context, I agree this shouldn’t be merged as-is.

The main risk with a standalone landmark guide is that it becomes a second source of truth that can drift from the already-merged layout implementation. The useful parts of this PR are not really the prose, but the contributor checklist:

1. Pages should use the shared global layout landmarks consistently (`header`, `nav`, `main`, `footer`).
2. If a page introduces repeated landmark regions, they should be distinguishable with an accessible name, e.g. `aria-label`.
3. Touched components should continue to be migrated from `packages/components` to `src/components` where appropriate.

Instead of a separate guide, those points can be captured directly in #4826 as acceptance criteria/contributor notes, or linked from existing documentation if maintainers want a lighter-weight reference.

Since the PR is closed, I won’t push further on this approach. If a minimal follow-up is preferred later — just a short pointer in the existing contributor docs without duplicating implementation details — that seems like the safer path.

PR #4868, opened 2026-08-28
chore: add signup captcha

mariojsnunes:
## PR Checklist

## What is the new behavior?

Adds cloudflare turnstile to the signup form. Chosen because:
1. Supabase has a native integration with it
2. Seamless UX for users, just an additional click
3. Free
4. Good analytics
5. Works well locally and on CI (we use a non-interactive "always passes" test key)

image

image

image

- [x] No


davehakkens:
Does this effect supporters that sign up?

PR #4870, opened 2026-08-29
feat(research): show contributor avatars with a contributors modal

### LINKED ISSUE #4866: Research Contributors UI improvement
### Problem

Current UI is too crowded and brings too much information to the header.

Image

### Proposal
Show only up to 3 contributors avatars and bring the contributors list to a modal.

Image

### More details:

When only 1 contributor, show the avatar and the username

Image

When 2 contributors, show 2 avatars

Image

When 3+ contributors, show 3 avatars and the number of contributors

Image

Hover state

Image

Contributors modal

Image

Modal

Image

Mobile

Image

## Documentation
All details and specifications are on this Figma file:
https://www.figma.com/design/2NpVYccYfQmdMctlFZS71D/Research?node-id=2858-3877&p=f&t=1KnCTZj6RJfQVTyU-0

tisankanj:

- [x] ✨ Feature

## What changed

The research header listed every contributor inline. It now shows up to 3 avatars and an `N contributors` button that opens a modal listing everyone. One contributor still renders as avatar plus username, and none renders nothing.

Contributors also carry their photo now. `getUsersByUsername` selected the `photo` column, but `getCollaborators` called `Author.fromDB(user)` without it, so every avatar fell back to an initial. It now passes the public URL through `ImageServiceServer`, matching `CommentFactory.createAuthor`.

Since the last review: the modal border is 2px, the hover is softblue `#e2edf7` behind `#97cdeb` at a 4px radius sampled from the Figma node, the font override is reverted so the row uses the component default, and the code comments are gone per `AGENTS.md`. `src/components/ui/avatar.tsx` and its story are no longer in this branch: #4895 added the same component to master, so this branch takes that one and only the Research files remain.

## Screenshots

Pending. The preview for this commit is sat at `waiting` on the protected environment, and the deployment that is currently live is the older build, so a screenshot of it would not show the state of this branch.

https://github.com/ONEARMY/community-platform/actions/runs/35053233430

Once someone with write access releases it I will attach before, after and 375px.

## Issue

Closes #4866

## Database

## Checklist

- [ ] I ran the app and checked this change myself

The first box is unticked on purpose. I verified the earlier build in a browser, but not this exact commit, for the reason above. `bun run test:unit` is 437 passing across 70 files, and `tsc --noEmit` is clean.

---
<table><tr>
<td></td>
<td><strong>Tisankan Jeyakumar</strong><br/>
Developed and verified</td>
</tr></table>


analordello:
Hey! It's looking great, I have just some small feedback:
1. The font in “with” should be inter
2. the spacing between “with” and the usernames should be 5px
3. The contributor's avatars should be 25px, not 30px.
PR:
image
Figma:
image

4. The hover interaction is different from the design in figma

Current
image

Figma
image

5. Why is the modal so different from the Figma proposal? It's not using the same fonts, border, and avatar size.
image
image

6. When I click on the modal, the first user is automatically selected
image

@rascal-sl @mariojsnunes

rascal-sl:
Thanks @analordello, all six addressed in f9d7b8e.

**1. Font.** You were right, and the cause is worth naming: the page around this row is theme-ui, whose body font is `'Inter', Arial, sans-serif` (`packages/themes/src/common/commonStyles.ts:50`), while the newer Tailwind component library sets `--font-sans: 'Geist Variable'` (`src/styles/ui-globals.css:12`). My component was the only text in that row rendering in the library's font instead of the page's. The label and the modal now use Inter explicitly.

**2. Spacing.** 5px between "with" and the contributors, and between the avatars and the count.

**3. Avatars.** 25px in the header row. They were 24px plus a 2px ring on each side, which is where the 30px you measured came from.

**4. Hover.** Now a bordered pill instead of an underline.

**5. Modal.** Black border, a divided header, tighter corners, and 32px avatars in the list, rather than the component library's default card. On the "why so different": the modal was built on the shared `Dialog` and I took its defaults as-is instead of styling it to the design, which is my mistake rather than a deliberate choice.

One limit worth flagging: I cannot open the Figma file, so I matched your screenshots. If any of the type scale, border weight or row spacing is still off against the file, tell me the values and I will set them exactly rather than keep eyeballing it.

**6. The focus bug.** Real, and thanks for catching it. Base UI moves focus to the first tabbable element in the popup by default, which was the first contributor's link, so the row rendered as selected. `initialFocus` now points at the popup itself, which keeps the dialog keyboard reachable and Escape-dismissable without singling out a person.

That one has a regression test. Removing the fix fails it:

```
× does not put focus on a contributor when the modal opens
  Tests  1 failed | 5 passed (6)
```

Full suite is green:

```
$ bun run test:unit -- --run
Test Files  61 passed (61)
     Tests  340 passed (340)
```

Still not verified in a running browser, since I do not have a local Supabase instance, so the visual side is worth a second look on the preview once CI redeploys.

benfurber:
Hi @rascal-sl, thanks so much for your work on this. @mariojsnunes will review the code properly, but I did notice it we'd appreciate some changes to the component location. It wasn't documented in the issue but we'd appreciate if you could follow our new [component creation guide](https://github.com/ONEARMY/community-platform/blob/master/src/components/ui/README.md).

rascal-sl:
Thanks @benfurber, that is a fair catch and it is done in `eb315d7`.

I had hand-rolled the avatar inside `ResearchContributors.tsx`, which is exactly the kind of thing the guide says should be a library component. It is now `src/components/ui/avatar.tsx`, built on `@base-ui/react/avatar`, with a story alongside it like the other primitives. The feature file just composes it:

```tsx
<Avatar aria-hidden="true" className={cn(className, 'ring-2 ring-background')}>
  {contributor.photo ? <AvatarImage alt="" loading="lazy" src={contributor.photo.publicUrl} /> : null}
  <AvatarFallback>{name.charAt(0).toUpperCase()}</AvatarFallback>
</Avatar>
```

A few notes on how it follows the guide:

- **No new dependency.** `@base-ui/react` was already there and ships `avatar`, so this is a new file only, no lockfile change.
- **Typed like the existing primitives.** `AvatarPrimitive.Root.Props` and friends, matching how `dialog.tsx` types its parts.
- **It also fixes a real gap.** The old markup rendered a bare ``, so a photo URL that 404s left a broken image. `AvatarFallback` now covers both the no-photo and failed-load cases, and there is a story for the broken-image case.
- **`aria-hidden` kept on the avatar.** The name is always rendered beside it and the trigger has its own `aria-label`, so it is decorative. That also keeps the existing test that counts the three stacked avatars honest.

I left `ResearchContributors.tsx` itself under `src/pages/Research/Content/`, since `src/components/ui/` currently holds only generic primitives and this one is Research-specific. If you meant the whole component should move as well, say so and I will move it.

Verified: `tsc --noEmit` clean, `biome check` clean, and the Research content suite passes 15 of 15 including the avatar-count and focus tests. Storybook entry is `ui/Avatar`.

analordello:
Hey @rascal-sl , I tested the [figma link](https://www.figma.com/design/2NpVYccYfQmdMctlFZS71D/Research?node-id=2858-3877&p=f&t=1KnCTZj6RJfQVTyU-0) and it was working for me, what problem did you face?

Please let me know (tag me) when the fixes are deployed to preview, so I can do the last check.

Thanks for your work :)

rascal-sl:
Thanks @analordello. Two answers.

**The Figma link.** It returns HTTP 403 for me:

```
$ curl -sIL "https://www.figma.com/design/2NpVYccYfQmdMctlFZS71D/Research?node-id=2858-3877"
status=403
```

So it is not public. It opens for you because your account has access to that project, which is why it looked fine on your side. If you can set the file to "anyone with the link can view", I will match the spec exactly. Otherwise the fastest path is you giving me the four numbers I was eyeballing: header type size, border weight, row gap, and the avatar sizes in the list. I would rather set them than keep guessing from screenshots.

**The preview.** I cannot trigger it. The `Deploy Fly PR Preview` run for `eb315d7` is sitting at `status=waiting` on the protected `preview` environment:

```
env=preview  current_user_can_approve=false
```

It is a `pull_request_target` workflow with an environment gate, so a maintainer has to approve the deployment before it builds. Once you or @benfurber release it, it will deploy `eb315d7` and I will tag you.

Everything else on the PR is green: Build, Storybook, Lint, Unit tests, all four e2e-electron shards, and Cypress at 119 passing.

Since your last look I also moved the avatar into `src/components/ui/avatar.tsx` per @benfurber's note, so the modal and the row now use the shared primitive rather than markup local to this feature.

analordello:
@rascal-sl the link is opening for me even when I open on an incognito window. Strange. But let's try a different approach. Do you have a Figma account? I'll share it with you. I know it won't make much difference now since you're already done with the fixes, but I still want to solve this problem for future contributions.

rascal-sl:
You are right and I was wrong. The file is public, and no account share is needed. Sorry for sending you looking.

The cause was on my side. My first check used `curl` with its default user agent, and Figma's CDN blocks that, so I got a 403 and read it as a permissions problem. It is a bot block. Same URL, same machine, only the user agent changed:

```
curl -I "https://www.figma.com/design/2NpVYccYfQmdMctlFZS71D/Research?node-id=2858-3877"
status=403

curl -I -A "Mozilla/5.0 ... Chrome/141.0.0.0 Safari/537.36" <same url>
status=200
```

I have since opened it in a real browser and the canvas renders fine.

So for future contributions the answer is simply that I need to open Figma in a browser rather than fetch it, and the "I cannot open the file" caveat on my earlier review reply should be struck. That caveat was wrong, and any place I said I was matching screenshots because the file was unavailable was me working around a problem I had created.

I would still rather you gave the last check on the preview than have me re-derive the spec now, since you have already reviewed the visual details once. The preview is still waiting on the environment approval I mentioned, so it is with you or @benfurber whenever suits.

analordello:
Sure! I'll be waiting on the preview 😄

rascal-sl:
One thing so you are not waiting on something that will not arrive: the preview will not build by itself. The run for `eb315d7` has been sat at `waiting` since 13:54 because the `preview` environment needs a deployment approval, and I cannot give it from a fork.

https://github.com/ONEARMY/community-platform/actions/runs/33970247942

Whoever has write access can hit "Review deployments" there and approve `preview`. It should come up shortly after, and I will tag you once it is live.

analordello:
Yeah I know, I'm aware of it 😅 
No worries 🤙

benfurber:
@rascal-sl - Sorry for the delay. Released the preview now.

mariojsnunes:
@analordello please check again :)

src/pages/Research/Content/ResearchContributors.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { useRef, useState } from 'react';
+import { Avatar, AvatarFallback, AvatarImage } from '@/components/ui/avatar';
+import { Dialog, DialogContent, DialogHeader, DialogTitle } from '@/components/ui/dialog';
+import { cn } from '@/lib/utils';
+
+const MAX_VISIBLE_AVATARS = 3;
+
+// The surrounding page is theme-ui, whose body font is Inter, while this
+// component library defaults to Geist. Match the page rather than the library.
+const FONT_FAMILY = 'font-[Inter,Arial,sans-serif]';
```
mariojsnunes: this should not be needed, it should already be applied globally

src/pages/Research/Content/ResearchContributors.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+interface AvatarProps {
+  contributor: Author;
+  className: string;
+}
+
+// Decorative: the name is always rendered next to it, and the trigger carries
+// its own aria-label, so the avatar adds nothing for a screen reader.
```
mariojsnunes: not sure I understand this comment. why isn't the aria-label used?

mariojsnunes (changes requested):

src/pages/Research/Content/ResearchContributors.tsx
rascal-sl: It is needed, and my earlier note had the reason wrong. Measured on the preview a moment ago:

```
html      "Geist Variable", sans-serif        ui-globals.css:133  html { @apply font-sans }
body      "Varela Round", Arial, sans-serif   GlobalStyles.tsx:7
Username  Inter, Arial, sans-serif            Username.tsx:36  sx={{ fontFamily: 'body' }}
```

So the global page font is Varela Round. Inter is not applied globally at all. It reaches the usernames because `Username` opts into the theme-ui `body` font token, which is Inter. Removing the class here drops this row to Varela Round sitting directly beside Inter usernames, which is the mismatch @analordello reported first time round.

You are right that a hardcoded font stack does not belong in a feature file, though. `7391bd0a` adds a `--font-body` token to `ui-globals.css`, mapped to the same Inter stack `oa-themes` uses, and this line is now just `font-body`. That follows the pattern the library README already describes for tenant colours, where this library maps its tokens onto the values the legacy theme reads rather than keeping a second set.

If you would rather the shared stylesheet stayed untouched, say so and I will put the literal back here instead.

src/pages/Research/Content/ResearchContributors.tsx
rascal-sl: Fair, that comment was badly written. It was explaining the `aria-hidden` on the `Avatar` below it and never actually said so.

A label would make a screen reader announce the same person twice, because the name is already text right next to the avatar everywhere this renders:

- the single-contributor row, where `Username` sits beside it
- the modal list, same thing
- the trigger button, which carries its own `aria-label` of "Show all N contributors" and so replaces its contents for a screen reader anyway

So the image is decorative. Hiding it is what stops the repetition, and an `aria-label` on it would cause the repetition.

Reworded in `7391bd0a` to say that directly.

rascal-sl:
Thanks @mariojsnunes, both addressed in `7391bd0a`. Details are in the two threads, short version here.

**The font.** It is not applied globally, and my earlier explanation of why was wrong as well. Measured on the running preview:

```
html      "Geist Variable", sans-serif        ui-globals.css:133
body      "Varela Round", Arial, sans-serif   GlobalStyles.tsx:7
Username  Inter, Arial, sans-serif            Username.tsx:36
```

The page font is Varela Round. Inter is not global at all, it reaches the usernames because `Username` opts into the theme-ui `body` token. So the label does have to opt in, but you are right that the stack should not be written into a feature file. There is now a `--font-body` token in `ui-globals.css` and the component just uses `font-body`.

**One change you did not ask for, from the same root cause.** The modal was rendering in two fonts. `DialogTitle` carries `font-heading`, which aliases `--font-sans`, so the header was Geist while the contributor list below it was Inter:

```
title  "Geist Variable", sans-serif
list   Inter, Arial, sans-serif
```

That is the second half of @analordello's point 5 and I had only fixed the first half. The title is now on `font-body` too. Say the word if you would rather that were a separate PR.

**The avatar comment.** It was explaining the `aria-hidden` below it and never said so, which is why it read as a puzzle. A label there would make a screen reader say the person's name twice, since the name is already text beside the avatar in every place this renders. Reworded to say that.

Verified: `tsc --noEmit` clean, `biome check` clean, Research content suite 15 of 15 including the avatar-count and focus tests.

@analordello the preview will not pick this up on its own. The run for `7391bd0a` is sat at `waiting` on the protected `preview` environment again:

https://github.com/ONEARMY/community-platform/actions/runs/34458884428

Once someone with write access approves it, I will tag you.

analordello:
Hey @rascal-sl 

For the font tokens and aliases issues, my recommendation is the following:
- Instead of having only one --font-sans token, have a --font-varela token and a --font-inter token
- Create 3 font tokens(or styles): font-title (which aliases --font-varela), font-label (for labels in buttons and interactive components, which aliases --font-varela too) and font-body (which aliases --font-inter)

This would be a separate issue though, but once it is set we'll ensure consistency and more likely resemble the old theme-ui structure.


Here's other things I noticed:

1. The border width should be 2px, instead of 1px.
image

2. The hover state does not yet resemble the intendend design

Preview:
image

Design:
image

Exact figma file: https://www.figma.com/design/2NpVYccYfQmdMctlFZS71D/Research?node-id=2858-6113&t=UM9IEuvPWCw03Suu-0

mariojsnunes:
Let's keep the font out of this PR, just use whatever is the default of the component.
@analordello we can align it in a different issue.

@rascal-sl AI isn't good at UI. if you don't put the minimum effort, it's going to be a waste of time for all of us. and please, keep the messages short. AI default mode is verbose, it's not useful.

rascal-sl:
Done in `e8f20ea3`.

1. Modal border is 2px.
2. Hover is `softblue` `#e2edf7` behind a `#97cdeb` border at a 4px radius, sampled off the Figma node you linked. Both already exist in `oa-themes` as `softblue` and `highlightHover`, and the theme's own quiet buttons use the same pair. Left as literals since this library has no token for them.
3. Font reverted, token included. The row is back on the component default, so it will render Varela Round until the token issue lands.

Verified the hover against the design rather than by eye: fill `#e2edf7`, border `#95d4ec` in your screenshot after compression, radius 4px.

tisankanj:
Rebased on master, `d047206b`.

1. The conflict was `src/components/ui/avatar.tsx`. #4895 added the same component, so I dropped mine and took master's. This branch is now only the Research files.
2. Code comments removed, per the `AGENTS.md` that landed since my last push.
3. PR description rewritten to the new template, under the word cap.
4. Screenshots still owed. The preview for this commit is at `waiting`, and the deployment currently live is the older build, so it does not show the hover and border fixes. Release it and I will attach before, after and 375px.

https://github.com/ONEARMY/community-platform/actions/runs/35053233430

`bun run format` clean, `bun run test:unit` 437 passing, `tsc --noEmit` clean.

---
<table><tr>
<td></td>
<td><strong>Tisankan Jeyakumar</strong><br/>
Verified review</td>
</tr></table>

analordello:
Hey! Everything is all good, except one small detail:

You did adjust the border width, but could you also adjust the separator width to 2px so it stays consistent with the platform?
image

tisankanj:
Done in `22e37dab`.

Good catch. Measured on the preview before changing it, the two really were different:

```
dialog outer border   2px
header separator      1px
```

The header was still `border-b`, so only the outer border picked up the earlier change. It is `border-b-2` now and both are 2px.

`bun run format` clean, `bun run test:unit` 437 passing, `tsc --noEmit` clean.

The preview for this commit is back at `waiting`. Release it and I will attach the before, after and 375px screenshots the description still owes.

https://github.com/ONEARMY/community-platform/actions/runs/35305901909

---
<table><tr>
<td></td>
<td><strong>Tisankan Jeyakumar</strong><br/>
Verified review</td>
</tr></table>

mariojsnunes:
I still see a few UI issues, but I will merge them and fix myself.

PR #4900, opened 2026-09-19
chore: migrate alert to new UI library

### LINKED ISSUE #4885: [UI Migration] Alert
## Context
We're migrating components from `oa-components` (theme-ui) into `src/components/ui/` (shadcn generated on Base UI, styled with Tailwind v4), following the strangler-fig approach described in the [[Platform UI README](https://github.com/ONEARMY/community-platform/blob/master/src/components/ui/README.md)](https://github.com/ONEARMY/community-platform/blob/master/src/components/ui/README.md). This is not a big-bang rewrite — this ticket migrates **one component only**. `oa-components` is not being deleted as a whole; only the piece migrated in this ticket is removed.

The goal is near-identical UI and behaviour. If the new implementation can't (or shouldn't) match the old one exactly, we've tried to note it here. If you need to make a call to make a minor change, please note why in your PR.

Reference: [[Alert in the legacy Storybook](https://storybook.onearmy.earth/?path=/story/components-alert--info)](https://storybook.onearmy.earth/?path=/story/components-alert--info) · [[Alert in shadcn/ui](https://ui.shadcn.com/docs/components/alert)](https://ui.shadcn.com/docs/components/alert)

## Scope
- **Old component:** `packages/components/src/Alert/Alert.tsx`
- **New component:** `src/components/ui/alert.tsx`
- **Base UI / shadcn primitive(s) used:** `Alert` (`bunx shadcn@latest add alert`) — this is plain markup styled with `cva`, not wired to a Base UI primitive, so there's no interaction/accessibility behavior to reconcile, only styling and variants

## Known differences
- **Variant coverage:** the old `Alert` supports `info` / `error` / `success` / `warning` via a single `variant` prop. shadcn's `Alert` ships only `default` / `destructive`. This ticket needs to extend the component's `cva` variants to add `success`, `warning`, and `info` (map `error` → `destructive`) rather than dropping any of the old states — flag in your PR the Tailwind colors chosen for each, since `success`/`warning`/`info` aren't tenant/DB-driven (only `primary`/`accent` are) and will be static rather than theme-able per tenant.
- **API shape:** the old `Alert` is a single flat component (e.g. `<Alert variant="info">Message</Alert>`). shadcn's `Alert` is composed of `Alert` + `AlertTitle` + `AlertDescription` (and optionally an icon as a direct child, and `AlertAction` for a trailing button). This means call sites need their JSX restructured, not just their import path swapped — decide per call site whether the old content maps to just `AlertDescription`, or a `AlertTitle` + `AlertDescription` split, and note any call sites where that split isn't obvious.
- **Icon:** shadcn's `Alert` expects an icon as a direct child (any `svg`, no dedicated prop) which shifts the layout into a two-column grid automatically. Check whether the old `Alert` rendered a per-variant icon (info/warning/error/success glyphs) — if so those will need porting as new SVGs under `src/components/ui/icons/` following the existing convention, since none of the 6 already-ported icons cover this.

## Screenshots / video
Attach to the PR, not just this ticket description, so reviewers see them in context:
- [ ] Old component — default state
- [ ] New component — default state
- [ ] Old vs new — any other relevant states (hover, focus, disabled, error, open/expanded, loading, dark mode if applicable, each tenant theme if colors are involved)
- [ ] If the interaction itself changed (animation, transition, keyboard behavior), a short screen recording of old vs new

## Acceptance criteria
- [ ] Implementation follows the conventions in the [[Platform UI README](https://github.com/ONEARMY/community-platform/blob/master/src/components/ui/README.md)](https://github.com/ONEARMY/community-platform/blob/master/src/components/ui/README.md) (component location, Base UI primitives, Tailwind theming, icon pattern)
- [ ] Story added to the **new** Platform UI Storybook (`bun run storybook:ui`, port 6008), not the legacy one
- [ ] Old component removed from `packages/components/` (source file, its story, any icon/asset used only by it)
- [ ] All call sites updated to import from `src/components/ui/` instead of `oa-components`
- [ ] Screenshots/video attached per above

DevShoaib78:

- [ ] 🐛 Bugfix
- [ ] ✨ Feature
- [x] ♻️ Refactoring
- [ ] ⚡️ Performance
- [ ] 🧪 Tests
- [ ] 🔧 Tools / CI
- [ ] 📝 Documentation
- [ ] 📦 Dependencies

## What changed

Migrates `Alert` to `src/components/ui/alert.tsx`, with a story and a unit test, and switches the seven call sites in `src/`.

To flag: `packages/components/src/Alert/Alert.tsx` does not exist. The legacy Alert is theme-ui's own, styled by the `alerts` variants in `packages/themes`; `oa-components` held only a story, which is what this removes. The theme block stays, since `Banner` and `CreateReply` still use it.

The new one is the shadcn `base-nova` Alert plus `success`, `warning`, `info`; `failure` maps to `destructive`. Built from the registry item by hand, as the CLI shells out to `bun add` and fails first. Base tuned for parity: 5px radius, 15px padding, centred 16px text, no border except `default`, inherited description colour, 10px paragraph gap. Colours are the legacy hex values, so static rather than tenant-themed — your call.

`LibraryDescription` is the only call site that splits into title and description; the rest are description only. All `data-cy` hooks preserved. In `_.email-preferences` a `<br>` became two paragraphs.

## Screenshots

Legacy Storybook, then new, at 1280px:

| Before | After |
|---|---|
|  |  |
|  |  |

At 375px, wrapping text centres and a short line sits left, in both:

| Before | After |
|---|---|
|  |  |

Font differs between the two Storybook harnesses only; in the app both inherit the body font.

## Issue

Closes #4885

## Database

## Checklist

- [ ] I ran the app and checked this change myself

First box unticked deliberately: the app needs Supabase and I have no Docker here, so the evidence is Storybook. Everything else ran clean — tsc, biome, the new test, 80 tests across the page folders touched, both builds.

Written with AI assistance (Claude Code); I read the diff and ran the checks.


src/components/ui/alert.tsx
```diff
@@ -0,0 +1,70 @@
...
+const alertVariants = cva(
+  "group/alert relative grid w-full justify-items-start gap-0.5 rounded-[5px] p-3 text-center text-base has-data-[slot=alert-action]:relative has-data-[slot=alert-action]:pr-18 has-[>svg]:grid-cols-[auto_1fr] has-[>svg]:gap-x-2 *:[svg]:row-span-2 *:[svg]:translate-y-0.5 *:[svg]:text-current *:[svg:not([class*='size-'])]:size-4",
+  {
+    variants: {
+      variant: {
+        default: 'border bg-card text-card-foreground',
+        destructive: 'bg-[#f58d8e] text-[#f4f6f7]',
+        success: 'bg-[#00c3a9] text-[#f4f6f7]',
+        warning: 'bg-[#fee77b] text-[#1b1b1b]',
+        info: 'bg-[#e2edf7] text-[#1b1b1b]',
```
mariojsnunes: these need to be added to the theme

src/components/ui/alert.tsx
```diff
@@ -0,0 +1,70 @@
+import { cva, type VariantProps } from 'class-variance-authority';
+import type * as React from 'react';
+import { cn } from '@/lib/utils';
+
+const alertVariants = cva(
+  "group/alert relative grid w-full justify-items-start gap-0.5 rounded-[5px] p-3 text-center text-base has-data-[slot=alert-action]:relative has-data-[slot=alert-action]:pr-18 has-[>svg]:grid-cols-[auto_1fr] has-[>svg]:gap-x-2 *:[svg]:row-span-2 *:[svg]:translate-y-0.5 *:[svg]:text-current *:[svg:not([class*='size-'])]:size-4",
```
mariojsnunes: use rounded-1

src/components/ui/alert.tsx
DevShoaib78: Done in 292a223. The four pairs are now theme variables mapped through `@theme inline` like the rest, and the component uses `bg-alert-success` / `text-alert-success-foreground` and so on, so there is no hex left in it.

I kept them separate from the existing `--destructive` rather than reusing it, because that one is a more saturated red used by buttons and validation, and reusing it would have changed the failure alert from the legacy pink. The values are the legacy `alerts` palette from `packages/themes` converted to oklch to match the file, and I checked each variant in Storybook afterwards to confirm they render exactly as before. The legacy theme has no dark variants for these, so dark mode inherits them.

Happy to rename them to plain `--success` / `--warning` / `--info` if you would rather they were global semantic tokens instead of alert-scoped.

mariojsnunes: just global --color-warning, etc. no "alert".

DevShoaib78: Renamed in ba41547 — `--success`, `--warning` and `--info` are global now, with their `-foreground` pairs, and the `alert-` prefix is gone.

One consequence worth your eye: the destructive variant now uses the existing global `--destructive`, since there can only be one, plus a new `--destructive-foreground`. That changes the failure alert from the legacy soft pink (`#f58d8e`) to the theme's destructive red — a real visual change from the old component, which I checked in Storybook rather than assuming. The other three are unchanged.

If you would rather keep the legacy pink, the options are to set `--destructive` to it globally, which would also repaint the destructive button and validation states, or to give the alert its own token after all. Happy either way, just say which.

mariojsnunes:
hey @DevShoaib78 , we are almost there!

sorry, we just merged a more strict linter for tailwind usage, can you take a look at the errors?

DevShoaib78:
Thanks — done, all 10 `lint:design` errors are gone.

**The four lint fixes**

- `rounded-[5px]` → `rounded-s`. Same 5px, and it's already how `ResearchContributors` and `SupporterCTA` spell it.
- `has-[>svg]:grid-cols-[auto_1fr]` → a `grid-cols-alert` `@utility` next to `z-above-header`.
- `AlertTitle` and `AlertDescription` now own their typography through `size` variants (`lg` on the title, `sm`/`xs` on the description), so the three call sites pass a prop instead of a className. Same approach you took with `TableRow` in 707ae96.
- `SettingsPageMapPin`'s `flex items-center gap-1` moves to a plain wrapper inside the description, since the description owns its spacing.

**One thing the master merge exposed**

This branch had been adding `--success` / `--warning` / `--info` to `:root` (the legacy `alerts` palette). #4915 added the same three tokens to the same block, lower down — so they were declared twice and master's values were winning silently. Column B is what that actually renders today:

alert status tokens

So I dropped this branch's `:root` block entirely and moved the variants onto the tint convention `Badge` and `Button` already use — `bg-<status>/10` with `dark:bg-<status>/20` (column C). That way the alert reads the same tokens as everything else, stays close to the old pale surface, keeps body text at ~16:1 on the tint, and gains a dark mode the legacy palette never had. `ui-globals.css` is now just the one `@utility`.

Happy to go the other way if you'd rather keep the exact legacy colours — that would mean alert-scoped tokens (`--alert-info`, etc.) so they don't collide with the badge ones. Your call; this seemed like the one that fits what you just merged.

**Checks** — `lint:design` clean on every file this PR touches, `biome ci` clean over 849 files, the alert unit tests pass (5, one new for the size variants), and `tsc` is clean on the component plus a file exercising all the new prop combinations. No bun on this machine, so those ran from a pinned local install of the same oxlint 1.85.0 / @shadcn/lint 0.1.5 / biome 2.5.0 / tailwind 4.3.3 the repo uses; the image above is the branch's own `ui-globals.css` compiled through the Tailwind CLI rather than Storybook.

PR #4914, opened 2026-09-22
chore: migrate LinkifyText to platform UI

### LINKED ISSUE #4909: [UI Migration] LinkifyText
## Context
We're migrating components from `oa-components` (theme-ui) into `src/components/ui/` (shadcn generated on Base UI, styled with Tailwind v4), following the strangler-fig approach described in the [[Platform UI README](https://github.com/ONEARMY/community-platform/blob/master/src/components/ui/README.md)](https://github.com/ONEARMY/community-platform/blob/master/src/components/ui/README.md). This is not a big-bang rewrite — this ticket migrates **one component only**. `oa-components` is not being deleted as a whole; only the piece migrated in this ticket is removed.

The goal is near-identical UI and behaviour. If the new implementation can't (or shouldn't) match the old one exactly, we've tried to note it here. If you need to make a call to make a minor change, please note why in your PR.

Reference: [[LinkifyText in the legacy Storybook](https://storybook.onearmy.earth/?path=/story/components-linkifytext--default)](https://storybook.onearmy.earth/?path=/story/components-linkifytext--default)

## Scope
- **Old component:** `packages/components/src/LinkifyText/LinkifyText.tsx`
- **New component:** `src/components/ui/linkify-text.tsx` — keeps using the `linkify-react` dependency directly (this is utility-level URL detection, not something shadcn/ui has or needs to replace), restyled with Tailwind anchor classes
- **Base UI / shadcn primitive(s) used:** none — `linkify-react` continues to do the detection; no shadcn primitive is involved

## Known differences
- **Rendering the detected links:** rather than writing a one-off `<a>` inside `linkify-react`'s `render`/component prop, use the already-migrated `ExternalLink` component (`src/components/ui/external-link.tsx`) for that render function if detected URLs are always treated as external — this keeps styling and the `rel="noopener noreferrer"` security attributes consistent with every other external link in the app instead of duplicating them here. Flag in the PR if you land on plain anchors instead and why.
- The 'supports-mentions' version is not used anymore (so doesn't need moving over)

## Screenshots / video
Attach to the PR, not just this ticket description, so reviewers see them in context:
- [ ] Old component — default state
- [ ] New component — default state
- [ ] Old vs new — any other relevant states (hover, focus, disabled, error, open/expanded, loading, dark mode if applicable, each tenant theme if colors are involved)
- [ ] If the interaction itself changed (animation, transition, keyboard behavior), a short screen recording of old vs new

## Acceptance criteria
- [ ] Implementation follows the conventions in the [[Platform UI README](https://github.com/ONEARMY/community-platform/blob/master/src/components/ui/README.md)](https://github.com/ONEARMY/community-platform/blob/master/src/components/ui/README.md) (component location, Base UI primitives, Tailwind theming, icon pattern)
- [ ] Story added to the **new** Platform UI Storybook (`bun run storybook:ui`, port 6008), not the legacy one
- [ ] Old component removed from `packages/components/` (source file, its story, any icon/asset used only by it)
- [ ] All call sites updated to import from `src/components/ui/` instead of `oa-components`
- [ ] Screenshots/video attached per above

Cory-Kim:

- [ ] 🐛 Bugfix
- [ ] ✨ Feature
- [x] ♻️ Refactoring
- [ ] ⚡️ Performance
- [x] 🧪 Tests
- [ ] 🔧 Tools / CI
- [ ] 📝 Documentation
- [x] 📦 Dependencies

## What changed

Migrates `LinkifyText` from the legacy `oa-components` package to Platform UI.

- Reimplements URL detection with `linkify-react` and Tailwind styling
- Opens detected URLs safely with `target="_blank"` and `rel="noopener noreferrer"`
- Updates all application call sites to use the new component
- Removes the legacy component and Storybook story
- Adds Platform UI stories and regression tests
- Preserves URL handling inside the legacy `CommentBody`

`ExternalLink` has not yet been migrated on `master` (#4907), so this uses a plain anchor while preserving the required styling and security attributes.

### Before

Legacy LinkifyText Storybook

### After

Platform UI LinkifyText Storybook

### After — mobile, 375px

LinkifyText at 375px mobile width

## Issue

Closes #4909

## Database

## Checklist

- [x] I ran the app and checked this change myself


mariojsnunes:
heey @Cory-Kim, welcome!

we just merged a more strict linter for tailwind usage, can you take a look at the errors?

